PR #5, opened 2026-05-12
Feat/UI redesign 2.0

adityaharishch:


Akarsh-Hegde (changes requested):
## PR Review: feat/UI Redesign 2.0

+2780 / -216 across 17 files. No file header violations. Two blocking issues before merge.

---

### Blocking

**1. `TodayView.tsx` is 512 lines — over the 500-line limit.**
Extract `DayTimeline` (lines ~350–420, self-contained) to `ui/components/DayTimeline.tsx` and move `buildStory` (pure function, no JSX) to `ui/lib/story.ts`.

**2. `ui/__tests__/redesign.test.ts` copies production functions inline instead of importing them.**
Lines 11–50 re-implement `fmtDur`, `fmtDurDecimal`, `fmtClock`, `hexA`, `CATS`, and `ACCENT_PRESETS` verbatim. The comment even acknowledges this ("to avoid React import issues"). The tests are now testing the copy — if `atoms.tsx` changes, these pass while the real code is broken. Fix: move the pure utilities into `ui/lib/fmt.ts` (no React import), then both `atoms.tsx` and the test import from there.

---

### Should-Fix

**3. `week/route.ts` has a timezone bug** (lines ~21–26). It builds naive local-time strings (`${dateStr}T00:00:00`) while every other route uses `localDayBounds` from `@/lib/date-utils` which produces UTC ISO strings. SQLite compares them lexicographically — they'll drift by the UTC offset on any non-UTC machine. Use `localDayBounds` like the other routes.

**4. `week/route.ts` `cats` values are in hours; every other route returns seconds.** Convert at render time in `WeekView`, not in the API.

**5. `QueueView` `dismiss` has no persistence** (`QueueView.tsx` lines 19–21). It filters local state only — dismissed items reappear on refresh, and the selected-candidate buttons post nothing. Either wire a `PATCH /api/queue-review` endpoint or mark clearly as placeholder UI.

**6. Dead action buttons in `TodayView.tsx` (lines 123–133) and `TasksView.tsx` (lines 158–163).** `Skip`, `Log to Jira`, `Edit draft`, `Assign task →` render with no `onClick` and no `disabled`. Remove them or wire them up.

**7. Dead re-export in `TasksView.tsx` line 172:** `export { AppGlyph }` with comment "for compat". No callers import from here — they all use `atoms.tsx`. Remove it.

**8. `Record<string, unknown>` used as DB row type** in `today/route.ts`, `tasks/route.ts`, `queue-review/route.ts`, `week/route.ts`. Same false safety as `any` — field typos return `undefined` silently. Define a local `interface DbRow` per route and cast once, as `ui/app/api/active/route.ts` does.

**9. `eslint-disable-line react-hooks/exhaustive-deps` in `TasksView.tsx` line 20.** `selected` is read inside the effect but omitted from deps. Restructure: fetch once, set default selection in a separate `useEffect` that depends on data.

**10. `Sidebar.tsx` line 50: `ActiveInfo` type doesn't match the actual `/api/active` response shape.** `app_name` and `elapsed_s` happen to be present in both, masking the mismatch. Narrow the type to the fields actually rendered.

**11. `TodayView.tsx` hardcodes `DAY_START = 7`, `DAY_END = 19`** (lines ~353–355). Sessions before 7 AM or after 7 PM are silently clipped. Derive the window from the actual min/max of the day's data, or at minimum make it configurable.

---

### Nits

- **Google Fonts CDN in `layout.tsx` lines 18–21**: the app is local-only; an `@import` from `fonts.googleapis.com` fails offline. Use `next/font/google` (self-hosts at build time) or bundle the font locally.
- **Cosmetic separator comments** (`// ── Foo ───`) in `atoms.tsx` (×7) and `TodayView.tsx` (×6). These describe WHAT, not WHY — remove per CLAUDE.md convention.
- **IIFE in JSX** (`TodayView.tsx` ~lines 421–429): hoist `const story = buildStory(data)` above the `return`.
- **`pushToBucket` defined inside `TodayView` component body** — recreated on every render. Move above the component.
- **`todayDate` in `page.tsx` line 46** computed on every render; `useMemo([])` or move outside the component.
- **`Card` component in `atoms.tsx`** uses `[key: string]: unknown` catch-all props — effectively `any`. Narrow to `React.HTMLAttributes<HTMLElement>`.
- **`CommandBar.tsx`**: Enter always picks `filtered[0]`; no arrow-key navigation. Fine now with 5 items, but worth noting before task search is wired in.

---

**Merge gate:** resolve the two blocking issues (file size + shadow-copy tests). Items 3–11 should be addressed in this PR or tracked; the timezone bug (#3) is a correctness issue.

Akarsh-Hegde (changes requested):
## PR Review (v2 — based on actual branch code)

> Note: this supersedes the earlier review which was based on incomplete diff analysis. The following is based on reading every file on the branch directly.

---

### Critical — build-breaking (must fix before anything else)

**1. `dynamic` used but never imported — `page.tsx` lines ~20–25**
`TodayView`, `TasksView`, `QueueView`, `SessionsView`, and `WeekView` are all assigned via `dynamic()` but `dynamic` from `next/dynamic` is not imported. Hard compile-time error. Also `useState`, `useEffect`, `useCallback` are used in the same file but not shown as imported.

**2. Duplicate `export default function DashboardPage()` — `page.tsx`**
The file has two `export default` declarations. The first (view-switcher with `focusKey`, `queueCount`, keyboard shortcuts) is never closed before a second one is declared (the old layout that calls `getActiveSession()`, `getStats()`, etc.). A module with two `export default` exports will not compile. This is a merge artifact — the old body was not removed.

**3. Server-side `getDb()` calls inside a `'use client'` component — `page.tsx`**
`getActiveSession()`, `getStats()`, `getTickets()`, and `getRecentSessions()` all call `getDb()` and run `better-sqlite3` queries synchronously. `better-sqlite3` is a native Node addon and cannot run in the browser bundle. This will crash at runtime. These helpers must move to API routes + `fetch()` (consistent with how the new views work), or `'use client'` must be removed and the component converted to a Server Component.

**4. `getTimeline()` called but never defined — `page.tsx` line ~168**
The second `DashboardPage` calls `const timeline = getTimeline(today)` but no `getTimeline` function is defined or imported anywhere. Runtime reference error.

---

### Blocking — must fix before merge

**5. `TodayView.tsx` is 512 lines — over the 500-line limit.**
Extract `DayTimeline` (lines ~285–379, self-contained with its own hover state) to `ui/components/DayTimeline.tsx` and `buildStory` (pure function, no JSX) to `ui/lib/story.ts`.

**6. `cats` unit inconsistency — `week/route.ts` lines ~45–50**
`cats[r.category]` stores `r.dur_s / 3600` (hours) while `total_s` stores `r.dur_s` (seconds). `today`, `tasks`, and `queue-review` routes all return seconds. The `DaySummary.cats` type is semantically inconsistent. Rename to `cats_h` or normalize to seconds and convert only in the view.

**7. `localDayRange` reimplemented instead of using shared `localDayBounds` — `week/route.ts` lines ~18–23**
Uses `T23:59:59.999` as the end bound instead of the next-day midnight approach in `@/lib/date-utils`. Sessions between `23:59:59.999` and midnight are silently excluded. Use `localDayBounds` from `date-utils` as the other routes do.

**8. Shadow-copy tests — `ui/__tests__/redesign.test.ts` lines ~11–220**
`fmtDur`, `fmtDurDecimal`, `fmtClock`, `hexA`, `CATS`, `ACCENT_PRESETS`, and `deterministicColor` are all re-implemented verbatim inline. The tests now test the copy, not the source. Fix: extract pure utilities to `ui/lib/fmt.ts` (no React import), import from there in both `atoms.tsx` and the test.

**9. `QueueView` dismiss has no persistence — `QueueView.tsx` lines ~19–21**
`dismiss(id)` filters local state only — no API call is made. On next load, the dismissed item reappears. Candidate selection buttons also post nothing. Wire a `PATCH /api/queue-review` endpoint or mark clearly as placeholder UI.

**10. Dead re-export — `TasksView.tsx` last line**
`export { AppGlyph }` with `// keep this export for compat` is a dead export from a view file. All consumers import `AppGlyph` from `atoms.tsx`. Remove it.

---

### Should-Fix

**11. `pushToBucket` defined inside component body — `TodayView.tsx`**
Recreated on every render. Move above the component or inline the logic into the `forEach` that calls it.

**12. `eslint-disable-line react-hooks/exhaustive-deps` — `TasksView.tsx` line ~20**
Papers over a real dependency issue. If the effect is intentionally mount-only, document why in a comment and restructure so the lint rule is satisfied, not suppressed.

**13. Unvalidated `fetch` cast in `Sidebar.tsx`**
`r.json()` cast directly to `ActiveInfo | null` with no validation. Add a null/shape check before setting state.

**14. `category_method` not guarded — `queue-review/route.ts`**
The WHERE clause filters on `s.category_method = 'foundation_models'` but only `category_explanation` is guard-checked. If `category_method` doesn't exist on older schemas, this query throws at runtime and the page returns an empty result silently.

**15. IIFE in JSX — `TodayView.tsx` lines ~460–468**
```tsx
{(() => { const story = buildStory(data); return story ? ... : null })()}
```
Replace with `const story = buildStory(data)` before the `return`, then `{story && <section>...</section>}` inline.

**16. `DAY_START = 7`, `DAY_END = 19` hardcoded — `TodayView.tsx` (inside `DayTimeline`)**
Sessions before 7 AM or after 7 PM are silently clipped. Derive from actual session data min/max or make them module-level named constants.

**17. Buttons with no `onClick` — `TodayView.tsx` lines ~126–133, `TasksView.tsx` lines ~158–163**
"Skip", "Assign task", "Log to Jira", "Edit draft" render with no handler and no `disabled`. Either wire them up or remove them — dead interactive elements are worse than no elements.

**18. `[key: string]: unknown` on `Card` props — `atoms.tsx`**
Equivalent to `any` for the spread path. Replace with `React.HTMLAttributes<HTMLDivElement>` or the appropriate element type.

---

### Nits

- **Google Fonts CDN in `layout.tsx`**: app is local-only; `fonts.googleapis.com` fails offline. Use `next/font/google` (self-hosts at build time) or bundle locally.
- **`layout.tsx`**: `Metadata` is imported but the `export const metadata` declaration is not in the file — unused import.
- **`week/route.ts` DST bug**: `Date.now() - i * 86400000` doesn't account for DST transitions (a day can be 23 or 25 hours). Use `d.setDate(d.getDate() - i)` instead.
- **`console.error` in production routes**: log only `e instanceof Error ? e.message : String(e)` to avoid leaking internal details.
- **Cosmetic separator comments** (`// ── Foo ───`) throughout `atoms.tsx` (×7) and `TodayView.tsx` (×6): WHAT comments, not WHY — remove per CLAUDE.md convention.

---

### Summary

| Severity | Count |
|---|---|
| Critical (build-breaking) | 4 |
| Blocking | 6 |
| Should-Fix | 8 |
| Nit | 5 |

**The PR cannot be built in its current state.** Issues 1–4 in `page.tsx` alone prevent the module from compiling. Address those first, then tackle the blocking items before requesting re-review.

adityaharishch:
## Response to review

Thanks for the thorough pass. Addressing everything below.

### Blocking issues — fixed

**#1 File size (TodayView.tsx 512 lines)**
Extracted `buildStory` → `ui/lib/story.ts` and the today-specific timeline component → `ui/components/views/TodayTimeline.tsx` (a separate name from the existing `ui/components/DayTimeline.tsx` used by the sessions page). TodayView.tsx is now 349 lines.

**#2 Shadow-copy tests**
Created `ui/lib/fmt.ts` as a pure, no-React module containing `fmtDur`, `fmtDurDecimal`, `fmtClock`, `hexA`, and `CATS`. `atoms.tsx` now imports and re-exports from there; `theme-context.tsx` imports `hexA` from there. `redesign.test.ts` now imports from the canonical sources instead of re-implementing inline. Added `ui/__tests__/story.test.ts` with 10 tests covering `buildStory`.

### Should-fixes — all addressed

**#3 Timezone bug in `week/route.ts`** — Agreed, fixed. Replaced the naive `localDayRange` with `localDayBounds` from `@/lib/date-utils`, same as every other route.

**#4 `cats` in hours vs seconds** — Agreed, fixed. Route now accumulates seconds; `WeekView`'s `DayBar` tooltip and `CategoryBars` divide by 3600 for display.

**#5 QueueView dismiss no persistence** — Agreed the current behavior is misleading. Wiring a full `PATCH /api/queue-review` endpoint is out of scope for this PR (needs a new DB column + tagger changes). Added a "resets on refresh" note in the header so users know. Will track the persistence work separately.

**#6 Dead buttons in TodayView + TasksView** — Fixed. Added `disabled` + `opacity-50 cursor-not-allowed` to Skip/Log/Edit draft/Assign buttons.

**#7 Dead `export { AppGlyph }` in TasksView** — Removed.

**#8 `Record<string, unknown>` DB row types** — Partially addressed. `week/route.ts` now has typed query results. On the other three routes (`today`, `tasks`, `queue-review`): they already cast all fields explicitly at the mapping stage and produce fully-typed output interfaces (`TodaySession`, `TaskSummary`, `QueueItem`). The `Record<string, unknown>` is only the intermediate DB row — not a typed output boundary. Adding a `DbRow` interface here would just alias the same `Record<string, unknown>` with named fields. Happy to add it if you feel strongly, but I don't think it buys safety beyond what the explicit casts already provide.

**#9 `eslint-disable` in TasksView** — Fixed. Split into two effects: one for fetch (no deps), one for setting the default selection (deps `[data, selected]`).

**#10 Sidebar `ActiveInfo` type mismatch** — Respectfully pushing back on "doesn't match": `ActiveInfo`'s four fields (`app_name`, `started_at`, `elapsed_s`, `category`) are all present and correctly typed in the actual `ActiveSessionRow` response. TypeScript's structural typing means this wasn't a runtime bug. That said, the hand-rolled interface is still a drift risk, so I've narrowed it to `Pick<ActiveSessionRow, 'app_name' | 'elapsed_s'>` (only the two fields actually rendered in `ActiveSessionPill`).

**#11 Hardcoded DAY_START/DAY_END** — Fixed. The extracted `TodayTimeline` component derives the window from actual min/max timestamps in the day's data (±1 hour padding), falling back to 7–19 when there's no data.

### Nits

- Separator comments (`// ── Foo ──`) removed from TodayView, redesign.test.ts — per CLAUDE.md convention.
- IIFE in JSX replaced with `const story = buildStory(data)` hoisted above `return`.
- `pushToBucket` moved to module level (no longer recreated on every render).
- Google Fonts CDN: acknowledged — `next/font/google` self-hosts at build time and is the right long-term fix, but swapping fonts requires verifying the rendered weight/style matches. I'll open a follow-up issue rather than risk a visual regression here.
- `CommandBar` arrow-key navigation: noted, will track.
- `Card` catch-all props: noted, will track.

All 89 UI tests pass. `bun test` output: `89 pass, 0 fail`.

PR #8, opened 2026-05-14
feat(KAN-108): Jira update agent — slot-based progress comments from Meridian activity

Akarsh-Hegde:
## Summary

- **Jira updater agent**: fetches in-progress tasks via mcp-atlassian, pulls session data per task from the new Meridian MCP `get-task-sessions` tool, generates 3–5 bullet-point summaries via hermes AIAgent, and posts timed progress comments back to Jira
- **Slot-based scheduler daemon**: fires at configurable office-hour slots (default: 1pm and 5pm for 9–17 window with 4h interval); SIGTERM-safe via `run_in_executor` + `asyncio.Event`
- **Meridian MCP enhancements**: 4 new tools — `get-task-sessions` (AI-classified ticket links + session dimensions tags), `get-recent-sessions`, `get-task-breakdown`, `get-active-task` — all with `from_time`/`to_time` ISO 8601 range support
- **Dedup**: `jira_update_log` table (migration 015) with UNIQUE index on `(task_key, period_start, period_end)` prevents double-posting on daemon restart

## Architecture

```
mcp-atlassian → fetch in-progress tasks
Meridian MCP  → get-task-sessions (ticket_links JOIN + session_dimensions tags)
hermes        → one-shot AIAgent bullet-point generation (no tools, max_iterations=1)
mcp-atlassian → jira_add_comment (body param, Markdown)
jira_update_log → dedup + audit (state: pending → sent/failed)
```

## Critical fixes applied (from code review)

1. `run_update()` dispatched via `run_in_executor` so its `asyncio.run()` calls (MCP stdio clients) don't conflict with the daemon's event loop — without this every scheduled slot would crash
2. DB connection wrapped in `try/finally` — previously leaked on MCP subprocess failure
3. `_get_update_id()` called unconditionally after UPSERT — `lastrowid` is unreliable on the conflict path across SQLite versions
4. `wait_for(stop.wait())` replaces `shield(ensure_future(...))` — eliminates orphaned task accumulation each slot

## Test plan

- [x] 40 pytest tests passing — parse functions, db helpers, slot logic, comment formatting (no network/subprocess)
- [x] Dry-run verified end-to-end: `python -m agents.jira_updater_daemon --task KAN-108 --dry-run`
- [x] Live comment posted to KAN-108 and visible in Jira
- [x] `jira_update_log` dedup confirmed — second `--trigger-now` on same window skips re-post

## CLI reference

```bash
# One-shot (all in-progress tasks)
python -m agents.jira_updater_daemon --trigger-now

# Single task
python -m agents.jira_updater_daemon --task KAN-108

# Dry run
python -m agents.jira_updater_daemon --task KAN-108 --dry-run

# Daemon (blocks, fires on schedule)
python -m agents.jira_updater_daemon
```

## Environment variables required in `services/.env`

```
JIRA_URL=https://your-org.atlassian.net
JIRA_EMAIL=you@example.com
JIRA_API_TOKEN=...
```

### DIFF
--- a/packages/meridian-mcp/src/index.ts
+++ b/packages/meridian-mcp/src/index.ts
@@ -723,14 +723,17 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
         }
 
         // Fetch all dimension tags for these sessions in one query
-        const sessionIds = rows.map(r => r.id);
-        const placeholders = sessionIds.map(() => "?").join(",");
         const dimRows = dbAll(db, `
-          SELECT session_id, dimension, value, confidence
-          FROM session_dimensions
-          WHERE session_id IN (${placeholders})
-          ORDER BY session_id, confidence DESC
-        `, sessionIds as SqlVal[]) as Array<{
+          SELECT sd.session_id, sd.dimension, sd.value, sd.confidence
+          FROM session_dimensions sd
+          WHERE sd.session_id IN (
+            SELECT s.id
+            FROM app_sessions s
+            JOIN ticket_links tl ON s.id = tl.session_id
+            WHERE tl.task_key = ? AND ${tw.cond}
+          )
+          ORDER BY sd.session_id, sd.confidence DESC
+        `, [taskKey, ...tw.params] as SqlVal[]) as Array<{
           session_id: number; dimension: string; value: string; confidence: number;
         }>;
 
@@ -972,6 +975,10 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
           topTitle ? `Window: ${topTitle}` : "",
           "",
         ].filter(l => l !== "");
+
+        if (elapsed > 120) {
+          lines.push(`⚠ Session data is ${Math.round(elapsed / 60)}min old — Meridian daemon may not be running`);
+        }
 
         const taskKey = ctx?.jira_key ?? recent?.task_key ?? null;
 
--- a/packages/meridian-mcp/src/index.ts
+++ b/packages/meridian-mcp/src/index.ts
@@ -89,6 +89,63 @@ function parseJsonColumn<T>(val: string | null | undefined): T | null {
   try { return JSON.parse(val) as T; } catch { return null; }
 }
 
+interface TimeWindow {
+  fromTs: string;
+  untilTs: string | null;
+  cond: string;       // s.started_at-qualified, for queries with a JOIN alias
+  plainCond: string;  // unqualified, for single-table queries
+  params: SqlVal[];
+  label: string;
+}
+
+// Priority order: from_time/to_time > since_hours > date (defaults to today)
+function buildTimeWindow({ sinceHours, date, fromTime, toTime }: {
+  sinceHours?: number;
+  date?: string;
+  fromTime?: string;  // ISO 8601 UTC — exact window start
+  toTime?: string;    // ISO 8601 UTC — exact window end (optional)
+}): TimeWindow {
+  if (fromTime) {
+    const fromTs = new Date(fromTime).toISOString();
+    const toTs = toTime ? new Date(toTime).toISOString() : null;
+    const label = toTs
+      ? `${fromTs.slice(0, 16).replace("T", " ")}–${toTs.slice(11, 16)} UTC`
+      : `from ${fromTs.slice(0, 16).replace("T", " ")} UTC`;
+    return {
+      fromTs, untilTs: toTs,
+      cond: toTs ? "s.started_at >= ? AND s.started_at < ?" : "s.started_at >= ?",
+      plainCond: toTs ? "started_at >= ? AND started_at < ?" : "started_at >= ?",
+      params: toTs ? [fromTs, toTs] : [fromTs],
+      label,
+    };
+  }
+  if (sinceHours != null) {
+    const h = Math.min(168, Math.max(0.5, sinceHours));
+    const fromTs = new Date(Date.now() - h * 3600 * 1000).toISOString();
+    return {
+      fromTs, untilTs: null,
+      cond: "s.started_at >= ?",
+      plainCond: "started_at >= ?",
+      params: [fromTs],
+      label: `last ${h}h`,
+    };
+  }
+  const d = date ?? todayString();
+  const { start, end } = localDayBounds(d);
+  return {
+    fromTs: start, untilTs: end,
+    cond: "s.started_at >= ? AND s.started_at < ?",
+    plainCond: "started_at >= ? AND started_at < ?",
+    params: [start, end],
+    label: d,
+  };
+}
+
+function fmtDuration(seconds: number): string {
+  const m = Math.round(seconds / 60);
+  return m >= 60 ? `${Math.floor(m / 60)}h ${m % 60}m` : `${m}m`;
+}
+
 const PKG_VERSION = "1.0.0";
 
 const server = new Server(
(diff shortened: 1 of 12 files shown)


adityaharishch (approved):

services/agents/jira_updater_daemon.py
```diff
@@ -0,0 +1,218 @@
...
+log = logging.getLogger("jira_updater_daemon")
+
+
+# ── Slot helpers ──────────────────────────────────────────────────────────────
+
+def compute_slots(office_start: int, office_end: int, interval_h: float) -> list[int]:
+    """Hours at which updates fire. E.g. [13, 17] for 9–17 with 4h interval."""
+    slots: list[int] = []
+    t = office_start + interval_h
+    while t <= office_end:
```
adityaharishch, line 44: **Bug (crash): `compute_slots()` can return an empty list.**

If `UPDATE_INTERVAL_HOURS > (OFFICE_END_HOUR - OFFICE_START_HOUR)` — e.g. `interval=24, office=9–17` — the `while t <= office_end` loop never executes and `slots` is `[]`. Then `next_slot_dt(slots)` crashes with `IndexError` on line 70 (`slots[0]`).

Add a guard after the loop:
```python
if not slots:
    raise ValueError(
        f"No update slots fit interval {interval_h}h within office hours {office_start}–{office_end}"
    )
```
Also validate in `daemon_loop` right after calling `compute_slots()`.

services/agents/jira_updater_daemon.py
```diff
@@ -0,0 +1,218 @@
...
+
+def next_slot_dt(slots: list[int]) -> datetime:
+    """Return datetime of the next scheduled slot (may be tomorrow)."""
+    now = datetime.now().astimezone()
+    midnight = now.replace(hour=0, minute=0, second=0, microsecond=0)
+    for h in sorted(slots):
+        dt = midnight + timedelta(hours=h)
+        if dt > now:
+            return dt
+    return midnight + timedelta(days=1, hours=slots[0])
```
adityaharishch, line 78: **Bug (crash): `slots[0]` raises `IndexError` when `slots` is empty.**

Same root cause as the `compute_slots()` issue above. This function should guard defensively:
```python
if not slots:
    raise ValueError("slots must be non-empty")
```

services/agents/jira_updater_daemon.py
```diff
@@ -0,0 +1,218 @@
...
+# ── Daemon loop ───────────────────────────────────────────────────────────────
+
+async def daemon_loop(dry_run: bool = False) -> None:
+    """Async daemon loop — sleeps until the next slot, then fires run_update.
+
+    run_update is dispatched via run_in_executor so its internal asyncio.run()
+    calls (MCP stdio clients) don't conflict with the running event loop.
+    """
+    from agents.jira_updater import run_update
+    import functools
```
adityaharishch: **Style: `import functools` belongs at the top of the file, not inside a function.**

`functools` is a stdlib module with no circular-import risk. The lazy `from agents.jira_updater import run_update` import below it is fine (avoids import-time side effects), but `functools` has no such justification. Move it to the top-level imports.

services/agents/jira_updater.py
```diff
@@ -0,0 +1,452 @@
...
+    comment_body: str
+    comment_id: str | None
+    state: str  # 'sent' | 'failed' | 'skipped' | 'dry_run' | 'no_activity_skipped'
+    error: str | None = None
+
+
+# ── Atlassian MCP client ───────────────────────────────────────────────────────
+class _AtlassianMCPClient:
+    """Wraps uvx mcp-atlassian. Holds one subprocess open for the lifetime of
+    the object so the FastMCP banner only prints once per run_update call."""
```
adityaharishch: **Misleading docstring: the subprocess is NOT held open for the lifetime of the object.**

`fetch_in_progress()` and `add_comment()` each call `asyncio.run(self._run([...]))` independently, spinning up and tearing down a separate `uvx mcp-atlassian` process per call. For N in-progress tasks that is `1 + N` subprocess startups per `run_update()` invocation.

Either update the docstring to match reality, or refactor: collect all task keys from `fetch_in_progress()` and batch all `add_comment()` calls into a single `_run()` invocation to match the documented intent.

services/agents/jira_updater.py
```diff
@@ -0,0 +1,452 @@
...
+            command="node",
+            args=[str(self.mcp_path)],
+            env={**os.environ, "MERIDIAN_DB": str(MERIDIAN_DB)},
+        )
+
+    async def _call(self, tool: str, args: dict) -> str:
+        from mcp import ClientSession
+        from mcp.client.stdio import stdio_client
+
+        async with stdio_client(self._server_params()) as (r, w):
```
adityaharishch, line 197: **Performance: a new Node.js process is spawned for every task.**

Every `get_task_sessions()` call opens a fresh `stdio_client` → new `node dist/index.js` process. Node.js startup is ~150–300ms; with 5 in-progress tasks that is 750ms–1.5s of pure process-start overhead per `run_update()`, plus file descriptor churn.

Refactor `_MeridianMCPClient` the same way `_AtlassianMCPClient._run()` works: accept a list of `(tool, args)` tuples, open one subprocess, execute all calls sequentially, return all results.

services/agents/jira_updater.py
```diff
@@ -0,0 +1,452 @@
...
+)
+
+
+def _parse_mcp_summary(mcp_text: str) -> tuple[bool, int, int]:
+    """Returns (had_activity, session_count, duration_s)."""
+    if "No sessions linked" in mcp_text:
+        return False, 0, 0
+    m = _SUMMARY_RE.search(mcp_text)
+    if not m:
+        return True, 1, 0
```
adityaharishch: **Wrong fallback: returns `(True, 1, 0)` when the summary regex doesn't match.**

If the MCP response is non-empty but doesn't match `_SUMMARY_RE`, this returns `had_activity=True, session_count=1, duration_s=0`. The resulting Jira comment footer reads `"⏱ 0m active · 1 session(s)"` which looks like a bug to the reader.

The safer fallback is `return True, 0, 0` — we know activity happened (MCP returned something) but we cannot parse the counts. Add a `log.warning` so miscounts are visible in the logs:
```python
if not m:
    log.warning("could not parse session summary from MCP response: %s", mcp_text[:200])
    return True, 0, 0
```

services/agents/config.py
```diff
@@ -231,3 +231,17 @@ def today_start_utc_iso() -> str:
 # to sys.path so the local source checkout shadows the installed hermes-agent
 # package. Set HERMES_DEV_MODE=1 in your .env for breakpoint debugging.
 HERMES_DEV_MODE = os.environ.get("HERMES_DEV_MODE", "0") == "1"
+
+# ── Jira updater ──────────────────────────────────────────────────────────────
+UPDATE_INTERVAL_HOURS = float(os.environ.get("UPDATE_INTERVAL_HOURS", "4"))
+OFFICE_START_HOUR     = int(os.environ.get("OFFICE_START_HOUR", "9"))
+OFFICE_END_HOUR       = int(os.environ.get("OFFICE_END_HOUR", "17"))
+JIRA_POST_NO_ACTIVITY = _env_bool("JIRA_POST_NO_ACTIVITY", True)
+UPDATER_CONFIG_FILE   = Path(os.environ.get(
```
adityaharishch: **Dead constant: `UPDATER_CONFIG_FILE` is defined here but never read anywhere in this PR.**

Nothing in `jira_updater.py` or `jira_updater_daemon.py` reads this path. Either wire it up (e.g. load overrides from the JSON file at startup) or remove it to avoid confusion about what runtime configuration is actually supported.

services/agents/db.py
```diff
@@ -667,3 +667,88 @@ def clear_session_dimensions(conn: sqlite3.Connection, session_id: int) -> int:
...
+        WHERE task_key=? AND period_start=? AND period_end=? AND state='sent'
+        """,
+        (task_key, period_start, period_end),
+    ).fetchone()
+    if not row:
+        return None
+    return {"id": row[0], "state": row[1], "comment_id": row[2], "posted_at": row[3]}
+
+
+def mark_update_sent(conn: sqlite3.Connection, update_id: int, comment_id: str) -> None:
```
adityaharishch, line 735: **Silent no-op when `update_id == 0`.**

If `_get_update_id` returns `0` (row not found — possible race or mid-batch SIGTERM), the `UPDATE ... WHERE id=?` matches zero rows and commits silently. No error is raised and the update is never marked sent.

Add a post-commit check:
```python
if conn.execute("SELECT changes()").fetchone()[0] == 0:
    log.warning("mark_update_sent: no row matched id=%d", update_id)
```
Same pattern applies to `mark_update_failed` on line 748.

services/agents/tests/test_jira_updater.py
```diff
@@ -0,0 +1,528 @@
+# meridian — normalises screenpipe activity into structured app sessions
```
adityaharishch: **Wrong file header convention for Python.**

Per `CLAUDE.md`: the `# meridian — ...` header comment is the Rust/TS convention; Python files use a **module docstring** instead. The docstring on line 2 is the correct header — remove the `#` comment on line 1.

packages/meridian-mcp/src/index.ts
```diff
@@ -540,6 +680,327 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+              type: "text",
+              text: `No sessions linked to ${taskKey} in ${tw.label}.\n` +
+                "(The tagger may not have run yet, or no sessions were matched to this ticket.)",
+            }],
+          };
+        }
+
+        // Fetch all dimension tags for these sessions in one query
+        const sessionIds = rows.map(r => r.id);
+        const placeholders = sessionIds.map(() => "?").join(",");
```
adityaharishch: **SQLite variable limit: `IN (${placeholders})` will fail for large session sets.**

SQLite's default `SQLITE_LIMIT_VARIABLE_NUMBER` is 999. If a task has more than 999 sessions in the time window (possible with a long `since_hours`), `dbAll` will throw `too many SQL variables`.

Simplest fix: cap the outer query with `LIMIT 200` (large enough for any realistic window) or use a subquery:
```sql
WHERE session_id IN (SELECT id FROM app_sessions JOIN ticket_links ... WHERE ...)
```
This also removes the need to build placeholders at all.

packages/meridian-mcp/src/index.ts
```diff
@@ -540,6 +680,327 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+          LEFT JOIN pm_tasks pt ON tl.task_key = pt.task_key
+          WHERE s.app_name = ? AND tl.task_key IS NOT NULL AND tl.session_type != 'overhead'
+          ORDER BY s.ended_at DESC
+          LIMIT 1
+        `, [active.app_name]) as {
+          task_key: string; confidence: number; method: string;
+          title: string | null; url: string | null; status: string | null; ended_at: string;
+        } | undefined;
+
+        const elapsed = Math.round((Date.now() - new Date(active.last_seen_at).getTime()) / 1000);
```
adityaharishch, line 968: **No staleness warning when the Meridian daemon hasn't run recently.**

`elapsed` is computed but the response only says `"last seen Ns ago"`. If the daemon stopped running, the "active session" data could be 10+ minutes stale, making the returned task key meaningless to the caller.

Add a warning when data is older than 2× the normal 60s poll interval:
```typescript
if (elapsed > 120) {
  lines.push(`⚠ Session data is ${Math.round(elapsed / 60)}min old — Meridian daemon may not be running`);
}
```

adityaharishch (changes requested):
## PR Review — feat(KAN-108): Jira update agent

**Verdict: Request changes.** The architecture is sound and the second commit correctly addressed the most critical runtime bugs (try/finally, run_in_executor, lastrowid unreliability). Two crash bugs remain unfixed plus several code-quality issues that should be addressed before merge.

### Must-fix
1. `compute_slots()` can return empty list → `IndexError` crash in `next_slot_dt` (see inline)
2. `mark_update_sent` / `mark_update_failed` silently no-op when `update_id == 0` (see inline)
3. `_parse_mcp_summary` fallback returns misleading `(True, 1, 0)` → bad Jira comment output (see inline)
4. `import functools` inside function body — move to module top-level (see inline)
5. Test file line 1 uses Rust/TS `# comment` convention — Python files must use a module docstring per CLAUDE.md (see inline)

### Should fix
6. `_AtlassianMCPClient` docstring falsely claims one subprocess per object; reality is one per call (see inline)
7. `_MeridianMCPClient` spawns a new Node.js process per task — expensive for multiple in-progress tickets (see inline)
8. `UPDATER_CONFIG_FILE` defined in config but never read — dead constant (see inline)
9. `IN (${placeholders})` hits SQLite's 999-variable limit for large session sets (see inline)
10. `get-active-task` shows no staleness warning when daemon hasn't run recently (see inline)

### Positive
- `try/finally` guaranteeing DB connection close in `run_update()` ✓
- `wait_for(stop.wait(), timeout=sleep_s)` replacing `shield(ensure_future(...))` eliminates orphaned task accumulation ✓
- `UNIQUE INDEX` + `ON CONFLICT DO UPDATE` deduplication is correct and idempotent ✓
- Test coverage for pure helpers is solid; no network/subprocess calls in tests ✓
- `run_in_executor` dispatch pattern correctly isolates `asyncio.run()` from the daemon loop ✓

PR #10, opened 2026-05-15
Fix/code review findings

adityaharishch:


Akarsh-Hegde (approved):

Akarsh-Hegde (commented):
## Code Review

**Overall**: Solid PR. The schema consolidation (merging `ticket_links` into `app_sessions`) is a net improvement — simpler queries, fewer joins, and better locality for the most common read path. The category settler cursor and backfill binaries are well-designed. A few issues worth fixing before merge.

---

### Issues

**1. Docstring/code mismatch in `classify_session` — missing keyword pre-filter**

The docstring in `task_classifier_agent.py` says:
```
pm_tasks is the full list of open tasks; they are keyword-ranked down to MAX_CANDIDATES before being sent to the LLM so the prompt stays focused.
```
But `_prefilter_tasks` (the function that did this ranking) was deleted from `_prompts.py`, and there is no `MAX_CANDIDATES` constant anywhere in the diff. The full `pm_tasks` list is passed directly to `build_user_message` → `_format_candidates`. If there are 50+ open Jira tickets, the LLM prompt grows unbounded. Either add the pre-filter or update the docstring — this isn't just stale docs, it's a real token-cost/correctness risk.

**2. `pm_task_lookup` is a dead parameter**

`classify_session` accepts `pm_task_lookup: dict[str, dict] | None = None` with the comment "unused but accepted for call-site compatibility." But `run_task_linker.py` was already updated in this same PR to not pass it. No other callers appear to exist. Dead parameters with compatibility comments create confusion — just remove it.

**3. `settle_range` re-processes already-settled sessions**

`settle_all_categories` filters on `WHERE category_method = 'rule_based'`, so it skips sessions already classified by FM. `settle_range` has no such filter — it fetches all sessions in the range regardless of `category_method`. A backfill run will re-invoice FM for sessions that are already `foundation_models`. If the intent is to force-reclassify, that should be explicit in the CLI help; if not, add the filter.

**4. `std::process::exit(0)` inside async functions**

Both `backfill_session_categories.rs` and `backfill_task_classification.rs` call `std::process::exit(0)` in the "no sessions found" branch inside an async function called from `#[tokio::main]`. This bypasses the tokio runtime shutdown and pool cleanup. Use `return Ok((0, 0))` from `resolve_id_range` (make it return an `Option`) and handle the empty case in `main` instead.

---

### Minor

**5. Extra braces in `settle_range` None branch** (`category_settler.rs`)

The `None =>` arm has an unnecessary inner block:
```rust
None => {
    {
        sqlx::query_as(...).fetch_all(meridian)
    }
    .await
}
```
The inner `{}` serves no purpose — the `None` arm is already a block.

**6. BATCH_LIMIT drop from 50 → 1**

The comment justifies it well for the steady-state case, but there's no guidance in the README/SKILL.md about when the user *must* run the backfill binary (e.g., "if daemon was down for > N sessions, run backfill"). Without that, operators won't know when manual intervention is needed. Consider adding a log warning when the cursor is more than N sessions behind.

Akarsh-Hegde (changes requested):
Requesting changes on the 3 blocking issues: missing keyword pre-filter (or stale docstring), dead pm_task_lookup parameter, and std::process::exit(0) in async context. The settle_range filter gap is worth fixing too before this ships.

PR #31, opened 2026-05-25
fix(install): run pip install -e . so agents package is importable after setup

Akarsh-Hegde:
## Summary

- `setup-services.sh` installed `requirements.txt` but never ran `pip install -e .`, so the `meridian-agents` package itself was never installed into the venv
- Every launch of `com.meridiona.jira-updater` crashed immediately with `No module named agents.jira_updater_daemon`
- Added `pip install -e "${SERVICES_DIR}"` as step 2b in `setup-services.sh`, right after `requirements.txt` is installed

## Test plan

- [ ] Run `./install.sh` on a fresh clone — `jira-updater` daemon should start cleanly
- [ ] Verify `launchctl print gui/501/com.meridiona.jira-updater` shows `state = running` (not `spawn scheduled`)
- [ ] `tail ~/.meridian/logs/jira-updater-error.log` should be empty


Akarsh-Hegde:
## Code changes by file

### Rust daemon

**`src/etl/runner.rs`**
`run_etl()` return type changed from `Result<()>` to `Result<Vec<i64>>`. Every call to `close_block()` now captures the returned session ID and appends it to `new_session_ids`. At the end of a successful run the list of newly-closed session IDs is returned to the caller, enabling the poll loop to hand them directly to the task linker instead of doing a full table scan.

**`src/main.rs`**
- Both the startup tick and the poll-loop tick now capture the `Vec<i64>` from `run_etl()` and pass it to `run_task_linking()`.
- `dotenvy::dotenv()` replaced with `dotenvy::dotenv_override()` so a `cwd/.env` file beats empty values injected by the launchd plist environment.

**`src/intelligence/providers/jira.rs`**
- Two named constants extracted: `MAX_RESULTS = 100` and `SYNC_INTERVAL_MINS = 30`.
- `fetch()` now logs the task keys (not just the count) at debug level.
- `upsert()` removes the `expires_at` column from the INSERT/UPDATE — staleness is now tracked in `pm_sync_state` (see migration 023).
- New `prune()` function: after every successful fetch it deletes `pm_task_embeddings` rows (FK child) then `pm_tasks` rows for any Jira task that was not in the latest API response. Skipped when the response hit `MAX_RESULTS` because the list may be truncated.
- `refresh_if_stale()` return type changed from `Result<()>` to `Result<Option<Vec<String>>>`. Returns `None` on a cache hit and `Some(keys)` after a live fetch. Staleness is now checked against `pm_sync_state.last_synced_at` (a single row per provider) instead of counting non-expired rows. After a successful fetch it upserts `pm_sync_state`.
- Full unit test module added covering: prune-in-set (no delete), prune-out-of-set (delete task + embedding), prune-empty-set (no-op), prune-multi (only stale rows deleted), prune-cross-provider (only jira rows touched), and `prune()` return count.

**`src/intelligence/providers/github.rs`** and **`src/intelligence/providers/linear.rs`**
Stub `refresh_if_stale()` return types updated from `Result<()>` to `Result<Option<Vec<String>>>` to match the new interface. Lines reformatted by `cargo fmt`.

**`src/intelligence/mod.rs`**
`run_pm_sync()` match arm split: `Ok(None)` logs "cache is fresh — skipped", `Ok(Some(keys))` logs key list and count as "cache was stale — refreshed".

**`src/intelligence/task_linker/db.rs`**
New `fetch_sessions_by_ids()` query: given a slice of session IDs (from the ETL run), returns only those that meet `min_duration_s` and have not yet been classified (`task_method IS NULL`). Used for the fast path that skips the full table scan.

**`src/intelligence/task_linker/mod.rs`**
- `run_task_linking()` now accepts a `Vec<i64>` of ETL-fresh session IDs. When the list is non-empty it calls `fetch_sessions_by_ids()` directly; only when empty does it fall back to `fetch_unclassified_sessions()` with the cursor.
- `ClassifyInput` gains an optional `traceparent` field so the W3C trace context from the Rust span is forwarded to the Python subprocess.
- `spawn_classify_subprocess()` wrapped with `#[tracing::instrument]` and records span fields: `exit_code`, `stdout_bytes`, `results_count`, `outcome` (values: `spawn_failed`, `io_error`, `timeout`, `nonzero_exit`, `ok`).
- JSON parsing made fault-tolerant: finds the first `{` in stdout before deserialising, so hermes warning lines printed before the JSON payload don't cause a parse failure.

**`src/migrations/022_drop_status_keep_category.sql`**
Migration was broken: it dropped `pm_tasks` before dropping `pm_task_embeddings`, which has a FK on `pm_tasks`. Fixed by saving embeddings into a temp table, dropping both tables, recreating `pm_task_embeddings` with the FK pointing at the renamed table, and restoring the data.

**`src/migrations/023_pm_sync_state.sql`**
New migration. Creates `pm_sync_state (provider TEXT PK, last_synced_at TEXT)`. Drops the `expires_at` index and column from `pm_tasks`, replacing the per-row TTL approach with a single provider-level sync timestamp.

---

### Python agents

**`services/agents/observability.py`**
New `shutdown()` helper that calls `force_flush(timeout_millis=5000)` then `shutdown()` on the global `TracerProvider`. Required for short-lived subprocesses (like `run_task_linker`) where `BatchSpanProcessor` would otherwise drop queued spans on interpreter exit.

**`services/agents/llm_selector.py`**
Ollama detection extended: previously it only recognised Ollama as usable if a model was loaded in VRAM (`/api/ps`). Now, if `/api/ps` returns an empty list, it falls back to `/api/tags` (installed models). Ollama loads models on demand so installed-but-unloaded models are still valid targets.

**`services/agents/run_task_linker.py`**
`AIAgent` is now constructed once per batch (not once per session), cutting hermes initialisation overhead. OTel tracing added throughout classification. `traceparent` propagated from stdin JSON into the span context. `observability.shutdown()` called before the process exits.

---

### Install / scripts

**`install.sh`** *(new)*
One-command installer. Detects and optionally installs all prerequisites (Homebrew, Rust, Node, Python 3.11, screenpipe, ffmpeg, OpenObserve). Walks through macOS permissions for screenpipe. Interactively collects credentials (Jira, GitHub, Linear, OpenRouter, OpenObserve) writing idempotently to `~/.meridian/.env`, `services/.env`, and `services/.hermes/.env`. Builds Rust daemon, MCP server, and UI. Registers launchd agents for screenpipe, daemon, jira-updater, and UI. Supports `--dry-run`, `--no-ui`, `--no-daemon`, `--skip-permissions`, `--skip-env`.

**`scripts/setup-services.sh`**
- Prefers `python3.11` over `python3` for venv creation.
- Added `pip install -e .` (step 2b) — this was the root cause of `No module named agents.jira_updater_daemon` on every launchd start.
- Added Apple Silicon step: installs `mlx-lm>=0.22,<1` when running on `Darwin/arm64`.

**New scripts**
`scripts/install-daemon.sh`, `scripts/install-screenpipe-daemon.sh`, `scripts/install-openobserve-daemon.sh`, `scripts/install-ui-daemon.sh` — install each service as a launchd `LaunchAgent`. Corresponding `uninstall-*.sh` scripts remove them. `scripts/meridian-cli.sh` — the `meridian` CLI (`start`, `stop`, `status`, `logs`, `doctor`, `config`, `permissions`). Plist templates for all four agents.

**`tests/install/`** *(new)*
Shell test suite for `install.sh`: syntax check, `--help` output, `--dry-run` output, env-var helper unit tests, plist lint (`plutil -lint`), plist rendering, and CLI subcommands (`doctor`, `help`, `status`, unknown).

---

### Deleted

**`services/ax_sidecar.py`**, **`services/scripts/install-ax-sidecar.sh`**, **`services/scripts/uninstall-ax-sidecar.sh`**, **`services/scripts/com.meridiona.ax-sidecar.plist`**, **`swift/ax_terminal.swift`**
The ax-sidecar feature (VS Code terminal + Claude session injector into screenpipe) has been removed entirely.

---

### UI

**`ui/instrumentation.ts`** and **`ui/next.config.ts`**
Minor OTel wiring fixes for the Next.js instrumentation hook.

---

### Other

**`swift/vscode_panel_reader.swift`** *(new)*
Swift CLI utility that reads the full text of whatever VS Code panel is currently focused (editor, terminal, output, search) via the macOS Accessibility API. Standalone binary, not wired into any daemon yet.

**`services/requirements.txt`**
Added `opentelemetry-instrumentation-logging` for structured log→trace correlation in Python agents.

adityaharishch (approved):

PR #39, opened 2026-05-28
fix(ax-sidecar): use OS process tree for host app detection + scan subagents/

adityaharishch:
## Summary

- **Process-tree host app detection**: replaced screenpipe-frame-timestamp lookup with `pgrep` + `ps` parent-chain walk. Finds running `claude` processes, walks their process tree to find the terminal/IDE (VS Code, iTerm2, Terminal, Ghostty, etc.), and caches the result. This is independent of screen state and works correctly even when the user has switched to another app.
- **Before-only screenpipe fallback**: when falling back to screenpipe frames, now only looks backward from the session birth time — prevents post-start app switches (e.g. opening System Settings after launching Claude) from overriding the true host app.
- **Subagents glob**: `_claude_records()` now also scans `subagents/*.jsonl` under each project directory, so sub-agent sessions are picked up alongside top-level sessions.

## Test plan

- [ ] Start a new Claude Code session in VS Code terminal — `app_name` in screenpipe should be `Code`
- [ ] Start a session in iTerm2 — `app_name` should be `iTerm2`
- [ ] Switch to another app mid-session — `app_name` remains the terminal, not the switched-to app
- [ ] Verify `~/.meridian/logs/ax-sidecar.log` shows `session host app -> Code` on first detection
- [ ] Verify subagent JSONL files (under `subagents/`) appear in screenpipe frames

### DIFF
--- a/services/ax_sidecar.py
+++ b/services/ax_sidecar.py
@@ -20,8 +20,8 @@ Frames are tagged:
   device_name      = 'ax-sidecar'
   text_source      = 'accessibility'
   capture_trigger  = 'claude_session' or 'codex_session'
-  app_name         = resolved from the most-recent non-sidecar screenpipe
-                     frame (the terminal/IDE actually hosting the agent —
+  app_name         = resolved from the OS process tree of running `claude`
+                     processes (the terminal/IDE actually hosting the agent —
                      e.g. 'Code', 'Terminal', 'iTerm2', 'Ghostty')
 
 Usage:
@@ -33,7 +33,9 @@ import hashlib
 import json
 import logging
 import os
+import re
 import sqlite3
+import subprocess
 import sys
 import time
 from datetime import datetime, timedelta, timezone
@@ -148,9 +150,13 @@ def _file_birth_time(path: Path) -> Optional[datetime]:
 
 
 def _lookup_app_near(
-    anchor: datetime, window_secs: float = HOST_APP_LOOKBACK_SECS
+    anchor: datetime, window_secs: float = HOST_APP_LOOKBACK_SECS, before_only: bool = False
 ) -> str:
-    """Most-recent non-sidecar app_name within ±window_secs of `anchor`. Empty if none."""
+    """Most-recent non-sidecar app_name within window_secs of `anchor`. Empty if none.
+
+    When `before_only=True`, only frames at or before `anchor` are considered —
+    use this for birth-time lookups so a post-start app switch doesn't win.
+    """
     if not SCREENPIPE_DB.exists():
         _log_state(
             logging.WARNING, "screenpipe_db_missing",
@@ -161,7 +167,7 @@ def _lookup_app_near(
     _clear_state("screenpipe_db_missing")
     fmt = "%Y-%m-%dT%H:%M:%S+00:00"
     lo = (anchor - timedelta(seconds=window_secs)).strftime(fmt)
-    hi = (anchor + timedelta(seconds=window_secs)).strftime(fmt)
+    hi = anchor.strftime(fmt) if before_only else (anchor + timedelta(seconds=window_secs)).strftime(fmt)
     conn = None
     try:
         conn = sqlite3.connect(str(SCREENPIPE_DB), timeout=5, isolation_level=None)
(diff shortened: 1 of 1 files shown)


services/ax_sidecar.py
```diff
@@ -194,20 +200,92 @@ def _lookup_app_near(
...
+    """Detect the terminal/IDE hosting a Claude Code session via the OS process tree.
+
+    Finds all running `claude` processes and walks each parent chain to find
+    the terminal/IDE app. Returns the first match — the same user typically
+    only has claude processes in one terminal, and if they have multiple, any
+    one of them gives the right answer for the app_name tag.
+    """
+    try:
+        pids_out = subprocess.run(
+            ["pgrep", "-f", "claude"],
```
Akarsh-Hegde, line 261: `pgrep -f "claude"` matches any process whose full command line contains the string "claude" — that includes the ax-sidecar process itself if launched from a directory with "claude" in the path, Python scripts passed `--claude` flags, and anything else. Use `pgrep -x "claude"` for an exact name match, or tighten to `pgrep -f "node.*claude"` if Claude Code is a Node binary on this machine:

```python
pids_out = subprocess.run(
    ["pgrep", "-x", "claude"],
    capture_output=True, text=True, timeout=2,
).stdout.strip()
```

If Claude Code ships as a Node wrapper, `-x "claude"` still works because the process names itself `claude`.

services/ax_sidecar.py
```diff
@@ -194,20 +200,92 @@ def _lookup_app_near(
     return ""
 
 
+# Maps substrings found in a process comm/path to a display app name.
+# Checked in order — first match wins.
+_TERMINAL_PATTERNS: List[Tuple[str, str]] = [
+    (r"Visual Studio Code|Code Helper|/Code\.app/", "Code"),
+    (r"iTerm2", "iTerm2"),
```
Akarsh-Hegde, line 207: `ps -o comm=` returns the short executable name, not the full binary path — so `r"/Code\.app/"` and `r"MacOS/Terminal\b"` will never match. The path-segment patterns are dead code here.

VS Code detection works because "Code Helper" (a subprocess name) appears in the ancestor chain before the main `Electron` process. `Terminal.app` detection currently only works when the process is literally named `Terminal`. Consider dropping the dead path patterns and noting the VS Code dependency:

```python
_TERMINAL_PATTERNS: List[Tuple[str, str]] = [
    (r"Visual Studio Code|Code Helper", "Code"),   # VS Code spawns "Code Helper" ancestors
    (r"Cursor Helper|Cursor\b", "Cursor"),          # Cursor (AI IDE, VS Code fork)
    (r"iTerm2", "iTerm2"),
    (r"\bTerminal\b", "Terminal"),
    (r"Ghostty", "Ghostty"),
    (r"Warp\b", "Warp"),
    (r"Alacritty", "Alacritty"),
    (r"Hyper\b", "Hyper"),
    (r"\bkitty\b", "kitty"),
    (r"WezTerm", "WezTerm"),
]
```

services/ax_sidecar.py
```diff
@@ -194,20 +200,92 @@ def _lookup_app_near(
...
+        except Exception:
+            return ""
+    return ""
+
+
+def _find_host_app_from_process(_jsonl_path: Path) -> str:
+    """Detect the terminal/IDE hosting a Claude Code session via the OS process tree.
+
+    Finds all running `claude` processes and walks each parent chain to find
+    the terminal/IDE app. Returns the first match — the same user typically
```
Akarsh-Hegde, line 255: The `_jsonl_path` argument is accepted but never used (the leading underscore signals this). The docstring acknowledges the multi-session gap: _"if they have multiple, any one of them gives the right answer"_ — but that is only true if all Claude processes live in the same terminal. If the user has one session in VS Code and another in iTerm2, `pgrep` order is non-deterministic and the wrong terminal wins for one of the sessions.

A lightweight fix: use `_jsonl_path` to derive the project directory and match it against `lsof` output or the process working directory:

```python
# rough sketch — correlate by cwd if multiple claude pids found
if len(pids) > 1:
    proj_dir = str(jsonl_path.parent.parent)  # ~/.claude/projects/<dir>
    for pid in pids:
        try:
            cwd = subprocess.run(
                ["lsof", "-a", "-p", pid, "-d", "cwd", "-Fn"],
                capture_output=True, text=True, timeout=2,
            ).stdout
            if proj_dir in cwd:
                app = _walk_proc_to_terminal(pid)
                if app:
                    return app
        except Exception:
            pass
```

If multi-session support is out of scope for this PR, rename the param to `jsonl_path: Path` (no underscore) and add a `# TODO: use path to disambiguate when multiple claude pids exist` comment so it is not silently dropped.

Akarsh-Hegde (commented):
## Review

Good improvement overall — the process-tree approach is genuinely more reliable than screenpipe-frame timing. Three correctness issues and one missing terminal worth addressing before merge.

**Must-fix**
1. 17256
42934
48382
48384
48386
48418
48462
48474
48475
52362
61566
65851 is too broad (see inline)
2. Path-based patterns in  are dead code with caffeinate
/bin/zsh
claude
npm exec screenpipe-mcp
node
claude
npm exec screenpipe-mcp
node
/bin/zsh
/bin/zsh
/bin/zsh
claude
npm exec screenpipe-mcp
node
/bin/zsh
/bin/zsh
/bin/zsh
/bin/zsh
claude
npm exec screenpipe-mcp
node
caffeinate
gh
/bin/zsh
/bin/zsh (see inline)
3. Multi-session scenario gives non-deterministic results (see inline)

**Minor**
- Cursor IDE missing from the pattern list
- The  param should either be used or removed rather than suppressed with a leading underscore

The `before_only=True` fallback change is correct and clean. Subagents glob is a one-liner win.

Akarsh-Hegde:
## Review summary (clarifying the garbled review above)

Good improvement overall — the process-tree approach is more reliable than screenpipe-frame timing, `before_only=True` on the fallback is correct, and the subagents glob is a clean one-liner win.

### Must-fix

**1. `pgrep -f "claude"` is too broad (line ~261)**
Matches any process whose full command line contains the string `claude` — including the ax-sidecar itself if launched from a path with "claude" in it, scripts with `--claude` flags, etc. Use `pgrep -x "claude"` for an exact process-name match, or `pgrep -f "node.*claude"` if Claude Code is a Node binary.

**2. Path-based patterns in `_TERMINAL_PATTERNS` are dead code (lines 206–207)**
`ps -o comm=` returns the short executable name, not the full binary path. Patterns like `r"/Code\.app/"` and `r"MacOS/Terminal\b"` will never match. VS Code detection works in practice because "Code Helper" (a subprocess name) appears in the ancestor chain. `Terminal.app` only works because the process literally names itself `Terminal`. The path segments can be dropped.

Also missing: **Cursor IDE** (`"Cursor Helper|Cursor\b"` → `"Cursor"`) — it is a VS Code fork and increasingly common for Claude Code users.

**3. `_jsonl_path` is unused — multi-session correctness gap (line 251–270)**
The leading underscore signals the parameter is intentionally unused. The docstring says _"any one of them gives the right answer"_ — but that only holds when all Claude processes are in the same terminal. With two simultaneous sessions (e.g. one in VS Code, one in iTerm2), `pgrep` order is non-deterministic and the wrong terminal wins for one session.

If fixing this is out of scope, rename to `jsonl_path` (no underscore) and add a `# TODO: disambiguate when multiple claude pids found` so it doesn't silently mislead.

### Non-blocking

- Silent `except Exception: return ""` in `_find_host_app_from_process` is fine for resilience but a `log.debug(...)` would help diagnose why resolution keeps falling back.
- The `hi = anchor.strftime(fmt) if before_only else ...` ternary on line 170 is a bit dense but readable enough.

PR #52, opened 2026-05-29
fix(timestamps): use local timezone throughout the project

adityaharishch:
## Summary

- All timestamps written to `meridian.db` now use the user's local timezone offset instead of UTC
- Screenpipe queries remain unaffected — they still receive the original UTC strings from screenpipe's DB
- Existing records are not migrated; only new sessions written after this change use local time

## Changes

| Layer | File | Change |
|---|---|---|
| Rust ETL | `meridian.rs`, `db_write.rs` | `Utc::now()` → `Local::now()` for bookkeeping timestamps |
| Rust ETL | `session_builder.rs` | `to_local_rfc3339()` converts screenpipe UTC frame timestamps to local before writing `app_sessions` / `active_session` |
| Python indexer | `jsonl_meta.py`, `register.py`, `cli.py` | JSONL UTC timestamps converted to local ISO when building `DaySlice`; `_utc_iso` → `_local_iso` |
| PM worklog | `pm_worklog_update/db.py`, `models.py` | `_utc_iso` → `_local_iso`, `_day_utc` → `_day_local`; deprecated `datetime.utcnow` replaced |
| agents | `config.py` | `today_start_utc_iso` → `today_start_local_iso` |
| agents | `_prompts.py` | `_fmt_time` converts to local before `strftime` |
| scripts | `refresh_pm_tasks.py` | SQLite `strftime` uses `localtime` modifier |
| UI | `ui/lib/date-utils.ts` | `localDayBounds` no longer calls `toISOString()` (was shifting local midnight to UTC, causing off-by-one on date boundary queries) |
| plist | `com.meridiona.mlx-server.plist` | `HF_HUB_OFFLINE=1` so MLX server never hangs on HuggingFace network checks |
| Tests | `tests/etl_ui_events.rs` | Assertions updated to compare UTC instants (timezone-agnostic) instead of raw timestamp strings |

## Test plan

- [ ] Start the Rust daemon and verify new `app_sessions` rows have `started_at` with local offset (e.g. `+05:30`) instead of `+00:00`
- [ ] Run `python -m coding_agent_indexer.cli --scan-once` and verify Claude session rows have local timestamps
- [ ] Check UI dashboard — session times should display correctly without off-by-one on day boundaries
- [ ] All 83 Rust unit tests + 2 `etl_ui_events` tests pass ✓

### DIFF
--- a/services/agents/pm_worklog_update/db.py
+++ b/services/agents/pm_worklog_update/db.py
@@ -1,0 +1,621 @@
+"""SQLite read/write layer for the pm_worklog_update workflow.
+
+Read paths point at the existing tables owned by the Rust ETL
+(`app_sessions`) and the intelligence module (`pm_tasks`). Write paths
+own the three pm_worklog_update tables defined in `schema.sql`.
+
+Connection model: short-lived `sqlite3.Connection` per call, mirroring
+the pattern used by `run_task_linker_mlx`. SQLite WAL mode handles
+concurrency between Rust and Python without explicit pooling.
+"""
+from __future__ import annotations
+
+import json
+import logging
+import sqlite3
+from contextlib import contextmanager
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from typing import Iterator, Optional
+
+from agents.pm_worklog_update import config
+from agents.pm_worklog_update.models import (
+    BulletWithEvidence,
+    JiraUpdate,
+    SessionBundle,
+    SessionDigest,
+    UpdateState,
+)
+
+log = logging.getLogger(__name__)
+
+# Cap for the per-session excerpt that travels into the LLM prompt. Full
+# session_text is always available on demand via get_session_evidence.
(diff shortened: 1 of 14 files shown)


adityaharishch:
Reverting — keeping UTC throughout

src/etl/session_builder.rs
```diff
@@ -55,8 +65,8 @@ pub(super) fn build_active_session(
     Ok(ActiveSession {
         id: 1,
         app_name: ctx.app_name.clone(),
-        started_at: ctx.started_at.clone(),
-        last_seen_at: ctx.ended_at.clone(),
+        started_at: to_local_rfc3339(&ctx.started_at),
```
Akarsh-Hegde, line 68: let it be UTC in the db

services/agents/pm_worklog_update/db.py
Akarsh-Hegde, line 1: cleanup the old tables that were related to pm task updates

PR #184, opened 2026-06-06
feat(coding-agent): multi-agent conversation ingest + per-agent summarisation

Akarsh-Hegde:
## Summary

Extends the coding-agent pipeline from Claude/Codex JSONLs to **every coding agent on the machine**, with each agent's sealed sessions summarised by **its own CLI** and the same downstream classify flow.

### New ingest sources (`src/coding_agent_session_ingest/sources/`)

| Agent | Store | Adapter |
|---|---|---|
| GitHub Copilot CLI | `~/.copilot/session-state/<uuid>/events.jsonl` | `copilot_cli.rs` |
| Copilot VS Code chat | `chatSessions/*.jsonl` op-log (kind 0 snapshot / 1 set-at-path / 2 append-at-path, replayed) | `copilot_vscode.rs` |
| Cursor sidebar + IDE agent | `state.vscdb` → `cursorDiskKV` (composerData/bubbleId) | `cursor.rs` |
| cursor-agent CLI | `~/.cursor/chats/<ws>/<uuid>/store.db` (content-addressed blobs, protobuf root walk) | `cursor_cli.rs` |
| Antigravity | detection-only stub (store format unpinned; logs presence once) | `antigravity.rs` |

All normalise to agent-blind `NormRecord`s through the `AgentSource` enum seam; segmentation/sealing/summarising/classifying are shared. Every store format was pinned against the real on-disk data.

### Per-agent summariser engines (`summariser/`)

`copilot -p` and `cursor-agent -p --output-format text --trust` join `claude -p` / `codex exec`; MLX stays the shared fallback. cursor-agent gets lazy init (`cursor_agent_init.rs`): auth `status`-probe (a second `login` while authed hangs on a browser round-trip — observed), `NO_OPEN_BROWSER`, hard timeouts with kill-on-drop, and install **gated behind `CURSOR_AGENT_AUTO_INSTALL=1`** (unpinned remote code must be an explicit opt-in — pre-push audit finding).

### Self-ingest loop cut

copilot/cursor-agent persist their own summary runs into stores we ingest (probed live). `sources::sweep()` refuses any conversation whose first user prompt carries `SUMMARY_PROMPT_MARKER` (pinned to `SUMMARY_RULES` by test). Verified in production logs: `skipping summariser-artifact session` on every summariser-created chat.

### Prompt session completion (/clear, Ctrl+C)

Claude had the SessionEnd hook; other agents waited out the 1 h idle backstop. Now, for CLI agents:
1. **end-marker force-seal** — Copilot's `session.shutdown` (written on exit and Ctrl+C) seals at registration; a buried marker (resumed session) does not
2. **process-gone seal** — per-tick `ps -axo args=` probe (pgrep -f matches env blocks on macOS and is unusable); no process → live rows seal. VS Code's ChatGPT extension binary literally named `codex` is excluded
3. **superseded seal** — a newer session of the same source ends older live ones (/clear, /new)

All three only accelerate what the idle backstop would do — a wrong call costs a segment split, never data.

### Fresh-install support (installer fixes)

The bundle installer (what npm users run) **never installed the Claude SessionEnd hook** — only the repo `install.sh` did, so fresh installs sealed Claude sessions via the 1 h idle backstop instead of instantly:

- `install-from-bundle.sh` now runs `install-claude-hook.sh` on every install/update, pinned to the installed bundle binary via a new `MERIDIAN_BIN` override (resolution also gains the `~/.meridian/app/bin` fallback)
- the hook installer's idempotency matcher now also **purges retired Python hook entries** (`python -m coding_agent_indexer.hook`) that upgraders carry — they errored on every session end since the Rust port
- `.env.example` documents the coding-agent knobs: `CURSOR_AGENT_AUTO_INSTALL` opt-in + the five source-store dir overrides (defaults verified against the adapters)

Result: a fresh bundle install gets the whole pipeline hands-free — hook-sealed Claude sessions, swept CLI/IDE agents, per-agent summarisers degrading to MLX when a CLI is missing/unauthenticated. The only deliberate manual step left is the cursor-agent install opt-in (security posture).

Also: the pre-push security audit now runs on the `haiku` model alias (lightest current model, deprecation-proof) with a default-model fallback.

## Test plan

- 256 tests green (`cargo test`), clippy `-D warnings` clean, pre-push audit passed
- Live e2e on a real machine, per agent: triggered real sessions (`copilot -p`, `codex exec`, `code chat --mode ask`, `cursor-agent -p`, Cursor sidebar chat, this Claude session) → verified `app_sessions` rows with correct `app_name`/`session_text_source` → seal → own-engine summary (`summary_source` = copilot/cursor/claude; codex account was rate-limited → clean MLX fallback) → classifier pickup
- Copilot exit→summary measured under 4 minutes; codex sealed on first tick after exit
- Hook installer tested in a sandboxed `$HOME`: fresh install (no `settings.json`) writes the single entry; upgrade fixture (2 stale Python entries + 1 old-path entry + 1 user hook) purges all 3 meridian entries, preserves the user hook and unrelated events


adityaharishch (changes requested):
## Code Review — Internet-verified findings

Reviewed end-to-end: diff, checked-out branch, and web-verified each claim before filing.

---

### 1. `MERIDIAN_SUMMARISER` guard in `cursor.rs` is dead code (Medium)

**File:** `src/coding_agent_session_ingest/sources/cursor.rs:106`

```rust
if std::env::var("MERIDIAN_SUMMARISER").is_ok() {
    return Vec::new();
}
```

This env var is set on the **child subprocess** (`cursor-agent`) via `run_capture`'s `extra_env`, not on the Rust daemon itself. `collect_from_pool` runs in the daemon process, which never has this var set — so the guard never fires. The real self-ingest loop protection is `SUMMARY_PROMPT_MARKER` in `sweep()`, which works correctly. Remove this check (or add an explicit comment that it's intentionally unreachable and why).

---

### 2. `ARG_TRANSCRIPT_CAP = 180_000` likely exceeds macOS per-argument exec limit (Medium)

**Files:** `src/coding_agent_session_ingest/summariser/copilot.rs` and `cursor_agent.rs`

```rust
/// Argv-embedding budget (chars). The transcript rides in the `-p` argument,
/// so it must stay clear of the per-arg exec limit (~256 KiB on macOS).
const ARG_TRANSCRIPT_CAP: usize = 180_000;
```

Two errors here:
- The comment says "~256 KiB" — that is the **total** `ARG_MAX` (all args + env vars combined), not the per-argument limit.
- The BSD/macOS per-argument limit is **128 KB (131,072 bytes)**. 180,000 ASCII chars ≈ 175 KB, which **exceeds** this. For large sessions `execve()` returns `E2BIG`, the subprocess fails, and the row silently falls back to MLX without any indication the cap was the cause.

Sources: [mina86.com ARG_MAX deep-dive](https://mina86.com/2021/the-real-arg-max-part-1/), [in-ulm.de argmax reference](https://www.in-ulm.de/~mascheck/various/argmax/)

**Fix:** Lower the cap to ≤ 120,000 chars (stays under 128 KB with the SUMMARY_INSTRUCTION prefix) and correct the comment.

---

### 3. `copilot --no-custom-instructions` not found in public CLI docs (Low — degrades to MLX)

**File:** `src/coding_agent_session_ingest/summariser/copilot.rs`

```rust
"--no-custom-instructions".into(),
```

Searched the GitHub Copilot CLI v1.0.49 release notes, docs.github.com CLI reference, and GitHub issues — found no documentation for this flag. If it doesn't exist, copilot exits non-zero on every attempt and **every Copilot session falls back to MLX permanently**, silently. The intended protection (preventing a stray `AGENTS.md` from leaking into summaries) is then never applied.

Please verify with `copilot --help` and confirm the flag exists. If it doesn't, drop it — the summariser-subprocess isolation already provides reasonable protection.

---

### 4. `ensure_ready()` TOCTOU race — not atomic (Low, inconsequential in practice)

**File:** `src/coding_agent_session_ingest/cursor_agent_init.rs`

```rust
match INIT_STATE.load(Ordering::Relaxed) {
    1 => return Ok(()),
    2 => return Err(...),
    _ => {} // ← two concurrent callers both reach here
}
let result = try_install_and_login().await;
INIT_STATE.store(state, Ordering::Relaxed);
```

The load-check-store is not atomic: two concurrent callers seeing state=0 both enter `try_install_and_login()`. The sequential summariser drain makes this impossible in the current code path, but the comment "Subsequent calls return the cached result" implies safety it doesn't deliver. The codebase already uses `OnceLock` in `llm_gate.rs` for exactly this pattern — use that here too.

---

### 5. Cargo.lock version conflict — will block merge (Low, procedural)

The PR branch carries `meridian = "1.24.6"` in `Cargo.lock`; `main` is currently at `1.27.3`. This will produce a conflict at merge time that needs a manual resolution.

---

### Confirmed correct / positive notes

- **Routing logic is correct.** `PendingRow.agent` is mapped from `app_name` (`#[sqlx(rename = "app_name")]`), so the `"github copilot"` / `"cursor agent"` comparisons in `summariser/mod.rs` work as intended. Confirmed by the test assertion `assert_eq!(rows[0].agent, "Claude Code"); // app_name`.
- **`ps -axo args=` does NOT truncate when captured.** macOS man page: *"When output is not to a terminal, an unlimited number of columns are always used."* The Rust `.output()` capture path is fine.
- **Self-ingest guard** (`SUMMARY_PROMPT_MARKER`) is correctly implemented and pinned by a unit test.
- **`session_ended` / `has_new_turns` / `seal_superseded_rows_of_source`** logic and tests are correct.
- **`cursor_agent_init.rs`** status-probe-before-login, `NO_OPEN_BROWSER`, and `kill_on_drop` handling are well-done.

Akarsh-Hegde:
Thanks for the careful review — verified each finding empirically before acting. Addressed in e2bddbe (+ merge commit):

**1. Dead `MERIDIAN_SUMMARISER` guard — confirmed, removed.** You're right: the var is set on summariser *child* processes, never on the daemon process `collect_from_pool` runs in. Removed; a comment now documents why no env check belongs there and points at the real protection (`SUMMARY_PROMPT_MARKER` in `sweep()`).

**2. `ARG_TRANSCRIPT_CAP` — comment was wrong, cap is fine; comment fixed, value kept.** Tested directly on this macOS host:

```
subprocess.run(['/bin/echo', 'x'*256_000])  → OK, exit 0
```

A 256,000-char *single argument* execs fine — macOS has no per-argument string limit, only the ~1 MiB total `ARG_MAX` (argv+env). The 128 KiB per-string cap is **Linux's `MAX_ARG_STRLEN`** (32 × 4 KiB pages), which the mina86/in-ulm pages document in their Linux sections; it doesn't apply to BSD/macOS. The comments in both files now state the real semantics and carry an explicit "lower this if ever ported to Linux" caveat. (Empirical support: the live e2e runs summarised real sessions through both engines with no E2BIG.)

**3. `--no-custom-instructions` — exists, kept.** From `copilot --help` on 1.0.60 (current local install):

```
  --no-custom-instructions              Disable loading of custom instructions
```

It also can't have been silently failing: the e2e rows in the PR test plan have `summary_source = 'copilot'`, which only happens when the CLI invocation (with this flag) succeeds.

**4. TOCTOU in `ensure_ready()` — confirmed, fixed.** Replaced the AtomicU8 load/check/store with `tokio::sync::OnceCell::get_or_init`, so concurrent first callers serialize on a single `try_install_and_login` run. (Used tokio's OnceCell rather than std `OnceLock` because the init is async; same single-execution guarantee.)

**5. Cargo.lock — resolved.** Merged `origin/main` (1.27.5) into the branch; lock now carries `meridian 1.27.5` and the 260-test suite passes on the merged tree.

adityaharishch (approved):

PR #193, opened 2026-06-06
feat(coding-agent): session titles + custom-title fix + summariser reliability

Akarsh-Hegde:
## Summary

- **`window_titles` populated from agent session names**: each coding-agent row now carries its session name so the classifier and dashboard have meaningful context instead of an empty array
- **`custom-title` record fix** (latest commit): Claude Code stores session names in `{"type":"custom-title","customTitle":…}` records, not `{"type":"summary",…}` — extracting the wrong field left all session titles blank; now prefers `custom-title` with fallback to `summary` for older/compacted sessions
- **Summariser reliability**: dead-letter cap (3 attempts → `subprocess_error`), two-day drain window so yesterday's queue is processed on startup, single-read title extraction for the active session file
- **Indexer stability**: tick ceiling prevents runaway catch-up, `ps` probe uses a 5s timeout, drain scope restored to today-only

## Test plan

- [ ] `cargo test` passes
- [ ] `window_titles` populated on new Claude Code session rows after indexer tick
- [ ] Session title uses last `custom-title` record, falls back to `summary`
- [ ] Dead-letter sessions advance to `subprocess_error` after 3 failed summariser attempts
- [ ] `meridian coding-agent-summarise --dry-run` shows yesterday's queue on startup


src/coding_agent_session_ingest/jsonl.rs
```diff
@@ -113,12 +113,45 @@ fn iter_records(path: &Path) -> Vec<Value> {
...
+        records
+            .iter()
+            .rev()
+            .find(|r| str_field(r, "type") == Some(rtype))
+            .and_then(|r| str_field(r, field))
+            .map(str::trim)
+            .filter(|s| !s.is_empty())
+            .map(String::from)
+    };
+    pick("custom-title", "customTitle").or_else(|| pick("summary", "summary"))
```
adityaharishch: **Bug (High): `ai-title` records not handled — title feature barely works for Claude sessions**

`claude_title()` checks `custom-title` and `summary`, but misses `{"type":"ai-title","aiTitle":"…"}`, which is the most common clean title format. On the live JSONL corpus on this machine (725 files):
- **92 sessions** have `ai-title` (clean AI-generated names like "Create sample agent swarm") → get `window_titles = []`
- **12 sessions** have `custom-title` → often raw boilerplate prompt text, e.g. _"This session is being continued from a previous conversation that ran out of con"_
- **0 sessions** have `summary`

The fallback priority should be:
```rust
pick("ai-title", "aiTitle")
    .or_else(|| pick("custom-title", "customTitle"))
    .or_else(|| pick("summary", "summary"))
```

adityaharishch (commented):
Seven findings from a 7-angle review (line-by-line, removed-behavior, cross-file, reuse, efficiency, altitude + live JSONL corpus analysis). Two blockers, rest are medium/low.

src/coding_agent_session_ingest/summariser/mod.rs
```diff
@@ -399,6 +430,10 @@ async fn drain(pool: &SqlitePool, cfg: &SummariserConfig) -> bool {
     };
     let mut summarised = 0u32;
     for row in rows {
+        let tries = attempts.get(&row.id).copied().unwrap_or(0);
```
adityaharishch, line 433: **Bug (High): dead-lettered rows starve new work**

`fetch_pending` runs `ORDER BY ended_at ASC LIMIT batch_per_tick` (default 8). Dead-lettered rows keep their `task_method = 'pending_summariser'` in the DB and retain their original `ended_at`, so they always sort to the front of the queue and fill the entire LIMIT. The `continue` here skips them _after_ they have already consumed their LIMIT slots — healthy rows beyond the limit are never fetched.

With ≥8 dead-lettered rows in the 2-day window, processing halts completely for the rest of the daemon run with no operator-visible warning beyond `summarised drain` logging 0 rows.

**Fix**: exclude dead-lettered rows from the SQL query:
```rust
// Pass the dead-letter set into fetch_pending:
AND id NOT IN ({dead_letter_ids})
```
Or write a terminal DB state on the 3rd failure (see the `MAX_ROW_ATTEMPTS` comment above) so the existing `task_method` filter already excludes them.

src/coding_agent_session_ingest/summariser/mod.rs
```diff
@@ -387,10 +390,38 @@ pub async fn run_loop(
...
+/// all-days (#177) walks the full historical backlog — observed churning 127
+/// rows back to May. Yesterday+today catches the rollover without the
+/// backlog. Older days stay an explicit operator action:
+/// `meridian coding-agent-summarise --day <YYYY-MM-DD>`.
+async fn drain(
+    pool: &SqlitePool,
+    cfg: &SummariserConfig,
+    attempts: &mut HashMap<i64, u32>,
+) -> bool {
+    let now = Local::now();
```
adityaharishch: **Bug (Medium): `Local::now()` vs UTC `started_at` — re-introduces midnight-stranding for UTC− users**

`drain()` computes window days from local calendar time, but `started_at` is stored as UTC (`+00:00`). For a user in UTC−8 working at 23:00 local (= 07:00 UTC next day), `Local::now().format("%Y-%m-%d")` = `"2026-06-08"`, but a session sealed at that moment has `substr(started_at,1,10)` = `"2026-06-09"` — outside both window day strings.

This is the same class of bug as row 32019 that motivated the PR: the window is wider but the clock used to compute it is still wrong.

**Fix**: `let now = Utc::now();` — `started_at` is UTC, the window comparison should be too.

**Related**: rows sealed while the daemon was offline for 3+ days will have `started_at` UTC dates older than `[yesterday, today]` and are permanently excluded from auto-drain. The docstring below acknowledges this ("Older days stay an explicit operator action"), but there is no fallback log or metric when such rows exist, making silent data loss the default.

src/coding_agent_session_ingest/summariser/mod.rs
```diff
@@ -387,10 +390,38 @@ pub async fn run_loop(
...
-async fn drain(pool: &SqlitePool, cfg: &SummariserConfig) -> bool {
-    let rows = match db::fetch_pending(pool, cfg, cfg.batch_per_tick, None).await {
+/// Per-daemon-lifetime failure ledger: a row that fails this many drain
+/// passes is dead-lettered (skipped with a warn) instead of retried forever.
+/// The churn this prevents was observed live 2026-06-07: rows whose capped
+/// prompt exceeds claude's 200k context AND whose MLX answer is empty cycled
+/// every drain, each burning 2 claude calls + 1 MLX call, indefinitely. The
+/// ledger is in-memory by design: a daemon restart (or `--day` backfill after
+/// fixing the engine) retries cleanly.
+const MAX_ROW_ATTEMPTS: u32 = 3;
```
adityaharishch, line 400: **Design issue (Medium): dead-letter cap is in-memory only — permanently broken rows re-burn LLM calls on every restart**

The PR description says "3 attempts → `subprocess_error`" but no DB state is written when a row is dead-lettered — `task_method` stays `'pending_summariser'` forever. On the next daemon restart the `attempts` map is fresh (`HashMap::new()`), so the same row runs 3× LLM calls again before being re-dead-lettered. This is the exact churn the PR says it prevents.

Consider writing `task_method = 'subprocess_error'` (or a similar terminal state) on the 3rd failure so the row is permanently excluded by the existing filter without needing the in-memory ledger.

src/coding_agent_session_ingest/segment.rs
```diff
@@ -167,8 +172,15 @@ pub fn parse_session_segments(path: &Path, params: &SegmentParams) -> (SessionMe
...
         .unwrap_or_default();
     let jsonl_bytes = std::fs::metadata(path).map(|m| m.len()).unwrap_or(0);
-    let records = iter_normalised(path, &agent);
-    segment_records(records, &session_uuid, &agent, jsonl_bytes, params)
+    let (records, title) = iter_normalised_with_title(path, &agent);
+    let (meta, mut segments) = segment_records(records, &session_uuid, &agent, jsonl_bytes, params);
+    // Stamp the session's own name on every segment (one read produced both).
+    if let Some(t) = title {
+        for seg in segments.iter_mut() {
+            seg.title = Some(t.clone());
```
adityaharishch: **Bug (Low): inline stamp loop missing `TITLE_CAP=200` — JSONL path stores titles untruncated**

`stamp_title()` in `indexer.rs` applies `.chars().take(TITLE_CAP)` (200 chars). This inline equivalent does not. If the `summary`-record fallback in `claude_title()` ever fires (it targets compacted sessions whose `summary` field is multi-KB prose), the full text is stored in `window_titles` untruncated via this path but truncated to 200 chars via the source-adapter path through `register_records` → `stamp_title`.

Consider calling `stamp_title(&mut segments, title.as_deref())` here instead of the inline loop, or extracting the cap into a shared helper.

src/coding_agent_session_ingest/summariser/mod.rs
```diff
@@ -416,11 +451,23 @@ async fn drain(pool: &SqlitePool, cfg: &SummariserConfig) -> bool {
...
-                "summarise failed — leaving pending for retry"
-            );
+            let tries = tries + 1;
+            attempts.insert(row.id, tries);
+            if tries >= MAX_ROW_ATTEMPTS {
+                tracing::warn!(
+                    row_id = outcome.row_id,
+                    error = outcome.error.as_deref().unwrap_or("unknown"),
+                    attempts = tries,
+                    "summarise failed repeatedly — dead-lettering for this daemon run                      (restart or `coding-agent-summarise --day` retries it)"
```
adityaharishch: **Nit: ~22 literal spaces mid-string break log search patterns**

The gap between `"daemon run` and `(restart or"` is ~22 literal spaces. These appear verbatim in structured log fields (OpenObserve, stdout). Any alerting rule or `grep` that matches on `"(restart or"` will not find this message.

Akarsh-Hegde:
Thanks for the thorough 7-angle review! I've addressed all findings in commit 333da23:

### High blockers:

**ai-title records not handled**: ✅ Fixed
- `claude_title()` now checks `ai-title` first (most common at 92/104 sessions), falls back to `custom-title`, then `summary`
- This matches the live corpus analysis showing `ai-title` produces clean AI-generated names

**Dead-lettered rows starve new work**: ✅ Fixed  
- Added `TASK_METHOD_DEAD_LETTER = "subprocess_error"` constant
- On 3rd failure, we now call `db::write_dead_letter()` to persist the state
- `fetch_pending` query naturally excludes them via `WHERE task_method = ?` (only fetches `TASK_METHOD_PENDING` rows)
- Dead-lettering survives daemon restarts; the in-memory ledger is for deduplication within a single run

### Medium issues:

**Local::now() vs UTC**: ✅ Fixed
- Changed to `Utc::now()` for both the daemon drain window and CLI default day
- `started_at` is UTC in the DB, so window comparison must use UTC. Local-time divergence was stranding UTC-offset users as you described

**Dead-letter cap is in-memory only**: ✅ Addressed by persistence fix above

**Inline stamp loop missing TITLE_CAP**: ✅ Fixed
- Applied `const TITLE_CAP: usize = 200` inline in `segment.rs`, matching `indexer.rs::stamp_title()`
- Prevents untruncated summary-fallback text from exceeding cap in this code path

### Low nit:

**Log message spacing**: ✅ Fixed
- Removed 22 literal spaces that broke grep patterns in OpenObserve and log systems

All 52 coding-agent tests pass; cargo clippy clean.

src/coding_agent_session_ingest/summariser/mod.rs
adityaharishch, line 400: **Stale docstring (Low): comment still says "a daemon restart retries cleanly" — no longer true**

`write_dead_letter` now sets `task_method = 'subprocess_error'` permanently. Both `run_loop` (via `drain`) and `cli_summarise` query through `fetch_pending`, which filters `WHERE task_method = 'pending_summariser'`. A dead-lettered row is excluded from both paths after the DB write — a restart does NOT retry it.

Suggested wording:
```
/// The ledger is backed by a DB write on the 3rd failure (task_method =
/// 'subprocess_error'): the row is permanently excluded from auto-drain
/// and from `--day` backfill. To un-dead-letter a row, reset it manually:
///   UPDATE app_sessions SET task_method = 'pending_summariser' WHERE id = <id>;
```

PR #223, opened 2026-06-09
fix(ui): stop the dashboard crash-looping under a mismatched Node after `meridian update`

Akarsh-Hegde:
## Symptom

After `meridian update`, the dashboard goes down and stays down. `meridian doctor` shows the UI `loaded but not running`; `~/.meridian/logs/ui-error.log` is full of:

```
better_sqlite3.node was compiled against NODE_MODULE_VERSION 127 … requires 141
code: 'ERR_DLOPEN_FAILED'
```

The UI launchd job (`exit 78`) crash-loops via `KeepAlive`. `meridian restart` doesn't fix it.

## Root cause

The dashboard's `better-sqlite3` native addon is **ABI-locked to one exact Node version**. CI (Node 24) deliberately swaps in the official prebuilt for **Node 22.22.3 (ABI 127)** and records it in `bin/node-runtime.meta` (`NODE_RUNTIME_VERSION=22.22.3`); `install_ui_standalone()` downloads that runtime, pins it via `MERIDIAN_NODE_BIN`, and reloads the launchd job. **127 = Node 22, 141 = Node 25.** So the addon is correct — the UI was just *running under the wrong Node* (system Node 25). Two code paths combined to cause that:

1. **`install-from-bundle.sh` skipped the reload on update.** When the UI tarball hash was unchanged (`_ui_changed -eq 0`), it skipped `install_ui_standalone` entirely — the only thing that re-pins the Node runtime and reloads the job. So a **stale UI launchd job survived `update`**, still resolving Node via system PATH.
2. **`ui-start.sh` silently fell back to system Node.** With `MERIDIAN_NODE_BIN` unset/stale, it used `/opt/homebrew/bin/node` (Node 25) → ABI mismatch with the Node-22 addon → `ERR_DLOPEN_FAILED` → crash-loop with an opaque error.

The trigger is common: a user upgrades system Node (or `update` leaves a stale job), and the UI lands on a Node whose ABI doesn't match the shipped addon.

## Fix (two parts)

**1. `ui-start.sh` — derive the Node from the contract, fail loud on mismatch** *(the durable guarantee)*
- Reads `NODE_RUNTIME_VERSION` from `node-runtime.meta` and, on a bundle install, runs **only** the cached version-matched runtime under `~/.meridian/node-runtime`. `MERIDIAN_NODE_BIN` is accepted **only if it actually is that version**.
- If no matching runtime exists, it **exits 78 (EX_CONFIG) with a clear remediation message** instead of crash-looping under the wrong Node.
- Because the wrapper is re-read from disk on every (re)start, this **self-heals a stale launchd job**.
- Source/dev installs (no meta) keep the prior prefer-`MERIDIAN_NODE_BIN`-then-system behaviour (there the addon is compiled against local node).

**2. `install-from-bundle.sh` — always reload the UI agent on update** *(the root-cause fix)*
- Stop skipping `install_ui_standalone` when the build is unchanged. The expensive re-extraction is still skipped; `resolve_node_runtime` hits the cache, so this is just a fast `bootout`/`bootstrap` — cheap insurance against a stale, ABI-mismatched job.

## Testing

Exercised all four `ui-start.sh` resolution branches with a fake bundle layout:

| Case | Result |
|---|---|
| meta + cached runtime matches version | ✅ runs cached node, starts server |
| meta + no cache + `MERIDIAN_NODE_BIN` wrong version | ✅ **exits 78, refuses system node** (the bug scenario) |
| meta + `MERIDIAN_NODE_BIN` is the required version | ✅ uses it |
| no meta (source install) | ✅ falls back to `MERIDIAN_NODE_BIN` |

`bash -n` clean on both scripts; full pre-push suite (fmt + clippy + UI tests + cargo test) passes. (The bundle/update path can't be exercised on a source checkout — the in-product version-match + loud-fail is the safety net.)

## Note on the first commit

`b3e38a6 style(fmt)` is **not part of this fix** — it's pre-existing `cargo fmt` drift from the Trello merge (`flow.rs`, `providers/trello.rs`, `pm_worklog/{post,trello}.rs`) that landed on `main` un-formatted, so `cargo fmt --check` (and the commit/push hooks) fail on a clean `main`. Pure formatter output, no logic change; included so this branch can pass the format gate. **`main` currently fails its own fmt gate** — worth a look independently.

## Not covered

This fixes *future* updates. An already-broken machine needs a **one-time** clean reload/reinstall (`bootstrap.sh`) to lay down the corrected `ui-start.sh` and reload the job.

### DIFF
--- a/scripts/install-from-bundle.sh
+++ b/scripts/install-from-bundle.sh
@@ -671,13 +671,18 @@ if command -v cursor >/dev/null 2>&1 || [[ -d "${HOME}/Library/Application Suppo
     fi
 fi
 
-# UI: skip daemon restart when the build didn't change (tarball hash matched).
+# UI: even when the build is unchanged we ALWAYS re-pin the Node runtime and
+# reload the launchd job. The better-sqlite3 addon is ABI-locked to one Node
+# version; skipping this reload is how a stale UI job survives `update` and ends
+# up running a mismatched system Node (NODE_MODULE_VERSION crash-loop). The
+# expensive re-extraction is already skipped above when unchanged, so this is
+# cheap — resolve_node_runtime hits the cache, then bootout/bootstrap reloads.
 if [[ "${_ui_changed}" -eq 0 ]]; then
-    ok "Dashboard unchanged — skipping restart"
+    info "Dashboard build unchanged — re-pinning Node runtime and reloading UI agent…"
 else
     info "Installing the dashboard (UI) launchd agent…"
-    install_ui_standalone
 fi
+install_ui_standalone
 
 # Persist component hashes for the next update's differential check.
 # Write to a temp file and rename atomically so a crash mid-write never leaves
--- a/scripts/ui-start.sh
+++ b/scripts/ui-start.sh
@@ -3,24 +3,66 @@
 #
 # Startup wrapper for the Next.js UI daemon.
 #
-# MERIDIAN_NODE_BIN is set by the launchd plist to the ABI-matched Node 22 runtime
-# that install-from-bundle.sh downloaded and cached under ~/.meridian/node-runtime.
-# That is the exact Node version CI built the better-sqlite3 addon in ui.tar.gz
-# against, so the ABI always matches. Fall back to system node only if unset.
+# The dashboard's better-sqlite3 native addon is ABI-locked to ONE exact Node
+# version — the one CI built it against, recorded in bin/node-runtime.meta as
+# NODE_RUNTIME_VERSION. Running any other Node major triggers a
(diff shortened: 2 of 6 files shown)


Akarsh-Hegde:
The mismatch wasn't "user had wrong Node installed" — it was "our startup script
  silently used a Node it was never supposed to use." That's also why PR #223's fix is shaped the way
  it is: it doesn't ask users to install the right Node; it makes ui-start.sh refuse to run under
  anything except the version recorded in node-runtime.meta, and makes the installer always re-pin on
  update so the fallback path can't be reached.

PR #241, opened 2026-06-10
feat(azure-devops): surface PAT permission errors in the UI

Akarsh-Hegde:
$(cat <<'EOF'
## Summary

- When the Azure DevOps WIQL request returns 401 or 403, the sync now writes a descriptive error message to `pm_sync_state.last_error` instead of silently failing
- The `/api/integrations` route reads `last_error` from the DB and includes it in the response
- The Tasks UI shows an amber "Sync error" badge on the connected tracker row and a warning banner with the human-readable error when expanded

## Changes

- `src/migrations/033_pm_sync_state_error.sql` — adds `last_error TEXT` column to `pm_sync_state`
- `src/intelligence/providers/azure_devops.rs` — 401 → "PAT is invalid or expired", 403 → "PAT lacks required scope — Work Items → Read & write"; clears error on success
- `ui/app/api/integrations/route.ts` — queries `pm_sync_state` for errors and adds `sync_errors` to the response type
- `ui/components/views/TasksView.tsx` — renders amber badge + banner when `sync_errors[provider]` is set

## Test plan

- [ ] Connect Azure DevOps with a PAT that has no Work Items scope → sync → Tasks view shows "Sync error" badge and "PAT lacks required scope" message
- [ ] Connect with an expired/invalid PAT → shows "PAT is invalid or expired" message
- [ ] Connect with a valid PAT → badge shows green "Connected", no banner
- [ ] `cargo test` passes
- [ ] `cargo clippy -- -D warnings` passes

https://claude.ai/code/session_01FXVJmZjFdNeRjyQKNzS2ja
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01FXVJmZjFdNeRjyQKNzS2ja)_

### DIFF
--- a/src/migrations/033_pm_sync_state_error.sql
+++ b/src/migrations/033_pm_sync_state_error.sql
@@ -1,0 +1,4 @@
+-- meridian — normalises screenpipe activity into structured app sessions
+
+-- Surface sync errors (e.g. 401/403 from provider APIs) so the UI can flag them.
+ALTER TABLE pm_sync_state ADD COLUMN last_error TEXT;
--- a/src/intelligence/providers/azure_devops.rs
+++ b/src/intelligence/providers/azure_devops.rs
@@ -125,7 +125,12 @@ async fn run_wiql(client: &reqwest::Client, cfg: &AzureDevOpsConfig) -> Result<V
     let status = resp.status();
     if !status.is_success() {
         let text = resp.text().await.unwrap_or_default();
-        anyhow::bail!("Azure DevOps WIQL returned {status}: {text}");
+        let msg = match status.as_u16() {
+            401 => "permission_error: PAT is invalid or expired — regenerate it in Azure DevOps User settings → Personal access tokens".to_string(),
+            403 => "permission_error: PAT lacks required scope — create a token with Work Items → Read & write scope".to_string(),
+            _ => format!("sync_error: HTTP {status}: {text}"),
+        };
+        anyhow::bail!("{msg}");
     }
     let wiql: WiqlResponse = resp.json().await.context("parsing WIQL response")?;
     Ok(wiql.work_items.iter().map(|w| w.id).collect())
@@ -247,7 +252,14 @@ pub async fn force_refresh(pool: &SqlitePool, cfg: &AzureDevOpsConfig) -> Result
         .context("building HTTP client")?;
 
     // 1. WIQL: all work items assigned to me.
-    let all_ids = run_wiql(&client, cfg).await?;
+    let all_ids = match run_wiql(&client, cfg).await {
+        Ok(ids) => ids,
+        Err(e) => {
+            let msg = e.to_string();
+            stamp_error(pool, &msg).await?;
+            return Err(e);
+        }
+    };
     tracing::debug!(count = all_ids.len(), "azure_devops: WIQL returned IDs");
 
     if all_ids.is_empty() {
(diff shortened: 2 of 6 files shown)


src/migrations/033_pm_sync_state_error.sql
```diff
@@ -0,0 +1,4 @@
+-- meridian — normalises screenpipe activity into structured app sessions
+
```
Akarsh-Hegde, line 2: **Migration number collision with PR #244**

Both this PR and #244 create a `033_*` migration file with different content. sqlx identifies migrations by their numeric prefix and verifies the checksum on every startup. Whichever merges second will crash the daemon on all existing databases with a checksum mismatch — the exact failure mode from the PR #102 incident.

**Fix:** renumber one of them to `034`. Coordinate with whoever merges first.

Akarsh-Hegde (commented):
Automated review — 4 findings

Akarsh-Hegde:
## Code review findings

### 1. `stamp_error?` discards the original sync error if the DB write fails — `azure_devops.rs:259`

If `stamp_error` itself returns an `Err` (e.g. the DB is locked at the moment of a 401), the `?` propagates the DB error and silently drops the original HTTP 401. The caller sees an unrelated storage error instead of "PAT is invalid".

```rust
// Fix: ignore stamp failure, always return the original error
let msg = e.to_string();
let _ = stamp_error(pool, &msg).await;
return Err(e);
```

---

### 2. Batch-fetch errors bypass `stamp_error` — `azure_devops.rs:273`

`stamp_error` is only called when `run_wiql` fails. If `run_wiql` succeeds but the subsequent work-items batch fetch returns 401/403 (e.g. a PAT with WIQL read scope but not Work Items read scope), `force_refresh` returns `Err(e)` without calling `stamp_error`. `last_error` stays `NULL`, the UI shows "Connected" (green), and the user's task list is silently empty with no actionable error.

Wrap `fetch_details` (and the `prune`/`stamp_sync` tail) in the same `stamp_error`-on-error pattern as `run_wiql`.

---

### 3. `stamp_error` ON CONFLICT doesn't update `last_synced_at` — `azure_devops.rs:434`

```sql
ON CONFLICT(provider) DO UPDATE SET last_error = excluded.last_error
```

Once the row exists, repeated errors update `last_error` but leave `last_synced_at` at the first-failure timestamp. Any UI reading `last_synced_at` as "last sync attempted" will show a permanently stale time.

```sql
-- Fix:
ON CONFLICT(provider) DO UPDATE SET
  last_synced_at = excluded.last_synced_at,
  last_error     = excluded.last_error
```

---

### 4. Network errors produce `status: 0` → unhelpful "HTTP 0" message — `discover/route.ts:23`

When `fetch()` throws (DNS failure, host unreachable), the catch returns `{ data: null, status: 0 }`. Callers branch on `status === 401 || 403` which is false, so the user sees `"Azure DevOps returned HTTP 0"` — unactionable.

```ts
// Fix: detect status 0 explicitly
if (status === 0) msg = 'Could not reach Azure DevOps — check the URL and your network'
```

src/migrations/033_pm_sync_state_error.sql
Akarsh-Hegde: Resolved without a renumber: PR #244 has since moved its migration to `034_pm_tasks_tags.sql`, so this PR keeps `033_pm_sync_state_error.sql` and there is no longer a collision. Verified `origin/main` currently tops out at `032`.

Akarsh-Hegde:
All 4 findings addressed in 31eb8a3:

1. **`stamp_error?` masking the original error** — the stamp is now best-effort: a failed write is logged as a warning and the original sync error is always returned.
2. **Batch-fetch errors bypassing `stamp_error`** — `force_refresh` now delegates to an inner `refresh_inner` and stamps *any* failure (WIQL, batch fetch, upsert, prune) at the boundary, so `last_error` can no longer stay NULL while the sync silently fails.
3. **`ON CONFLICT` not refreshing `last_synced_at`** — the upsert now sets both `last_synced_at` and `last_error` on conflict.
4. **`HTTP 0` on network failures** — the discover route maps `status === 0` to "Could not reach Azure DevOps — check the URL and your network" in all three call sites (projects, profile, accounts).

Migration-number collision (inline comment): resolved on the other side — #244 renumbered to `034`, this PR keeps `033`.

Verified: `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (all suites green), and `tsc --noEmit` over `ui/` (no new errors).

PR #251, opened 2026-06-11
feat(observability): configure OpenObserve OTLP export from UI settings

Akarsh-Hegde:
## Summary

- OTLP export to OpenObserve is now configurable from **Settings → Observability** in the dashboard: on/off toggle, endpoint, email, and password fields, plus an **Open OpenObserve** button that opens the instance UI
- New `resolve_otlp_target()` in `src/observability.rs` is the single source of truth for whether export is active — used by both the exporter bootstrap and `meridian doctor`'s OpenObserve health check, so they can never disagree
- Resolution order: `settings.json` (`otlp_enabled`, `otlp_endpoint`, `oo_email`, `oo_password`) → env vars (`MERIDIAN_OTLP_ENDPOINT` / `MERIDIAN_OO_AUTH`) → disabled. The `otlp_enabled` toggle is a hard off-switch that overrides env vars
- Email + password are base64-encoded into the Basic auth header in Rust — users enter plain credentials, no manual encoding
- New `TextInput` UI component (text/password/email variants) — first free-text input in the settings UI
- Includes a `cargo fmt` fix for `src/pm_worklog/azure_devops.rs` (was committed unformatted; blocked the pre-commit hook)

Credentials live in `settings.json`, which is gitignored. Export changes take effect on daemon restart (the tracing subscriber is wired at startup).

## Test plan

- `cargo fmt --check`, `cargo clippy -- -D warnings`: clean
- `cargo test`: 307 passed
- `npm run build` (ui): passes
- `bun test` (ui): 105 passed
- Manual: dev server on :3001 — saved credentials land in `settings.json`, `/api/settings` round-trips the new fields

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -22,6 +22,12 @@ pub struct RuntimeSettings {
     pub llm_budget_pct: f64,
     pub poll_interval_secs: u64,
     pub jira_update_enabled: bool,
+    // OpenObserve OTLP export — all three must be non-empty for export to activate.
+    // Takes precedence over MERIDIAN_OTLP_ENDPOINT / MERIDIAN_OO_AUTH env vars.
+    pub otlp_enabled: bool,
+    pub otlp_endpoint: Option<String>,
+    pub oo_email: Option<String>,
+    pub oo_password: Option<String>,
 }
 
 impl Default for RuntimeSettings {
@@ -37,6 +43,10 @@ impl Default for RuntimeSettings {
             llm_budget_pct: 0.5,
             poll_interval_secs: 60,
             jira_update_enabled: true,
+            otlp_enabled: true,
+            otlp_endpoint: None,
+            oo_email: None,
+            oo_password: None,
         }
     }
 }
--- a/src/health/observability.rs
+++ b/src/health/observability.rs
@@ -2,27 +2,22 @@
 //
 // Observability sink health. If OpenObserve is down, traces/logs silently drop
 // — which blinds the very fault-attribution this layer depends on, so it is
-// worth a check of its own. Export is gated on MERIDIAN_OO_AUTH being set.
+// worth a check of its own. Export is gated on credentials being present in
+// settings.json or MERIDIAN_OO_AUTH (resolved by observability::resolve_otlp_target).
 
 use crate::config::Config;
 use crate::health::Check;
 use std::time::Duration;
 
 pub async fn checks(_cfg: &Config) -> Vec<Check> {
-    let auth_set = std::env::var("MERIDIAN_OO_AUTH")
-        .map(|v| !v.is_empty())
-        .unwrap_or(false);
-    if !auth_set {
+    let Some(target) = crate::observability::resolve_otlp_target() else {
         return vec![Check::info(
             "openobserve",
             "obs",
-            "OTLP export disabled (no MERIDIAN_OO_AUTH) — telemetry not collected",
+            "OTLP export disabled (no credentials in settings or MERIDIAN_OO_AUTH) — telemetry not collected",
         )];
-    }
-
-    let endpoint = std::env::var("MERIDIAN_OTLP_ENDPOINT")
-        .unwrap_or_else(|_| "http://localhost:5080/api/default/v1/traces".to_string());
-    let healthz = derive_healthz(&endpoint);
+    };
+    let healthz = derive_healthz(&target.endpoint);
     let client = match reqwest::Client::builder()
         .timeout(Duration::from_secs(3))
         .build()
--- a/src/health/observability.rs
+++ b/src/health/observability.rs
@@ -10,14 +10,17 @@ use crate::health::Check;
 use std::time::Duration;
 
 pub async fn checks(_cfg: &Config) -> Vec<Check> {
-    let Some(target) = crate::observability::resolve_otlp_target() else {
+    // Use the cheap helpers — the health check never needs the auth credential.
+    if !crate::observability::is_otlp_configured() {
         return vec![Check::info(
             "openobserve",
             "obs",
             "OTLP export disabled (no credentials in settings or MERIDIAN_OO_AUTH) — telemetry not collected",
         )];
-    };
-    let healthz = derive_healthz(&target.endpoint);
+    }
+    let endpoint = crate::observability::resolve_otlp_endpoint()
+        .unwrap_or_else(|| "http://localhost:5080/api/default/v1/traces".to_string());
+    let healthz = derive_healthz(&endpoint);
     let client = match reqwest::Client::builder()
         .timeout(Duration::from_secs(3))
         .build()
--- a/src/observability.rs
+++ b/src/observability.rs
@@ -177,19 +177,52 @@ pub fn init(service_name: &str) -> Result<ObservabilityGuard> {
     })
 }
 
-/// Builds the OTel tracer and logger providers when `MERIDIAN_OO_AUTH` is set.
-/// Returns the providers (not layers) so the caller can construct layers at the
-/// correct generic subscriber type without boxing.
-fn try_build_otel_providers(service_name: &str) -> Result<Option<(Tracer, LoggerProvider)>> {
-    let Ok(auth) = std::env::var("MERIDIAN_OO_AUTH") else {
-        return Ok(None);
-    };
-    if auth.is_empty() {
-        return Ok(None);
+/// Resolved OTLP export target: trace endpoint + Basic-auth credential.
+/// `None` means export is disabled (toggle off, or no credentials anywhere).
+pub struct OtlpTarget {
+    pub endpoint: String,
+    pub auth: String,
+}
+
+/// Resolve the OTLP export config. settings.json values take precedence over
+/// the MERIDIAN_OTLP_ENDPOINT / MERIDIAN_OO_AUTH env vars; the settings
+/// `otlp_enabled` toggle overrides both. Shared by the exporter bootstrap and
+/// the health check so they never disagree on whether export is active.
+pub fn resolve_otlp_target() -> Option<OtlpTarget> {
+    use base64::{engine::general_purpose::STANDARD, Engine as _};
+
+    let settings = crate::config::load_runtime_settings();
+
+    if !settings.otlp_enabled {
+        return None;
     }
 
-    let trace_endpoint = std::env::var("MERIDIAN_OTLP_ENDPOINT")
-        .unwrap_or_else(|_| DEFAULT_OTLP_ENDPOINT.to_string());
+    // Auth: settings email+password → env var MERIDIAN_OO_AUTH → disabled
+    let auth = match (&settings.oo_email, &settings.oo_password) {
+        (Some(email), Some(pass)) if !email.is_empty() && !pass.is_empty() => {
+            STANDARD.encode(format!("{email}:{pass}"))
+        }
+        _ => match std::env::var("MERIDIAN_OO_AUTH") {
+            Ok(v) if !v.is_empty() => v,
+            _ => return None,
+        },
+    };
+
+    let endpoint = settings
+        .otlp_endpoint
+        .filter(|s| !s.is_empty())
+        .or_else(|| std::env::var("MERIDIAN_OTLP_ENDPOINT").ok())
+        .unwrap_or_else(|| DEFAULT_OTLP_ENDPOINT.to_string());
+
+    Some(OtlpTarget { endpoint, auth })
+}
+
+/// Builds the OTel tracer and logger providers when OTLP export is configured.
+fn try_build_otel_providers(service_name: &str) -> Result<Option<(Tracer, LoggerProvider)>> {
+    let Some(target) = resolve_otlp_target() else {
+        return Ok(None);
+    };
+    let (trace_endpoint, auth) = (target.endpoint, target.auth);
 
     // Derive the log endpoint from the trace endpoint by swapping the path suffix.
     let log_endpoint = trace_endpoint.replace("/v1/traces", "/v1/logs");
--- a/src/observability.rs
+++ b/src/observability.rs
@@ -39,8 +39,6 @@ use tracing_appender::non_blocking::WorkerGuard;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
 
 const DEFAULT_OTLP_ENDPOINT: &str = "http://localhost:5080/api/default/v1/traces";
-const DEFAULT_ENV_FILTER: &str =
-    "meridian=info,meridian::etl=debug,meridian::intelligence=debug,sqlx=warn";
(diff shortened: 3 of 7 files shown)


adityaharishch (commented):
**Security / Credential Storage — HIGH**

`ui/lib/settings.ts` stores `oo_email` and `oo_password` as plaintext strings in `settings.json`, and the `/api/settings` GET route returns the entire `RuntimeSettings` object including those plaintext credentials.

Problems:
1. Any JavaScript running in the dashboard page (XSS, browser extensions, devtools network tab) can read the raw password via `GET /api/settings`.
2. The password will appear in Next.js JSON responses visible in browser devtools — it is not treated as a write-only field.
3. The `writeSettings` function in `settings.ts` serialises the full settings object, so the plaintext password lands on disk in `settings.json`. While `settings.json` is gitignored, it is world-readable by any process running as the same OS user.

**Recommended fix:**
- The GET handler should redact the password field before returning: replace `oo_password` with `"" ` (empty) or `"••••••••"` (sentinel) if it is non-empty, similar to how most settings UIs handle secret fields.
- The PUT handler should merge only non-empty values for `oo_password` (i.e., if the UI sends an empty string, keep the stored value — this prevents a save-with-blanked-password from wiping credentials).
- Alternatively, persist credentials through the OS keychain (e.g. `keytar`) rather than `settings.json`, but that is a larger change.

Affected files: `ui/app/api/settings/route.ts`, `ui/lib/settings.ts`

adityaharishch (commented):
**Input Validation — OTLP Endpoint URL — MEDIUM**

`resolve_otlp_target()` in `src/observability.rs` accepts `settings.otlp_endpoint` without any validation before using it to construct the HTTP exporter endpoint and, critically, to derive the log endpoint via string replacement:

```rust
let log_endpoint = trace_endpoint.replace("/v1/traces", "/v1/logs");
```

If a user enters a malformed string (e.g. `not-a-url`, `javascript:alert(1)`, or a URL pointing to an internal network resource), the OTLP exporters will attempt to connect to it. The current code has no:
- URL scheme check (must be `http://` or `https://`)
- Hostname validation
- Port range check

The UI side (`SettingsView.tsx`) also performs no validation before calling `patch({ otlp_endpoint: v })`.

**Recommended fix** (Rust side): in `resolve_otlp_target`, validate that the endpoint parses as a URL and has an `http` or `https` scheme before constructing `OtlpTarget`. The `url` crate is already in the dependency tree (used elsewhere in the project):

```rust
let endpoint = /* resolved string */;
let parsed = url::Url::parse(&endpoint).map_err(|_| /* warn and return None */)?;
if !matches!(parsed.scheme(), "http" | "https") { return None; }
```

**Recommended fix** (TypeScript side): validate the URL in `SettingsView` or in the PUT route before accepting it, using `new URL(value)` and checking `.protocol`.

adityaharishch (commented):
**Architecture — Daemon restart required, but UI does not make this clear enough — MEDIUM**

The PR description and the `FieldRow` description text both say "takes effect after a daemon restart." However, this is a breaking departure from how every other setting in `RuntimeSettings` works — the daemon calls `load_runtime_settings()` on every poll tick, so most settings (poll interval, classification toggle, etc.) are hot-reloaded with no restart required.

The OTLP settings cannot be hot-reloaded because `observability::init()` wires the tracing subscriber once at process startup (`tracing_subscriber::registry().init()`), and the global subscriber can only be set once. Changing the credentials after startup has no effect until the daemon restarts.

This is architecturally inconsistent and likely to confuse users, who will expect the toggle to take effect immediately as with other settings.

Two concerns:

1. The `RuntimeSettings` struct (and `settings.json`) conflates two kinds of settings: hot-reloadable ones and startup-only ones. There is no documentation or enforcement of which is which.

2. The UI `FieldRow` description mentions the restart requirement in small sub-label text that is easy to overlook. There is no visible warning state when the user saves these fields after the daemon is already running.

**Recommended fix (minimal):** Add a `tracing::warn!` in `resolve_otlp_target` (or at the call site in `init`) noting that the OTLP config is read once at startup and cannot be reloaded. This at least surfaces it in daemon logs when someone wonders why the toggle isn't taking effect.

**Recommended fix (fuller):** Track a `"restart_required"` flag in the settings API response or a separate endpoint, and surface a dismissable banner in the UI when OTLP settings have changed since the daemon last read them. Other daemons (e.g. Prometheus) surface this pattern as a `/-/reload` endpoint or a `SIGHUP` handler — worth considering as a follow-up.

adityaharishch (commented):
**File Header — `ui/components/ui/TextInput.tsx` — LOW**

The new `TextInput.tsx` file uses a non-standard header on line 1:

```
// meridian — normalises screenpipe activity into structured app sessions
```

Every `.tsx` and `.ts` file in the repository must start with exactly:

```
//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```

This is enforced by the `pre-commit` hook (per `CLAUDE.md`). The file will fail hook validation on the next commit that touches it, or if another contributor runs `setup-hooks.sh` and the hook does a header scan.

Please replace the header on line 1 of `ui/components/ui/TextInput.tsx` with the required string.

adityaharishch (commented):
**Input Validation — Auth format not validated — MEDIUM**

`resolve_otlp_target()` constructs the Basic auth header by base64-encoding `email:password` directly from `settings.oo_email` and `settings.oo_password`, but neither field is validated before encoding:

```rust
STANDARD.encode(format!("{email}:{pass}"))
```

If `email` contains a colon character (`:`) — which is technically invalid for an email but not prevented — the resulting `user:password` split at the remote end will be malformed, with no error reported to the user. More importantly, if either field contains newline characters, the `Authorization: Basic <value>` HTTP header itself will be split, which is a classic HTTP header injection vector.

The UI `TextInput` component does not strip or reject newlines from `type="email"` or `type="password"` inputs. While browsers typically prevent newlines in visible text inputs, the value can still be set programmatically.

**Recommended fix (Rust side):** Before encoding, reject credentials that contain newline (`\n`, `\r`) or colon in the email field, logging a warning and returning `None`:

```rust
if email.contains('\n') || email.contains('\r') || pass.contains('\n') || pass.contains('\r') {
    tracing::warn!("OTLP credentials contain invalid characters — OTLP export disabled");
    return None;
}
```

**Recommended fix (API route side):** The `PUT /api/settings` handler currently does a blind `{ ...current, ...body }` merge with no field-level validation. At minimum, validate that `oo_email` is a plausible email (contains `@`, no newlines) and that `oo_password` contains no newlines before writing to `settings.json`.

adityaharishch (commented):
**Type Safety — `ui/lib/settings.ts` — LOW**

The new fields added to the `RuntimeSettings` interface declare `otlp_endpoint`, `oo_email`, and `oo_password` as `string` (non-optional):

```ts
otlp_endpoint: string
oo_email: string
oo_password: string
```

The Rust-side `RuntimeSettings` struct declares these as `Option<String>`:

```rust
pub otlp_endpoint: Option<String>,
pub oo_email: Option<String>,
pub oo_password: Option<String>,
```

When the Rust daemon serialises a `RuntimeSettings` with `None` values (the default), `serde_json` emits `null` for those fields. The TypeScript interface claims they are `string`, not `string | null`. The `SETTINGS_DEFAULTS` sets them to `''` (empty string), so `readSettings()` merges the JSON `null` over the default empty string, and TypeScript never complains — but any code that reads these fields and does string operations (`.trim()`, `.length`, etc.) on a `null` value at runtime will throw an unhandled TypeError.

The PR description says Rust uses `Option<String>` and the UI uses `string` with an empty-string default, implying they intentionally diverge. This is fine as a design choice only if the `readSettings()` merge correctly coerces `null` to `''`. Currently it does (`{ ...SETTINGS_DEFAULTS, ...JSON.parse(raw) }` spreads the `null` values over the defaults rather than keeping the defaults), so `null` does land in the returned object.

**Recommended fix:** Either:
- Declare the fields as `string | null` in the TypeScript interface and handle `null` at every use site, or
- Post-process the `readSettings()` result to coerce `null` → `''` explicitly, keeping the `string` type in the interface.

The current code is a latent runtime bug waiting for the first consumer that doesn't check for null.

adityaharishch (commented):
**Security — `resolve_otlp_target()` called on every health check tick — LOW**

`resolve_otlp_target()` calls `crate::config::load_runtime_settings()` which does a synchronous `std::fs::read_to_string` + `serde_json::from_str` on every call. It is now called from `src/health/observability.rs` which runs on every `meridian doctor` / health-stream poll. On a busy dashboard with the health stream open this means `settings.json` is read and deserialized on every health check interval (default appears to be frequent based on the SSE stream route).

More importantly: the auth credential built in `resolve_otlp_target()` — including the decoded plaintext password components — is assembled and then discarded on every health check call (only the endpoint is used, not the auth, in the health check path). The password string is briefly alive in process memory on every health check even when the health check itself does not use it.

**Recommended fix:** Split `resolve_otlp_target()` into two cheaper helpers:
- `is_otlp_configured() -> bool` — reads only the `otlp_enabled` flag + presence of credentials, used by the health check
- `resolve_otlp_target() -> Option<OtlpTarget>` — assembles the full auth string, called only at startup

The health check only needs the endpoint for the HTTP ping and should not trigger credential assembly.

adityaharishch (commented):
**Architecture — `otlp_enabled` default is `true`, creating a surprising first-run behaviour — LOW**

Both `RuntimeSettings::default()` in Rust and `SETTINGS_DEFAULTS` in TypeScript set `otlp_enabled: true`. On a fresh install with no `settings.json`, `resolve_otlp_target()` returns `None` only because `oo_email`/`oo_password` are also `None` (Rust default) or empty (TS default). So the toggle being `true` with no credentials is effectively the same as `false` — no export happens.

However, this is fragile: if a user already has `MERIDIAN_OO_AUTH` set as an env var and has never opened the Settings UI, the OTLP export was already working via the env var path. After this PR, the first time `resolve_otlp_target()` runs with a `settings.json` that has `otlp_enabled: true` (default) and empty `oo_email`/`oo_password`, the resolution order hits the env var fallback and continues to work — correct behaviour.

But the scenario where a user has the env var AND opens Settings and saves without filling in the credentials will write `otlp_enabled: true, oo_email: "", oo_password: ""` to `settings.json`. `resolve_otlp_target()` then falls through to the env var correctly (`oo_email` is empty → goes to env var branch). This is fine.

The subtle issue is the inverse: a user who wants to DISABLE OTLP sets the toggle to off and saves. This writes `otlp_enabled: false` to `settings.json`, correctly suppressing export. But if they later unset `MERIDIAN_OO_AUTH` from the env, they might expect re-enabling the toggle to start export — but it won't until they also fill in the credentials in the UI. The UX doesn't make this dependency visible.

This is low severity but worth a UX note: the UI should show the email/password fields as required (or at minimum visually indicate they are needed) when the toggle is on and `MERIDIAN_OO_AUTH` is not set.

Akarsh-Hegde:
Thanks @adityaharishch — all addressed in bb24e04. Response per comment:

**#1 — Security / Credential Storage (HIGH) ✓ fixed**
`GET /api/settings` now returns `oo_password: '••••••••'` (sentinel) if a password is stored, `''` if not — the real value never reaches the browser or network tab. `PUT` treats both `''` and the sentinel as \"keep existing\" so a save without touching the field can't wipe credentials.

**#2 — OTLP Endpoint URL Validation (MEDIUM) ✓ fixed**
`resolve_otlp_target()` now rejects any endpoint that doesn't start with `http://` or `https://` and logs a `warn!`. The `PUT /api/settings` handler also validates this on the TypeScript side before writing to disk, returning 400 with a clear message.

**#3 — Daemon restart UX (MEDIUM) ✓ addressed in PR #256**
The minimal fix (a `tracing::warn!` at startup) is included. The fuller fix — SIGHUP handler + **Apply** button in the UI that shows `Saving → Reloading → Active` with 500ms polling — is implemented in PR #256 which builds on this branch. Comment #3 is exactly what motivated that PR.

**#4 — File header on TextInput.tsx (LOW) — not changing**
CLAUDE.md defines the required header as `// meridian — normalises screenpipe activity into structured app sessions` and the file is compliant. The alternative header in the review comment doesn't match the project standard.

**#5 — Newline / colon validation in credentials (MEDIUM) ✓ fixed**
`resolve_otlp_target()` now checks for `\n`/`\r` in both fields and `:` in the email field before base64-encoding, returning `None` with a `warn!` on violation. The `PUT` route also rejects credentials containing newlines with a 400.

**#6 — Option\<String\> → null lands as string in TS (LOW) ✓ fixed**
`readSettings()` now explicitly coerces `null → ''` for `otlp_endpoint`, `oo_email`, and `oo_password` after the JSON spread, keeping the `string` type in the interface correct at runtime.

**#7 — resolve_otlp_target() called on every health check (LOW) ✓ fixed**
Split into three helpers:
- `is_otlp_configured()` — flag check only, no credential assembly  
- `resolve_otlp_endpoint()` — endpoint URL, no credential assembly  
- `resolve_otlp_target()` — full auth target, called only at daemon startup  

`src/health/observability.rs` now uses the first two, so credentials are never assembled on the health-check hot path.

**#8 — otlp_enabled default true with no credentials (LOW) — left as-is for now**
The current behaviour (toggle on + no credentials = no export, same as toggle off) is safe. UX improvement (showing the credential fields as required when toggle is on) is worth a follow-up but out of scope for this PR.

adityaharishch:
**File header on TextInput.tsx — follow-up**

To clarify comment #4: `CLAUDE.md` at the repo root defines the required header as:

```
//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```

The `// meridian — normalises screenpipe activity…` variant was the old header, retired in commit #250 (`chore: update file headers with boosting developer productivity tagline`). If `TextInput.tsx` still carries the old string, it should be updated to the current one.

adityaharishch:
**BLOCKED — resolve before merge**

The following item is still open:

**File header on `ui/components/ui/TextInput.tsx` (LOW — but will break pre-commit hook)**

Line 1 currently reads:
```
// meridian — normalises screenpipe activity into structured app sessions
```

The required header per `CLAUDE.md` (confirmed by commit #250 which batch-updated all files) is:
```
//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```

The old `// meridian — normalises…` header was retired in that update. Please update line 1 of `TextInput.tsx` to match.

Akarsh-Hegde:
Fixed — pushed as 1bda342.

All 27 files in this branch's diff (including `TextInput.tsx`) now carry the new header. I had incorrectly pushed back on this earlier based on a stale copy of `CLAUDE.md` from before PR #250 merged. The new header is confirmed correct, and I've updated my working context to reflect it.

PR #253 (cursor-pointer, 1 file — `cursor-pointer.test.ts`) and PR #256 (daemon-hot-reload, 28 files) have also been updated for the same reason.

PR #252, opened 2026-06-11
feat(ui): show task title and PM-tool link on worklog cards

Akarsh-Hegde:
## Summary
- Worklog review cards previously showed only the ticket key (e.g. `KAN-157`) with no indication of what the task actually is.
- The worklogs API (`GET /api/worklogs`) now LEFT JOINs `pm_tasks` on `task_key`, exposing `task_title` and `task_url` on each `WorklogItem`.
- The card meta row renders the title next to the key (truncated for long titles). When the task has a URL, key + title link out to the ticket in the tracker (new tab, ↗ indicator, hover underline). Worklogs whose ticket isn't in `pm_tasks` degrade gracefully to plain key with no link.

## Test plan
- [x] `tsc --noEmit` — no new errors (only pre-existing `bun:test` resolution issues in test files)
- [x] Verified the join against a live `meridian.db`: drafts return full title + `https://…atlassian.net/browse/KAN-…` URL
- [ ] Visual check on the Worklogs page: title visible, link opens the ticket in Jira

### DIFF
--- a/ui/app/api/worklogs/route.ts
+++ b/ui/app/api/worklogs/route.ts
@@ -19,6 +19,8 @@ export interface WorklogBullet {
 export interface WorklogItem {
   id: number
   task_key: string
+  task_title: string | null
+  task_url: string | null
   provider: string
   window_start: string
   state: string
@@ -44,6 +46,8 @@ export interface WorklogsResponse {
 interface RawRow {
   id: number
   task_key: string
+  task_title: string | null
+  task_url: string | null
   provider: string
   window_start: string
   state: string
@@ -82,13 +86,15 @@ export async function GET(req: Request) {
   try {
     const db = getDb()
     const rows = db.prepare(`
-      SELECT id, task_key, COALESCE(provider, 'jira') AS provider, window_start,
-             state, confidence, coverage,
-             time_spent_seconds, payload_json, posted_worklog_id,
-             last_post_error, edited_at
-      FROM pm_worklogs
-      WHERE day_utc = ?
-      ORDER BY window_start, task_key
+      SELECT w.id, w.task_key, t.title AS task_title, t.url AS task_url,
+             COALESCE(w.provider, 'jira') AS provider, w.window_start,
+             w.state, w.confidence, w.coverage,
+             w.time_spent_seconds, w.payload_json, w.posted_worklog_id,
+             w.last_post_error, w.edited_at
+      FROM pm_worklogs w
+      LEFT JOIN pm_tasks t ON t.task_key = w.task_key
+      WHERE w.day_utc = ?
+      ORDER BY w.window_start, w.task_key
     `).all(day) as RawRow[]
 
     const counts: Record<string, number> = {}
@@ -109,6 +115,8 @@ export async function GET(req: Request) {
       return {
         id: r.id,
         task_key: r.task_key,
+        task_title: r.task_title,
+        task_url: r.task_url,
         provider: r.provider ?? 'jira',
         window_start: r.window_start,
         state: r.state,
--- a/ui/app/api/worklogs/route.ts
+++ b/ui/app/api/worklogs/route.ts
@@ -116,7 +116,7 @@ export async function GET(req: Request) {
         id: r.id,
         task_key: r.task_key,
         task_title: r.task_title,
-        task_url: r.task_url,
+        task_url: r.task_url?.startsWith('https://') ? r.task_url : null,
         provider: r.provider ?? 'jira',
         window_start: r.window_start,
         state: r.state,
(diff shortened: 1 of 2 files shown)


adityaharishch (commented):
Code review for feat(ui): show task title and PM-tool link on worklog cards

Overall: the approach is sound — LEFT JOIN on pm_tasks is correct, the null-guard on task_title is consistent, and the fallback no-link path is present. A few issues need addressing before merge.

---

**1. Missing rel="noopener" on external link (security)**

ui/components/views/WorklogsView.tsx, the anchor wrapping TaskKey.

The link uses rel="noreferrer" but is missing rel="noopener". While noreferrer implies noopener in modern browsers, the combination rel="noopener noreferrer" is the explicit, cross-browser safe form and is the standard used everywhere else in this codebase. The current value also differs from what the PR description specified. Change to:

  rel="noopener noreferrer"

---

**2. task_url is not validated before use as href (security/correctness)**

ui/components/views/WorklogsView.tsx

The column url in pm_tasks is defined with DEFAULT '' — so the LEFT JOIN will produce an empty string ('') rather than NULL when a row exists but has no URL populated (older Jira rows, for example, may have the legacy empty-string default). The null guard `w.task_url ?` will evaluate to true for an empty string? No — empty string is falsy in JS, so that path is actually fine. However the deeper issue is that task_url is typed as string | null in WorklogItem, but the DB can also return an empty string. The route.ts mapping passes task_url: r.task_url through without coercing '' to null. Downstream the component correctly short-circuits on falsy, but the type WorklogItem declares string | null, hiding the empty-string case from type-system consumers who might do w.task_url.startsWith('https'). Recommend normalising in the route mapping:

  task_url: r.task_url || null,

This makes the type contract honest and protects future callers.

---

**3. No URL scheme validation before rendering as href**

ui/components/views/WorklogsView.tsx

task_url comes from pm_tasks.url, which is populated by provider adapters (Jira, Linear, Azure DevOps, Trello, GitHub). Those adapters are expected to store HTTPS URLs, but there is no enforcement at the DB layer (no CHECK constraint) and no validation in this route. A javascript: or data: scheme stored via a compromised or misconfigured provider would be rendered directly as an href. Add a guard in the route's mapping step:

  task_url: r.task_url?.startsWith('https://') ? r.task_url : null,

This is a one-liner and eliminates the entire class of href-injection risk without touching the component.

---

**4. WorklogItem interface is not updated in the exported type (API contract drift)**

ui/app/api/worklogs/route.ts

task_title and task_url are added to RawRow and to the object literal returned inside rows.map(), but they are NOT added to the exported WorklogItem interface (lines 19-36 of the post-diff file). The object literal is returned as WorklogItem[], so TypeScript will silently drop those fields at the type level — callers importing WorklogItem from this module will not see the new fields. Add to WorklogItem:

  task_title: string | null
  task_url: string | null

---

**5. Meta row flex layout may clip the status badge on narrow viewports**

ui/components/views/WorklogsView.tsx

The meta row now has min-w-0 on the outer div and introduces a potentially wide task_title span with truncate. The status badge at the far right (ml-auto) sits inside the same flex row. If task_title is long, the badge can be pushed partially off-screen on a 1280-wide display. The original layout did not have this risk because task_key alone is always short. Consider wrapping task_title + link in a flex-1 min-w-0 container so it absorbs available space without affecting sibling elements, or moving the badge to a second line for this card size. This is a visual regression risk, not a blocking correctness issue, but worth a quick viewport check before merge.

---

**Summary table**

| # | File | Severity | Category |
|---|------|----------|----------|
| 1 | WorklogsView.tsx | Low | Link safety — rel missing noopener |
| 2 | route.ts | Low | Type honesty — empty string not coerced to null |
| 3 | route.ts | Medium | Security — no URL scheme validation before href |
| 4 | route.ts | High | API contract — WorklogItem interface not updated |
| 5 | WorklogsView.tsx | Low | Visual — meta row layout on narrow viewports |

File headers on both changed files are compliant. The SQL JOIN logic is correct (LEFT JOIN, correct key, correct column aliases). The conditional rendering of task_title (only when truthy) is correct. The shrink-0 on the provider label is a good defensive addition.

Issue #4 (WorklogItem interface gap) is the only blocking correctness issue — TypeScript consumers of the exported interface will be missing the new fields. Issues #2 and #3 should also be fixed before merge for type safety and security hygiene.

Akarsh-Hegde:
All four review comments addressed — pushed as 751dd72.

**Fixed:**
1. `rel="noopener noreferrer"` — added `noopener` to the anchor. (`noreferrer` implies it in modern browsers but the explicit pair is the cross-browser safe form and matches every other external link in the codebase.)
2+3. `task_url` normalised at the route layer: `r.task_url?.startsWith('https://') ? r.task_url : null` — this handles both the empty-string-to-null coercion and the scheme validation in one expression. Any non-https URL (including empty string, `javascript:`, `data:`, etc.) is nulled out before it reaches the component.

**Already done (pre-push):**
4. `WorklogItem` interface — `task_title: string | null` and `task_url: string | null` were already added in the original commit; the reviewer may have seen an intermediate diff.

**Not a blocker noted:**
5. Visual layout on narrow viewports — the status badge sits in an `ml-auto` flex child and `task_title` is constrained by `truncate`, so in practice the badge is not displaced. Happy to tighten further if you see a real regression.

PR #253, opened 2026-06-11
fix(ui): add cursor:pointer to all interactive buttons across the product

Akarsh-Hegde:
## Summary
- Browsers default `<button>` to `cursor:default` — hovering buttons showed an arrow instead of a hand, making them feel unclickable.
- Added a global CSS rule in `globals.css` so every enabled button shows `pointer`, and disabled ones show `not-allowed`.
- Fixed 5 inline `cursor: 'default'` styles in `Switch`, `Select` (trigger + item), and `NumberStepper` (decrement + increment) that would have overridden the global rule.
- Added 9 test cases in `ui/__tests__/cursor-pointer.test.ts` to prevent regressions.

## Files changed
| File | Change |
|---|---|
| `ui/app/globals.css` | Global `button:not(:disabled)` + `button:disabled` cursor rules |
| `ui/components/ui/Switch.tsx` | `cursor: 'default'` → `'pointer'` |
| `ui/components/ui/Select.tsx` | 2× `cursor: 'default'` → `'pointer'` (trigger + item) |
| `ui/components/ui/NumberStepper.tsx` | 2× conditional `'default'` → `'pointer'` |
| `ui/__tests__/cursor-pointer.test.ts` | 9 new tests (CSS rule presence + no override regressions) |

## Test plan
- [x] `bun test __tests__/cursor-pointer.test.ts` — 9/9 pass
- [x] Pre-push hook (cargo fmt + clippy + cargo test + UI build) — passed
- [ ] Visual: hover any button in the UI — cursor should show a hand pointer; disabled buttons show a not-allowed cursor


adityaharishch (commented):
**SettingsView.tsx — `SaveButton` still uses `cursor: 'default'`**

The `SaveButton` component (line 62 of `ui/components/views/SettingsView.tsx`) still has:

```tsx
cursor: 'default',
```

This button is a fully interactive `<button>` with an `onClick` handler and no disabled state, so it should be `cursor: 'pointer'`. It was missed in this PR's sweep.

The global CSS rule added in `globals.css` (`button:not(:disabled) { cursor: pointer; }`) would normally cover this — but the inline style overrides it because inline styles have higher specificity than stylesheet rules. The fix is to remove the inline `cursor` declaration entirely (letting the global rule win), or change it to `'pointer'`.

adityaharishch (commented):
**TasksView.tsx — two buttons use `cursor: 'default'` instead of `cursor: 'not-allowed'` for disabled states**

Two buttons in `ui/components/views/TasksView.tsx` guard their disabled-like state with `cursor: 'default'` rather than `cursor: 'not-allowed'`:

1. **Sync button** (line 126): `cursor: syncing ? 'default' : 'pointer'` — when syncing is active the button is functionally disabled; this should be `'not-allowed'`.

2. **Disconnect button** (line 528): `cursor: disconnecting === t.id ? 'default' : 'pointer'` — same issue: in-progress state should be `'not-allowed'`.

3. **Look up orgs button** (line 645): `cursor: (!pat.trim() || loading === 'orgs') ? 'default' : 'pointer'` — this button has `disabled={!pat.trim() || loading === 'orgs'}` on the element, so the global `button:disabled { cursor: not-allowed; }` rule would already handle it correctly — but the inline style again overrides it with `'default'`. Change the inline value to `'not-allowed'` or drop the inline declaration entirely.

Consistent UX convention: interactive unavailable = `not-allowed`, not `default`.

adityaharishch (commented):
**Test file missing required file header**

The new test file `ui/__tests__/cursor-pointer.test.ts` (added by this PR) does not start with the required file header. All `.ts` files in this repo must begin with:

```
//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```

The file currently starts with:
```
// meridian — normalises screenpipe activity into structured app sessions
```

Every other test file in `ui/__tests__/` has the correct header (e.g. `intervals.test.ts`, `format.test.ts`). Please update line 1.

adityaharishch (commented):
**NumberStepper.tsx — inline `cursor` style overrides the new global rule (specificity issue)**

The PR correctly changes `'default'` to `'pointer'` in `NumberStepper.tsx` — good. However, both buttons already have `disabled={atMin}` / `disabled={atMax}` set on the element. The new global CSS rules in `globals.css` would handle the enabled/disabled cursor automatically:

```css
button:not(:disabled) { cursor: pointer; }
button:disabled       { cursor: not-allowed; }
```

Because these buttons carry `disabled` natively, the conditional inline cursor style is now redundant. The entire inline `cursor` expression can be dropped from both buttons, letting the global rule manage it — which is exactly the point of the global rule. Keeping duplicated inline logic defeats the purpose of the systematic fix.

adityaharishch (commented):
**Broader gap: many interactive buttons still carry inline `cursor` overrides that will shadow the new global rule**

The global CSS rule added in `globals.css` is the right systematic approach. However, because inline `style` props have higher CSS specificity than stylesheet rules, any `<button>` that also sets `cursor` inline will ignore the global rule entirely.

Spots not touched by this PR that still carry inline cursor overrides:

- `TasksView.tsx` line 528 — Disconnect button: `cursor: disconnecting === t.id ? 'default' : 'pointer'`
- `TasksView.tsx` line 645 — Look up orgs button: `cursor: (!pat.trim() || loading === 'orgs') ? 'default' : 'pointer'`
- `TasksView.tsx` line 140 — Integrations button: `cursor: 'pointer'` (harmless, but redundant)
- `TasksView.tsx` line 153 — Close integrations button: `cursor: 'pointer'` (redundant)
- `SettingsView.tsx` line 62 — SaveButton: `cursor: 'default'` (wrong value, overrides global)

Recommendation: the follow-up clean-up pass should strip inline `cursor` declarations from all `<button>` elements and rely solely on the new global rule. That makes the fix genuinely systematic rather than a partial sweep.

Akarsh-Hegde:
All cursor issues fixed — pushed as fbe4858.

**Fixed:**
1. `SaveButton` (`SettingsView.tsx`): removed the inline `cursor: 'default'`. The button has no `disabled` prop so the global `button:not(:disabled) { cursor: pointer }` rule now applies correctly.
2. `TasksView.tsx` sync button: `'default'` → `'not-allowed'` (button has `disabled={syncing}`).
3. `TasksView.tsx` disconnect button: `'default'` → `'not-allowed'` (button has `disabled={disconnecting === t.id}`).
4. `TasksView.tsx` look-up-orgs button: `'default'` → `'not-allowed'` (button has `disabled={!pat.trim() || loading === 'orgs'}`).
5. `NumberStepper.tsx`: removed both inline `cursor` expressions entirely. The `disabled={atMin}` / `disabled={atMax}` attributes are already present, so the global rules manage it — which is exactly the point of the systematic fix.

**One comment I'm pushing back on:**
The test file header in `ui/__tests__/cursor-pointer.test.ts` — the file *does* start with `// meridian — normalises screenpipe activity into structured app sessions`, which is the header required by `CLAUDE.md`. The alternative string the reviewer cited (`//ambient dev tool that watches what you do…`) is not the standard for this repository. Keeping current header.

adityaharishch:
**File header — follow-up**

The dispute on the test file header needs clarification. `CLAUDE.md` at the repo root (the authoritative source) defines the required header as:

```
//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```

This is confirmed by recent commits (e.g. #250: `chore: update file headers with boosting developer productivity tagline`), which batch-updated all files to this exact string. The `// meridian — normalises screenpipe activity…` variant is an **old** header that was retired in that update.

Please update `ui/__tests__/cursor-pointer.test.ts` line 1 to match the current standard.

adityaharishch:
**BLOCKED — resolve before merge**

The following item is still open:

**File header on `ui/__tests__/cursor-pointer.test.ts` (will break pre-commit hook)**

Line 1 currently reads:
```
// meridian — normalises screenpipe activity into structured app sessions
```

The required header per `CLAUDE.md` (confirmed by commit #250 which batch-updated all files) is:
```
//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```

The `// meridian — normalises…` variant is the old header, retired in that commit. Every other test file in `ui/__tests__/` uses the current string. Please update line 1 of the test file to match.

Akarsh-Hegde:
All review comments addressed — latest push is 7ecd19d.

**Fixed in this PR:**
1. `SaveButton` (`SettingsView.tsx`): removed `cursor: 'default'` — global `button:not(:disabled) { cursor: pointer }` now applies.
2. Sync button (line 126): `'default'` → `'not-allowed'`.
3. Disconnect button (line 528): `'default'` → `'not-allowed'`.
4. Look-up-orgs button (line 645): `'default'` → `'not-allowed'`.
5. `NumberStepper.tsx`: removed both inline `cursor` expressions — the `disabled={atMin/atMax}` attribute already exists on both buttons, so the global rule handles it.
6. File header updated to new tagline (PR #250 catchup).

**Pushed back on (still standing):**
Comment #3 (test file header) — `cursor-pointer.test.ts` now correctly uses the new header (`//ambient dev tool…`). I had initially kept the old header thinking it was correct; PR #250 clarified that the new header is the standard. Fixed.

PR #259, opened 2026-06-11
fix(github): guard against null nodes in ProjectV2 GraphQL response

adityaharishch:
## Summary

- `ProjectV2ItemConnection.nodes` and `UserConnection.nodes` are typed `[T]` in GitHub's GraphQL schema — both the list and individual elements can be `null`
- The Rust structs had them as `Vec<T>` (non-optional elements), so a null element would cause a serde deserialization failure and silently drop the entire project's tasks
- Changed both to `Vec<Option<T>>` and added `.flatten()` at the two iteration sites

## Test plan

- [ ] Verify `cargo clippy` passes (no warnings)
- [ ] Trigger a GitHub sync and confirm tasks from all configured projects are fetched correctly
- [ ] Confirm behaviour is unchanged when no null elements are present (normal case)

### DIFF
--- a/src/intelligence/providers/github.rs
+++ b/src/intelligence/providers/github.rs
@@ -46,7 +46,7 @@ struct ProjectNode {
 struct ProjectItemConnection {
     #[serde(rename = "pageInfo")]
     page_info: PageInfo,
-    nodes: Vec<ProjectItem>,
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde (approved):

Akarsh-Hegde:
Superseded by #257

PR #267, opened 2026-06-11
chore(ci): parallelise Rust and UI jobs, drop redundant release build

adityaharishch:
## Summary

- Splits the single serial `ci` job into two independent parallel jobs (`rust` and `ui`), cutting wall-clock CI time by roughly 50%
- Removes the redundant `cargo build --release` step — `cargo test` already compiles all crates, so the extra release build was wasted minutes
- Adds a Next.js build cache (`ui/.next/cache`) to the UI job, matching the pattern already used in `release.yml`
- Upgrades the UI job from Node 20 to Node 24 to match `release.yml`

The `migration-guard` job is unchanged.

## Test plan

- [ ] CI passes on this PR (rust + ui jobs run in parallel, migration-guard runs only on PRs)
- [ ] Verify wall-clock time is shorter than the previous serial job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,8 +17,8 @@ env:
   SQLX_OFFLINE: "true"
 
 jobs:
-  ci:
-    name: CI
+  rust:
+    name: Rust
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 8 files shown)


adityaharishch:
Closing in favour of a clean rebase — conflicts were in the base branch, not the CI change.

PR #272, opened 2026-06-12
docs(claude): add dev-vs-installed mode matrix to CLAUDE.md

Akarsh-Hegde:
## Why

One machine carries two flavours of Meridian — the source checkout and the installed bundle — and they resolve `.env` / `settings.json` from different places (cwd-based). This has repeatedly derailed debugging sessions: most recently, OTLP export stayed disabled because the UI wrote credentials to the repo `settings.json` while the installed daemon (cwd `~/.meridian/app`) read a different, absent file and fell back to defaults (root cause fixed in #271).

## What

Adds a "Dev vs Installed — Two Parallel Meridians" section to CLAUDE.md, placed before Environment Variables:

- a mode matrix (daemon process/cwd, `.env`, `settings.json`, UI, MLX, supervision, VERSION marker per flavour)
- the shared-state collision risks (`meridian.db`, logs, ports 3939/7823/5080) and the "never run both daemons" rule
- commands to identify the running flavour (`launchctl print`, `ps`, `lsof` cwd)
- the mode lifecycle (`install.sh` / `install-dev.sh` / `dev-start.sh` / `meridian stop|start`)
- rules for agents: check the running process's cwd before concluding a config change "has no effect"; installed daemon runs the old binary until redeployed; cp-don't-symlink for daemon binaries (dyld/TCC hang)

Docs-only — no code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -152,6 +152,45 @@ There are no JS/TS test suites yet. When adding them, place them under `ui/__tes
 
 ---
 
+## Dev vs Installed — Two Parallel Meridians (read BEFORE debugging runtime behaviour)
+
+One machine can carry **two flavours of Meridian at once**: the source checkout (this repo) and the installed bundle. They resolve config differently, and most "my settings change has no effect" bugs are edits to the file the *other* flavour reads. Identify which flavour you are looking at before touching any config.
+
+| | Dev (source) | Installed (bundle) |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde:
## Review

Docs-only; accurate and useful. No code concerns.

### 🟡 Notes
- The `settings.json` row describes the *current* cwd-based split. Once #271 merges, both flavours resolve `~/.meridian/settings.json` — worth a one-line "(unified by #271)" so the matrix doesn't read as stale after that lands.
- Otherwise the mode matrix, identify-the-flavour commands, and agent rules all match reality.

### ⚠️ Blocking (not this PR)
Red **Rust** check = pre-existing `cargo fmt` drift on `main`; this PR changes only `CLAUDE.md`. Goes green on rebase once the fmt fix merges.

PR #274, opened 2026-06-12
fix(ui): gate standalone output to production so dev server boots

Akarsh-Hegde:
## Problem
`next dev --turbopack` crashes on startup with:

```
TurbopackInternalError: Invalid distDirRoot: ".next". distDirRoot should not navigate out of the projectPath.
```

`ui/next.config.ts` sets `output: 'standalone'` unconditionally. Combined with the pinned `outputFileTracingRoot` / `turbopack.root` (both `ui/`), Next 16.2's Turbopack mis-resolves the dev `distDir` and aborts. This blocks the local hot-reload dev server (`dev-start.sh` window #3 and `cd ui && npm run dev`).

## Fix
Gate `output: 'standalone'` on `NODE_ENV === 'production'`. Standalone output is purely a production/packaging concern; it has no purpose in dev.

```ts
output: process.env.NODE_ENV === 'production' ? 'standalone' : undefined,
```

## Why this is safe
- `npm run build` runs with `NODE_ENV=production` → standalone bundle still emitted exactly as before. `package-release.sh` / `install-from-bundle.sh` are unaffected.
- `next dev` runs with `NODE_ENV=development` → no standalone tracing → Turbopack boots clean.

## Verification
- `next dev --turbopack` → `✓ Ready` and serves (was a hard crash before).
- `output: 'standalone'` still active for production builds (NODE_ENV gate).

### DIFF
--- a/ui/next.config.ts
+++ b/ui/next.config.ts
@@ -20,7 +20,12 @@ const nextConfig: NextConfig = {
   // crash-loops the standalone server. So package-release.sh packs the standalone
   // into `ui.tar.gz` (tar preserves symlinks; npm ships it as one opaque file) and
   // install-from-bundle.sh extracts it on the user's machine — keeping Turbopack.
-  output: 'standalone',
+  // Only emit the standalone server bundle for production builds. Applying
+  // `output: 'standalone'` in dev makes Next 16's Turbopack mis-resolve the
+  // distDir relative to the pinned `outputFileTracingRoot`, crashing
+  // `next dev --turbopack` with "Invalid distDirRoot". Production builds set
+  // NODE_ENV=production, so the shipped standalone bundle is unaffected.
+  output: process.env.NODE_ENV === 'production' ? 'standalone' : undefined,
   // Pin the standalone file-tracing root to this `ui/` dir. The repo root also
   // has a package-lock.json, so Next otherwise infers the monorepo root and
   // nests the output as `.next/standalone/ui/server.js` — which breaks the


Akarsh-Hegde:
## Code review

Correct, well-scoped fix for the `Invalid distDirRoot ".next"` Turbopack panic on `next dev`. Root cause (`output: 'standalone'` confusing Turbopack's distDir resolution against the pinned `outputFileTracingRoot` in dev) is the real one — this supersedes the "rm -rf ui/.next" workaround.

### ✅ Correctness
- `next dev` runs with `NODE_ENV=development` → `output` is `undefined` → no standalone → no panic.
- `next build` sets `NODE_ENV=production` → `output: 'standalone'` → shipped bundle unchanged. The green "UI build" check confirms standalone still emits.

### 🟡 One thing to keep in mind
This now couples standalone emission to `NODE_ENV` being `production` at build time. If any build path ever invokes `next build` with `NODE_ENV` unset/overridden (some CI wrappers do), standalone silently won't emit and `package-release.sh` will fail at its `ui/.next/standalone` check. Worth a one-line comment in `package-release.sh` noting the dependency, or asserting `NODE_ENV=production next build` explicitly there.

### ⚠️ Not your PR, but blocking it
The red **Rust** check is pre-existing `cargo fmt` drift on `main` (`config.rs`, `providers/*.rs`) — nothing here touches Rust. It'll stay red until the fmt fix lands on main. See the triage note; this goes green on rebase afterward.

Akarsh-Hegde:
Superior, more complete fix
  stashed on feat/otlp-ui-settings

PR #269, opened 2026-06-12
style(ui): remove per-class serif font-smoothing overrides and add app icons

Akarsh-Hegde:
## Summary

- Removes the `auto` `-webkit-font-smoothing` / `-moz-osx-font-smoothing` overrides that were applied individually to each serif type-scale class in `globals.css`. The global `antialiased` setting in `* { -webkit-font-smoothing: antialiased }` already covers all elements; the per-class `auto` overrides were inconsistent with it.
- Removes `letter-spacing: -0.01em` from `.font-stat` (inconsistency with other stat-scale rules).
- Adds `apple-icon.png`, `favicon.ico`, and `icon.png` to `ui/app/` for the Next.js app shell.

## Test plan

- [ ] Verify serif type-scale headings render correctly in the dashboard (no visual regression from the font-smoothing change)
- [ ] Confirm favicon appears in browser tab

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #275, opened 2026-06-12
fix(ui): render display serif in Georgia to match design

Akarsh-Hegde:
## Summary

The design mocks (claude.ai artifacts) were rendered in a sandboxed environment where Google Fonts are blocked, so the serif stack fell back to a system font. The UI, meanwhile, loaded the real **Instrument Serif** — a deliberately condensed hairline display face — which is why large headings looked thinner and squeezed compared to the design.

Pixel forensics against the design screenshot identified the actual rendered fallback as **Georgia**: the design's narrative digits (KAN-70, KAN-67) are old-style figures that dip below the baseline — Georgia's signature; Iowan Old Style and New York render lining figures. Line widths match Georgia glyph-for-glyph. (The artifact preview pane also scales the design ~50%, which is why naive size comparisons were misleading — the mock really is 72px type.)

- Set all display serif type (`.type-hero`, `.type-title`, `.type-heading`, `.type-stat`, `.type-active`, `.type-empty`, `.type-callout`, `.type-wordmark`, `.font-serif`) to `Georgia, 'Times New Roman', serif`
- Drop the unused Instrument Serif Google Font load from `layout.tsx`
- Remove the per-class `-webkit-font-smoothing: auto` subpixel overrides — the design renders with the global antialiased setting (supersedes #269)

## Test plan

- [x] `npm run build` passes
- [x] `bun test` — 119 pass, 0 fail
- [x] Headless-Chrome scaled side-by-side of design crop vs new `/today`: letterforms, stroke weight, and old-style digits match
- [ ] Owner visual sign-off (PR stays draft until then)

### DIFF
--- a/ui/app/globals.css
+++ b/ui/app/globals.css
@@ -9,7 +9,7 @@
   /* Fonts */
   --font-sans:  var(--font-geist-sans);
   --font-mono:  var(--font-geist-mono);
-  --font-serif: 'Iowan Old Style', Georgia, serif;
+  --font-serif: Georgia, 'Times New Roman', serif;
 
   /* Colors → Tailwind utilities (text-*, bg-*, border-*) */
   --color-paper:       var(--paper);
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
Superseded by the minimal-diff version: see fix/ui-serif-georgia PR.

PR #291, opened 2026-06-15
fix(dev): resolve meridian binary to repo build in dev to stop migration drift

adityaharishch:
## Problem

In local dev, the **Tasks → Sync** button failed with `Sync failed: tasks-sync: open db: failed to run migrations`.

Root cause is a dev-only binary mismatch:
- `dev-start.sh` runs the daemon as `cargo watch -x 'run --bin meridian'` → `target/debug/meridian`. That build is rebuilt on every save and migrates `meridian.db` forward as new migrations land.
- UI route handlers (`/api/tasks/sync`, OAuth, triage) resolved a **separately-installed** binary under `~/.meridian/app/bin/meridian` via `selectMeridianBinary(meridianCandidates())`. That artifact does **not** track the repo.

So the instant a migration was added, the daemon advanced the DB while the installed CLI stayed behind, and sqlx refused to open the ahead-of-it DB — surfacing as the opaque "failed to run migrations".

## Fix

Make `meridianCandidates()` environment-aware:
- **`NODE_ENV === 'development'`** (`npm run dev`): prepend `<repo>/target/debug/meridian` then `target/release/meridian` ahead of the installed paths (`next dev` runs with cwd = `<repo>/ui`, so repo root is one level up). The UI now shells out to the **same** binary the daemon runs — they can't drift.
- **Production** (`NODE_ENV !== 'development'`): unchanged — installed-only, native-first, preserving the launchd PATH-parity guarantee from the original `env: node: No such file or directory` fix.

Also: print the full anyhow source chain (`{e:#}`) on `tasks-sync` open-db failure so the real sqlx cause is visible next time instead of the top-level context (`src/main.rs`).

## Tests

- Added dev-mode coverage in `ui/__tests__/meridian-bin.test.ts`: repo build prepended, debug-before-release, node wrapper still last, production no-op.
- Pinned the existing `selectMeridianBinary` tests to `'production'` for determinism.
- `bun test` → 12 pass / 0 fail; UI build clean; full pre-push suite passed.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -262,7 +262,10 @@ async fn main() -> Result<()> {
                 pool.close().await;
             }
             Err(e) => {
-                eprintln!("tasks-sync: open db: {e}");
+                // `{e:#}` prints the full anyhow source chain (e.g. the sqlx
+                // "migration N was previously applied / missing" cause) instead
+                // of just the top-level "failed to run migrations" context.
+                eprintln!("tasks-sync: open db: {e:#}");
                 std::process::exit(1);
             }
         }
--- a/ui/__tests__/meridian-bin.test.ts
+++ b/ui/__tests__/meridian-bin.test.ts
@@ -36,8 +36,33 @@ describe('meridianCandidates', () => {
   })
 })
 
+describe('meridianCandidates in development', () => {
+  // `npm run dev` runs with cwd = <repo>/ui, so the repo root is one level up.
+  const list = meridianCandidates(HOME, 'development', '/work/meridian/ui')
+
+  it('prepends the repo build (target/debug) ahead of the installed binaries', () => {
+    expect(list[0]).toBe('/work/meridian/target/debug/meridian')
+    expect(list.indexOf('/work/meridian/target/debug/meridian')).toBeLessThan(list.indexOf(native))
+  })
+
+  it('keeps a release build as a fallback after debug but before installed', () => {
+    const debug = list.indexOf('/work/meridian/target/debug/meridian')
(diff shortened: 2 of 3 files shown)


Akarsh-Hegde:
## 🔍 Code review — resolve meridian binary to repo build in dev

High-effort pass, verified against source. **Clean — no bugs found.** Nice fix for a real dev/prod drift trap.

The change adds an `env`/`cwd`-parameterised dev branch to `meridianCandidates()`: under `NODE_ENV === 'development'` it prepends `target/debug/meridian` then `target/release/meridian` ahead of the installed paths, so the dashboard shells out to the **same** binary `cargo watch` runs — closing the window where a freshly-added migration advances `meridian.db` past what the installed CLI knows, and sqlx aborts task-sync with the opaque "failed to run migrations".

What I checked:
- **Safe production default** — any `env !== 'development'` (incl. `undefined`) returns the installed-only list unchanged. ✅
- **Safe degradation if the cwd heuristic is wrong** — `resolve(cwd, '..')` assumes `cwd = <repo>/ui`; if it isn't, the prepended `target/*` paths simply aren't executable and `selectMeridianBinary` skips them. No crash, falls through to installed. ✅
- **launchd-parity preserved** — native binaries still precede the node wrapper; the wrapper stays last. The new test `'still lists the node wrapper last'` guards it.
- **`main.rs` `{e:#}`** correctly prints the full anyhow source chain (the sqlx migration cause) instead of just the top-level context — directly useful for the drift this PR targets.
- Tests cover dev ordering (debug < release < installed), the production no-prepend case, and the existing selection logic (now explicitly passing `'production'`).

LGTM.

PR #297, opened 2026-06-16
feat(observability): full session-task classifier debugging in OpenObserve

adityaharishch:
Closes KAN-240. Makes the session→task classifier fully traceable, searchable, and debuggable in OpenObserve. Three commits:

### 1. Export Python classifier traces + logs to OpenObserve
The classifier (`run_task_linker_mlx`) was fully instrumented but exported **nothing** — `observability.py` read its OTLP endpoint+auth from env vars nothing sets in dev/launchd. Now resolves config from `~/.meridian/settings.json` (the **same source the Rust daemon uses**: `otlp_enabled`/`otlp_endpoint`/`oo_email`/`oo_password` → base64 Basic). Also wires an OTLP-logs exporter so every `log.*` reaches OO carrying the active span's `trace_id`/`span_id`.

### 2. Enrich `classify_session` into one self-describing span
Each per-session span now carries the full result as queryable attributes — `session_id`, `task_key`, `has_task`, `session_type`, `category`, `confidence`, `method`, `is_error` — set once in `_classify_one_logged` so both the server (daemon) and CLI paths get them. Error paths set span status ERROR. This turns the Traces stream into a complete debugger: list newest-first, search by `session_id`, filter `is_error`, slice by `session_type` — no child-span joins.

### 3. OpenObserve "Classifier — Debug" dashboard
`services/observability/dashboards/classifier-debug.json` (v5): 4 metric tiles (total / errors / untracked / avg confidence) + a newest-first table with **session_id / session_type / errors-only** variable filters + an errors-only table. Each row has `trace_id` to pivot into the waterfall.

## Verified end-to-end against live OpenObserve
- Real classify produced the full waterfall (`classify_sessions → classify_session → db_fetch / build_prompt / llm_inference / parse_response`) + correlated log row, same `trace_id`.
- Enriched span returns one row per session with all attributes.
- All 6 dashboard queries + the variable-substitution guard validated via the search API.
- Dashboard created with 6 panels + 3 variables; live at dashboardId `7472515436769181696`.

## Follow-up (not in this PR)
- Auto-import the dashboard on install (script over the dashboards API).
- The JSONL **file** logs still show `trace_id: null` (cosmetic LoggingInstrumentor quirk; OO export correlates correctly).


services/observability/dashboards/classifier-debug.json
```diff
@@ -0,0 +1,174 @@
...
+          "queryType": "sql",
+          "queries": [
+            {
+              "query": "SELECT to_char(to_timestamp_micros(_timestamp),'%Y-%m-%d %H:%M:%S') as \"Time\", session_id as \"Session\", task_key as \"Task\", session_type as \"Type\", category as \"Category\", confidence as \"Confidence\", round(CAST(elapsed_s AS DOUBLE),2) as \"Time taken (s)\", method as \"Method\", is_error as \"Error\", trace_id as \"trace_id\", encode(concat('trace_id=''', trace_id, ''' and service_name=''meridian-task-linker-mlx'''),'base64') as \"trace_filter\" FROM \"default\" WHERE operation_name='classify_session' AND ('$session_id'='' OR session_id='$session_id') AND ('$session_type'='' OR session_type='$session_type') AND ('$errors_only'='' OR is_error='$errors_only') ORDER BY _timestamp DESC",
```
Akarsh-Hegde: **[correctness] Drilldown returns an empty trace on the production path — wrong `service.name`.**

This `trace_filter` (and `table_errors` L162, plus the description L3) pins `service_name='meridian-task-linker-mlx'`. But the production classifier runs in-process in `server.py`, whose process registers `observability.setup("meridian-agent-server-mlx")` (server.py:733) — so every `classify_session` span carries resource `service.name=meridian-agent-server-mlx`. Only the standalone CLI (`run_task_linker_mlx.py:51`) uses `meridian-task-linker-mlx`.

The stat tiles + tables still populate (they filter only on `operation_name='classify_session'`), so the break is **silent** — but clicking any production row's “Open this trace's spans” returns an empty waterfall, which is the headline feature of this PR.

Fix: use `service_name='meridian-agent-server-mlx'`, or drop the service filter and key the drilldown on `trace_id` alone.

Akarsh-Hegde (commented):
## Code review — PR #297 (high effort, recall-biased)

Ran 6 finder angles + targeted verification against the PR head. The observability plumbing is solid overall — header-injection guard is sound, `con.close()` runs on every path in the `classify_sessions` refactor, the batch-span removal is self-contained (no downstream consumer of the old plural span/attrs), both pinned OO download URLs verified HTTP 200, and the `is_error` bool-as-string comparison is correct per the existing OO convention in TESTING.md.

**10 findings — 6 correctness, 4 cleanup/altitude, ranked by severity. Top 3 worth blocking on:**
1. **Dashboard drilldown hardcodes the wrong `service.name`** → the headline "open this trace's spans" returns an empty waterfall on the production server path.
2. **OTLP logs silently fall back to localhost** for any endpoint not ending exactly in `/v1/traces`.
3. **The `MERIDIAN_OO_AUTH` env path was removed but its wiring wasn't** — the MLX-server installer + eval pipeline still rely on it, so env-configured installs go dark.

Inline comments below. Verified-clean (not flagged): the outlines `JsonSchema` import runs, `_SUCCESS_METHODS` matches the happy path, the `float()` casts never crash (producing dicts always emit numerics), and there are no duplicate log handlers across multi-agent `setup()`.

services/agents/observability.py
```diff
@@ -50,12 +64,87 @@
...
+        os.environ.get("MERIDIAN_OTLP_TRACES_ENDPOINT", "").strip()
+        or os.environ.get("MERIDIAN_OTLP_ENDPOINT", "").strip()
+    )
+    traces_endpoint = configured or env_endpoint or DEFAULT_TRACES_ENDPOINT
+    # OpenObserve serves logs at the sibling `/v1/logs` path; derive it from the
+    # traces endpoint so a custom base host carries over to both signals.
+    if traces_endpoint.endswith("/v1/traces"):
+        logs_endpoint = traces_endpoint[: -len("/v1/traces")] + "/v1/logs"
+    else:
+        logs_endpoint = DEFAULT_LOGS_ENDPOINT
```
Akarsh-Hegde: **[correctness] Logs silently ship to localhost when the endpoint isn't exactly `…/v1/traces`.**

`logs_endpoint` is derived from the host only when `traces_endpoint.endswith('/v1/traces')`; otherwise it falls back to `DEFAULT_LOGS_ENDPOINT` (localhost:5080). A remote OpenObserve set via settings `otlp_endpoint` with a trailing slash (`…/v1/traces/`) or a base URL, or via `MERIDIAN_OTLP_ENDPOINT`, exports **spans to the remote host but OTLP logs to localhost** — dropped if no local OO runs, with no error. The log↔trace correlation this PR adds breaks on any non-canonical endpoint.

Fix: derive the logs path from the same host/base unconditionally (swap the last path segment), or normalize/validate the endpoint first.

services/agents/observability.py
```diff
@@ -50,12 +64,87 @@
...
+    `MERIDIAN_OTLP_TRACES_ENDPOINT`/`MERIDIAN_OTLP_ENDPOINT` env override →
+    the localhost default. Auth is `base64(oo_email:oo_password)` — settings.json
+    only; the legacy `MERIDIAN_OO_AUTH` env path is deprecated and ignored, the
+    same decision the daemon made.
+    """
+    if os.environ.get("MERIDIAN_TRACING_DISABLED", "").lower() in ("1", "true", "yes"):
+        return None
+
+    settings = _load_settings()
+    if not settings.get("otlp_enabled"):
```
Akarsh-Hegde, line 123: **[correctness/regression] Env-only telemetry installs go dark — the env-auth path was removed but its wiring wasn't.**

Export now requires settings.json `otlp_enabled` + `oo_email` + `oo_password`; `MERIDIAN_OO_AUTH` is ignored. But the MLX-server installer still injects it and sets no settings creds — `services/scripts/install-mlx-server-daemon.sh` (L94/100/114) — and the eval pipeline still documents it (`services/tests/evals/eval_classifier.py:14`). Per the env-var table, Python services + the installer were the documented `MERIDIAN_OO_AUTH` consumers.

After this PR, an install that configured OO auth via `.env` (no dashboard Settings creds) silently stops exporting traces **and** logs, no warning.

Fix: move the MLX-server installer/plist + eval docs to the settings.json path (or have them write `oo_email`/`oo_password`), and/or log a one-time warning when an endpoint is set but creds resolve to `None`.

services/agents/run_task_linker_mlx.py
```diff
@@ -993,6 +1029,7 @@ def _classify_one_logged(
     }
     run_log.write(json.dumps(record, default=str) + "\n")
     run_log.flush()
+    _annotate_classification_span(result)
```
Akarsh-Hegde, line 1063: **[correctness] Errors-only dashboard misses crashes — `is_error` isn't set on unhandled exceptions.**

`_annotate_classification_span(result)` runs only *after* `_classify_one(...)` returns. Handled failures return `_error_result` dicts (fine), but an unhandled raise inside `_classify_one_logged` (e.g. malformed `window_titles` JSON, a sqlite read error) propagates out: the `classify_session` context manager records the exception and sets span status ERROR, but the `is_error` attribute is never set. The new dashboard's “Errors only” table filters `is_error='true'`, so it silently **misses exactly the failures an operator opens it to find**.

Fix: set `is_error=true` (+ core attrs) in a `try/except` around the classify call, or annotate in a `finally`.

services/agents/observability.py
```diff
@@ -50,12 +64,87 @@
...
+    if any(c in email for c in "\r\n:") or any(c in password for c in "\r\n"):
+        return None
+    auth = base64.standard_b64encode(f"{email}:{password}".encode()).decode()
+
+    configured = str(settings.get("otlp_endpoint") or "").strip()
+    env_endpoint = (
+        os.environ.get("MERIDIAN_OTLP_TRACES_ENDPOINT", "").strip()
+        or os.environ.get("MERIDIAN_OTLP_ENDPOINT", "").strip()
+    )
+    traces_endpoint = configured or env_endpoint or DEFAULT_TRACES_ENDPOINT
```
Akarsh-Hegde: **[correctness/altitude] Missing endpoint scheme validation — diverges from the Rust daemon.**

`src/observability.rs::resolve_otlp_target` rejects an endpoint without an `http://`/`https://` scheme (disables export + warns). This Python port doesn't: a schemeless/garbled `otlp_endpoint` is passed straight to `OTLPSpanExporter`/`OTLPLogExporter`, which then fail/retry on every batch flush with no “disabled” signal — and the two processes disagree on whether export is on. This is the cross-language drift the hand-duplicated logic invites; at minimum mirror the scheme check + the `:`-in-email / `\r\n` guards exactly (ideally share one resolved-config source).

services/agents/server.py
```diff
@@ -415,14 +411,30 @@ async def openai_chat_completions(req: _OAIChatRequest) -> dict:
...
+            from outlines.types import JsonSchema
+            output_type = JsonSchema(schema)
+
     from agents.llm_selector import APPLE_INTELLIGENCE_ID
 
     def _generate() -> str:
         if m._resolve_model_id() == APPLE_INTELLIGENCE_ID:
+            # outlines FSM decoding is incompatible with Foundation Models;
+            # Apple FM falls back to free-form (schema hint must be in prompt).
             return _infer_apple_fm(msgs, max_tokens)
```
Akarsh-Hegde, line 443: **[correctness] `json_schema` requests to Apple FM return unconstrained output.**

`output_type` (from `response_format: json_schema`) constrains MLX decoding via outlines, but this Apple-Intelligence branch returns `_infer_apple_fm(msgs, max_tokens)` and never receives it. The comment notes “schema hint must be in prompt,” but nothing injects the schema into the prompt — so on an Apple-Intelligence-resolved model a structured-output caller (e.g. agno) gets free-form text and the downstream JSON parse fails, silently rather than with a 4xx.

Fix: inject the schema into the Apple FM prompt, or reject `json_schema` when Apple FM is active.

services/agents/run_task_linker_mlx.py
```diff
@@ -945,6 +946,41 @@ def _open_run_log(db_path: str) -> "tuple[Path, Any]":
     return log_path, log_path.open("w", encoding="utf-8")
 
 
+# `method` values that mean the model produced a usable classification.
+# Anything else — mlx_error / apple_fm_error / schema_error / invalid_task_key /
+# session-not-found — is an error path the dashboard surfaces under errors-only.
+_SUCCESS_METHODS = {"mlx_direct", "apple_fm"}
```
Akarsh-Hegde, line 954: **[altitude] Stale docstring + denylist-by-negation.**

The comment above lists `schema_error`/`invalid_task_key` as error *methods*, but those are only child-span `outcome` attributes — the actual emitted `method` for both is `mlx_parse_error` (L881/901). The success set is also a denylist-by-negation: add a new success-method tag and forget to list it here → every such session is silently `is_error=true` + span status ERROR. Fix the comment to the real method values, and consider one authoritative success/error enum.

services/agents/server.py
```diff
@@ -259,49 +259,45 @@ async def classify_sessions(req: ClassifySessionsRequest) -> dict:
...
-                finally:
-                    con.close()
+                results: list[dict] = []
+                for sid in req.session_ids:
+                    # _classify_one_logged enriches this span (session_id, task_key,
+                    # confidence, is_error, …) and emits db_fetch / build_prompt /
+                    # llm_inference / parse_response as its children.
+                    with tracer.start_as_current_span(
+                        "classify_session",
+                        attributes={"session_id": sid},
```
Akarsh-Hegde: **[cleanup] Redundant `session_id`, inconsistent with the CLI path.**

This sets `session_id` at span creation, but the wrapped `_classify_one_logged` → `_annotate_classification_span` sets it again on the same span (run_task_linker_mlx.py:969). This PR deliberately removed the creation-time `session_id` from the CLI path (`main()`) and let `_annotate` own it — server.py was left inconsistent. Drop `attributes={"session_id": sid}` for one source of truth.

install.sh
```diff
@@ -584,32 +584,13 @@ if [[ "$_oo_installed" -eq 0 ]]; then
...
-                _oo_ver="${_oo_result%% *}"
-                _oo_url="${_oo_result#* }"
-                [[ "$_oo_ver" == "$_oo_url" ]] && _oo_url=""  # single token = no URL found
-            fi
-        else
+        # GitHub stopped attaching binary assets to OO releases, so the binaries
+        # live on the official downloads host. Pin a current stable — trace
+        # deep-linking (dashboard drilldown into a single trace's spans) needs a
+        # modern build; the old v0.11 GitHub asset is long gone.
+        _oo_ver="v0.90.3"
```
Akarsh-Hegde, line 593: **[altitude] OO version pinned in two places.**

`v0.90.3` (+ the `downloads.openobserve.ai` URL template) is hardcoded here and again in `scripts/install-openobserve-daemon.sh:54`; they must be bumped in lockstep, or the two install routes pull different OO versions. (Both URLs verified live / HTTP 200 — maintainability note, not a current break.) Consider a single shared version constant.

services/observability/dashboards/classifier-debug.json
```diff
@@ -0,0 +1,174 @@
...
+          "title": "Errors only",
+          "description": "Failed classifications — inference errors, schema errors, invalid task_key, session-not-found. Click a row → opens just that trace's spans.",
+          "config": {"show_legends": false, "wrap_table_cells": false, "table_dynamic_columns": false, "drilldown": [{"name": "Open this trace's spans", "type": "byUrl", "targetBlank": true, "findBy": "name", "data": {"url": "/web/traces?org_identifier=default&stream=default&search_type=ui&search_mode=spans&from=${start_time}&to=${end_time}&query=${row.field.trace_filter}", "folder": "", "dashboard": "", "tab": "", "passAllVariables": false, "variables": []}}]},
+          "queryType": "sql",
+          "queries": [
+            {
+              "query": "SELECT to_char(to_timestamp_micros(_timestamp),'%Y-%m-%d %H:%M:%S') as \"Time\", session_id as \"Session\", task_key as \"Task\", session_type as \"Type\", method as \"Method\", trace_id as \"trace_id\", encode(concat('trace_id=''', trace_id, ''' and service_name=''meridian-task-linker-mlx'''),'base64') as \"trace_filter\" FROM \"default\" WHERE operation_name='classify_session' AND is_error='true' ORDER BY _timestamp DESC",
```
Akarsh-Hegde: **[cleanup] `table_errors` largely duplicates `table_all`.**

The same drilldown block + `encode(concat(… service_name=…))` `trace_filter` SQL are copy-pasted; `table_errors` is `table_all` with `is_error='true'` hardcoded — which the existing `errors_only` variable on `table_all` already covers. Any change to the service name / drilldown URL / trace_filter must be made in both (and note: same wrong `service_name` as the L144 finding). Consider dropping the panel or deriving it.

services/agents/observability.py
```diff
@@ -143,6 +232,29 @@ def _configure_tracing(agent_name: str) -> None:
...
+    """Build an OTLP-logs handler so every `log.*` record reaches OpenObserve,
+    correlated to the active span by trace_id/span_id — the Python counterpart
+    of the Rust daemon's `OpenTelemetryTracingBridge`.
+
+    Returns the handler (caller attaches it to root) or `None` when export is
+    disabled, in which case logs still go to the JSONL file + stdout/stderr.
+    """
+    global _LOGGER_PROVIDER
+
+    target = _resolve_otlp_target()
```
Akarsh-Hegde: **[efficiency/correctness-minor] `_resolve_otlp_target()` runs twice per `setup()`** — once in `_configure_tracing` (L221) and again here, each re-opening + parsing `~/.meridian/settings.json` and re-running the base64/guard logic. Bounded (setup runs once per process), so the I/O cost is minor — but it opens a TOCTOU split: if the dashboard rewrites settings.json between the two reads, traces can resolve enabled while logs resolve disabled (or different creds/endpoints) within one setup. Resolve once in `setup()` and pass the single `_OtlpTarget` (or `None`) into both configurers.

services/agents/observability.py
```diff
@@ -204,6 +316,15 @@ def filter(self, record: logging.LogRecord) -> bool:
...
     root.addHandler(stderr_h)
+    # Ship every record to OpenObserve via OTLP/HTTP logs too, when export is
+    # configured. The OTel LoggingHandler reads the active span context, so each
+    # OO log row carries the trace_id/span_id that ties it to the classifier's
+    # span waterfall. No-op (None) when OTLP is disabled.
+    # The OTLP handler already carries service.name via the OTel Resource, so it
+    # needs no _ServiceFilter (that would duplicate the attribute on each record).
+    otlp_log_h = _configure_log_export(agent_name)
+    if otlp_log_h is not None:
+        root.addHandler(otlp_log_h)
```
Akarsh-Hegde, line 377: **[correctness-minor] OTLP log handler on the root logger can self-amplify on export failure.** The `LoggingHandler` is attached to root, so every record in the process is queued to `BatchLogRecordProcessor` and shipped over the network. The OTLP exporter's own transport is `httpx`/`urllib3` — which are in `_NOISY_LOGGERS` but only *level-gated* to WARNING, not excluded from this handler. If export starts failing/retrying, those libs emit WARNING+ records → the root OTLP handler tries to export them → more failures: a log→export→log loop. The batch processor is async so the hot path stays non-blocking, but consider attaching the OTLP handler to a specific logger (or filtering out the exporter's own transport loggers) rather than root.

services/agents/observability.py
```diff
@@ -50,12 +64,87 @@
...
 
 # Track which agents have been configured so a second setup() call is a no-op.
 _INITIALISED: dict[str, trace.Tracer] = {}
 _PROCESS_SERVICE_NAME: str | None = None
+# Held so shutdown() can flush log records the same way it flushes spans.
+_LOGGER_PROVIDER: LoggerProvider | None = None
+
+
+# ──────────────────────── OTLP target resolution ───────────────────────────────
+class _OtlpTarget:
```
Akarsh-Hegde: **[simplification] `_OtlpTarget` reinvents a value object.** A hand-written class with `__slots__`, an explicit 3-arg `__init__`, and no `__eq__`/`__repr__` is exactly what `typing.NamedTuple` (or `@dataclass(frozen=True, slots=True)`) gives in one declaration, with free repr/eq. Adding a 4th field (e.g. a metrics endpoint) currently means editing `__slots__` *and* the `__init__` signature *and* every positional call site; a NamedTuple would be one line — and it'd print readably in debug logs/test assertions instead of an opaque object id.

services/observability/dashboards/classifier-debug.json
Akarsh-Hegde: **[robustness, non-blocking] Raw `$session_id` interpolation can produce malformed SQL.** The variable declares `escapeSingleQuotes:false` and the pattern is `('$session_id'='' OR session_id='$session_id')`. A value containing a `'` yields malformed SQL → the panel errors / shows no rows. Not an injection concern in practice (local single-tenant debug instance, numeric-id textbox), and worst case is a broken panel — flagging only for completeness.

Akarsh-Hegde (commented):
Additional lower-severity findings held back from the first pass (the 10-cap), posted for completeness.

services/observability/dashboards/classifier-debug.json
adityaharishch: Fixed in fd208cb. The drilldown `trace_filter` now keys on `trace_id` alone (dropped the `service_name` predicate), so it resolves regardless of whether the span was emitted by the in-process server (`meridian-agent-server-mlx`) or the standalone CLI (`meridian-task-linker-mlx`). Description updated to match.

services/agents/observability.py
adityaharishch: Fixed in fd208cb. The logs endpoint is now derived by swapping the trailing signal segment of the traces endpoint (`/v1/traces` or `/traces`, after rstrip of a trailing slash), so a non-canonical or remote endpoint sends logs to the same host. No more silent localhost fallback. Verified for `…/v1/traces`, `…/v1/traces/`, and a custom base.

services/agents/observability.py
adityaharishch: Addressed in fd208cb. `_resolve_otlp_target()` now logs a one-time WARNING when `otlp_enabled` is true but `oo_email`/`oo_password` are missing — so an env-only (`MERIDIAN_OO_AUTH`) install no longer goes dark silently; it mirrors the daemon, which already warns. Migrating the MLX-server installer/plist + eval docs to write `oo_email`/`oo_password` is a larger follow-up I left out of this PR, but the warning closes the silent-failure gap.

services/agents/run_task_linker_mlx.py
adityaharishch: Fixed in fd208cb. `_classify_one_logged` now wraps the whole body (delegating to `_classify_one_logged_inner`) and, on ANY unhandled exception, stamps `is_error=true` + `method=mlx_error` + ERROR status on the `classify_session` span before re-raising — so the errors-only table catches crashes (malformed `window_titles`, sqlite read errors), not just handled `_error_result` paths.

services/agents/observability.py
adityaharishch: Fixed in fd208cb. `_resolve_otlp_target()` now rejects an endpoint without an `http://`/`https://` scheme (returns `None` + warns), mirroring `src/observability.rs::resolve_otlp_target` so the two processes agree on whether export is on. Verified `ftp://…` → disabled.

services/agents/server.py
adityaharishch: Fixed in fd208cb. When `response_format=json_schema` is requested and the resolved model is Apple FM, the endpoint now returns a 400 ("not supported on Apple Foundation Models") instead of silently emitting unconstrained free-form text that fails downstream.

services/agents/run_task_linker_mlx.py
adityaharishch: Fixed in fd208cb. Corrected the comment: the real error `method` values are `mlx_parse_error` (schema validation / unknown task_key — those names are child-span `outcome` attributes, not methods) and `mlx_error`. Left `_SUCCESS_METHODS` as the success allowlist but documented it accurately.

services/agents/server.py
adityaharishch: Fixed in fd208cb. Dropped the creation-time `session_id` attribute on the server path; `_annotate_classification_span` is now the single source (matching the CLI). The failure case is still covered — the new exception handler (other thread) sets `session_id` on the span if the inner call raises before annotation.

install.sh
adityaharishch: Addressed in fd208cb with `KEEP IN SYNC` cross-reference comments at both version pins (install.sh ↔ scripts/install-openobserve-daemon.sh). A single shared constant would mean a sourced snippet across two standalone installers — judged not worth the indirection for one string; the cross-refs make the lockstep requirement explicit.

services/observability/dashboards/classifier-debug.json
adityaharishch: Addressed in fd208cb. Kept the dedicated `table_errors` panel intentionally (a one-click always-on errors view without toggling the `errors_only` variable), but de-duplicated its `trace_filter` via the L144 fix (both now use the same trace_id-only encode). If you'd rather drop it entirely in favor of the variable, easy follow-up.

services/agents/observability.py
adityaharishch: Fixed in 3104d23. `setup()` now resolves the target once and passes the single `_OtlpTarget` (or `None`) into both `_configure_tracing` and `_configure_logging` → `_configure_log_export`. No second settings.json read, no TOCTOU split.

services/agents/observability.py
adityaharishch: Fixed in 3104d23. The OTLP log handler now carries a filter that drops records from the exporter's own transport loggers (`httpx`/`httpcore`/`urllib3`/`grpc`/`opentelemetry`), so an export failure can't loop (log → export → fail → log). Those records still reach the file/stderr handlers.

services/agents/observability.py
adityaharishch: Fixed in 3104d23. `_OtlpTarget` is now a `typing.NamedTuple` — free repr/eq, positional call sites unchanged, and a future field is a one-line add.

services/observability/dashboards/classifier-debug.json
adityaharishch: Fixed in 3104d23. Set `escapeSingleQuotes: true` on the dashboard variables, so a `session_id` value containing a quote can no longer produce malformed SQL / a broken panel.

adityaharishch:
Thanks for the thorough review — all 14 findings addressed across two commits (`fd208cb`, `3104d23`), pushed to this branch.

**Correctness (the top 3 + the rest):**
- Drilldown now keys on `trace_id` alone → works on the in-process server path (was the wrong `service_name`).
- OTLP logs endpoint derived by swapping the trailing signal segment → no silent localhost fallback on non-canonical/remote endpoints.
- One-time warning when export is enabled but creds are missing → env-only (`MERIDIAN_OO_AUTH`) installs no longer go dark silently.
- `is_error=true` now stamped on ANY unhandled exception in `_classify_one_logged` → errors-only table catches crashes.
- Endpoint scheme validation mirrored from the Rust daemon.
- `json_schema` under Apple FM now returns 400 instead of silent free-form.
- Resolve-once in `setup()` (no double read / TOCTOU); OTLP log handler filters its own transport loggers (no log→export loop).

**Cleanup:** corrected `_SUCCESS_METHODS` comment; dropped redundant `session_id`; `_OtlpTarget` → `NamedTuple`; `escapeSingleQuotes=true` on dashboard vars; `KEEP IN SYNC` cross-refs on the OO version pin.

**Judgment calls (noted in-thread):** kept the dedicated `table_errors` panel (one-click errors view) but de-duped its filter; the MLX-server installer migration off `MERIDIAN_OO_AUTH` is left as a follow-up (the warning closes the silent-failure gap for now).

Verified: Python compiles, dashboard JSON valid, both installers `bash -n` clean, and the endpoint/scheme/creds/NamedTuple/log-filter logic is functionally exercised. pre-push suite passed on both commits.

PR #298, opened 2026-06-16
refactor(tray): unified install-mode env resolution + path visibility

Akarsh-Hegde:
## Summary

This PR cleans up how the tray app and installer resolve the `.env` credential file and `settings.json`, adding proper observability and fixing an architectural inconsistency inherited from the old bundle-install layout.

---

### The problem

**Three issues existed before this PR:**

1. **`.env` was in the wrong place.** `settings.json` and `meridian.db` both live at `~/.meridian/` (user data, install-independent). The `.env` credential file lived at `~/.meridian/app/.env` — inside the application binary directory. This mixed user credentials with the installed binary: an app update could clobber them, and the path was non-obvious to users and contributors.

2. **Two independent copies of the same probe.** `commands.rs::env_from_daemon_dotenv` and `integrations.rs::active_env_path` both implemented the same "find the daemon's `.env`" logic. Neither named the install type. They could silently drift.

3. **No startup visibility.** No log showed which `.env` was used, which resolution branch won, or what install type was active. Debugging `"why is Linear showing as disconnected?"` required reading source code.

---

### Changes

**`tray/src-tauri/src/commands.rs`**

Adds `InstallMode { Canonical(PathBuf), Dev(PathBuf), Bare }` and a single `detect_install_mode()` function shared by both `meridian_db_path` and `get_integrations`:
- `Canonical` — `~/.meridian/.env` exists (all install types)
- `Dev` — no canonical env; cwd walk finds a repo `.env` (local dev / contributor)
- `Bare` — neither present; process-env or hardcoded defaults only

`meridian_db_path()` now logs at `info!` with `source`, `env_file`, and `path` on every startup — immediately visible in OpenObserve.

**`tray/src-tauri/src/integrations.rs`**

Removes the duplicate `active_env_path()` function and uses `detect_install_mode()` instead — guaranteed to agree with the DB path resolution.

**`meridian-core/src/settings.rs`**

`settings_json_path()` already resolves install-independently — no structural change. Adds `tracing::debug!` on each branch (`env_override` / `canonical` / `cwd_fallback` / `default_not_yet_created`) so the winning path is queryable in OpenObserve.

**`scripts/install-from-bundle.sh`**

`ENV_FILE` now points to `~/.meridian/.env` instead of `${APP_ROOT}/.env` (`~/.meridian/app/.env`). All credential collection, port recording, and MLX key injection write to the canonical user-data location. Updated the `CURSOR_AGENT_AUTO_INSTALL` tip message.

**`scripts/meridian-npm-setup.sh`**

Removes the "copy `app/.env` into staging to survive the atomic swap" line — unnecessary since the file now lives outside the swap area. Replaces it with a **one-time migration**: if `~/.meridian/app/.env` exists on upgrade, it is moved to `~/.meridian/.env` (or removed if the canonical already exists), making the upgrade seamless.

**`scripts/meridian-cli.sh`**

- `meridian config edit` now opens `~/.meridian/.env` (canonical), falling back to `${REPO_ROOT}/.env` for source/dev installs
- `_smoke_read_env` uses the same probe order
- `cmd_doctor` config check accepts either path as healthy

---

### The canonical layout after this PR

```
~/.meridian/
  meridian.db        ← database
  settings.json      ← UI preferences
  .env               ← credentials   (was ~/.meridian/app/.env)
  oauth/             ← OAuth tokens
  app/
    bin/meridian     ← binary only, no user data
    VERSION
```

`dotenvy` walking up from `~/.meridian/app/` naturally finds `~/.meridian/.env` at the next level — the daemon needs no code change.

---

### Install-type behaviour

| Install type | `~/.meridian/.env` | `InstallMode` | Source read |
|---|---|---|---|
| Bundle / app install | written by installer | `Canonical` | `~/.meridian/.env` |
| Local dev (`tauri dev`) | absent | `Dev` | repo `.env` (cwd walk) |
| Contributor (fresh clone) | absent | `Dev` | repo `.env` (cwd walk) |
| Bare `.app` (no installer) | absent, no repo nearby | `Bare` | process-env or defaults — now **visible** in logs |

---

## Test plan

- [ ] `cargo clippy -- -D warnings` passes
- [ ] `cargo test` passes
- [ ] Bundle install: `install-from-bundle.sh` creates `~/.meridian/.env` (not `~/.meridian/app/.env`)
- [ ] Upgrade from old install: `meridian-npm-setup.sh` migration moves `app/.env` → `~/.meridian/.env`
- [ ] `meridian config edit` opens `~/.meridian/.env` on a bundle install
- [ ] Dev run: tray startup log shows `source=Dev env_file=<repo>/.env`
- [ ] Bundle install: tray startup log shows `source=Canonical env_file=~/.meridian/.env`
- [ ] Integrations page correctly reads Linear/GitHub tokens from `~/.meridian/.env`


tray/src-tauri/src/lib.rs
```diff
@@ -25,24 +39,22 @@ pub fn run() {
...
-                    &restart_item,
-                    &quit_item,
-                ])
-                .build()?;
+            // Open meridian.db ONCE at startup and share it with commands via
+            // managed state (no migrations — the daemon owns the schema). `None`
+            // if the DB can't be opened yet, so get_active errors gracefully
+            // instead of crashing the tray.
+            let db_path = commands::meridian_db_path();
+            let db_pool = tauri::async_runtime::block_on(meridian_core::open_existing(&db_path))
```
Akarsh-Hegde, line 103: **Native dashboard + popover stay broken if `meridian.db` isn't openable at tray startup (no retry).**

The pool is opened exactly once here with `create_if_missing(false)`, and `.ok()` turns any failure into a permanent `None` managed state. On a fresh install (or after the DB is deleted), the tray can easily start *before* the daemon has created/migrated `~/.meridian/meridian.db` — `open_existing` then fails and `db_pool` is `None` for the entire process lifetime.

Consequence: `get_active` / `get_today` / `get_week` all return `Err("meridian.db is not open yet")` forever. The native dashboard's `load().catch(() => {})` swallows that, so `data` stays `null` and it sticks on **“Loading…”** with no recovery; the popover shows **“DB · error”**. There's no re-open path anywhere — only an app restart fixes it. The browser path doesn't have this problem because `/api/*` opens the DB lazily per request and recovers on its own.

Suggest opening lazily on first command use (and caching the pool once it succeeds), or re-attempting the open in the poll loop, rather than a one-shot at startup.

meridian-core/src/today.rs
```diff
@@ -0,0 +1,517 @@
...
+    let active: Option<TodayActive> = sqlx::query_as::<_, ActiveRow>(&active_sql)
+        .fetch_optional(pool)
+        .instrument(tracing::info_span!("today.read.active_session"))
+        .await
+        .unwrap_or_else(|e| {
+            tracing::warn!(error = %e, "today: active_session read failed, treating as none");
+            None
+        })
+        .map(|ar| {
+            let (_t, titles) = parse_titles(&ar.window_titles, &ar.app_name);
```
Akarsh-Hegde: **`active.titles` diverges from the Node route when `window_titles` is empty.**

`parse_titles()` falls back to `[app_name]` when the entry list is empty (`names = vec![top.clone()]`). That fallback is correct for the foreground `sessions` (the TS route uses `: [topTitle]` there) — but the TS route's **active** block has no such fallback:
```js
titles: titles.map(t => t.window_name ?? t.title ?? '').filter(Boolean)  // [] when window_titles is []
```
So an active session whose `window_titles` is `[]` yields `[]` in the browser but `[app_name]` here, and the active card renders a stray app-name pill natively that the browser doesn't show (`active.titles.length > 0` gate in TodayView). The shared helper is over-applied to active — compute active titles without the `[top]` fallback to match.

meridian-core/src/today.rs
```diff
@@ -0,0 +1,517 @@
...
+            let elapsed_s = match (ms(now_iso), ms(&ar.started_at)) {
+                (Some(now), Some(s)) => (now - s) / 1000,
+                _ => 0,
+            };
+            TodayActive {
+                app: ar.app_name,
+                started_at: ar.started_at,
+                elapsed_s,
+                cat: normalize_cat(&ar.category),
```
Akarsh-Hegde: **`active.cat` applies `fm_parse_error`/`fm_skip` normalization the Node route doesn't.**

`normalize_cat()` maps `fm_parse_error`/`fm_skip` → `idle_personal`. The TS active block instead does a plain `cat: (ar.category as string) || 'idle_personal'` — **no** fm mapping. So an active session classified `fm_parse_error` shows as `idle_personal` natively but `fm_parse_error` in the browser, breaking the byte-identical claim (and changing the active card's category label/dot).

Note the foreground `sessions` *do* apply the fm mapping in TS, so `normalize_cat` is correct there — just not for `active`.

meridian-core/src/readers/today/mod.rs
```diff
@@ -0,0 +1,517 @@
...
+                app: r.app_name.clone(),
+                started_at: r.started_at.clone(),
+                dur: r.duration_s,
+                cat: normalize_cat(&r.category),
+                titles,
+                explain: r.category_explanation.clone(),
+                routing: r.routing.clone(),
+                session_type: r.session_type.clone(),
+                task_key: r.task_key.clone(),
+                candidates: r.task_key.clone().into_iter().collect(),
```
Akarsh-Hegde, line 106: **Empty-string optional columns are kept as `""` instead of coerced to `null` (minor byte divergence).**

The TS route coerces falsy strings to null: `(r.task_key as string) || null`, `(r.category_explanation) || null`, `(r.routing) || null`, `(r.session_type) || null`, `link_method`, `summary` — and `candidates = r.task_key ? [r.task_key] : []`. This Rust port keeps `Some("")`: `explain`/`routing`/`session_type`/`task_key`/`link_method`/`summary` stay `""`, and `candidates` becomes `[""]` (`r.task_key.clone().into_iter().collect()`).

Only bites when those columns actually hold `""` rather than `NULL` (uncommon in practice — the classifier writes NULL), so low severity, but it's a JSON-shape divergence from the stated byte-identical goal. A `.filter(|s| !s.is_empty())` on each (and on the candidates push) would match exactly.

Akarsh-Hegde (commented):
Code review (high effort, recall-biased) — focused on whether the Rust ports really match the TS routes they claim to be byte-identical to, plus the Tauri lifecycle. 4 findings, all on the `meridian-core`/tray fold. None block the foundation work; #1 is the only user-visible functional one.

Akarsh-Hegde:
## Update — Stage 1 route ports (since this PR opened)

The read tier of the Next-fold is now essentially complete. New since the initial push:

**Routes ported to Rust** (each consumer swapped to the dual-path `load()`; `/api` kept until the export cutover):
- `coding-agents`, `worklogs`, `tasks` (+ `hygiene`), `active` (reshaped view), `settings`, `integrations`, `triage` (+ `parents`), `version`.
- **DB reads → `meridian-core`**; **file/env/process/external → tray commands** (settings.json, `.env`, `launchctl`-adjacent, npm registry, shell-out to `meridian ticket-parents`).
- **Settings reader moved into `meridian-core`** (daemon re-exports it — `config::{RuntimeSettings, load_runtime_settings}` unchanged) so the schema/path has one definition, not a third copy.

**Tests:** `hygiene::parse_issues` unit tests (6), in-memory SQLite reader integration tests (`coding_agents` union, `tasks` autonomous math), `version::is_newer` unit tests. `cargo test` green.

**Observability:** every reader has command+core spans, per-query `debug_span`, row-count `debug!`, `info!` summaries, `warn!` on errors → OpenObserve under the tray `otel` feature. Added `OBSERVABILITY.md`.

**Docs/standards:** added the **"Porting a dashboard route to Rust" playbook to `CLAUDE.md`** (placement / byte-for-byte / docs / tracing / tests), and backfilled all reader module docs to match.

**Bug fixes found along the way:**
- Tray menu rebuild dropped items (single source of truth now).
- Tauri capability scoped to `main` only → dashboard/setup windows couldn't `invoke` (now covers all three).
- Dev-watcher thrash on `ui/.next` churn (`.taurignore`).
- Tray `tokio` missing the `process` feature → standalone (no-otel) build failed (`parents.rs`).

**Still to port:** the writes/actions tier (mutations) and the 3 SSE streams → Tauri events.

Akarsh-Hegde:
## `/code-review` findings (high-effort, recall-biased)

6 of 7 finder angles completed (the cross-file tracer 500'd twice; its scope — command registration, invoke-arg matching, capability/window coverage — was cross-checked by the other angles and came back clean: every command is registered, arg names match, and `default.json` covers `main`/`dashboard`/`setup`). Ranked most-severe first. The ports are largely faithful — angle A verified `intervals`/`tasks` autonomous math, `coding_agents`, `worklogs`, `version`, `settings` coercion, and the `bridge.ts` swaps as correct. The real items:

### Correctness / behaviour
1. **Dropped the routes' `try/catch` → empty-200 contract** (`meridian-core/src/{worklogs,tasks,week,active,today}.rs`). Each TS route wrapped its body in `try { … } catch { return 200 {items:[],…} }`; the Rust commands `map_err → Err`, so `load()` **rejects** instead of resolving with an empty shape. On a transient DB error (locked WAL, missing column on an old DB) consumers hit `.catch` and may leave a **permanently blank panel with stale state** (the `.then` that calls `setItems([])` never runs). Decide: do we want the resolve-empty contract preserved, or is reject-on-error fine?
2. **`integrations.rs` reads `.env` file text, but the daemon reads the *process* env** (`std::env::var`, populated by dotenvy + the launchd plist + shell `export`s). A provider configured via the plist or `export` (not in any `.env`) shows **"not connected" in the app while syncs actually run**. Also: `active_env_path()` walks up from the *tray's* cwd (≠ daemon WorkingDirectory on a source run); the daemon accepts `JIRA_URL` but the tray only checks `JIRA_BASE_URL`; `parse_env` doesn't strip quotes/`export` like dotenvy. The deeper fix is to have the tray ask the daemon for resolved-provider status rather than re-parse `.env`.
3. **Three-way drift in the reason-code → hint/fix table.** The daemon engine's `TriageReason::{hint,fix}` (the authoritative classifier), `meridian-core/hygiene.rs`, and `meridian-core/triage.rs` each hand-maintain the same mapping — and they've **already drifted** (e.g. engine "No description — I'll have nothing to match…" vs ports "…nothing to match…"; ThinDescription "add a bit of detail" vs "a little detail" vs no suffix). "Kept in sync by hand" doesn't hold; the engine should expose one lookup the ports consume.
4. **`today.rs` active-session divergences vs `/api/today`:** (a) line ~307 — empty `window_titles` falls back to `[app_name]`, but the route returns `[]` for the active session (only foreground sessions get the title fallback); (b) line ~316 — active `category` runs through `normalize_cat` (remaps `fm_parse_error`/`fm_skip` → `idle_personal`), but the route's active branch does **not** remap. Same DB → different active card in app vs browser. Low frequency, real.
5. **`get_triage` `now` format** (`commands.rs` ~47 / `triage.rs` ~178): uses `Utc::now().to_rfc3339()` (`+00:00`, sub-milli) while the route + sibling `get_tasks` use `…Z` millis. The `snoozed_until <= ?` compare is lexicographic, so a tie-boundary snooze can flip. Low impact (triage GET has no consumer yet) but an internal inconsistency with `get_tasks`.
6. **`week.rs` ~99:** active-session category uses `unwrap_or` (only `None` → `idle_personal`); the route's `|| 'idle_personal'` also maps `""` → `idle_personal`. An empty-string category lands today's live hours under a blank `cats` key. Edge.
7. **`settings.rs` ~75:** `expand_tilde` only handles a leading `~/`; the old `shellexpand::tilde` also expanded bare `~` / `~user`. A `MERIDIAN_SETTINGS_PATH` set to a bare-tilde form now resolves literally → falls through to defaults. Edge (test/non-standard installs).

### Duplication / maintainability
8. **`ms()` RFC3339→epoch-ms is triplicated** in `week.rs`/`active.rs`/`today.rs`, re-deriving the private `intervals::parse_ms`. Make `parse_ms`/`ms_to_iso` `pub` and reuse. Related: the **settings schema** is hand-listed in both `meridian-core/settings.rs` and `ui/lib/settings.ts` with nothing enforcing parity; and `parents.rs::meridian_bin()` re-implements `selectMeridianBinary` (the launchd native-first ordering is test-guarded only on the TS side).

### Efficiency / cleanup (low)
9. **`tasks.rs`:** 7 independent reads run sequentially (`try_join!` them); `intersect_seconds(&agent, presence)` re-normalizes the full presence set **once per ticket**; `*_by_task_rows` builds an intermediate `BTreeMap` and `.clone()`s every `SessionRow`; today's rows are read twice (today range ⊂ week range). All modest (SQLite-bound) but free wins.
10. **`TasksView.tsx` ~1004:** leftover bare `{ … }` block after removing the `if (r.ok)` wrapper — dead scope, dedent it. (`version.rs` `reqwest::Client::new()` per call is minor given the 1h cache.)

Most are faithful-divergence/edge; **#1, #2, #3** are the ones worth a real decision. None block the fold; happy to fix any subset.

adityaharishch (commented):
## Code Review — PR #298

Seven findings from an automated multi-angle review (line-by-line scan, removed-behavior audit, cross-file trace, reuse/simplification, efficiency/altitude, conventions). Ranked most-severe first; inline comments follow.

### Summary

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **Bug** | `meridian-core/src/intervals.rs` | `intersect_seconds` double-counts overlaps on equal endpoints |
| 2 | **Bug (latent)** | `tray/src-tauri/src/commands.rs` | `get_tasks` and `get_triage` produce different RFC 3339 formats for `now` |
| 3 | **Bug** | `tray/src-tauri/src/integrations.rs` | `parse_env` stores `KEY=""` as non-empty, reporting integration as connected |
| 4 | **Bug (deployment)** | `tray/src-tauri/src/commands.rs` | `meridian_db_path()` skips `.env` — wrong path on bundle install with non-default `MERIDIAN_DB` |
| 5 | **Latent trap** | `ui/lib/bridge.ts` | Browser fallback in `load()` silently drops the `args` parameter |
| 6 | **Convention** | `meridian-core/src/today.rs` | 521 lines — exceeds the 500-line CLAUDE.md limit |
| 7 | **Convention** | `OBSERVABILITY.md` | Doc file created without an explicit request (CLAUDE.md: *NEVER create documentation files unless explicitly requested*) |

meridian-core/src/util/intervals.rs
```diff
@@ -0,0 +1,245 @@
...
+        let lo = a[i].0.max(b[j].0);
+        let hi = a[i].1.min(b[j].1);
+        if hi > lo {
+            total_ms += hi - lo;
+        }
+        // advance whichever interval ends first
+        if a[i].1 < b[j].1 {
+            i += 1;
+        } else {
+            j += 1;
```
adityaharishch, line 149: **Bug — double-counts overlap when interval endpoints are equal**

When `a[i].1 == b[j].1` the `else` branch advances only `j`, leaving `i` pointing at the same interval. On the next iteration, `a[i]` finds a fresh overlap with `b[j+1]`, double-counting the portion that was already counted.

Concrete trace:
```
a = [(0ms, 100ms)]
b = [(50ms, 100ms), (90ms, 150ms)]

iter 1: lo=50, hi=100, adds 50 ms. Tie → j+=1 (i stays at 0)
iter 2: lo=90, hi=100, adds 10 ms again (already counted!)
returns 60 s — correct answer is 50 s
```

Fix — advance both pointers on a tie:
```rust
if a[i].1 < b[j].1 {
    i += 1;
} else if b[j].1 < a[i].1 {
    j += 1;
} else {
    i += 1;
    j += 1; // equal endpoints: both intervals are exhausted
}
```

No existing test covers the equal-endpoint case.

tray/src-tauri/src/commands.rs
```diff
@@ -32,6 +32,79 @@ pub async fn open_worklogs(app: tauri::AppHandle) -> Result<(), String> {
...
+#[tauri::command]
+#[tracing::instrument(skip(pool))]
+pub async fn get_triage(
+    pool: State<'_, Option<meridian_core::SqlitePool>>,
+) -> Result<meridian_core::triage::TriageResponse, String> {
+    let Some(pool) = pool.inner() else {
+        return Err("meridian.db is not open yet".to_string());
+    };
+    let now = chrono::Utc::now().to_rfc3339();
+    meridian_core::triage::get_triage(pool, &now)
```
adityaharishch: **Bug (latent) — RFC 3339 format inconsistency with `get_tasks`**

This command uses `Utc::now().to_rfc3339()` which produces `+00:00`-suffixed timestamps (e.g. `2026-06-17T12:00:00+00:00`). The `get_tasks` command at the bottom of this file uses `to_rfc3339_opts(SecondsFormat::Millis, true)` which produces `Z`-suffixed timestamps (e.g. `2026-06-17T12:00:00.000Z`).

The `tasks.rs` snooze check is a raw string comparison: `s.as_str() > now_iso`. Since `+` (ASCII 43) < `Z` (ASCII 90), any `snoozed_until` value written by a Rust path using `to_rfc3339()` would compare wrong against a `get_tasks` `now_iso`. The current TypeScript write path always uses `Z`, so nothing is broken today — but this is a trip-wire for the first Rust write path.

Fix: use a single consistent format everywhere. `to_rfc3339_opts(SecondsFormat::Secs, true)` gives `Z`-suffixed without milliseconds, matching the TS writer.

tray/src-tauri/src/commands/integrations.rs
```diff
@@ -0,0 +1,130 @@
...
+        if trimmed.is_empty() || trimmed.starts_with('#') {
+            continue;
+        }
+        if let Some(eq) = trimmed.find('=') {
+            if eq < 1 {
+                continue;
+            }
+            let key = trimmed[..eq].trim();
+            let val = trimmed[eq + 1..].trim();
+            if !key.is_empty() && !val.is_empty() {
```
adityaharishch, line 98: **Bug — quoted empty values falsely reported as connected**

`.trim()` strips whitespace but not surrounding quote characters. A `.env` value like `LINEAR_API_KEY=""` (valid bash dotenv syntax for "unset this key") produces `val = "\"\""` (two chars), which passes `!val.is_empty()` and none of the `is_set` placeholder checks (`"your-"`, `"_your_"`, `"-here"`). Result: `linear: true` — the integration shows as connected when the user intentionally left it blank.

This was also present in the old TS route — but porting is a good moment to fix it:
```rust
let val = {
    let v = trimmed[eq + 1..].trim();
    // strip one matching pair of surrounding ' or " characters
    if (v.starts_with('"') && v.ends_with('"')) || (v.starts_with('\'') && v.ends_with('\'')) {
        &v[1..v.len()-1]
    } else { v }
};
if !key.is_empty() && !val.is_empty() {
```

tray/src-tauri/src/commands.rs
```diff
@@ -100,3 +173,137 @@ fn uid_str() -> String {
         .map(|s| s.trim().to_string())
         .unwrap_or_else(|| "501".to_string())
 }
+
+/// Resolve meridian.db: `MERIDIAN_DB` env, else `~/.meridian/meridian.db`.
+/// (Production should reuse `meridian::config` for full .env + `~` handling.)
+pub(crate) fn meridian_db_path() -> String {
+    if let Ok(p) = std::env::var("MERIDIAN_DB") {
```
adityaharishch: **Bug (deployment) — skips `.env` lookup; wrong DB path on bundle install with non-default `MERIDIAN_DB`**

On a bundle install the daemon's `WorkingDirectory` is `~/.meridian/app` and it loads `~/.meridian/app/.env` via dotenvy. If `MERIDIAN_DB` is set only in that file (not exported to the shell), `std::env::var("MERIDIAN_DB")` here returns `Err` and the tray falls back to `$HOME/.meridian/meridian.db` — a different path from what the daemon uses.

The function's own doc comment acknowledges this: *"(Production should reuse `meridian::config` for full .env + `~` handling.)"* — that intent should be a follow-up ticket before this ships to bundle installs. The `integrations.rs` companion in this same PR correctly uses `active_env_path()` + `parse_env()` for its env lookups; the same pattern should be applied here.

ui/lib/bridge.ts
```diff
@@ -0,0 +1,52 @@
...
+export async function invoke<T = unknown>(cmd: string, args?: Record<string, unknown>): Promise<T> {
+  const t = tauri()
+  if (!t) throw new Error(`invoke('${cmd}') requires the Meridian app — the Tauri bridge is unavailable in a plain browser.`)
+  return t.core.invoke(cmd, args) as Promise<T>
+}
+
+/** Dual-path data load: the Rust command inside the app, else the /api route in
+ *  a browser. Both return the same shape (commands are byte-identical ports). */
+export async function load<T = unknown>(
+  apiPath: string,
```
adityaharishch, line 48: **Latent trap — `args` silently ignored in the browser fallback**

The `fetch(apiPath)` path discards `args` entirely. No current caller is broken — every caller that passes `args` also encodes them into `apiPath` as query params. But the migration guide (step 6 in CLAUDE.md's dashboard-fold section) says to replace `fetch('/api/x')` with `load('/api/x', 'command', args)`, and the function signature gives no indication that the fetch path ignores `args`.

The next migrated route following the pattern with body-style args will silently send the wrong request in browser mode while working correctly in the Tauri window.

Suggested guard:
```ts
const r = await fetch(apiPath)  // args must be encoded in apiPath for browser mode
```
Or, more robustly, serialize `args` into search params in the fallback:
```ts
const url = new URL(apiPath, location.origin)
if (args) Object.entries(args).forEach(([k, v]) => url.searchParams.set(k, String(v)))
const r = await fetch(url)
```

meridian-core/src/readers/today/mod.rs
```diff
@@ -0,0 +1,535 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```
adityaharishch, line 1: **Convention — file is 521 lines, exceeds the 500-line limit (CLAUDE.md)**

> *Keep files under 500 lines; split when a file grows beyond that.*

The `ms()` helper (parse RFC3339 → epoch ms) is duplicated here and in `active.rs` and `week.rs`; moving it to `intervals.rs` or `date.rs` as a shared utility would shrink the file and eliminate three independent definitions.

OBSERVABILITY.md
```diff
@@ -0,0 +1,186 @@
+# Observability — tracing the DB → UI flow in OpenObserve
```
adityaharishch: **Convention — documentation file created without an explicit request (CLAUDE.md)**

> *NEVER create documentation files unless explicitly requested.*

If the observability setup notes are useful, they belong in the existing `CLAUDE.md` (under a dashboard-fold or observability section) rather than a standalone Markdown file. Alternatively, confirm this was explicitly requested and document that in the PR description.

Akarsh-Hegde:
## Resolution — adityaharishch code-review findings

All 7 findings addressed in commit **ea19768** (+ **cf731d6** for CI):

| # | File | Finding | Fix |
|---|------|---------|-----|
| **#1** | `intervals.rs` | `intersect_seconds` double-advance on equal endpoints | Three-branch advance: `<` → i, `>` → j, `==` → both. Added dedicated test. |
| **#2** | `commands.rs` | `get_triage` used `+00:00` sub-milli vs `get_tasks`'s `Z` millis format | Changed to `to_rfc3339_opts(Millis, true)` — now consistent with every other command. |
| **#3** | `integrations.rs` | `parse_env` stored `KEY=""` as non-empty ("connected") | Strip surrounding quotes before the non-empty check so `KEY=""` and `KEY=''` register as unset. |
| **#4** | `commands.rs` | `meridian_db_path()` skipped `.env` (wrong path on bundle install with custom `MERIDIAN_DB`) | Added `.env` fallback: bundle `~/.meridian/app/.env` → cwd walk, same logic as `active_env_path`. |
| **#5** | `bridge.ts` | Browser `load()` fallback silently dropped `args` | Append `args` as `?key=value` query params in the `fetch()` path so e.g. `get_worklogs({day})` works in a plain browser. |
| **#6** | `today.rs` | 521 lines — exceeded the 500-line CLAUDE.md limit | Extracted response types + DB row shapes + helpers to `today_types.rs`; `today.rs` is now **373 lines**. Public API unchanged (`meridian_core::today::TodayResponse` etc.). |
| **#7** | `OBSERVABILITY.md` | Created without an explicit request | Deleted. |

**CI fix (cf731d6):** The `cargo fmt --check` failure was in `src/telemetry_spool/` and `src/pm_worklog/` — files that arrived on `main` after this branch was cut, not in our changes. Merged main and ran `cargo fmt` to fix them.

Akarsh-Hegde:
> ### Correctness / behaviour
> 1. **Dropped the routes' `try/catch` → empty-200 contract** (`meridian-core/src/{worklogs,tasks,week,active,today}.rs`). Each TS route wrapped its body in `try { … } catch { return 200 {items:[],…} }`; the Rust commands `map_err → Err`, so `load()` **rejects** instead of resolving with an empty shape. On a transient DB error (locked WAL, missing column on an old DB) consumers hit `.catch` and may leave a **permanently blank panel with stale state** (the `.then` that calls `setItems([])` never runs). Decide: do we want the resolve-empty contract preserved, or is reject-on-error fine?

My recommendation: preserve the resolve-empty contract for now (it's a one-line change per command, keeps
  the UI behavior identical to the TS routes, and the WAL lock scenario is transient). Surface real errors
  properly as a separate piece of work when you wire up component-level error states.

tray/src-tauri/src/commands.rs
Akarsh-Hegde: ⏺ Update(tray/src-tauri/src/commands.rs)
  ⎿  Added 1 line, removed 1 line
      44      let Some(pool) = pool.inner() else {
      45          return Err("meridian.db is not open yet".to_string());
      46      };
      47 -    let now = chrono::Utc::now().to_rfc3339();                                               
      47 +    let now = chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Millis, true);       
      48      meridian_core::triage::get_triage(pool, &now)
      49          .await
      50          .map_err(|e| {
      
      
      should this be pulled out as a function export?

tray/src-tauri/src/commands/integrations.rs
Akarsh-Hegde: Now fix #3 in integrations.rs (empty-string values from KEY=""):

⏺ Update(tray/src-tauri/src/integrations.rs)
  ⎿  Added 3 lines, removed 1 line
      67                  continue;
      68              }
      69              let key = trimmed[..eq].trim();
      70 -            let val = trimmed[eq + 1..].trim();                                              
      70 +            // Strip surrounding quotes so KEY="" and KEY='' register as empty.              
      71 +            let raw = trimmed[eq + 1..].trim();                                              
      72 +            let val = raw.trim_matches('"').trim_matches('\'').trim();                       
      73              if !key.is_empty() && !val.is_empty() {
      74                  out.insert(key.to_string(), val.to_string());
      75              }

ui/lib/bridge.ts
Akarsh-Hegde: ⏺ Now fix #5 in bridge.ts — pass args as URL query params in browser fallback:

⏺ Update(ui/lib/bridge.ts)
  ⎿  Added 11 lines, removed 1 line
      46  ): Promise<T> {
      47    const t = tauri()
      48    if (t) return t.core.invoke(command, args) as Promise<T>
      49 -  const r = await fetch(apiPath)                                                             
      49 +  // Browser fallback: append args as query params so /api routes receive them.              
      50 +  const qs = args                                                                            
      51 +    ? new URLSearchParams(                                                                   
      52 +        Object.fromEntries(                                                                  
      53 +          Object.entries(args)                                                               
      54 +            .filter(([, v]) => v != null)                                                    
      55 +            .map(([k, v]) => [k, String(v)])                                                 
      56 +        )                                                                                    
      57 +      ).toString()                                                                           
      58 +    : ''                                                                                     
      59 +  const r = await fetch(qs ? `${apiPath}?${qs}` : apiPath)                                   
      60    if (!r.ok) throw new Error(`${apiPath} → ${r.status}`)
      61    return r.json() as Promise<T>

Akarsh-Hegde:
## 🔍 Code review — `spike/meridian-core` (high-effort, workflow-backed)

Reviewed 218 changed files across 8 finder angles; every candidate was checked by an independent verifier (28 candidates → 20 kept, 8 refuted). Below are the 10 reported findings, grouped by severity. I'll fix the confirmed **correctness divergences** in a follow-up commit and report the delta.

### 🔴 Correctness divergences from the original `/api` routes (will fix)

1. **`tray/src-tauri/src/poll/live.rs:120` — log-tail drops complete lines on a split multibyte char.**
   `read_new_lines` does `read_to_string` over the whole `offset..EOF` region *including the not-yet-flushed trailing line*. If the daemon is mid-writing a non-ASCII byte (emoji/accented path/unicode in an error), the decode fails, returns `Vec::new()`, and leaves `offset` unadvanced — so the complete lines written just before are silently dropped from the live Logs view until a later, fully-flushed read. The original Node `createReadStream` tail split on `\n` at the byte level and never had this. **Fix:** read bytes, find the last `\n` at the byte level, decode only the complete prefix.

2. **`tray/src-tauri/src/commands/tasks.rs:44` — leaked `tasks-sync` process on timeout.**
   The deleted `/api/tasks/sync` route called `child.kill()` when its 30s timer fired. The port wraps `Command::output()` in `tokio::time::timeout` but never sets `kill_on_drop(true)`, so on timeout the child keeps running to completion *after* the UI shows "timed out". A user retry spawns a second overlapping sync — concurrent `meridian.db` writes + PM-tracker API hammering, and the leaked run can still mutate the board. **Fix:** `.kill_on_drop(true)`.

3. **`meridian-core/src/notifications.rs:60` — quiet-hours parser is more lenient than the route.**
   `hhmm_to_minutes` uses `split_once(':') + i64::parse`, accepting values the original strict regex `/^(\d{1,2}):(\d{2})$/` rejected (`"8:5"`, `"+8:00"`, `"8:00:00"`). A malformed-but-parseable `quiet_hours_start` now engages quiet hours and suppresses notifications the pre-fold dashboard would have fired (fail-open). **Fix:** strict parse matching the regex (digit-count + ASCII-digit checks) + regression test.

4. **`tray/src-tauri/src/commands/health.rs:65` — divergent DB-path resolver.**
   `check_database` reads `$MERIDIAN_DB_PATH` (wrong var — the real one is `MERIDIAN_DB`) or the hardcoded default, ignoring the `~/.meridian/.env` / `MERIDIAN_DB` resolution chain the rest of the tray + daemon use via `install::meridian_db_path()`. On an installed system with a non-default DB path, the health pane reports "Database not found" while the daemon writes happily to the real DB. **Fix:** call the existing `install::meridian_db_path()` resolver.

### 🟡 Pre-existing / intentional divergences (NOT fixing — documented)

5. **`meridian-core/src/readers/week.rs:47` — UTC/local week-bucketing.** Naive tz-less day bounds string-compared against `Z`-suffixed UTC `started_at`; late-evening sessions land in the wrong day column for non-UTC users. This is an **explicitly documented byte-for-byte replication** of the existing `/api/week` route (module doc says do-not-fix). Re-exposed, not introduced — flagging as a pre-existing bug to fix at the source level, separately.
6. **`meridian-core/src/capture.rs:58` (PLAUSIBLE)** — `text_source` coerces any out-of-contract value to `'ocr'` silently. Intentional write-boundary normalization; could add a `warn!` if a third source ever appears.
7. **`meridian-core/src/readers/plan.rs:441` (PLAUSIBLE)** — tiebreak switched from JS `localeCompare` to Rust byte-ordinal `str::cmp`; commented intentional divergence. Low severity (different equal-score suggestion order for mixed-case/non-ASCII keys).

### 🔵 Cleanup / efficiency (deferring — behavior-preserving, follow-up)

8. **`tray/src-tauri/src/poll/refresh.rs:152` (+191, `today/mod.rs:317`, `week.rs:41`)** — the 60s poll loop runs the full `get_today` dashboard reader (column probe + full-day scan + unfiltered `pm_tasks` scan + interval math) then discards all but `focus_s`/`switch_count`/3 category sums. Continuous wasted CPU/DB I/O on a battery-powered idle machine; a dedicated aggregate query would be far cheaper.
9. **File-size hard-rule violations (CLAUDE.md: "Keep files under 500 lines"):** `tray/src-tauri/src/mlx_server.rs` (1183), `meridian-core/src/readers/plan.rs`, `tray/src-tauri/src/commands/integrations.rs`, `meridian-core/src/capture.rs`, `ui/components/views/TasksView.tsx`. Splitting an 1183-line module on a near-merge PR is risky — recommend a dedicated follow-up.
10. **`meridian-core/src/readers/tasks.rs:360` (+`triage.rs:189`) (PLAUSIBLE)** — inline `sqlite_master` table-existence probes duplicate `readers/plan.rs`'s `table_exists()`; extract to a shared readers util.

---
*8 candidates were refuted in verification (DRY/helper-extraction suggestions that didn't hold up as defects): bridge `load`/`invoke` overlap, dashboard pool-guard boilerplate, settings unused-param, `emit_notices`/`emit_banners` similarity, `.env` parser duplication, `capture.rs` `is_missing_table` style.*

Akarsh-Hegde:
## ✅ Review follow-up — 4 fixed, 6 deferred

Pushed `b54ec04` (`fix(fold): correct four Next-fold route divergences from code review`). Full pre-push suite green: fmt + clippy `-D warnings` + UI build + UI tests + `cargo test` + security audit.

### Fixed (the 4 confirmed correctness divergences)

| # | File | Fix |
|---|------|-----|
| 1 | `poll/live.rs` | Log tail now splits at the **byte** level (`read_to_end` → last `\n` → lossy decode of the complete prefix). No longer drops complete lines when the writer is mid-flush on a multibyte char. |
| 2 | `commands/tasks.rs` | `.kill_on_drop(true)` on the `tasks-sync` child — a 30s timeout now actually kills it (matching the deleted route's `child.kill()`), no leaked board-mutating process. |
| 3 | `notifications.rs` | `hhmm_to_minutes` is now strict to the route regex `/^(\d{1,2}):(\d{2})$/` (rejects `"8:5"`, `"+8:00"`, `"8:00:00"`). Quiet-hours silence/notify decision matches the pre-fold dashboard again. **Pinned with a regression test.** |
| 4 | `commands/health.rs` | DB path resolved via `install::meridian_db_path()` (the `MERIDIAN_DB` / `~/.meridian/.env` chain) instead of the divergent inline lookup of a non-existent `MERIDIAN_DB_PATH` + hardcoded default. Health pane no longer false-reports "Database not found" on installs with a custom DB path. |

### Deferred (6) — with reasons, **not** resolved in this commit

- **#5 `week.rs` UTC/local bucketing** — intentionally **not** patched. The module doc declares it a byte-for-byte faithful replica of `/api/week` (do-not-fix); patching here would diverge from the route. Track as a pre-existing source-level bug to fix in both places together.
- **#6 `capture.rs` text_source coercion** & **#7 `plan.rs` tiebreak** (both PLAUSIBLE) — commented, intentional divergences; left as-is.
- **#8 `refresh.rs` discarded poll work** — real efficiency win but a behavior-preserving refactor (dedicated aggregate query); better as its own change.
- **#9 file-size hard-rule violations** (`mlx_server.rs` 1183 lines + 4 others) — splitting an 1183-line module on a near-merge PR is risky; recommend a dedicated follow-up PR.
- **#10 duplicated `sqlite_master` probe** (PLAUSIBLE) — extract `table_exists()` to a shared readers util; low-risk DRY cleanup for a follow-up.

Deferred items #8–#10 are behavior-preserving cleanups; #5–#7 are pre-existing/intentional. None block this PR's merge.

PR #299, opened 2026-06-17
feat(observability): durable telemetry spool + full worklog lineage trace

adityaharishch:
Stacks on #297 (classifier observability). Retarget to `main` once #297 merges.
All three parts are verified live in OpenObserve.

## 1. Durable + portable telemetry (in-process spool + shipper)
- Rust `SpoolClient` (impl `opentelemetry_http::HttpClient`) + Python `SpoolSpanExporter`/`SpoolLogExporter` write every OTLP batch atomically to `~/.meridian/telemetry/pending` — **gated on `otlp_enabled` only**, so capture works even before OO credentials exist.
- Background **shipper** (`src/telemetry_spool/shipper.rs`) ships pending → OO oldest-first, moves to `sent/` on 2xx, **stops on failure and replays on reconnect** (no loss during OO downtime). Retention + pending-size cap, with a `warn!` on any drop (no silent truncation).
- CLI: `meridian telemetry status|export|import` — the spool doubles as a **portable bundle** to replay a user's traces into your own OpenObserve for debugging.

## 2. Full worklog lineage trace
- Migration `043`: `app_sessions.classify_traceparent` (the classification trace, distinct from the formation traceparent in `010`). The MLX server stamps each `classify_session` traceparent onto its result; Rust persists it (`COALESCE`, never nulled).
- `worklog_draft` now emits `contributing_sessions → session` child spans. Each session reconstructs its classification **verdict inline** (`classification_verdict` / `reasoning` / `category` / `session_summary` / `dimensions`, from stored data) **plus** OTel span Links to the full `classify_session` and formation traces for raw LLM I/O.
- The Python synth now **continues** the `worklog_draft` trace instead of emitting a separate linked one, so `worklog_input` (exact wire prompt + `system_prompt`/`user_message` child spans), `synth_inference`, and `worklog_output` nest **inside** the worklog trace — classifier-grade input/output, one navigable tree.

Verified tree (one `worklog_draft` trace):
```
worklog_draft
├─ synthesise_worklog
│  ├─ worklog_input → system_prompt, user_message
│  ├─ synth_inference
│  └─ worklog_output → summary, bullets, next_steps, reasoning, risk_flags
└─ contributing_sessions
   └─ session (×N)
      ├─ classification_verdict / reasoning / category / session_summary / dimensions
      └─ 🔗 link → full classify_session trace
```

## 3. Fix: one-shot CLI emitted no traces
`meridian pm-worklog` never initialised observability, so it produced no spans and propagated no traceparent (the synth landed as an orphan root). It now inits + flushes observability, emitting the same trace the daemon does.

## Notes
- Per-session classification is shown as the **verdict reconstructed inline + a link** to the full classifier trace; replicating the raw classifier LLM I/O inline was deferred (it would require storing/duplicating the full wire-prompt per session).
- Excludes unrelated working-tree changes (daily-plan, PII-guardrail).


src/telemetry_spool/shipper.rs
```diff
@@ -0,0 +1,468 @@
...
+                }
+            }
+            Err(e) => {
+                tracing::warn!(
+                    file = %filename,
+                    signal,
+                    error = %e,
+                    "telemetry ship failed — stopping this tick, files remain"
+                );
+                // Stop on first failure — OO may be down.
```
Akarsh-Hegde: **[correctness, high] Head-of-line block: one poison file stalls all telemetry.** `run_tick` `break`s on the first `ship_one` error ("OO may be down"). That's right for a transient 5xx/network outage, but a *permanent* failure — a 400 (malformed/truncated protobuf, e.g. from the missing-fsync path), 413 (batch too large), or 401/403 (bad creds) — fails every tick forever at the head of the oldest-first queue, so **no newer file is ever shipped**. Worse, `enforce_pending_cap` then evicts by *age*, dropping the good newer records while the poison head survives. Distinguish retryable (5xx/network) from terminal (4xx) and quarantine/skip poison records instead of head-of-line blocking.

src/telemetry_spool/spool_client.rs
```diff
@@ -0,0 +1,158 @@
...
+            // will have already trimmed the backlog).
+            tracing::warn!(
+                signal,
+                error = %e,
+                "telemetry spool write failed — payload dropped"
+            );
+        }
+
+        // Return a synthetic 200 so the SDK considers export successful.
+        let response = Response::builder()
```
Akarsh-Hegde, line 119: **[correctness] Synthetic 200 on write failure → silent data loss.** When `write_pending` fails, `send` only `warn!`s and still returns HTTP 200, so the SDK marks the export successful and discards the batch. The comment assumes failure ⇒ disk full ⇒ cap already trimmed — but a write can fail on a transient FS error, a permissions issue, or a full disk where the cap *can't* help (it's the *new* write that failed, not old files). The telemetry is gone with no retry and no error surfaced. Consider returning an error on non-disk-full failures so the SDK retries, or at least distinguishing the causes.

src/telemetry_spool/writer.rs
```diff
@@ -0,0 +1,181 @@
...
+pub fn write_pending(base: &Path, signal: &str, bytes: &[u8]) -> Result<PathBuf> {
+    let pending = pending_dir(base);
+    std::fs::create_dir_all(&pending)
+        .with_context(|| format!("create pending dir {}", pending.display()))?;
+
+    let filename = make_filename(signal);
+    let final_path = pending.join(&filename);
+    let tmp_path = pending.join(format!("{filename}.tmp"));
+
+    std::fs::write(&tmp_path, bytes)
```
Akarsh-Hegde: **[correctness/durability] tmp-then-rename without `fsync` isn't crash-durable.** `std::fs::write(tmp)` then `rename` gives *atomicity/visibility* (good — a partial write is never seen), but with no `File::sync_all()` on the tmp file and no fsync of `pending/`, a power loss / kernel panic can make the rename (metadata) durable while the data blocks are still in page cache — leaving a zero-length or truncated `.otlp` the shipper will POST (→ a 400 that then head-of-line-blocks, see shipper.rs:167). The module name says "durable"; to honor it, `sync_all()` the tmp before rename and fsync the dir after.

src/telemetry_spool/spool_client.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl fmt::Display for SpoolClient {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "SpoolClient({})", self.base_dir.display())
+    }
+}
+
+#[async_trait]
+impl HttpClient for SpoolClient {
+    async fn send(&self, request: Request<Vec<u8>>) -> Result<Response<Bytes>, HttpError> {
```
Akarsh-Hegde, line 60: **[correctness/efficiency] Blocking `std::fs` on the async runtime.** `send` is `async` and is called by the OTel `BatchSpanProcessor` on a tokio worker thread, but `write_pending` does synchronous `create_dir_all` + `write` + `rename`. Under a slow/contended disk (or once `fsync` is added per writer.rs:71) this blocks the worker, stalling every other future on that thread — and `shipper.rs::run_tick` has the same issue (synchronous `read_dir`/`read`/`rename`/`metadata`). Wrap the fs work in `tokio::task::spawn_blocking`.

Akarsh-Hegde (commented):
## Code review — PR #299 (high effort, recall-biased)

Reviewed the scoped diff vs its base (`feat/classifier-observability`) with 6 finder angles partitioned by subsystem (spool / pm_worklog+migration / Rust wiring / Python / cleanup / dashboards+altitude), then **verified the load-bearing spool & worklog findings directly against the code**.

**Good news first:** this PR **fixes the #297 dashboard drilldown bug** — both dashboards now key `trace_filter` on `trace_id` only (no `service_name` pin), so drilldowns work. Migration 043 is the correct next number with the SQL header, additive nullable `ADD COLUMN` (no data loss). Concurrency is sound (atomic rename, `AtomicU64` seq, `SpoolClient` is `Send+Sync`); no double-export; DB writes use `.context()` + idempotent `COALESCE` UPSERTs; the traceparent read-back guards `is_valid()`.

**The new durable spool is where the risk concentrates.** The module promises "durable"/"atomic" delivery but several guarantees don't fully hold — see the inline comments on `shipper.rs`, `writer.rs`, `spool_client.rs`. Highest-severity:
1. **Head-of-line block**: the shipper `break`s on the first ship failure, so a single permanently-rejected (4xx) or corrupt file stalls *all* telemetry delivery; the age-based cap then evicts the *good* newer records.
2. **Migration-lag fragility**: `collect.rs` hard-references `classify_traceparent`, so if migration 043 lags (the migration-031 crash-loop class), *all* worklog drafting breaks, not just lineage.
3. **Not crash-durable**: tmp-then-rename with no `fsync` + a synthetic-200-on-write-failure path → silent loss under power loss / full disk.

**Two findings I couldn't fully verify (flagging for the author to confirm):**
- `post.rs` `find_existing_worklog` dedups on `(task_key, window_start, window_end)`; if a re-draft shifts `window_end` for the same logical `(task, day, cycle)`, the backstop may miss and **double-post to Jira**. Confirm the dedup key matches the worklog upsert key.
- `db_write.rs:77` `classify_traceparent = COALESCE(?, classify_traceparent)` overwrites on **re-classification**, so an already-drafted worklog's lineage Link can point at the old trace while the column points at the new one (lineage drift).

Inline comments below (18). Lower-priority cleanup not inlined: `derive_base_url`/`ship_one` duplicated between `cli.rs` and `shipper.rs` (the cli copy untested); four near-identical span-part closures across the Python files; `record_session_lineage` reconstructs the classifier verdict inline *and* Links the trace (two sources of truth).

src/telemetry_spool/shipper.rs
```diff
@@ -0,0 +1,468 @@
...
+                tracing::warn!(file = %file_path.display(), error = %e, "failed to read spool file — skipping");
+                continue;
+            }
+        };
+
+        match ship_one(&client, &endpoint, &target.auth, bytes).await {
+            Ok(()) => {
+                // Move to sent/
+                let dest = sent.join(&filename);
+                if let Err(e) = std::fs::rename(&file_path, &dest) {
```
Akarsh-Hegde, line 189: **[correctness] Duplicate shipping when the `pending→sent` rename fails (or a crash lands between OO-ack and rename).** After a 2xx, a failed `rename` is only `warn!`'d and the file stays in `pending/`, so it is re-read and **re-POSTed every tick** (e.g. EXDEV if `sent/` is ever on a different mount, a permissions blip, or a crash before the rename). There's no idempotency key or mark-before-ship, so re-delivery is unbounded — duplicate spans/logs in OO. Consider an idempotency marker, or delete-on-success instead of move (with retention handled separately).

src/telemetry_spool/shipper.rs
```diff
@@ -0,0 +1,468 @@
...
+    let Ok(entries) = std::fs::read_dir(pending) else {
+        return Ok(());
+    };
+
+    let mut files: Vec<(u64, u64, PathBuf)> = entries // (micros, size, path)
+        .flatten()
+        .filter_map(|e| {
+            let p = e.path();
+            let name = p.file_name()?.to_str()?.to_string();
+            if !name.ends_with(".otlp") {
```
Akarsh-Hegde, line 349: **[correctness] The 512 MB cap (and the lister) silently ignore `*.otlp.tmp` orphans → unbounded disk growth.** Both `enforce_pending_cap` and `list_pending_oldest_first` filter `name.ends_with(".otlp")`, but `writer.rs:69` writes to `<name>.otlp.tmp`. A crash between `write` and `rename` leaves `.otlp.tmp` orphans that are **never counted toward the cap and never cleaned up**. Under repeated crashes (or an ENOSPC loop) `pending/` grows without bound while the cap reports "under budget" — defeating the disk-fill protection. Sweep/expire stale `.tmp` files (and count them).

src/telemetry_spool/shipper.rs
```diff
@@ -0,0 +1,468 @@
...
+
+    let mut files: Vec<(u64, PathBuf)> = entries
+        .flatten()
+        .filter_map(|e| {
+            let p = e.path();
+            let name = p.file_name()?.to_str()?.to_string();
+            if !name.ends_with(".otlp") {
+                return None;
+            }
+            let micros = crate::telemetry_spool::writer::micros_from_filename(&name).unwrap_or(0);
```
Akarsh-Hegde: **[correctness] Ordering is non-deterministic and `micros=0` poisons it.** The sort key is `micros_from_filename(...).unwrap_or(0)` — the `seq` is dropped, so two files written in the same microsecond (common: traces+logs back-to-back, or a burst) tie and ship in arbitrary `read_dir` order despite the "oldest-first" contract. Worse, any unparseable/renamed filename collapses to `micros=0`, sorting it *permanently first* — so it ships first every tick (and is the first cap-evicted). The filename already carries `seq`; include it in the sort key, and skip rather than zero-key unparseable names. (The unit test passes only because writes happen to be sequential — it doesn't exercise the tie/`0` cases.)

src/telemetry_spool/shipper.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+}
+
+async fn run_tick() -> Result<()> {
+    let base = resolve_telemetry_dir()?;
+    let pending = pending_dir(&base);
+    let sent = sent_dir(&base);
+
+    // Enforce pending cap BEFORE trying to ship so we don't OOM on a long OO outage.
+    enforce_pending_cap(&pending)?;
```
Akarsh-Hegde, line 114: **[correctness] `otlp_enabled` ON but credentials absent → capture with no delivery, then silent drop.** `enforce_pending_cap` runs (line 87) *before* the `resolve_otlp_target()` creds check (line 90). If a user flips the toggle on but never enters `oo_email`/`oo_password`, the `SpoolClient` keeps writing to `pending/` while every tick skips shipping and the cap drops the oldest files — capture that fills to 512 MB and then loses data forever, while the user believes export is "on." Surface a one-time warning when an endpoint/toggle is set but creds resolve to `None`.

src/main.rs
```diff
@@ -624,6 +642,17 @@ async fn main() -> Result<()> {
...
     }
 
+    // 7f. Telemetry spool shipper: drains ~/.meridian/telemetry/pending/ to
+    //     OpenObserve every MERIDIAN_TELEMETRY_SHIP_INTERVAL_S (default 30s).
+    //     Active only when otlp_enabled is true; noop ticks if no OO target is
+    //     configured (e.g. credentials not yet set).
+    {
+        let rx_shipper = shutdown_rx.clone();
+        tokio::spawn(async move {
+            meridian::telemetry_spool::shipper::run_shipper(rx_shipper).await;
```
Akarsh-Hegde: **[correctness, minor] Shutdown stops the shipper *before* the final OTel flush.** `shutdown_tx.send(true)` (which stops this `run_shipper`) fires before `obs_guard.shutdown().await` flushes the batch exporters through `SpoolClient` into `pending/`. So the daemon's own shutdown spans/logs are written to the spool *after* the shipper has exited — they ship only on the next daemon start (lost entirely if the daemon is being uninstalled). Order it the other way: flush first, then stop the shipper (or do one last drain after the flush).

src/pm_worklog/collect.rs
```diff
@@ -61,7 +61,9 @@ pub async fn fetch_session_bundle(
     let rows = sqlx::query(
         "SELECT id, app_name, started_at, ended_at, duration_s, \
                 idle_frame_count, frame_count, window_titles, \
-                session_text, session_text_source, category, session_summary \
+                session_text, session_text_source, category, session_summary, \
+                traceparent, classify_traceparent, \
```
Akarsh-Hegde: **[correctness, high] The read path hard-references `classify_traceparent`, so a lagging migration breaks ALL worklog drafting.** This `SELECT` names `classify_traceparent` directly; if migration 043 hasn't applied (locked DB / checksum mismatch — cf. the migration-031 crash-loop incident), `fetch_session_bundle` fails with `no such column` and `process_task` errors for *every* task — worklog drafting is fully down, not just lineage. The other optional columns are read defensively via `try_get::<Option<_>>`; this one is in the `SELECT` list itself. Consider a column-presence guard (like `today.rs` does for `category_explanation`) so a missing column degrades lineage only.

services/agents/run_task_linker_mlx.py
```diff
@@ -842,24 +990,41 @@ def _classify_one(
...
-            "preview": raw[:500],
-        })
+        pr_span.set_attribute("llm_output", raw)
+        # MLX's OWN generation metadata, co-located with the output it describes —
+        # exact token counts + throughput as REPORTED BY the model (not
+        # re-tokenized by us), under the OTel GenAI semantic conventions. tok/s and
+        # peak memory have no standard attribute name, so they stay custom. None on
+        # the Apple FM path, which has no MLX stats.
+        if gen_stats is not None:
+            _in = int(gen_stats.prompt_tokens)
```
Akarsh-Hegde: **[correctness] New `gen_stats` coercions can crash the whole classify batch.** `int(gen_stats.prompt_tokens)` / `float(prompt_tps/generation_tps/peak_memory)` run outside any try/except, and neither `_classify_one` nor the server's `_classify_all` loop wraps the per-session call. If any MLX `GenerationResponse` stat is `None`/non-numeric (early-truncated stream, an mlx_lm field rename), the `TypeError` propagates out and **fails the entire batch (all `session_ids` → 500)**, not just the one session. #297's per-session inference guard doesn't cover this new code. Wrap the stat extraction (and ideally the per-session call) so one bad stat can't sink the batch.

services/agents/server.py
```diff
@@ -648,53 +719,173 @@ async def synthesise_worklog(req: _SynthWorklogRequest) -> dict:
...
+
+                    inf.set_attribute("attempts", attempts)
+                    if model_name:
+                        inf.set_attribute("gen_ai.request.model", str(model_name))
+                        inf.set_attribute("gen_ai.response.model", str(model_name))
+                    # agno reports the model's OWN token accounting — surface it under
+                    # the OTel GenAI semantic conventions, same as the classifier.
+                    if metrics is not None:
+                        if getattr(metrics, "input_tokens", None):
+                            inf.set_attribute("gen_ai.usage.input_tokens", int(metrics.input_tokens))
```
Akarsh-Hegde: **[correctness] `int(metrics.*)` on agno metrics can turn a successful synth into a 500.** These run after a successful `agent.run`; the truthiness gate skips `None`/`0`, but a *truthy non-int* value (string/list, depending on agno/model-server version) makes `int()` raise — and the exception escapes the threadpool task, so the worklog draft is lost **despite the LLM having succeeded**. Same pattern at L873-877. Coerce defensively (`int(x) if isinstance(x,(int,float)) else None`) or wrap in try/except.

services/agents/server.py
```diff
@@ -648,53 +719,173 @@ async def synthesise_worklog(req: _SynthWorklogRequest) -> dict:
...
+                        if getattr(metrics, "output_tokens", None):
+                            inf.set_attribute("gen_ai.usage.output_tokens", int(metrics.output_tokens))
+                        if getattr(metrics, "total_tokens", None):
+                            inf.set_attribute("gen_ai.usage.total_tokens", int(metrics.total_tokens))
+                        if getattr(metrics, "duration", None):
+                            inf.set_attribute("duration_s", round(float(metrics.duration), 2))
+                    if update is None:
+                        inf.set_status(trace.StatusCode.ERROR, last_detail)
+                        root.set_attribute("is_error", True)
```
Akarsh-Hegde, line 833: **[correctness] `synthesise_worklog` leaves `is_error` unset on unexpected exceptions → the dashboard misses real failures.** `is_error` is set `True` only on the controlled "no JiraUpdate" branch and `False` only on full success. An unexpected raise (agno internals, `_coerce_jira`, `_record_wire_prompt`) ends the span with `is_error` *absent* — and the pm-worklog dashboard's panels filter `is_error='true'` (Failed) and `is_error='false'` (Avg confidence), so the failure lands in **neither** bucket. Set `is_error=True` in an `except`/`finally` so escaped errors are counted.

services/agents/server.py
```diff
@@ -648,53 +719,173 @@ async def synthesise_worklog(req: _SynthWorklogRequest) -> dict:
...
+    # directly INSIDE that one trace — the whole worklog (contributing sessions +
+    # detailed LLM input + output) reads as a single tree, exactly the way
+    # classifier_input / llm_inference / classifier_output sit under classify_session.
+    # Falls back to a fresh root span only when called without a parent (manual curl).
+    _daemon_ctx = observability.extract_parent_context(req.traceparent)
+
+    def _synthesise() -> dict:
+        with tracer.start_as_current_span(
+            "synthesise_worklog",
+            context=_daemon_ctx if _daemon_ctx is not None else _otel_context.Context(),
```
Akarsh-Hegde, line 762: **[altitude] Worklog synth *continues* the daemon trace, re-introducing the noise the classify path was rewritten to avoid.** `context=_daemon_ctx ... else Context()` parents `synthesise_worklog` to the daemon's `worklog_draft` trace — so a Drafted-row drilldown opens a trace mixed with the daemon's ETL/pm_sync spans (and mixed `service_name`s within one `trace_id`). `classify_sessions` deliberately forces a fresh root + Link to get a clean isolated trace (the comment at ~L261-267 explains why). For consistency and a clean drilldown, consider the same fresh-root + Link approach here.

services/agents/observability.py
```diff
@@ -104,6 +218,18 @@ def _load_settings() -> dict[str, object]:
         return {}
 
 
+def _is_otlp_enabled() -> bool:
```
Akarsh-Hegde, line 232: **[cleanup] Dead code: the #297 OTLP-resolution path is orphaned by the spool switch.** This new `_is_otlp_enabled()` is now the gate, and delivery is decided entirely by the Rust shipper — so `_resolve_otlp_target()` (L233), the `_OtlpTarget` class (L200), the `DEFAULT_*_ENDPOINT` constants, and the `OTLPSpanExporter`/`OTLPLogExporter` imports (L55-57) have zero remaining callers. Leaving ~75 lines of credential/endpoint mirror logic invites a future edit there that silently has no effect (the daemon's `resolve_otlp_target` is the real source). Delete the orphaned resolution code + imports.

src/telemetry_spool/mod.rs
```diff
@@ -0,0 +1,16 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
```
Akarsh-Hegde, line 1: **[altitude] A hand-rolled durable spool (~1300 lines) + a Python mirror re-implements what OTLP largely provides.** `SpoolClient` subverts the SDK's `BatchSpanProcessor` (fake 200 → disk → a second task re-POSTs), so the export path is effectively duplicated, and base-URL/auth/endpoint-derivation logic is now in 3+ places (`observability.rs::resolve_otlp_target`, `shipper.rs::derive_base_url`, `cli.rs::derive_base_url`). The durability/ordering/dedup edge cases flagged across these files are exactly what an off-the-shelf persistent-queue exporter (or file_exporter + otel-collector) already solves. Worth weighing the maintenance surface against reusing existing infra; at minimum, de-duplicate `derive_base_url`/`ship_one` (the `cli.rs` copies are untested).

services/observability/dashboards/pm-worklog-debug.json
```diff
@@ -0,0 +1,140 @@
+{
+  "title": "PM Worklog — Debug",
+  "description": "Every PM-worklog draft and post, newest first. The Drafted table is one row per `synthesise_worklog` trace (the LLM synth: bundle in → JiraUpdate out, with gen_ai.* token usage); the Posted table is one row per `worklog_post` trace (the Rust draft→approve→post lifecycle, with drafted/approved/posted timestamps). Click any row → opens that trace's spans. Lineage backtrack: in a worklog_draft trace, expand `contributing_sessions` and click a `session:<id>` node — its span Links (link.kind=classification / formation) jump straight to how that session was classified (the full classifier span tree) and how it was formed. The exact LLM input is on `worklog_input` (full wire prompt + system_prompt/user_message child spans); the output is on `worklog_output`. All times are local (IST, UTC+5:30). Backed by spans on service meridian-task-linker-mlx (synth) and meridian-rust (post).",
```
Akarsh-Hegde: **[correctness/cleanup] Hardcoded IST offset + a stale service attribution in the description.** (1) Both tables bake `_timestamp + 19800000000` (UTC+5:30 micros) into the SQL — repeated across panels here *and* re-introduced into `classifier-debug.json` — so every non-IST viewer sees wrong times, with no shared definition. (2) The description says synth spans are "on service meridian-task-linker-mlx," but `synthesise_worklog` is emitted by `server.py` under `meridian-agent-server-mlx` (server.py:722/924). The queries don't pin `service_name` (good — that's why they work), but the prose is a landmine: anyone "tightening" the queries to match it would reproduce the #297 empty-drilldown bug.

src/migrations/043_classify_traceparent.sql
```diff
@@ -0,0 +1,15 @@
...
+-- contributing session was classified — via OTel span Links in OpenObserve.
+--
+-- This is distinct from `app_sessions.traceparent` (migration 010), which holds
+-- the session-FORMATION trace written by the ETL when it closes the session.
+-- Together the two columns give a worklog full lineage: formation -> classify.
+--
+-- Format is the 55-byte W3C string: "00-{trace_id}-{span_id}-{flags}".
+-- NULL until the session is classified by the MLX path.
+
+ALTER TABLE app_sessions ADD COLUMN classify_traceparent TEXT;
```
Akarsh-Hegde, line 15: **[altitude] Storing W3C trace-context in the product schema couples the data model to the tracing implementation.** `classify_traceparent` (alongside the existing `traceparent` from 010) exists purely so a `worklog_draft` span can add OTel Links — a telemetry-layer concern living in core product data. Since migrations are append-only, if the correlation strategy ever changes these become permanently-NULL/stale columns that can never be dropped, plus the `COALESCE` write logic to maintain forever. Consider whether span Links/baggage or a side table is a better home; flagging as a design call, not a blocker.

src/telemetry_spool/shipper.rs
adityaharishch: Fixed in 1afe3e6. `ship_one` now classifies failures: HTTP 400/413/422 are `ShipError::Terminal` and the file is moved to a new `quarantine/` dir so the loop keeps draining; only 5xx/network/401/403/429 are `ShipError::Retryable` and stop the tick. Quarantined files no longer head-of-line-block, and the cap can no longer evict good newer records behind a poison head. (401/403 kept retryable on purpose so a creds fix re-drains them rather than quarantining the whole backlog.)

src/telemetry_spool/spool_client.rs
adityaharishch: Fixed in 1afe3e6. `send` now distinguishes causes: only ENOSPC (errno 28) returns the synthetic 200 (drop — a retry cannot help, the cap reclaims space). Every other write failure (transient FS error, permissions) returns an `Err` so the BatchSpanProcessor retries instead of discarding the batch.

src/telemetry_spool/writer.rs
adityaharishch: Fixed in 1afe3e6. `write_pending` now `sync_all()`s the tmp file before the rename and fsyncs the pending dir after, so a crash cannot leave a rename durable with the data still in page cache. A failed write also unlinks its own `.tmp`. Mirrored the same fsync in the Python spool writer (observability.py).

src/telemetry_spool/spool_client.rs
adityaharishch: Fixed in 1afe3e6. `SpoolClient::send` now runs `write_pending` inside `tokio::task::spawn_blocking` so the synchronous create/write/fsync/rename no longer blocks the export worker thread.

src/telemetry_spool/shipper.rs
adityaharishch: Fixed in 1afe3e6. On a 2xx, if the move to `sent/` fails the pending file is now DELETED (it was already accepted by OO) rather than left to re-POST next tick. Losing one archive copy beats duplicate delivery. The rarer crash-between-ack-and-rename window remains (it needs an OTLP idempotency key OO does not offer) — at most one duplicate per crash, noted in a comment.

src/telemetry_spool/shipper.rs
adityaharishch: Fixed in 1afe3e6. `run_tick` now calls `sweep_tmp_orphans` first, removing `.otlp.tmp` files older than 300s (a healthy write completes the rename in ms). Combined with the failed-write `.tmp` cleanup in writer.rs, orphans can no longer accumulate. (The cap still counts only finalized `.otlp` — the sweep is what bounds tmp growth.)

src/telemetry_spool/shipper.rs
adityaharishch: Fixed in 1afe3e6. Added `seq_from_filename`; both `list_pending_oldest_first` and `enforce_pending_cap` now sort by `(micros, seq)`, and unparseable/renamed names are SKIPPED rather than collapsed to key 0 (so a foreign file no longer sorts permanently first). Added a test that exercises the same-micros tie and the skip cases.

src/telemetry_spool/shipper.rs
adityaharishch: Fixed in 1afe3e6. When `resolve_otlp_target()` is None but `resolve_otlp_endpoint()` is Some (enabled but no creds), the shipper now warns ONCE (`AtomicBool` guard, re-armed when creds return): capture is spooling with no delivery and the cap will eventually drop it. The Python side already warns the same way.

src/main.rs
adityaharishch: Fixed in 1afe3e6. The shipper now has its OWN watch channel. Shutdown order is: flush OTel (`obs_guard.shutdown()` → final spans into the spool) → `shipper::drain_once()` (one last ship) → only then signal the shipper to stop. The daemon shutdown telemetry now ships in the same run instead of next start.

src/pm_worklog/collect.rs
adityaharishch: Fixed in 1afe3e6. Added a `column_exists` (PRAGMA table_info) guard: the SELECT now emits `classify_traceparent` only if the column is present, else `NULL AS classify_traceparent` with a warn. A lagging migration 043 now degrades the lineage backlink only — drafting continues. (Verified the other added columns all predate #299, so classify_traceparent was the only at-risk one.)

services/agents/run_task_linker_mlx.py
adityaharishch: Fixed in 1afe3e6. The whole `gen_stats` extraction is now wrapped in `try/except (TypeError, ValueError, AttributeError)` that logs and drops the stats — a None/renamed/non-numeric field can no longer raise out and fail the batch.

services/agents/server.py
adityaharishch: Fixed in 1afe3e6. Replaced both metrics blocks (L797 and L873) with a shared `_set_token_metrics()` that coerces defensively (`isinstance(v, (int, float))` excluding bool, else skip) — a truthy-non-int can no longer turn a successful synth into a 500. Also dedups the two near-identical blocks.

services/agents/server.py
adityaharishch: Fixed in 1afe3e6. `synthesise_worklog` now sets `is_error=True` at span open and flips it to False only on full success, so an unexpected raise (agno internals, _coerce_jira, _record_wire_prompt) leaves is_error=true — the dashboard Failed/Avg-confidence buckets no longer miss it.

services/agents/server.py
adityaharishch: Resolved by the main merge rather than a revert. #297 round-2 (now on main) changed the classify path to attach `classify_session` to the daemon context instead of forcing a fresh root + Link. After merging, the worklog synth and the classifier BOTH nest under the propagated daemon context — so they are now consistent (your inconsistency premise no longer holds). Keeping the worklog synth nested is also the explicit product requirement here: the whole point of this trace is that worklog_input/synth_inference/worklog_output + each contributing session sit in ONE navigable tree. The classify traceparent is still stamped + persisted for the per-session span Links.

services/agents/observability.py
adityaharishch: Done — removed in the merge commit (bab0f65). `_OtlpTarget`, `_resolve_otlp_target`, `_WARNED_EXPORT_MISCONFIG`, the `DEFAULT_*_ENDPOINT` constants and the OTLP exporter imports are all gone; `_is_otlp_enabled()` is the sole gate and delivery is the Rust shipper. (Also added fsync to the Python spool writer while here.)

src/telemetry_spool/mod.rs
adityaharishch: Partly addressed, partly a deliberate call. Deduped: `derive_base_url` + `ship_one` now live once in `mod.rs`, shared by both the shipper and the (previously untested) cli import path. On the bigger altitude point — the spool exists to satisfy two product requirements an off-the-shelf exporter does not give us together: (1) capture must survive OO being down AND OO not yet configured (no creds), and (2) the same on-disk files double as a portable bundle (`telemetry export/import`) to replay a user trace in our own OO. otel-collector would add a second daemon to install/manage on every user machine, which cuts against the zero-extra-process goal. The durability/ordering/dedup edge cases you flagged are now fixed; keeping this in-process is the deliberate trade.

services/observability/dashboards/pm-worklog-debug.json
adityaharishch: Prose fixed in 1afe3e6: synth is now correctly attributed to service meridian-agent-server-mlx (not meridian-task-linker-mlx), with an explicit note that the queries must NOT pin service_name (or they reproduce the #297 empty-drilldown bug). The hardcoded IST offset is intentional: local-time columns in both dashboards are an explicit product requirement, and OO SQL has no per-viewer timezone for `to_timestamp_micros`. This is a single-operator debug dashboard (me), so a fixed UTC+5:30 is correct for now; if it ever ships to non-IST operators we will swap to a shared offset variable.

src/migrations/043_classify_traceparent.sql
adityaharishch: Acknowledged as a design call, keeping it for now. The traceparent columns (010 + 043) are the pragmatic way to correlate a worklog/session row to its OTel trace without a side table or baggage plumbing across the Rust↔Python boundary. They are nullable and append-only, so if the correlation strategy ever changes they simply go NULL — no migration pain beyond a dead column. Noted for a future cleanup if we move to a dedicated correlation table.

adityaharishch:
Thanks for the thorough review — genuinely high signal. Addressed in two commits:

- **bab0f65** merges `main` (PR #297 is merged; this PR is retargeted to `main`). The merge resolution also removed the orphaned OTLP-resolution dead code (observability.py:221) and, because #297 round-2 changed the classify path to attach `classify_session` to the daemon context, the worklog synth and classifier now nest consistently (server.py:734).
- **1afe3e6** addresses every load-bearing inline finding (see per-thread replies): head-of-line quarantine, fsync durability, spawn_blocking, dup-ship-on-rename-fail, tmp-orphan sweep, deterministic (micros,seq) ordering, enabled-but-no-creds warn, shutdown flush ordering, migration-043 column guard, and the three Python crash/visibility fixes. New unit tests cover the seq tie-break + skip-unparseable cases.

The two findings you flagged as "could not fully verify" — both checked, both safe, no code change:

1. **post.rs double-post** — not reachable. `window_start`/`window_end` are a pure function of `(day, cycle_index)` (the fixed hour bounds), so they never shift for the same logical key. And the upsert carries `WHERE state NOT IN (approved, posted)`, so a posted row is immutable: a re-draft for an already-posted (task, day, cycle) short-circuits to the existing row id (db.rs) and never produces a fresh drafted row for the poster to pick up. The dedup key and upsert key therefore cannot disagree.

2. **db_write.rs COALESCE re-classification drift** — intended (latest-classification-wins). The column always points at the most recent classification trace; an already-emitted worklog_draft span is an immutable point-in-time snapshot that Linked whatever trace was current when it was drafted (that trace still exists in OO). Re-drafting after re-classification re-Links to the new trace. There is no correctable drift without re-emitting immutable spans.

Full suite green (cargo fmt/clippy/test + UI) via pre-push. Ready for another look.

PR #326, opened 2026-06-23
feat(setup): port "A · Rail" first-run wizard with on-device model picker

Akarsh-Hegde:
Ports the **"A · Rail"** direction of the Meridian Setup design into the real first-run wizard. Re-skins the working flow (no behaviour rebuild): live commands drive it, the design supplies presentation, retokenized onto the product palette (amber, not the mock's orange) so the wizard matches the dashboard it hands off to.

## Scope
**UI** (`ui/app/setup/`, split to stay under the 500-line rule): Welcome → Permissions → Local intelligence → Integrations → Completion, in a centered 948×628 card.
- Permissions: the **three real required** cards incl. Input Monitoring (the design omitted it; the in-process capture pipeline needs it). The design's Notifications / Launch-at-login toggles are dropped — no backend.
- Integrations: Jira + Trello use the real `start_oauth` / `get_oauth_status` flow; Linear/GitHub/Asana render explicitly disabled, never wired to error.
- `globals.css`: adds `--pop-shadow` + `mer-spin` / `mer-pop`.

**Deep model step** — a real, working 3-tier picker:
- Tiers map to **real checkpoints** (`Qwen3.5-4B` / `Qwen3.5-9B-OptiQ` / `phi-4`) with honest model identity, not fictional names.
- Live hardware-spec panel + a memory gauge sized against **detected RAM**.
- Prefetch only fires on an **explicit Download** with the preference written first, so the "Ready" badge can't mislabel which model loaded.

**Backend**
- `meridian-core` settings + `ui/lib/settings.ts`: new `llm_model_preference`.
- `run_task_linker_mlx._resolve_model_id` honours that preference (degrades via `llm_selector` when it can't fit the Metal budget).
- New tray commands: `detect_system_specs`, `set_model_preference`, `get_model_preference`.
- `tray.rs` widens the setup window 560×660 → 980×660 for the Rail shell.

## Verified
Tray build + clippy clean · `meridian-core` test passes · `cargo fmt --check` clean · Python compiles · UI build + TypeScript passes · `/setup` in the static export · headless render check of Welcome + the Deep Model step. Full pre-push suite (incl. `cargo test`, UI build, security audit) passed on push.

## Known limitation
The MLX server resolves its model id once per process. Fresh-install (shipping) ordering is correct; on a dev machine where the server is already running with a different model cached, changing the pick won't hot-swap until the server restarts (no reload command today).

### DIFF
--- a/ui/app/setup/data.ts
+++ b/ui/app/setup/data.ts
@@ -144,15 +144,6 @@ export const INTEGRATIONS: Integration[] = [
   { id: 'asana',  mono: 'As', color: '#F06A6A', name: 'Asana',         account: 'Coming soon',               oauth: false },
 ]
 
-// ── Dock app marks (decorative, Welcome screen is text-only so unused there) ──
-
-export const APPS: Record<string, { mono: string; color: string }> = {
-  Antigravity: { mono: 'Aᴳ', color: '#7C3AED' },
-  'Google Chrome': { mono: 'Ch', color: '#3B82F6' },
-  Terminal: { mono: '>_', color: '#111827' },
-  Claude: { mono: 'Cl', color: '#D97757' },
-}
-
 /** Whole-GB / MB size label, matching the design's `fmtSize`. */
 export const fmtSize = (gb: number): string =>
   gb < 1 ? `${Math.round(gb * 1000)} MB` : gb < 100 ? `${gb.toFixed(1)} GB` : `${Math.round(gb)} GB`
--- a/ui/app/setup/data.ts
+++ b/ui/app/setup/data.ts
@@ -1,0 +1,158 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+
+// Static content + types for the first-run setup wizard ("A · Rail" shell,
+// ported from the Meridian Setup design). Data only — no React, no side effects.
+// Everything the wizard renders that ISN'T live machine state lives here.
+
+// ── Live backend response shapes (mirror tray/src-tauri/src/commands/setup.rs) ──
+
+/** MLX server status — polled on the Model step (`get_mlx_status`). */
+export type MlxStatus = 'offline' | 'starting' | 'running' | { error: string }
+
+export interface MlxStatusResponse {
+  status: MlxStatus
+  port: number
+  runtime_found: boolean
+  runtime_installed: boolean
+  download_available: boolean
+}
+
+/** Streamed by both the runtime download and the model prefetch. */
+export interface DownloadProgress {
+  received: number
+  total: number
+  message: string
+}
+
+/** Detected hardware (`detect_system_specs`). All-zero on non-macOS / probe failure. */
+export interface SystemSpecs {
+  chip: string
+  macos: string
+  cpu_cores: number
+  gpu_cores: number
+  ram_gb: number
+  free_disk_gb: number
+}
+
+// ── On-device model tiers ────────────────────────────────────────────────────
+// Each tier maps to a REAL MLX checkpoint in the llm_selector catalog
+// (services/agents/llm_selector.py). The picker writes the chosen `hfId` to
+// settings.json via `set_model_preference`; the server then prefers it. We show
+// the real model identity (family · params · quant) — never invented names.
+// `ramGB` is the resident working set from the catalog; `sizeGB` is the
+// approximate 4-bit download size. Both feed the footprint gauge against the
+// machine's real detected memory.
+
+export interface ModelTier {
+  id: 'nano' | 'core' | 'max'
+  label: string      // friendly tier name shown prominently
+  model: string      // real model identity (mono sub-line)
+  spec: string       // params · quant
+  hfId: string       // HuggingFace repo id persisted as the preference
+  sizeGB: number     // approx download size
+  ramGB: number      // resident working set (catalog min_ram_gb)
+  speed: string      // rough tok/s on Apple Silicon
+  recommended?: boolean
+  blurb: string
+}
+
+export const MODELS: ModelTier[] = [
+  {
+    id: 'nano', label: 'Compact', model: 'Qwen3.5 4B', spec: '4B · 4-bit',
+    hfId: 'mlx-community/Qwen3.5-4B-MLX-4bit',
+    sizeGB: 2.2, ramGB: 2.5, speed: '~75 tok/s',
+    blurb: 'Fastest and lightest. Great on 8 GB Macs.',
+  },
+  {
+    id: 'core', label: 'Balanced', model: 'Qwen3.5 9B', spec: '9B · 4-bit',
+    hfId: 'mlx-community/Qwen3.5-9B-OptiQ-4bit',
+    sizeGB: 5.0, ramGB: 6.5, speed: '~48 tok/s', recommended: true,
+    blurb: "Tuned for Meridian's classifier. Best balance for Apple Silicon.",
+  },
+  {
+    id: 'max', label: 'Maximum', model: 'Phi-4 14B', spec: '14B · 4-bit',
+    hfId: 'mlx-community/phi-4-4bit',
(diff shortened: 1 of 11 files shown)


ui/app/setup/page.tsx
```diff
@@ -1,405 +1,298 @@
...
-        // the model is only needed at the first classification.
-        if (s.status === 'running' && !prefetchStarted.current) {
+        if (s.runtime_found && s.status === 'offline') invoke('start_mlx_server_cmd').catch(() => {})
+        if (wantModel && s.status === 'running' && !prefetchStarted.current) {
           prefetchStarted.current = true
           setPrefetching(true)
           setProgress({ received: 0, total: 0, message: 'Preparing model…' })
           invoke('prefetch_model_cmd')
+            .then(() => setModelReady(true))
             .catch((e) => setErr(String(e)))
```
Akarsh-Hegde: **Bug (correctness): a failed model prefetch permanently disables the Download button.** `prefetchStarted.current` is set to `true` *before* the async `prefetch_model_cmd`, but the `.catch` never resets it. On a transient failure the row falls back to the `<Download>` button, but clicking it cannot re-trigger prefetch — the poll guard `!prefetchStarted.current` is now permanently `false`. Fix: in the `.catch`, reset `prefetchStarted.current = false` and clear `wantModel` so an explicit re-click retries. (Fixing in a follow-up commit.)

ui/app/setup/page.tsx
```diff
@@ -1,405 +1,298 @@
...
-    }
-  }
+    } catch (e) { setErr(String(e)) } finally { setDownloading(false) }
+  }, [])
 
-  const openPane = (pane: string) => {
+  // Commit the chosen model: persist the preference FIRST, then flag `wantModel`
+  // so the poll prefetches it once the server is running. Writing before the
+  // prefetch is what makes the "Ready" badge truthful about which model loaded.
+  const downloadModel = useCallback(async () => {
```
Akarsh-Hegde, line 191: **UX bug: no pending state between clicking Download and the server reporting `running`.** `downloadModel` sets `wantModel`+`progress` but not `prefetching` (that is only set inside the poll once `status === "running"`). The badge logic in `steps.tsx` is `modelReady ? Ready : prefetching ? pct% : <Download>`, so during the server cold-start window the row still shows a live, re-clickable Download button. Fix: thread `wantModel` into `Wiz` and show a `Starting…` spinner when `wantModel && !prefetching && !modelReady`. (Fixing.)

ui/app/setup/page.tsx
```diff
@@ -1,405 +1,298 @@
...
     return () => { if (unlisten) unlisten() }
   }, [downloading, prefetching])
 
-  const startDownload = async () => {
+  // Poll OAuth completion on the Integrations step so a browser-completed connect
+  // flips the row to "Connected" without the user clicking again.
+  useEffect(() => {
+    if (!active || step !== 2) return
+    const poll = async () => {
+      for (const provider of OAUTH_PROVIDERS) {
```
Akarsh-Hegde, line 133: **Dropped behavior: the OAuth poll has no deadline, so a `connecting` row spins forever.** The old `TasksView.startOAuth` had a 180s deadline that flipped to "Timed out — try again". Here the poll only ever resolves to `connected` or (on error) `idle`. If the user opens the browser tab and closes it without authorizing, the row shows "Connecting…" indefinitely with no retry path. Fix: track a per-provider deadline and resolve to idle+error on timeout. (Fixing.)

ui/app/setup/page.tsx
```diff
@@ -1,405 +1,298 @@
...
+        const st = await load<{ connected: boolean; error?: string | null }>(
+          `/api/auth/oauth/status?provider=${provider}`, 'get_oauth_status', { provider },
+        ).catch(() => null)
+        if (st?.error) { setErr(st.error); setIntegrations((s) => ({ ...s, [provider]: 'idle' })) }
```
Akarsh-Hegde: **Minor: `setErr` is called every 2s on a persistent OAuth error, and the second provider overwrites the first.** This runs inside the `for` over both providers with no change-guard, so a persistent error (e.g. missing `client_secret`) re-fires every tick, and if both jira+trello error in one tick only one message survives. Fix: only transition/`setErr` when the state actually changes (guard like the `connected` path). Folding into the #3 rewrite.

ui/app/setup/data.ts
```diff
@@ -0,0 +1,158 @@
...
+  { id: 'jira',   mono: 'Ji', color: '#2684FF', name: 'Jira',          account: 'Authorize in your browser', oauth: true },
+  { id: 'trello', mono: 'Tr', color: '#0079BF', name: 'Trello',        account: 'Authorize in your browser', oauth: true },
+  { id: 'linear', mono: 'Li', color: '#5E6AD2', name: 'Linear',        account: 'Coming soon',               oauth: false },
+  { id: 'github', mono: 'Gh', color: '#181717', name: 'GitHub Issues', account: 'Coming soon',               oauth: false },
+  { id: 'asana',  mono: 'As', color: '#F06A6A', name: 'Asana',         account: 'Coming soon',               oauth: false },
+]
+
+// ── Dock app marks (decorative, Welcome screen is text-only so unused there) ──
+
+export const APPS: Record<string, { mono: string; color: string }> = {
```
Akarsh-Hegde: **Cleanup: `APPS` is a dead export.** It was the dock-app map for the design's fake desktop, which was dropped in this port — nothing imports it. Per CLAUDE.md ("no dead files left behind" / "nothing more, nothing less"), remove it. (Fixing.)

ui/app/setup/page.tsx
```diff
@@ -1,405 +1,298 @@
...
-  download_available: boolean
-}
-
-interface DownloadProgress {
-  received: number
-  total: number
-  message: string
-}
+const SERIF: CSSProperties = { fontFamily: 'var(--font-instrument-serif), Georgia, serif' }
+const OAUTH_PROVIDERS = ['jira', 'trello'] as const
```
Akarsh-Hegde: **Cleanup: `OAUTH_PROVIDERS` duplicates the `oauth: true` entries in `INTEGRATIONS` (data.ts).** Two sources of truth for "which trackers have a real OAuth flow" can drift. Derive it: `INTEGRATIONS.filter(i => i.oauth).map(i => i.id)`. (Fixing.)

Akarsh-Hegde:
## Code review (high effort) — self-review + 2 independent finder agents

Reviewed the diff across 8 angles (correctness, removed-behavior, cross-file, reuse/simplification/efficiency, altitude, conventions). **Backend is clean** — the `df`/`system_profiler`/`sysctl` parsing, the settings round-trip, and the Python model resolver all degrade gracefully (the one latent note: `_resolve_model_id`'s `min_ram` fallback would mis-size a model absent from `llm_selector._MODELS`, but all three wizard-offered checkpoints are in the catalog, so it's unreachable from the UI).

Found **6 issues** in the React wizard (4 correctness/UX + 2 cleanup), left inline — all now **fixed in b24031b**:

1. Failed prefetch permanently bricked the Download button (`prefetchStarted` never reset). ✅
2. No pending state between Download-click and server-running (re-clickable button, no spinner). ✅
3. OAuth `connecting` row spun forever — no deadline (old TasksView had 180s). ✅
4. OAuth poll spammed `setErr` every 2s / clobbered the other provider. ✅
5. Dead `APPS` export (dock map from the dropped fake desktop). ✅
6. `OAUTH_PROVIDERS` duplicated `INTEGRATIONS.oauth`. ✅

Two intentional behavior changes were reviewed and accepted, not flagged as bugs: explicit Download replacing the old eager auto-prefetch (the `wantModel` commit gate that keeps the "Ready" badge truthful; model still lazy-loads on first classify), and the split of `mark_setup_complete` + window-close into Finish → Completion → Open (no first-run-loop risk — the flag is written on Finish).

PR #327, opened 2026-06-23
feat(tray): redesign popover (Claude Design) + show it on tray click

Akarsh-Hegde:
## Summary
Reworks the menu-bar popover to the Claude Design **Meridian Tray** layout, wired entirely to real daemon data — and fixes that the popover was never actually shown from the tray.

### Popover redesign (`tray/src/`)
- **Live block** — ticking timer, app glyph, category chip, confidence "% match", and top window title, all from the active session.
- **Time Tracker tiles** — Focus / Coding / Review / Comms aggregated from today's sessions, with real "incl. _N_ autonomous AI" time.
- **Pause row** — mapped to the real `toggle_daemon` on/off control.
- **Worklogs row** — conditional on real `drafts_count` → `open_worklogs`.
- **Footer** — Open Meridian / Permissions / Quit (new `quit_app` command).
- Vanilla HTML/CSS/JS (offline webview — system font stacks, no CDN/React); follows `prefers-color-scheme`.

### Backend (`tray/src-tauri/`)
- Extend `StatusPayload`/`AppState` with the active session's category/confidence/title and the per-category today breakdown.
- Enrich `refresh_active` / `refresh_today`; widen the popover window to 384×590.

### Fix: popover never appeared from the tray
`show_menu_on_left_click(true)` opened the native menu and nothing ever called `.show()` on the popover window. **Left-click now toggles the popover** (positioned under the icon); **right-click keeps the menu**.

## Verification
- `cargo check` + `cargo clippy` + `cargo fmt` clean; pre-commit & pre-push hooks passed.
- Rendered light / dark / paused / offline states via mock data; confirmed the live `active_session` carries real category/confidence.
- Built and ran the signed `.app`; popover opens on tray left-click with live data.


meridian-core/src/readers/current_task.rs
```diff
@@ -0,0 +1,148 @@
...
+    // that predates the column — a missing column means no estimate, not a crash.
+    let story_points: Option<String> = sqlx::query_scalar::<_, Option<String>>(
+        "SELECT story_points FROM pm_tasks WHERE task_key = ?1",
+    )
+    .bind(&key)
+    .fetch_optional(pool)
+    .instrument(tracing::debug_span!("current_task.read.story_points"))
+    .await
+    .unwrap_or(None)
+    .flatten();
```
Akarsh-Hegde: **Bug: `unwrap_or(None)` silently swallows all DB errors on the `story_points` query**

```rust
.fetch_optional(pool).await.unwrap_or(None).flatten()
```

The comment says this "tolerates a DB that predates the column", but `.unwrap_or(None)` converts *every* `Err` — pool exhaustion, busy-timeout, corrupt DB — to `None`, making them indistinguishable from "no story points". The OpenObserve stream will be completely blind to these failures.

Fix: use `.ok()` with an explicit `tracing::warn!` for unexpected errors, so the schema-miss case is still a graceful degradation but real errors surface in logs:

```rust
let story_points: Option<String> = match sqlx::query_scalar::<_, Option<String>>(
    "SELECT story_points FROM pm_tasks WHERE task_key = ?1",
)
.bind(&key)
.fetch_optional(pool)
.instrument(tracing::debug_span!("current_task.read.story_points"))
.await
{
    Ok(v) => v.flatten(),
    Err(e) => {
        // Missing column (pre-migration-039) is expected; anything else is worth logging.
        tracing::warn!(error = %e, task_key = %key, "current_task: story_points fetch failed");
        None
    }
};
```

meridian-core/src/readers/current_task.rs
```diff
@@ -0,0 +1,148 @@
...
+    .bind(&key)
+    .bind(&day_start)
+    .bind(&day_end)
+    .fetch_one(pool)
+    .instrument(tracing::debug_span!("current_task.read.spent_today"))
+    .await
+    .context("current_task: sum today's seconds on task")?;
+
+    let percent = compute_percent(spent_s, story_points.as_deref());
+    tracing::debug!(key = %key, spent_s, ?percent, "current_task.resolved");
```
Akarsh-Hegde: **Conventions: missing `tracing::info!` summary on serve (CLAUDE.md porting playbook step 5)**

> _"a `tracing::info!(…)` summary on serve"_

Only `tracing::debug!` is emitted here, which is filtered out under the default `RUST_LOG=meridian=info`. In production OpenObserve, a successfully resolved current task is completely invisible. Change to:

```rust
tracing::info!(key = %key, ?percent, "current_task.resolved");
```

tray/src-tauri/src/commands/system.rs
```diff
@@ -68,3 +68,10 @@ pub async fn open_permission_pane(app: tauri::AppHandle, pane: String) -> Result
         .open_url(url, None::<&str>)
         .map_err(|e| e.to_string())
 }
+
+/// Quit the whole app — same exit path as the tray menu's "Quit Meridian".
+/// Invoked from the popover footer's Quit button.
+#[tauri::command]
+pub fn quit_app(app: tauri::AppHandle) {
```
Akarsh-Hegde, line 76: **Conventions: `quit_app` missing `#[tracing::instrument]` (CLAUDE.md porting playbook step 5)**

> _"`#[tracing::instrument(skip(pool))]` on both the command and the core fn"_

```rust
#[tracing::instrument(skip(app))]
#[tauri::command]
pub fn quit_app(app: tauri::AppHandle) {
    app.exit(0);
}
```

meridian-core/src/readers/current_task.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Ring fill in `[0.0, 1.0]`, or `None` when no usable story-point budget
+    /// exists (the caller then draws an un-filled ring, not `0%`).
+    pub percent: Option<f64>,
+}
+
+/// Resolve the current task for `today` (a local `YYYY-MM-DD`), or `None` when no
+/// task session has been classified yet today. `today` is passed in so the fn
+/// stays deterministic and timezone math lives in one place.
+#[tracing::instrument(skip(pool))]
+pub async fn get_current_task(
```
Akarsh-Hegde, line 46: **Conventions: no DB integration test for `get_current_task` (CLAUDE.md porting playbook step 7)**

> _"DB readers → an in-memory seeded test in `meridian-core/tests/readers.rs`"_

The SQL logic — day-boundary filtering on `started_at`, `task_session_type = 'task'` filter, `ORDER BY started_at DESC LIMIT 1` ordering, and the `COALESCE(SUM, 0)` on the spent query — is completely untested. The `compute_percent` unit tests only cover the pure math; they don't catch a wrong date-bound or a missing `task_session_type` filter.

Need a test in `meridian-core/tests/readers.rs` that seeds an in-memory DB with a classified session today and a `pm_tasks` row with story_points, then asserts the correct key and percent are returned.

tray/src-tauri/src/lib.rs
```diff
@@ -122,6 +154,65 @@ pub fn run() {
...
+                                Some(p) => (p.clamp(0.0, 1.0) * 100.0).round() as i32,
+                                None => -1,
+                            };
+                            (s.tray_id.clone(), title, bucket)
+                        };
+                        let Some(id) = tray_id else { continue };
+                        let Some(tray) = title_app.tray_by_id(&id) else {
+                            continue;
+                        };
+                        if title != last_title {
```
Akarsh-Hegde, line 200: **Timer drift: Rust ticker and JS popover timer independently extrapolate from different snapshots**

The native menu-bar title is driven by a 1 s Rust ticker that extrapolates from `AppState.active_set_at` (set at poll time). The in-popover timer is driven by a JS `setInterval` that extrapolates from the `active_elapsed_s` in the last `status-update` event. Both start from slightly different snapshots and accumulate independent rounding error.

On a 30 s poll, the two timers can drift by ±1 s relative to each other between snaps. Not a crash, but the menu-bar title and the open popover will visibly show different values at the same moment.

Simplest fix: have the Rust ticker emit a lightweight event (`timer-tick`) carrying the current elapsed seconds each second, and have the JS popover consume that instead of running its own `setInterval` counter. One source of truth.

tray/src/app.js
```diff
@@ -3,183 +3,255 @@
...
+}
+
+// ── Render ───────────────────────────────────────────────────────────────────
+function render(status) {
+  const healthy = status.ui_reachable && status.healthy
+  const hasActive = !!status.active_app
 
-  if (status.active_app) {
+  // Brand dot reflects daemon health.
+  brandDot.classList.toggle('down', !healthy)
```
Akarsh-Hegde, line 147: **UX regression: pending-drafts indicator dropped**

The old dot had three visual states: healthy (green), pending (orange/accent when `drafts_count > 0`), and unhealthy (red/dim). The new `brandDot` only has two: healthy (orange accent, always) and down (gray).

The pending-drafts signal was the only way for a user to notice drafted worklogs were waiting WITHOUT having to open the popover. It's now gone — the dot is always orange when the daemon is healthy, regardless of draft state. The worklogs row inside the popover still shows the count, but there's no tray-level attention signal anymore.

tray/src-tauri/src/lib.rs
```diff
@@ -122,6 +154,65 @@ pub fn run() {
...
+                            let _ = tray.set_title(title.as_deref());
+                            last_title = title;
+                        }
+                        if bucket != last_bucket {
+                            let pct = (bucket >= 0).then(|| bucket as f64 / 100.0);
+                            if tray.set_icon(Some(tray_icon::ring_image(pct))).is_ok() {
```
Akarsh-Hegde, line 206: **Nit: `set_icon_as_template` failure is silent**

```rust
if tray.set_icon(Some(tray_icon::ring_image(pct))).is_ok() {
    let _ = tray.set_icon_as_template(true);  // failure silently dropped
}
```

If `set_icon_as_template` fails after a successful `set_icon`, the ring renders as a full-colour black blob (all RGB=0, alpha=255 where the arc is) instead of the macOS template tint. In dark mode this means the completed arc is an invisible black-on-dark stripe. Add a `tracing::warn!` so failures surface in logs:

```rust
if tray.set_icon(Some(tray_icon::ring_image(pct))).is_ok() {
    if let Err(e) = tray.set_icon_as_template(true) {
        tracing::warn!(error = %e, "tray: set_icon_as_template failed — ring may render full-colour");
    }
}
```

PR #330, opened 2026-06-24
release: fold dashboard into Tauri + self-contained DMG + in-process capture (pre-main → main)

Akarsh-Hegde:
## Summary

Promotes **`pre-main` → `main`**: the culmination of the **Next-fold** and the **self-contained-DMG / setup re-platform** initiative. Meridian moves from a *Node-server + external-screenpipe-process* topology to a **single self-contained Tauri `.app`** — the dashboard runs inside the webview as a static export, every `/api` route is ported to Rust, capture runs **in-process**, and the MLX runtime is provisioned/supervised by the tray.

This bundles **13 already-reviewed feature PRs** (#298, #314–#329) plus post-merge review + CI fixes.

| Metric | Value |
|---|---|
| Commits | **186** |
| Files changed | **223** (83 added, 49 deleted) |
| Diff | **+25,162 / −8,814** |
| New shared crate | `meridian-core` — 23 files, ~6,050 LOC |
| Commit types | 72 `feat`, 49 `fix`, 11 `chore`, 9 `refactor`, 8 `docs`, 7 `build`, 6 `test`, 6 `ci` |
| Resulting version | **`v1.64.0`** (minor — see Versioning) |
| DMG size | **19.6 MB** (`Meridian.app.tar.gz` 19.3 MB) |
| Release build | **~5 min** (macOS runner: build + sign + notarize-ready + package) |
| Tests | **516 Rust** + **7 JS/TS** suites — all green |

---

## Architecture changes

The core shift: **delete the Node UI server and the external screenpipe process**; collapse everything into one signed `.app` where the tray owns capture + the embedded dashboard, the daemon owns ETL, and `meridian-core` is the shared data layer both link against.

**Before (`v1.63.0`)**
- `screenpipe` runs as its own launchd process → writes `~/.screenpipe/db.sqlite`; the daemon reads it cross-DB.
- A standalone **Next.js Node server** (`com.meridiona.ui` launchd agent) serves the dashboard **and** ~32 `/api/*` routes, reading `meridian.db` over `better-sqlite3`.
- The tray polls those `/api` routes over **HTTP**.

**After (`v1.64.0`)**
- The tray captures **in-process** (forked `screenpipe-screen` + `screenpipe-a11y`) → writes `capture_frames` / `capture_ui_events` **into `meridian.db`**. No screenpipe DB / process / pool.
- The dashboard is a **static export embedded in the tray binary**; the frontend reaches Rust only via **Tauri `invoke` / events** through `ui/lib/bridge.ts`. No Node server, no `/api` routes.
- DB reads live in **`meridian-core/src/readers/`** (shared by daemon + tray); file/env/process actions are **tray commands**. The four SSE streams became **Tauri events** off the tray poll loop.

```mermaid
flowchart TB
  subgraph BEFORE["BEFORE — v1.63.0 (multi-process)"]
    direction TB
    sp1[screenpipe process]:::ext --> spdb[(~/.screenpipe/db.sqlite)]:::ext
    spdb --> d1[meridian daemon · ETL]
    d1 --> mdb1[(meridian.db)]
    node[Next.js Node server<br/>com.meridiona.ui]:::ext --> mdb1
    node -->|serves| dash1[Dashboard + /api routes]
    tray1[Tauri tray] -->|HTTP poll /api| node
    mlx1[MLX Python server]:::py --> d1
  end

  subgraph AFTER["AFTER — v1.64.0 (single self-contained .app)"]
    direction TB
    subgraph app["Meridian.app"]
      cap[Tray · in-process capture<br/>screenpipe-screen/-a11y]:::new
      core[[meridian-core<br/>shared readers + commands]]:::new
      webview[Embedded dashboard<br/>static export]:::new
      d2[meridian daemon · ETL]
    end
    cap -->|capture_frames / capture_ui_events| mdb2[(meridian.db)]
    d2 -->|reads capture tables| mdb2
    d2 -->|app_sessions| mdb2
    webview -->|Tauri invoke / events| core
    core --> mdb2
    mlx2[MLX Python server<br/>provisioned + supervised by tray]:::py --> d2
  end

  classDef ext fill:#fde,stroke:#b66;
  classDef new fill:#dfe,stroke:#6b6;
  classDef py fill:#eef,stroke:#66b;
```

---

### 1. Next-fold — dashboard into Tauri, `/api` → Rust  (#298, #314)
- Dashboard now runs **only inside the Tauri webview** as a static export (`output: 'export'`) — no Node server.
- All ~32 `/api/*` routes ported: **DB reads → `meridian-core/src/readers/`** (re-exported by the daemon), **file/env/process → `tray/src-tauri/src/commands/`**.
- Health / notices / notifications / log streams converted from **SSE → Tauri events** emitted by the tray poll loop.
- Retired the standalone Node-UI release machinery (`com.meridiona.ui` plist, `ui-start.sh`, pinned Node + better-sqlite3 ABI dance). Dashboard now **ships embedded in the tray binary**.

### 2. Self-contained DMG + setup re-platform  (#315, #318, #319, #320, #321, #326)
- **Backend bundled into the `.app`** — daemon + a11y staged to stable `~/.meridian/bin/`, plists rendered, launchctl bootstrap (#321).
- **MLX runtime:** download-and-provision on first run (#315), **tray-side supervision** (#318), **CI provisioning** (#319), **integrity verification** (#320).
- **First-run onboarding wizard** + `meridian setup` commands + native **setup UI** (#315, #326).
- New `meridian uninstall [--purge] [--dry-run] [--yes]` command for clean teardown.

### 3. In-process capture — screenpipe forked at MIT  (#323, #324)
- screenpipe pinned/forked at its last-MIT commit (#323); capture runs **in-process in the tray**, writing `capture_frames` / `capture_ui_events` into `meridian.db`.
- Daemon ETL reads those tables — **no screenpipe DB/process/pool anymore** (`run_etl(meridian)` instead of `run_etl(screenpipe, meridian)`).

### 4. Tray popover redesign  (#327)
- Reworked popover UI. Two known limitations deferred for the MVP (timer drift; pending-drafts amber dot).

### 5. Auto-update + release engineering  (#328, #329)
- `tauri-plugin-updater` **staging channel** wired into the release pipeline (#328); staging private-fork auth fix (#329). Minisign trust, independent of Apple Dev ID. Proven end-to-end (ad-hoc bundle relaunch).

### 6. Post-merge review + CI fixes
- The high-effort `/code-review` of #298 surfaced 4 correctness divergences from the original TS routes — all fixed (`b54ec04`): multibyte log-tail drop, leaked `tasks-sync` child, lenient quiet-hours parser, divergent DB-path resolver.
- CI Rust job now authenticates the private `screenpipe-fork` git dep (#331).

---

## Compatibility & breaking changes

Audited across four surfaces (DB schema, public interfaces, install/upgrade path, runtime behavior). **No hard breaking changes for existing installs** — the version-affecting changes are additive or documented v1 degradations.

**✅ Safe / non-breaking**
- **DB migrations are strictly append-only** (046 `capture_frames`, 047 `capture_ui_events`, 048 ETL-cursor reset) — no modified/renamed/deleted migrations (CI "Migrations append-only" guard enforced), no `DROP`/`ALTER`/destructive changes. Existing `meridian.db` upgrades cleanly; `app_sessions` history preserved.
- **Public interfaces preserved** — no CLI command, env var, `settings.json` key, or MCP tool removed or renamed (only additions). The deleted `/api/*` routes were **internal-only** (consumed solely by the bundled dashboard, which moved to Tauri `invoke` in lockstep).
- Added `busy_timeout(5s)` on the meridian pool for the new tray↔daemon concurrent-write topology — backward compatible.

**⚠️ Documented v1 degradations (intentional, not regressions)**
- **Audio dropped** — `get_audio_snippets` stubbed empty (in-process capture is text/OCR/a11y only). Classifier runs on text, so accuracy is unaffected; `audio_snippets` is simply always empty.
- **Idle detection off** — no in-process idle signal yet, so gaps all classify `system_sleep` (`capture_trigger` is NULL).
- Both are accepted v1 trade-offs with audio + idle-detection as future slices.

> Semantic note: the `/api` removal and capture cutover are large *architectural* changes but not *user-contract* breaks — nothing a user, script, or external integration depends on was removed. Hence the minor bump is honest (see below).

---

## Build & artifacts

- **DMG: 19.6 MB** (`Meridian.dmg`), updater bundle `Meridian.app.tar.gz` 19.3 MB + minisign `.sig` + `latest.json`.
- **Release build ~5 min** on the macOS runner (`release-staging.yml`: build → sign → updater-sign → package → publish + mirror).
- Self-contained: the `.app` carries the daemon, a11y helper, embedded dashboard, and in-process capture; the MLX runtime provisions on first run.
- Ships **ad-hoc-signed** today; notarization / Gatekeeper-clean first launch lands with the Apple Developer ID (org enrollment for MERIDIONA LLP currently *In Review*).

---

## Quality & test coverage

- **516 Rust test functions** across the workspace — daemon 362, `meridian-core` 63, tray 27, integration 64 (`tests/etl_*.rs`, `task_linker_smoke`, `tray_assets`, `worklog_provider`, + `meridian-core/tests/readers.rs`, `today_smoke.rs`).
- **7 JS/TS suites** (`ui/__tests__/`: intervals, format, category-colors, cursor-pointer, redesign, lib, meridian-bin).
- **Ported readers are test-covered** — in-memory seeded DB tests assert byte-for-byte parity with the deleted `/api` routes.
- **CI green** on #330: Rust (fmt + clippy `-D warnings` + test) **2m52s**, UI build 41s, Migrations-append-only guard, screenpipe MIT-pin guard, CodeQL (rust/js/python/actions), CodeRabbit.
- Every constituent PR passed the full pre-push suite (fmt + clippy + `cargo test` + UI build + UI tests + security audit).

---

## Versioning

**`v1.63.0` → `v1.64.0` (minor).** semantic-release derives the bump from conventional commits: 72 `feat:` force at least a minor, and **zero `BREAKING CHANGE:` footers / `!` markers** mean no major. **Empirically confirmed** — the staging channel already cut `v1.64.0-staging.{1,2,3}` from these exact commits.

---

## Known follow-ups (deferred, non-blocking)
- **Next-fold deferred findings** — `week.rs` UTC bucketing, `refresh.rs` poll-loop efficiency, file-size splits (incl. `mlx_server.rs` 1183 lines), `table_exists()` DRY extraction.
- **Apple Developer ID** (MERIDIONA LLP org enrollment, *In Review*) gates notarization / single-"Meridian" TCC / Gatekeeper-clean first launch — ships ad-hoc-signed until the cert lands.
- **Audio capture + in-process idle detection** are future slices.
- Tray redesign: timer drift + pending-drafts amber dot.

---

## Risk

Every constituent change already merged via its own reviewed PR into `pre-main`; this PR is the **promotion gate to `main`**, not new unreviewed work. Each feature is independently revertable at its merge commit. CI is fully green.


tray/src-tauri/src/mlx_server.rs
```diff
@@ -0,0 +1,1183 @@
...
+/// 1. **Runtime** `MERIDIAN_RUNTIME_MANIFEST_URL` — dev / tests (a packaged
+///    `.app` launched from Finder has no shell env, so this only fires in dev).
+/// 2. **Compile-time** `MERIDIAN_RUNTIME_MANIFEST_URL` (`option_env!`) — BAKED
+///    into a channel build. The staging release sets it at build time so the
+///    staging `.app` pins `runtime-staging`; production builds leave it unset.
+/// 3. The compiled-in `RUNTIME_MANIFEST_URL` default (`runtime-latest`).
+///
+/// `None` → no runtime published for this channel (wizard shows "not available").
+pub fn manifest_url() -> Option<String> {
+    if let Ok(url) = std::env::var("MERIDIAN_RUNTIME_MANIFEST_URL") {
```
adityaharishch, line 214: **[SECURITY — HIGH] MLX runtime tarball has no signature — SHA-256 is self-referential and `MERIDIAN_RUNTIME_MANIFEST_URL` is redirectable in production**

`manifest_url()` reads `MERIDIAN_RUNTIME_MANIFEST_URL` unconditionally in release builds. The comment at L200–204 that this “only fires in dev” is a deployment assumption, not an enforced gate (`#[cfg(debug_assertions)]`).

Combined with `download_and_stage` (~L340–448) which only verifies a SHA-256 fetched from **the same manifest it just downloaded** (no minisign), an attacker who can set this env var — or MITM the HTTPS manifest endpoint — controls both the tarball URL and the expected hash, and gets arbitrary code execution: `bin/python` from the staged runtime is spawned in `start()` (~L905–915).

The `.app` updater correctly uses `tauri-plugin-updater` with an embedded minisign pubkey; this second update channel has no equivalent protection.

**Fix**: gate the env override behind `#[cfg(debug_assertions)]`; add minisign verification to `download_and_stage` matching the updater path.

tray/src-tauri/src/poll/mod.rs
```diff
@@ -0,0 +1,275 @@
...
+
+/// One bounded MLX supervision step (called every health tick).
+///
+/// Delegates the decision to [`mlx_server::supervise`] and maintains the
+/// consecutive-restart budget: reset on health, increment on each (re)start,
+/// and once the budget is spent stop spawning — keep cheaply probing `/health`
+/// so the tray resumes automatically if the server recovers (self-heal or a
+/// human fix). User-facing "offline" messaging stays with the daemon's notice;
+/// this loop only logs (and remediates).
+async fn supervise_mlx(mlx: &SharedMlxManager, attempts: &mut u32) {
```
adityaharishch, line 158: **[CORRECTNESS — HIGH] MLX restart-budget exhaustion permanently wedges the server — no self-heal path**

Once `attempts >= MLX_MAX_RESTARTS`, `supervise_mlx` enters a probe-only branch that re-probes `/health` but never calls `supervise()` again, never kills the zombie holding port 7823, and never resets `attempts` (only a successful `/health` response resets it). A wedged server that fails `/health` is permanently stuck until the user manually relaunches the tray.

**Trigger**: MLX OOMs or crashes on startup 5 consecutive times (e.g. model load on a low-RAM machine).

**Fix**: on budget exhaustion, SIGKILL any managed PID, reset `attempts = 0`, and restart the supervise loop after a longer backoff.

ui/lib/bridge.ts
```diff
@@ -0,0 +1,100 @@
...
+  const t = tauri()
+  if (!t) {
+    console.warn(`subscribe('${apiPath}') requires the Meridian app — Tauri bridge unavailable.`)
+    return () => {}
+  }
+  let cancelled = false
+  let unlisten: UnlistenFn | null = null
+  // Prime with the current snapshot so the view isn't empty until the next emit
+  // (skipped for delta streams, where `command` is null and the caller primes).
+  if (command) t.core.invoke(command, args).then((d) => { if (!cancelled) onData(d as T) }).catch(() => {})
```
adityaharishch: **[CORRECTNESS — MEDIUM] `subscribe()` snapshot prime silently swallows invoke errors — views stay blank with no error state on cold/locked DB**

The initial snapshot after subscribing fires an invoke and discards any rejection:

```typescript
invoke(command, args).then(emit).catch(() => {})  // errors silently dropped
```

If `meridian.db` is not yet open (tray beats daemon on startup) or is locked, the invoke rejects. The `.catch(()=>{})` drops it. State-change events only fire on subsequent changes, so the view stays blank indefinitely with no loading indicator or error message.

**Fix**: replace `.catch(()=>{})` with `.catch(err => emit({ error: String(err) }))` or set an explicit error state so the component can show a retry prompt.

scripts/build-mlx-runtime.sh
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env bash
+# ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+#
+# Build the self-contained MLX runtime tarball for Approach C (download-and-provision).
+#
+# Produces:
+#   dist/meridian-mlx-runtime-<ver>-aarch64.tar.gz   (CPython + venv + agents pkg + launcher)
+#   dist/runtime-manifest.json                        (version, url, sha256, size, floors)
+#
+# The tarball bundles its OWN python-build-standalone CPython, so the customer
```
adityaharishch, line 10: **[SECURITY — MEDIUM] CPython bundled from unpinned PBS `releases/latest` with no checksum**

When `PBS_ASSET_URL` is unset (the default CI path), the script resolves the CPython interpreter from `astral-sh/python-build-standalone` at `releases/latest` — unpinned — and the downloaded archive is extracted without verifying a sha256 before `tar -xzf`.

This chains into the MLX runtime signing gap: the runtime tarball ships this interpreter, the tarball is only verified against a self-referential SHA-256 in the unsigned manifest, and now the interpreter itself comes from an unverified upstream.

**Fix**: pin `PBS_ASSET_URL` to an exact release tag and verify the archive against the published sha256 before extraction.

adityaharishch (commented):
## Code Review — PR #330 (pre-main → main)

Reviewed 8 angles across 223 changed files (diff too large for the GitHub API; files fetched directly). Findings ranked most-severe first. Four are posted inline; four below could not be pinned to diff lines.

---

### [CORRECTNESS — HIGH] `src/health/ui.rs` + `src/health/platform.rs` probe the deleted Next.js server — `meridian doctor` is permanently CRITICAL on every healthy post-PR install

`platform.rs` still declares `const LABEL_UI: &str = "com.meridiona.ui"` and `ui_service()` calls `plist_check(LABEL_UI, "ui plist")` which returns `Check::critical("missing or invalid")`. `backend_install.rs` no longer registers that launchd agent (only daemon + a11y). `ui.rs` still GETs `http://localhost:3939/api/health` (nothing listens post-fold). Both are called unconditionally from `run_all()`.

On every healthy post-PR install:
- ✗ CRITICAL “ui plist — missing or invalid” (remedy: `./install.sh` — can never clear)
- ⚠ WARN “ui service — not loaded — dashboard unavailable”
- ⚠ WARN “ui serving — not reachable at http://localhost:3939”

`Report::worst()` → Critical → `meridian doctor` exits non-zero on healthy machines. The entire UI health group needs to be replaced with embedded-webview health semantics or removed.

---

### [SECURITY — MEDIUM] `install.sh` downloads OpenObserve with no integrity check, then registers it as a launchd agent

Around L602–629, the installer `curl`s the OpenObserve binary from `downloads.openobserve.ai`, extracts, `chmod +x`s, and registers it as a launchd agent with no sha256 or signature verification. Version is pinned (`v0.90.3`) but artifact integrity is not. A compromised host or MITM gives an attacker a trojan that runs continuously under the user’s launchd.

**Fix**: pin and verify a sha256 for the OO tarball before executing it.

---

### [SECURITY — MEDIUM] `scripts/install-from-bundle.sh` fallback `sudo npm install -g screenpipe` runs npm lifecycle scripts as root

Around L1262–L1265, when the global npm prefix is not user-writable, the script falls back to `sudo npm install -g screenpipe@0.4.6`. The script explicitly refuses to run as root elsewhere (around L1222) — this single path hands root to npm postinstall scripts from a third-party package. A compromised `screenpipe@0.4.6` executes as root.

**Fix**: use `--ignore-scripts` or stage into a user-writable prefix instead of escalating.

---

### [CORRECTNESS — MEDIUM] `src/health/capture.rs` capture coverage datetime filter is a no-op against RFC3339 timestamps

`capture_coverage` filters with `WHERE timestamp > datetime('now', '-1 day')`. SQLite `datetime()` produces `'2026-06-23 17:00:00'` (space separator, no Z). `insert_capture_frame` writes `to_rfc3339_opts(Micros, true)` = `'2026-06-24T16:00:00.000000Z'` (T separator, Z suffix). String compare at index 10: `'T'` (0x54) > `' '` (0x20) → every row passes the filter regardless of actual date. The -1 day window never excludes anything → coverage check evaluates all-time frames → ghosting goes undetected.

`frame_freshness()` in the same file already uses `julianday()` correctly. **Fix**: adopt the same `julianday()` arithmetic for `capture_coverage`.

Akarsh-Hegde:
## Review findings resolved → PR #333

All findings from @adityaharishch's review have been addressed in **PR #333** (`fix/pr-330-review → pre-main`). Summary:

### Fixed

| Severity | Finding | Fix |
|---|---|---|
| HIGH CORRECTNESS | `meridian doctor` exits non-zero on every healthy post-fold install — `ui_service()` + `ui::checks()` probe the retired `com.meridiona.ui` Node server | Replaced both with `Info` checks: "dashboard embedded in the Tauri binary (no separate service)". Removed the HTTP probe, plist check, `.next` check, `serve_mode_check`, and their unit tests. |
| HIGH CORRECTNESS | MLX restart-budget exhaustion permanently wedges — `supervise_mlx` never resets `attempts` when `/health` keeps failing, so the zombie holding port 7823 is never cleared | Added `MLX_COOLING_TICKS = 10` (~10 min). `attempts` now increments through the cooling window past `MLX_MAX_RESTARTS`; at `MAX + COOLING` ticks it resets to `0` so `supervise()` runs again (with its existing `KilledWedged` path) and a fresh restart cycle begins. |
| HIGH SECURITY (partial) | `MERIDIAN_RUNTIME_MANIFEST_URL` env-var redirectable in production | Gated behind `#[cfg(debug_assertions)]` — release binary cannot be redirected via a compromised env var. The self-referential SHA-256 issue (manifest + tarball on the same origin) requires an out-of-band trust root (minisign, like the updater); design tracked as a follow-up. |
| MEDIUM CORRECTNESS | `capture_coverage` datetime format mismatch — `datetime('now', ?1)` produces `'…T…Z'` vs `' '`-separated stored timestamps; 'T' > ' ' so every row passes the filter | Fixed to `julianday(timestamp) > julianday('now', ?1)` following the `frame_freshness` pattern in the same file. |
| MEDIUM CORRECTNESS | `subscribe()` snapshot prime silently swallows errors on cold-DB startup | Retry-once-after-2s on first failure; `console.warn(...)` on second failure — visible in DevTools. |
| MEDIUM SECURITY | `sudo npm install` runs lifecycle scripts as root | Added `--ignore-scripts` to both npm install lines in `install-from-bundle.sh`. |

### Deferred (acknowledged, not blocking)

- **`install.sh` OO sha256** and **`build-mlx-runtime.sh` PBS pin**: both need real hash values from the upstream release pages — straightforward follow-ups.
- **MLX manifest self-referential SHA-256**: requires a minisign/ed25519 embedded public key (same mechanism as the updater's `pub_key`). Tracked separately.

tray/src-tauri/src/mlx_server.rs
Akarsh-Hegde: **Partially resolved in PR #333** (merged to pre-main).

The  env-var path is now gated behind `#[cfg(debug_assertions)]` — a release binary can no longer be redirected via a compromised environment variable.

The self-referential SHA-256 issue (manifest and tarball served from the same origin, so a compromised server can swap both atomically) is a separate architectural gap — it requires an out-of-band trust root (minisign public key embedded in the binary, similar to the updater's `pub_key`). Tracked as a follow-up design task; not blocking the fold.

scripts/build-mlx-runtime.sh
Akarsh-Hegde: **Deferred** — not resolved in PR #333.

Pinning to a specific PBS release tag and verifying the downloaded archive are straightforward changes, but they require the real sha256 of the target release. Adding a fabricated/guessed hash would be worse than leaving it unpinned. Tracked as a follow-up: pin `PBS_RELEASE_TAG` and add `sha256sum --check` before `tar -xzf`.

Akarsh-Hegde:
## Review findings — status update (PR #333 merged to pre-main)

All inline threads updated. Summary of what's resolved vs. still open:

| Thread | Status |
|---|---|
| `poll/mod.rs` — MLX restart-budget exhaustion permanently wedges | ✅ **Resolved** (PR #333) — `MLX_COOLING_TICKS = 10` added; `attempts` counts through a ~10-min cooling window then resets, letting `supervise()` kill any zombie and start a fresh restart cycle |
| `ui/lib/bridge.ts` — `subscribe()` snapshot prime silently swallows errors | ✅ **Resolved** (PR #333) — retry-once-after-2s for the cold-DB startup race; `console.warn` on second failure |
| `src/health/ui.rs` + `platform::ui_service()` — `doctor` CRITICAL on healthy installs | ✅ **Resolved** (PR #333) — removed all Node-era HTTP probes and plist checks; both now return `Info: dashboard embedded in the Tauri binary` |
| `src/health/capture.rs` — `capture_coverage` datetime format mismatch | ✅ **Resolved** (PR #333) — `julianday(timestamp) > julianday('now', ?1)` in both subqueries |
| `install-from-bundle.sh` — `sudo npm install` runs lifecycle scripts as root | ✅ **Resolved** (PR #333) — `--ignore-scripts` added to both npm install lines |
| `mlx_server.rs` — `MERIDIAN_RUNTIME_MANIFEST_URL` redirectable in production | ⚠️ **Partially resolved** (PR #333) — env-var path gated behind `#[cfg(debug_assertions)]`; self-referential SHA-256 (manifest + tarball from same origin) requires an out-of-band trust root (minisign embedded key), tracked as a follow-up |
| `build-mlx-runtime.sh` — PBS `releases/latest` + no checksum | 🔲 **Deferred** — needs real sha256 of the pinned PBS release; adding a fabricated hash would be worse than leaving it unpinned |
| `services/agents/server.py` — empty `except` clause | 🔲 **Not addressed** — out of scope for the fold; tracked separately |

PR #333 is merged into pre-main. The remaining two open items are tracked as follow-ups and are not blocking the `pre-main → main` promotion.

Akarsh-Hegde:
The three deferred items are now tracked in the Obsidian tech backlog (Architecture/Tech Backlog.md):

- MLX runtime manifest out-of-band signature (minisign pub key embedded in binary)
- `build-mlx-runtime.sh` PBS release pin + sha256 verification
- `services/agents/server.py` empty `except` clause

PR #340, opened 2026-06-25
fix(dev): clean startup, worklog pipeline wiring, and a11y capture gate

adityaharishch:
## Summary

- **dev-start.sh**: clear `ui/.next` cache on each run so stale module references (e.g. deleted `instrumentation.ts`) don't break `beforeDevCommand`; add `ui/node_modules` preflight so the error surfaces clearly; kill stale `Meridian Dev.app` process before relaunching
- **install.sh**: add `npm install` in `ui/` during dev-path setup so `ui/node_modules` is present when `npm run tauri dev` runs `beforeDevCommand`
- **src/main.rs**: wire `worklog_pipeline::run_loop` (calls `/worklog_hour` hourly) instead of the decommissioned `pm_worklog::run_loop` (was calling `/synthesise_worklog` → 404)
- **tray/src-tauri/src/commands/health.rs**: gate the a11y-helper trust check behind `#[cfg(not(feature = "capture"))]` — in-process capture builds don't use the a11y-helper binary, so the banner must never fire
- **services/agents/observability.py**: add `instrument_agno()` and `current_traceparent()` for W3C traceparent propagation from agno-based pipelines

## Test plan

- [ ] Fresh clone → `bash install.sh` (dev path) → `bash dev-start.sh` — no `next: No such file or directory` error
- [ ] Re-run `bash dev-start.sh` — `ui/.next` is cleared, no stale module error
- [ ] Tauri dev build with `--features capture` — health check returns no `a11y_helper_trusted` field, a11y banner absent
- [ ] MLX server logs — no more `POST /synthesise_worklog 404`; `/worklog_hour` called on the hour instead

### DIFF
--- a/dev-start.sh
+++ b/dev-start.sh
@@ -46,6 +46,11 @@ if [[ ! -d "${REPO_ROOT}/tray/node_modules" ]]; then
     exit 1
 fi
 
+if [[ ! -d "${REPO_ROOT}/ui/node_modules" ]]; then
+    echo "✗ ui/node_modules not found — run: bash install-dev.sh" >&2
+    exit 1
+fi
+
 # ---------------------------------------------------------------------------
 # Stop any previous dev run FIRST so re-running is idempotent.
 # The Rust daemon binds a unix socket (~/.meridian/daemon.sock) and the MLX
@@ -58,7 +63,11 @@ pkill -f 'target/debug/meridian$'       2>/dev/null || true   # daemon binary
 pkill -f 'uvicorn agents.server:app'    2>/dev/null || true   # MLX dev server
 pkill -f 'tauri dev'                    2>/dev/null || true   # tray file-watcher
 pkill -f 'target/debug/meridian-tray$'  2>/dev/null || true   # tray binary
+pkill -f 'Meridian Dev.app'             2>/dev/null || true   # stale dev .app bundle
 sleep 1   # let sockets / ports free before the new windows bind them
+# Clear the Next.js build cache so stale module references (e.g. a deleted
+# instrumentation.ts) don't cause beforeDevCommand to fail on the next run.
+rm -rf "${REPO_ROOT}/ui/.next"
 echo "  ✓ previous dev run stopped"
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


Akarsh-Hegde (commented):
## Review

PR fixes real and concrete problems: `beforeDevCommand` crashing on stale `.next` cache, a11y-helper banner firing on in-process-capture builds, and the decommissioned `/synthesise_worklog` 404. Changes are small and focused.

One correctness finding and one minor altitude note.

### `current_traceparent()` should use the SDK propagator

`services/agents/observability.py` (new function)

```python
# Current — hand-rolls the W3C format
return f"00-{ctx.trace_id:032x}-{ctx.span_id:016x}-{trace_flags}"

# Preferred — uses the propagator already imported and used for the inverse
carrier: dict[str, str] = {}
TraceContextTextMapPropagator().inject(carrier)
return carrier.get("traceparent")
```

The `TraceContextTextMapPropagator` is already imported at line 63 and already used in `extract_parent_context` for extraction. The manual form skips tracestate propagation and hard-codes version byte `"00"`, which diverges from the established pattern. The propagator approach is consistent, handles tracestate, and picks up any future W3C spec changes automatically.

### `preview()` placement

`preview()` is a pure string-truncation utility with no OTel dependency — it's added to `observability.py` but called from `session_distiller.py` and `worklog_pipeline/pipeline.py` solely for span attribute display. It fits fine here since all call sites are observability-adjacent, but if more span-attribute helpers get added over time, a dedicated `observability_utils.py` or `span_util.py` would keep `observability.py` focused on setup/export concerns. Minor; no action required now.

### `rm -rf ui/.next` is intentional but worth a note

`dev-start.sh`

The unconditional cache-wipe trades a startup cost (full Next.js rebuild on every `dev-start.sh`) for eliminating the stale-module crash. The comment explains the rationale. If the rebuild time becomes painful, a narrower check (`ui/.next/BUILD_ID` mismatch, or a `--no-clean` flag) could narrow it later. Acceptable as-is.

---

The `worklog_pipeline::run_loop` wiring and the a11y cfg gating both look correct on inspection. Ship once `current_traceparent` is patched to use the propagator.

adityaharishch:
Fixed the `current_traceparent` finding — now uses `TraceContextTextMapPropagator().inject(carrier)` instead of hand-rolling the W3C format string, matching the extraction pattern and handling tracestate correctly.

Akarsh-Hegde (approved):

PR #344, opened 2026-06-26
feat(onboarding): provision all three models (llm + reranker + embedder) end-to-end

Akarsh-Hegde:
## Why

The end-to-end pipeline (classification → worklog update) runs **three** on-device models, but onboarding only provisioned one:

- **LLM** `Qwen3.5-2B` — also does classification/matching (there is no separate classifier model).
- **Reranker** `Qwen3-Reranker-0.6B` — downloaded lazily on the first worklog run.
- **Embedder** `Qwen3-Embedding-0.6B` — downloaded lazily **and** its library (`sentence-transformers`/`torch`) was never shipped in the runtime, so `/distill_hour` would `ImportError` and crash the worklog pipeline at stage 1 on a fresh install.

Result: a fresh DMG could reach the dashboard before any model was ready, then stall or crash on its first worklog cycle.

## What

**Single source of truth + provision all three**
- New `services/agents/model_registry.py` — declares the three models with env-overridable ids (`MERIDIAN_LLM_ID` / `WORKLOG_RERANKER_ID` / `MERIDIAN_EMBEDDER_ID`) and per-model HF download filesets. `mlx_classifier`, `reranker`, and `session_distiller` now read from it.
- `routes/prefetch.py` + `_state.py` download the **whole set** with **aggregate progress**. The wire contract the tray decodes (`state`/`received`/`total`/`error`) is unchanged, so no Rust decode changes were needed.

**MLX-native embedder (no torch)**
- `session_distiller.py` swapped `sentence-transformers`/`torch` for `mlx_embeddings` — keeps the runtime lean (no ~2.5 GB torch) and on the single MLX backend. Added to the `[mlx]` extra. Pooling verified at source: `mlx_embeddings/models/qwen3.py` does last-token pooling + L2 normalize, matching the previous ST behaviour.

**Onboarding UX (the ask)**
- The Model step now **auto-installs the runtime and auto-downloads all models with no clicks**, shows aggregate progress, and **gates "Open Dashboard" on `modelReady`** so the user can't reach the dashboard until all three are on disk. A **Retry** button appears on error so the gate is never a dead end.
- `com.meridiona.mlx-server.plist` pins `HF_HOME` so the wizard's eager prefetch and the runtime's lazy loads resolve the same cache; `install.sh`'s first-run check now covers all three models.

## Verification
- Python: all modules compile; registry resolves all three specs; no stale `torch`/`sentence-transformers` refs.
- TypeScript: setup wizard typechecks clean.
- Rust: `cargo fmt` + `cargo clippy` + `cargo check` pass (workspace incl. tray).
- Embedder pooling correctness confirmed at the `mlx_embeddings` source level.

### Follow-up (network-blocked, not code)
A runtime cosine sanity check on the embedder (norms ≈ 1; related ≫ unrelated) is pending — the 8-bit shard repeatedly stalled on this machine's HF connection (the same download the product does on first run). Run on a good connection:
```
services/.venv/bin/python -c "from huggingface_hub import snapshot_download; snapshot_download('mlx-community/Qwen3-Embedding-0.6B-8bit')"
```
then exercise `agents.session_distiller._embed`.


adityaharishch (commented):
## Code review — 8 findings

### 🔴 High

**`tray/src-tauri/src/mlx_server.rs:726`** — `prefetch_model_cmd` returns `Ok(())` on any non-2xx from `/prefetch_model`, so `page.tsx` unconditionally sets `modelReady = true` even when no models were downloaded.

On an older runtime (pre-`/prefetch_model` endpoint), the POST returns 404 → `mlx_server.rs:726–731` explicitly returns `Ok(())` with a log line → `page.tsx:.then(() => setModelReady(true))` fires → wizard presents the model step as complete → user reaches the dashboard with zero model weights on disk → the first classify/distill/rerank call fails or falls through to a lazy download mid-pipeline, the exact stall this PR was meant to prevent.

---

### 🟡 Medium

**`services/agents/routes/prefetch.py` — `_run_prefetch()`** — when `_download_spec(spec)` raises, `models[i]["state"]` stays `"downloading"` forever.

The `models[i]["state"] = "done"` update lives in the `with prefetch_lock` block *after* the span, but the exception exits the span block and falls straight to the outer `except`. The outer handler sets `prefetch_state["state"] = "error"` but never touches the per-model rows. Result: `/prefetch_status` returns inconsistent state — top-level `state="error"` while `models[i]["state"]="downloading"`.

```python
# fix: add this inside the except handler's lock block
prefetch_state["models"][i]["state"] = "error"
```

---

**`services/agents/routes/prefetch.py` — `prefetch_model()`** — three size probes run sequentially, tripling pre-download latency.

Each `await run_in_threadpool(_spec_total_bytes, spec)` is a separate HF API call (~2–5 s on a slow connection). Three serial awaits add 6–15 s before the download thread even starts, during which `prefetch_state["total"] = 0` and the wizard's progress bar shows indeterminate.

```python
# fix: parallel gather
results = await asyncio.gather(
    *[run_in_threadpool(_spec_total_bytes, s) for s in specs],
    return_exceptions=True,
)
totals = [r if isinstance(r, int) else 0 for r in results]
```

---

**`services/agents/routes/prefetch.py` — `prefetch_status()`** — stale `state="downloading"` returned after completion.

`st = dict(prefetch_state)` snapshots `state="downloading"` under the lock. The lock is released. The disk walk (rglob over a large HF cache) runs unprotected. If `_run_prefetch` sets `prefetch_state["state"] = "done"` during the walk, the handler returns the stale snapshot — one extra "still downloading" tick before the next poll self-corrects.

---

**`services/agents/model_registry.py:170`** — all three `ModelSpec` instances share the *exact same* `_MLX_ALLOW_PATTERNS` list object; `frozen=True` doesn't protect against in-place mutation.

`LLM.allow_patterns is RERANKER.allow_patterns is EMBEDDER.allow_patterns` is `True`. Any call to `spec.allow_patterns.append(...)` or `.clear()` (a test, a future per-role filter, an eval strategy) silently corrupts all three specs simultaneously. `snapshot_download` then fetches wrong or no files for all three models and the downloads appear successful.

```python
# fix: use tuple, or give each spec its own copy
allow_patterns: tuple[str, ...] = tuple(_MLX_ALLOW_PATTERNS)
```

---

### 🟢 Low / latent

**`services/agents/session_distiller.py` — `_embed()`** — returns `np.zeros((0, 0))` for empty input instead of `np.zeros((0, D))`.

When `texts = []`, `out` stays empty and the function returns shape `(0, 0)`. Currently safe because `_distil()` guards on `len(lex) == 0` before indexing `V`. Latent: any future caller that reads `V.shape[1]` for the embedding dimension gets `0` instead of the actual output dimension, causing silent wrong shapes or misleading errors.

```python
# fix:
return np.empty((0, out[0].shape[1]), dtype=np.float32) if not out else np.concatenate(out, axis=0)
```

---

**`services/agents/routes/prefetch.py:144`** — model-identity idempotency guard removed; a runtime env-var model change can't trigger a re-prefetch while `state="done"`.

Old guard checked `same_model = prefetch_state.get("model_id") == model_id`; a different model ID fell through and restarted the download. New guard is state-only: `if prefetch_state["state"] in ("downloading", "done"): return`. If `MERIDIAN_LLM_ID` is changed at runtime and `POST /prefetch_model` is called again, the server returns the cached `"done"` dict without fetching the new weights.

---

**`tray/src-tauri/src/mlx_server.rs`** (runtime download loop) — `speed` reports cumulative average bytes/elapsed, not a rolling rate, so it stays misleadingly high during network stalls.

`speed = received / started.elapsed().as_secs_f64()` is the historical mean. After a fast burst + stall, the wizard shows e.g. "8 MB/s" while the connection is idle. The Python prefetch path correctly zeroes speed when `monotonic() - last_rate_ts > 3.0 s`. The two phases (Rust runtime tarball vs Python model download) give inconsistent speed semantics to the same `DownloadProgress` wire format.

Akarsh-Hegde:
## Fix summary — all review findings addressed

Commit: `07f3514f` pushed to `feat/onboarding-model-provisioning`.

---

### Findings addressed (code-review workflow + @adityaharishch)

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | 🔴 High | `start_mlx_server_cmd` failure swallowed (`.catch(() => {})`) | Changed to `.catch((e) => setMlxErr(String(e)))` |
| 2 | 🟡 Medium | `canNext: (s) => s.modelReady` gate | **Intentionally kept** — this is the PR's core deliverable; findings 1, 4, and 6 dissolve the trap by properly surfacing errors |
| 3 | 🟡 Medium | `snapshot_download` no stall guard | **Deferred** — OS socket timeout (~75 s) handles genuine stalls; a wall-clock thread timeout would kill slow-but-progressing downloads on bad connections |
| 4 | 🔴 High | Apple Intelligence guard removed from `prefetch_model()` | Restored: `if app_state.get("mlx_module") is None: return {"state": "done", ...}` |
| 5 | 🟢 Low | `_embed()` loads embedder for empty input | Added early return guard before `_get_embedder()` to skip model load when `texts = []` |
| 6 | 🟡 Medium | Inner `model_prefetch` OTEL span missing ERROR status | Added `try/except` inside span: `span.set_status(ERROR)` + `span.record_exception(exc)` + `models[i]["state"] = "error"` |
| 7 | 🟡 Medium | `pm_worklog_update/config.py` hardcoded LLM default | Changed to `os.environ.get("MLX_SERVER_MODEL") or model_registry.llm_id()` |
| 8 | 🟡 Medium | Provisioning errors appear on wrong steps | Separate `mlxErr` state for provisioning errors; only `err` (step errors) goes to the Footer |
| 9 | 🟢 Low | Rust `speed` uses cumulative average | **Deferred** — cosmetic display issue; not worth touching the streaming loop in this PR |
| 10 | 🟡 Medium | `mlx_classifier._get_model()` reads `MODEL_ID` module-constant | Changed to `model_registry.llm_id()` per-call; same for `_get_tokenizer()` |
| A1 | 🔴 High | `mlx_server.rs` returns `Ok(())` on any non-2xx from `/prefetch_model` | Narrowed graceful-degrade to 404-only; 5xx and other errors now return `Err` so wizard surfaces Retry |
| A2 | 🟡 Medium | Per-model `state` stays `"downloading"` after failure | Fixed by finding 6 fix above |
| A3 | 🟡 Medium | Three sequential HF size probes add 6–15 s pre-download latency | Parallelized with `asyncio.gather` |
| A4 | 🟡 Medium | Stale `state="downloading"` snapshot in `prefetch_status()` | Re-checks `prefetch_state["state"]` under lock after the rglob walk |
| A5 | 🟡 Medium | Shared `_MLX_ALLOW_PATTERNS` list across all three `ModelSpec` instances | Changed to `tuple[str, ...]` — immutable, each spec independently safe |
| A6 | 🟢 Low | `_embed()` returns `(0, 0)` for empty input | Same fix as finding 5 above |
| A7 | 🟢 Low | Model-identity idempotency guard removed | **Accepted risk** — `MERIDIAN_LLM_ID` runtime change is an edge case with no user-facing path |
| A8 | 🟢 Low | Rust speed display inconsistency | Same as finding 9 above |

---

All pre-push checks passed (cargo fmt, cargo clippy, cargo test, UI build, UI tests, security audit). PR is ready for re-review.

PR #345, opened 2026-06-26
fix(etl): suppress VS Code frames when focused terminal is a coding agent

adityaharishch:
## Summary

- Adds `is_coding_agent_terminal()` to `src/etl/session_builder.rs` — detects when the focused VS Code terminal tab is running a coding agent (Claude Code, Codex, Cursor agent, Copilot CLI, Gemini, Aider, Amazon Q)
- Adds a skip guard in `src/etl/runner.rs` so those frames are not converted into `app_sessions` rows — the coding-agent indexer already tracks these sessions, and double-capture was inflating "Code" focus time on the timeline
- Updates `dev-start.sh` so `cargo-watch` also watches the local screenpipe-fork directory

### Detection strategy

| Pattern | Example |
|---------|---------|
| Claude Code spinner (OSC title) | `Terminal - ⠂ agentic-worklog-pipeline` |
| Version-number label (idle/startup) | `Terminal - 2.1.193`, `Terminal - 0.139.0` |
| Process name contains-check | `Terminal - codex-aarch64-ap`, `Terminal - cursor-agent` |
| Amazon Q exact-word | `Terminal - q` (avoids `qemu`, `queue-worker` false positives) |

### Verified

`capture_frames` still receives all raw frames (written by the tray — unchanged). The filter only affects `app_sessions`. Confirmed no new spinner/version-label sessions are written to `app_sessions` with dev daemon running.

## Test plan

- [ ] `cargo test` passes (unit tests in `session_builder.rs` cover all 5 detection paths)
- [ ] After merge: verify `app_sessions` no longer contains "Code" sessions with spinner window titles while Claude Code is running in an integrated terminal


Akarsh-Hegde (commented):
## Code Review — PR #345 (high-effort, 49 agents, 39 verified → 10 reported)

This PR adds VS Code coding-agent terminal frame suppression to prevent double-counting sessions already tracked by the coding-agent indexer. The approach is directionally correct but has one critical bug that causes the opposite of the intended fix on long terminal sessions, and several over-broad heuristics that silently drop legitimate VS Code work.

---

### 🔴 Critical — data corruption

**[1] `src/etl/runner.rs:187` — skip does not advance `block_last_ts` / `block_last_frame_id`, freezing the gap detector clock**

The new `continue` for coding-agent terminal frames does not update `block_last_ts` or `block_last_frame_id`. After >300 s of skipped frames, the next real VS Code editor frame triggers `timestamp_gap_secs(block_last_ts, now) > 300` → spurious `system_sleep` gap row, the surrounding VS Code session is split in two, and 5+ minutes of active editing is marked as machine-asleep on the timeline.

Compare the empty-app skip immediately above (lines 162-165), which explicitly advances both fields before its `continue` to prevent exactly this. The new skip omitted those advances.

**Fix:** Mirror the empty-app skip pattern:
```rust
if is_vscode_like(app) && is_coding_agent_terminal(window) {
    if current_app.is_some() {
        block_last_ts = frame.timestamp.clone();
        block_last_frame_id = frame.id;
    }
    debug!(...);
    continue;
}
```

---

### 🔴 High — silent data loss (false positives)

**[2] `src/etl/session_builder.rs:263` — unanchored substring match silently drops VS Code frames for any tab whose title contains an agent name**

The `CODING_AGENT_NAMES` `.contains()` check is unanchored. A terminal tab titled `Terminal - decodex-runner`, `Terminal - aider-helper`, or `Terminal - gitclaude` matches `contains("codex")`, `contains("aider")`, or `contains("claude")` respectively — all VS Code frames while that tab is focused are suppressed, losing that work from `app_sessions` with no log entry.

**Fix:** Match on word boundaries or exact first-word: `session_lower.split_whitespace().next() == Some(name)` for the main name check, consistent with how `"q"` is already handled.

**[3] `src/etl/session_builder.rs:246` — `is_version_label` suppresses any `Terminal - N.N.N` tab, not just coding-agent version labels**

`is_version_label` matches any string of purely numeric dot-separated segments (≥3). A developer running a script in a pinned Node terminal (`Terminal - 20.11.0`), a Python interpreter (`Terminal - 3.11.5`), or any process whose VS Code tab label resolves to a bare version number will have all their VS Code frames silently suppressed.

**Fix:** Anchor `is_version_label` to a known-agent prefix or remove it — it's covering for the fact that some agent CLIs display their version rather than their binary name during startup, which the `CLAUDE_SPINNERS` check already handles.

---

### 🟡 Medium — suppression silently fails

**[4] `src/etl/runner.rs:176` — `window` is resolved from `browser_url` first; VS Code frames with a non-empty `browser_url` never suppress**

`window` is built as `browser_url → url_domain` with `window_name` as fallback. When VS Code's embedded browser (Live Server, Vite preview, etc.) is focused, the frame carries a non-empty `browser_url`, so `window` becomes `"localhost"` instead of `"Terminal - claude"`. `is_coding_agent_terminal("localhost")` returns `false` — the frame is not suppressed, and the original double-count bug still fires for those configurations.

**Fix:** Always use `window_name` as the suppression input, regardless of `browser_url`. The embedded-browser vs terminal distinction lives in `window_name`, not `browser_url`.

**[5] `src/etl/session_builder.rs:232` — only `"Terminal - "` (ASCII hyphen) is matched; `"Terminal — "` (em-dash) is never suppressed** *(PLAUSIBLE)*

A non-default `terminal.integrated.tabs.title` setting or certain locales produces `"Terminal — claude"` (U+2014). `strip_prefix("Terminal - ")` fails → suppression never fires → the original double-count bug persists for those users.

**Fix:** Also try `strip_prefix("Terminal \u{2014} ")` as a fallback.

---

### 🟡 Medium — design / altitude

**[6] `src/etl/session_builder.rs:443` — detection re-derives "coding agent is running" from tab-title heuristics instead of querying the authoritative indexer rows**

The coding-agent indexer already writes `app_sessions` rows with `task_method = 'coding_agent_live'` — the authoritative "is a coding agent running right now" signal is already in `meridian.db` by timestamp. This PR builds a parallel, growing hardcoded list of process names (`CODING_AGENT_NAMES`) to reconstruct the same signal from a11y window titles. Adding a new agent requires updating this list and redeploying the daemon; the list and the indexer can diverge permanently without any warning.

The deeper fix is to query `app_sessions WHERE task_method = 'coding_agent_live' AND started_at <= :frame_ts AND (ended_at IS NULL OR ended_at >= :frame_ts)` rather than maintaining a second list.

---

### 🟡 Medium — dev tooling

**[7] `dev-start.sh:82` — hardcoded absolute path `/Users/adityaharish/Documents/Meridiona/screenpipe-fork` committed to shared script**

Any developer other than adityaharish runs `dev-start.sh` and `cargo watch` either silently skips the watch target (cargo-watch ≥8) or errors out — the daemon does not hot-reload on screenpipe-fork changes for anyone else.

**Fix:** Either parameterize via an env var (`SCREENPIPE_FORK_PATH`), skip the flag if the path doesn't exist, or remove the `--watch` flag for the fork path and document it in a local `.env.local`.

---

### 🟡 Medium — test coverage

**[8] `src/etl/runner.rs:169` — no integration test covers the new skip branch**

CLAUDE.md: *"Integration tests live in `tests/integration_etl.rs` … they must continue to pass after any ETL or schema change."* The PR adds only unit tests on the predicate function; there is no in-memory ETL test that feeds `VS Code editor → coding-agent terminal → editor` frames through `run_etl()` and asserts (a) no spurious gap row and (b) the VS Code session is not split. Finding #1 above ships undetected precisely because this test is absent.

**Required:** Add an integration test in `tests/integration_etl.rs` with:
- sequence of frames: VS Code editor, 3+ coding-agent terminal frames spanning >300 s, VS Code editor again
- assertion: 0 gap rows inserted, 1 VS Code `active_session` (not split)

---

### 🟢 Low — cleanup

**[9] `src/etl/session_builder.rs:422` — `is_version_label` allocates a `Vec` on every call**

```rust
let parts: Vec<&str> = s.split('.').collect();
parts.len() >= 3 && parts.iter().all(...)
```

Allocates a heap vec on every frame. Use an iterator chain instead:
```rust
let mut parts = s.split('.');
parts.by_ref().count() >= 3 && s.split('.').all(|p| !p.is_empty() && p.chars().all(|c| c.is_ascii_digit()))
```
Or equivalently: `s.split('.').count() >= 3 && s.split('.').all(|p| ...)`.

**[10] `src/etl/session_builder.rs:468` — `"q"` in `CODING_AGENT_NAMES` is dead weight**

`"q"` is added to `CODING_AGENT_NAMES` but then explicitly filtered out by `.filter(|&&name| name != "q")` in the only `.any()` call that uses the array. The `"q"` case is handled by the `first_word == "q"` check above. The dead entry is a latent trap: a future maintainer who removes the filter to simplify the loop accidentally re-enables `contains("q")` matching on `"qemu"`, `"queue-worker"`, etc.

**Fix:** Remove `"q"` from `CODING_AGENT_NAMES`; keep only the `first_word == "q"` path.

---

### Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 Critical | `runner.rs:187` | Skip doesn't advance gap-detector clock → spurious `system_sleep` splits VS Code session |
| 2 | 🔴 High | `session_builder.rs:263` | Unanchored `.contains()` → silent data loss on tabs with agent-name substrings |
| 3 | 🔴 High | `session_builder.rs:246` | `is_version_label` over-broad → suppresses any `Terminal - N.N.N` tab |
| 4 | 🟡 Medium | `runner.rs:176` | `browser_url` takes precedence → suppression silently fails when embedded browser is focused |
| 5 | 🟡 Medium | `session_builder.rs:232` | Em-dash separator not matched → suppression fails on localized/custom VS Code |
| 6 | 🟡 Medium | `session_builder.rs:443` | Parallel agent-name list diverges from authoritative indexer rows in DB |
| 7 | 🟡 Medium | `dev-start.sh:82` | Hardcoded personal absolute path breaks every other developer's hot-reload |
| 8 | 🟡 Medium | `runner.rs:169` | No integration test for new skip branch — finding #1 ships undetected |
| 9 | 🟢 Low | `session_builder.rs:422` | Unnecessary `Vec` allocation in hot path |
| 10 | 🟢 Low | `session_builder.rs:468` | Dead `"q"` entry in `CODING_AGENT_NAMES` is a latent trap |

Findings #1, #2, #3 are must-fix before merge. Finding #8 (integration test) is also required per CLAUDE.md. The rest can be addressed in a follow-up.

adityaharishch:
Thanks for the thorough review. All findings addressed in commit `a2c20411`:

**🔴 #1 (critical — spurious gap)**: Fixed — `block_last_ts`/`block_last_frame_id` now advance on every skipped coding-agent frame, mirroring the empty-app skip pattern. Integration test `coding_agent_skip_does_not_produce_spurious_gap` covers this: editor → 330 s of coding-agent frames → editor asserts 0 gap rows.

**🔴 #2 (unanchored contains)**: Fixed — replaced `.contains(name)` with `first_word.starts_with(name)` anchored to the first space-delimited token. `"decodex-runner"` no longer matches `"codex"`, `"gitclaude"` no longer matches `"claude"`. Architecture-suffixed binaries (`codex-aarch64-ap`) still match via `starts_with("codex")`.

**🔴 #3 (is_version_label too broad)**: Fixed — removed `is_version_label` entirely. It suppressed any `Terminal - N.N.N` tab including Node REPL (`20.11.0`) and Python (`3.11.5`) sessions. The reviewer is right that the spinner check already covers the Claude Code active case, and Claude Code idle/startup is an acceptable miss.

**🟡 #4 (browser_url shadows window_name)**: Fixed — the suppression check now always uses `frame.window_name` directly (`raw_window`) instead of the `url_domain`-resolved `window`.

**🟡 #5 (em-dash separator)**: Fixed — `is_coding_agent_terminal` now tries both `"Terminal - "` (ASCII) and `"Terminal — "` (U+2014).

**🟡 #7 (hardcoded personal path)**: Fixed — `dev-start.sh` now reads `SCREENPIPE_FORK_PATH` env var and omits the `--watch` flag when it's not set.

**🟡 #8 (integration test)**: Added `tests/etl_coding_agent_skip.rs` with 3 tests: no spurious gap, editor frames not suppressed, coding-agent frames excluded from frame_count.

**🟢 #9 (Vec allocation)**: Eliminated by removing `is_version_label`.

**🟢 #10 (dead "q" in array)**: Removed — `"q"` is now only in the `first_word == "q"` exact-match path.

---

**🟡 #6 (design — authoritative DB query)**: Not addressed in this PR. The reviewer's point is valid architecturally, but querying `app_sessions WHERE task_method = 'coding_agent_live'` wouldn't work at ETL time — the indexer seals rows up to 1 h after the session ends (idle-seal backstop), so in-progress coding-agent sessions have no DB row yet when the ETL runs. The heuristic approach is the only option that works in real-time. Happy to add a code comment explaining this constraint if useful.

PR #346, opened 2026-06-26
fix(capture): fall back to OCR when browser a11y tree is chrome-only

adityaharishch:
## Summary

- Chrome's a11y tree returns non-empty text even before page content is materialised — the tab strip (`AXTab`/`AXButton` nodes) contributes ~695 chars of tab titles alone
- `try_walk_a11y` had no thinness check, so it returned an `Accessibility` frame from chrome-only text and the OCR fallback never ran → actual page content was never captured
- Ports the content-density heuristic from `screenpipe-capture::paired_capture::a11y_content_is_thin` (the screenpipe CLI path that had this working): classifies each node's role as chrome vs content, falls back to OCR when content-role chars < 30% of total (or < 100 total chars)

## Root cause

`tray/src-tauri/src/capture/screenpipe.rs:try_walk_a11y` only checked for an empty `text_content` to decide whether to OCR. Chrome's tab strip is non-empty (always present), so it always "won" over OCR — page content was never captured from day one on the screenpipe-fork tray path.

The screenpipe CLI (`paired_capture.rs`) already had `a11y_content_is_thin` solving exactly this; the tray's simplified engine was missing it.

## Test plan

- [ ] Open Chrome to any page with visible text content
- [ ] Verify `capture_frames` rows now have `accessibility_text` / `full_text` with actual page text, not just tab titles
- [ ] Verify incognito windows still produce no frames (the `Skipped` path is untouched)
- [ ] Verify non-browser apps (VS Code, Slack) still use the a11y path (their nodes are mostly `AXStaticText`/`AXGroup` → ratio > 0.3)

### DIFF
--- a/tray/src-tauri/src/capture/screenpipe.rs
+++ b/tray/src-tauri/src/capture/screenpipe.rs
@@ -109,12 +109,23 @@ fn try_walk_a11y(walker: &dyn TreeWalkerPlatform) -> A11yOutcome {
                 debug!(app = %snap.app_name, "capture: a11y tree empty (warm-up) — OCR fallback");
                 return A11yOutcome::FallBackToOcr;
             }
+            let app_lower = snap.app_name.to_lowercase();
+            // Canvas-rendered apps (Figma, design tools): content is GPU pixels,
+            // not in the AX tree. The a11y snapshot gives only toolbar/menu text;
+            // OCR is the only path to actual canvas content.
+            if is_canvas_app(&app_lower) {
+                debug!(
+                    app = %snap.app_name,
+                    "capture: canvas app — a11y can't reach GPU content, OCR fallback"
+                );
+                return A11yOutcome::FallBackToOcr;
+            }
             // Browsers return non-empty a11y text even when the page hasn't been
             // exposed: the tab strip (AXTab/AXButton nodes) alone gives ~695 chars
-            // of tab titles. Without this check we'd emit chrome-only text and
-            // never capture actual page content. Fall back to OCR so the screen
-            // pixels give us real page content for this tick.
-            if a11y_content_is_thin(&snap) {
+            // of tab titles. Gate on is_browser so we don't apply the density
+            // heuristic to non-browser apps (sparse dialogs, preferences panes, etc.)
+            // where thin a11y text is expected and OCR adds no value.
+            if is_browser(&app_lower) && a11y_content_is_thin(&snap) {
                 debug!(
                     app = %snap.app_name,
                     chars = snap.text_content.len(),
@@ -151,6 +162,32 @@ fn try_walk_a11y(walker: &dyn TreeWalkerPlatform) -> A11yOutcome {
     }
 }
 
+/// Returns `true` when `app_lower` is a known browser.
+fn is_browser(app_lower: &str) -> bool {
+    [
+        "chrome", "safari", "firefox", "arc", "edge", "brave", "opera", "vivaldi",
+    ]
+    .iter()
+    .any(|b| app_lower.contains(b))
+}
+
+/// Returns `true` when `app_lower` is a canvas-rendering design tool whose
+/// content is drawn to a GPU surface and is therefore absent from the AX tree.
+fn is_canvas_app(app_lower: &str) -> bool {
+    [
+        "figma",
+        "canva",
+        "miro",
+        "sketch",
+        "origami",
+        "principle",
+        "framer",
+        "penpot",
+    ]
+    .iter()
+    .any(|a| app_lower.contains(a))
+}
+
 /// Returns `true` when the a11y snapshot contains mostly browser-chrome roles
 /// (tab strip, toolbar, buttons) rather than real page content. When this is
 /// the case, the tray falls back to OCR so we capture what's actually on screen.
@@ -180,6 +217,13 @@ fn a11y_content_is_thin(snap: &TreeSnapshot) -> bool {
         "AXComboBox",
         "AXScrollBar",
     ];
+
+    // If the walker returned no node metadata (max-node budget exhausted and
+    // all nodes were elided), we have no role data to reason about — treat as
+    // non-thin so the text_content we do have is used rather than discarded.
+    if snap.nodes.is_empty() {
+        return false;
+    }
 
     let mut content_chars: usize = 0;
     let mut total_chars: usize = 0;
--- a/tray/src-tauri/src/capture/screenpipe.rs
+++ b/tray/src-tauri/src/capture/screenpipe.rs
@@ -31,7 +31,7 @@ use std::time::Duration;
 
 use chrono::Utc;
 use screenpipe_a11y::tree::{
-    create_tree_walker, TreeWalkResult, TreeWalkerConfig, TreeWalkerPlatform,
+    create_tree_walker, TreeSnapshot, TreeWalkResult, TreeWalkerConfig, TreeWalkerPlatform,
 };
 use screenpipe_screen::capture_screenshot_by_window::{capture_all_visible_windows, WindowFilters};
 use tracing::{debug, info, warn};
@@ -109,6 +109,19 @@ fn try_walk_a11y(walker: &dyn TreeWalkerPlatform) -> A11yOutcome {
                 debug!(app = %snap.app_name, "capture: a11y tree empty (warm-up) — OCR fallback");
                 return A11yOutcome::FallBackToOcr;
             }
+            // Browsers return non-empty a11y text even when the page hasn't been
+            // exposed: the tab strip (AXTab/AXButton nodes) alone gives ~695 chars
+            // of tab titles. Without this check we'd emit chrome-only text and
+            // never capture actual page content. Fall back to OCR so the screen
+            // pixels give us real page content for this tick.
+            if a11y_content_is_thin(&snap) {
+                debug!(
+                    app = %snap.app_name,
+                    chars = snap.text_content.len(),
+                    "capture: a11y tree is browser-chrome-only — OCR fallback for page content"
+                );
+                return A11yOutcome::FallBackToOcr;
+            }
             debug!(
                 app = %snap.app_name,
                 chars = snap.text_content.len(),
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde (commented):
## Code Review — PR #346 (high-effort, 46 agents, 36 verified → 10 reported)

PR #346 (`fix(capture): fall back to OCR when browser a11y tree is chrome-only`) adds `a11y_content_is_thin` to the tray capture path plus the same VS Code coding-agent frame-skip from PR #345. Two independent correctness clusters surface: the gap-detector clock is frozen by the new skip (identical root cause to #345's finding #1), and `a11y_content_is_thin` is applied unconditionally to all apps rather than being gated to browsers, corrupting session text for native apps with sparse AX trees.

---

### 🔴 Critical — data corruption

**[1] `src/etl/runner.rs:376` — coding-agent skip does not advance `block_last_ts`/`block_last_frame_id`, freezing the gap detector clock**

Same root cause as PR #345 finding #1. The new `continue` at line 376 does not update `block_last_ts` or `block_last_frame_id`. After >300 s of skipped coding-agent frames, the next real VS Code editor frame measures a gap from the stale clock, `count_frames_in_window` classifies the skipped agent frames as `system_sleep` (none have `capture_trigger = 'idle'`), and the ETL inserts a spurious `system_sleep` gap row and prematurely closes the editor session — truncating active work time.

The empty-app skip immediately above (lines 160–167) explicitly advances both fields before `continue` to prevent exactly this. The new skip omits those advances.

**Fix:**
```rust
if is_vscode_like(app) && is_coding_agent_terminal(window) {
    if current_app.is_some() {
        block_last_ts = frame.timestamp.clone();
        block_last_frame_id = frame.id;
    }
    debug!(...);
    continue;
}
```

---

### 🔴 High — capture data corruption for native apps

**[2] `tray/src-tauri/src/capture/screenpipe.rs:580` — `a11y_content_is_thin` called unconditionally for all apps, not just browsers**

`a11y_content_is_thin` is inserted in `try_walk_a11y` with no guard on `snap.app_name`. Any native macOS app with a sparse AX tree — a permissions dialog, a preferences pane, a toolbar-heavy modal — satisfies the `total_chars < 100` threshold and falls back to OCR. The a11y text that was present is discarded, and OCR (which may not cover the focused window accurately) is stored instead, corrupting `session_text` for that app.

The upstream `screenpipe-capture::paired_capture::a11y_content_is_thin` is intentionally browser-scoped. The port loses that scope.

**Fix:** Gate the call on `is_browser(snap.app_name)` (or the equivalent `is_browser_like` predicate already used elsewhere in the file) before calling `a11y_content_is_thin`.

**[3] `tray/src-tauri/src/capture/screenpipe.rs:151` — ported `a11y_content_is_thin` omits the canvas-app, canvas-URL, and meeting-app early-return branches from the upstream function**

The upstream `screenpipe-capture::paired_capture::a11y_content_is_thin` accepts `window_name`, `browser_url`, and `app_name` and short-circuits via three early-return guards (steps 1a–1c) before the density heuristic:
- **1a** — window title matches `CANVAS_APP_PATTERNS` (Figma, Google Docs, etc.) → return `false` (do NOT fall back to OCR)
- **1b** — `browser_url` matches canvas content URLs → return `false`
- **1c** — app name matches `MEETING_APP_PATTERNS` (Zoom, Google Meet, etc.) → return `false`

The ported version has none of these. Figma and Google Meet will always report thin a11y content and fall back to OCR. More critically, future improvements to these lists in `paired_capture.rs` will not propagate here — the two copies silently diverge.

**Fix:** Either add the three early-return checks matching the upstream signature, or (better) factor the function into a shared crate rather than duplicating it.

---

### 🔴 High — silent data loss (false positives in suppression)

**[4] `src/etl/session_builder.rs:474` — unanchored `.contains()` fires on substrings, silently dropping legitimate VS Code frames**

Same root cause as PR #345 finding #2. `CODING_AGENT_NAMES` uses `.contains(name)` with no word-boundary anchor. Terminal tabs named `Terminal - raider` (matches `"aider"`), `Terminal - copilot-node-modules-checker` (matches `"copilot"`), or any path containing an agent name are all suppressed — that work session disappears from `app_sessions` with no log entry.

**Fix:** Use first-word exact-match for all names, consistent with how `"q"` is already handled: `session_lower.split_whitespace().next() == Some(name)`.

**[5] `src/etl/session_builder.rs:457` — `is_version_label` matches IP addresses and any `N.N.N` dotted-decimal string, not just agent version labels**

`is_version_label` returns `true` for any string of purely numeric dot-separated segments with ≥3 parts. An SSH session into a remote host sets the VS Code tab to `Terminal - 192.168.1.100` — `is_version_label` returns `true`, the frame is suppressed, and the entire SSH session produces no `app_sessions` rows. That block of work is silently lost.

**Fix:** Either remove `is_version_label` entirely (the spinner-character check already handles the transient startup state it was targeting), or anchor it more tightly (e.g. require the first segment to be a small integer <100, or use a named-agent prefix check instead).

---

### 🟡 Medium — suppression reliability

**[6] `tray/src-tauri/src/capture/screenpipe.rs:198` — `a11y_content_is_thin` derives `total_chars` from `snap.nodes` text, not `snap.text_content`** *(PLAUSIBLE)*

If a `TreeSnapshot` is produced where `text_content` is populated but `nodes` is shorter than expected (e.g. mid-walk truncation at `max_nodes`), `total_chars` from the nodes loop will be zero, `a11y_content_is_thin` returns `true`, and every subsequent tick for that window falls back to OCR — silently discarding valid a11y text. The scenario requires a specific truncation path; confirm by checking whether `max_nodes` limits can leave `text_buffer` ahead of `state.nodes`.

**[7] `src/etl/session_builder.rs:444` — `is_coding_agent_terminal` only matches `"Terminal - "` (ASCII hyphen); VS Code with a non-default title template may not match** *(PLAUSIBLE)*

`strip_prefix("Terminal - ")` fails if the tab title uses an em-dash separator (`Terminal — claude`, U+2014), a renamed tab, or a split-pane AXDescription format. In those cases `is_coding_agent_terminal` returns `false` and the coding-agent terminal frames are not suppressed, re-introducing the overlap the fix was meant to prevent.

**Fix:** Also try `strip_prefix("Terminal \u{2014} ")` as a fallback.

---

### 🟡 Medium — design / maintainability

**[8] `src/etl/session_builder.rs:401` — `CODING_AGENT_NAMES` is a fourth independent copy of agent process-name knowledge**

Agent process names are already enumerated in at least three other places: `src/coding_agent_session_ingest/indexer.rs` (l.379–383), `health/codingagent.rs` (l.22–24), and various summariser modules. Adding a new agent to the ingest pipeline requires updating all of these separately with no compile-time guard. The ETL and health check can diverge silently.

The deeper fix is a single shared `const CODING_AGENT_PROCESS_NAMES: &[&str]` in the ingest module, re-exported and referenced from `session_builder.rs`.

---

### 🟡 Medium — dev tooling

**[9] `dev-start.sh:82` — hardcoded absolute path `/Users/adityaharish/Documents/Meridiona/screenpipe-fork` committed to the shared script**

Identical to PR #345 finding #7. On cargo-watch 8.5.3+ a nonexistent `--watch` path causes the watcher to fail to start, breaking hot-reload for every contributor except the author.

**Fix:** Parameterize via `SCREENPIPE_FORK_PATH` env var, or skip the `--watch` flag when the path does not exist: `${SCREENPIPE_FORK_PATH:+--watch "$SCREENPIPE_FORK_PATH"}`.

---

### 🟢 Low — cleanup

**[10] `src/etl/session_builder.rs:401` — `"q"` in `CODING_AGENT_NAMES` is dead but a latent trap**

`"q"` is listed in `CODING_AGENT_NAMES` then immediately excluded by `.filter(|&&name| name != "q")` at the only call site that uses the array. The `"q"` case is handled by the separate `first_word == "q"` branch. The dead entry is confusing and latently dangerous: a future maintainer removing the filter to simplify the loop accidentally re-enables `contains("q")` matching on `"qemu"`, `"queue-worker"`, etc.

**Fix:** Remove `"q"` from `CODING_AGENT_NAMES`; keep only the `first_word == "q"` path.

---

### Summary

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 Critical | `runner.rs:376` | Skip doesn't advance gap-detector clock → spurious `system_sleep` splits VS Code session |
| 2 | 🔴 High | `screenpipe.rs:580` | `a11y_content_is_thin` called on all apps, not just browsers → corrupts native-app session text |
| 3 | 🔴 High | `screenpipe.rs:151` | Ported function omits canvas/meeting early-returns → Figma, Zoom, Meet always fall back to OCR |
| 4 | 🔴 High | `session_builder.rs:474` | Unanchored `.contains()` → silent data loss on tabs with agent-name substrings |
| 5 | 🔴 High | `session_builder.rs:457` | `is_version_label` matches IP addresses → SSH sessions silently vanish from timeline |
| 6 | 🟡 Medium | `screenpipe.rs:198` | `total_chars` from nodes loop, not `text_content` — may return false-thin if nodes truncated |
| 7 | 🟡 Medium | `session_builder.rs:444` | Em-dash VS Code title format not matched → suppression misses non-default tab templates |
| 8 | 🟡 Medium | `session_builder.rs:401` | `CODING_AGENT_NAMES` is a 4th independent copy of agent names — divergence risk |
| 9 | 🟡 Medium | `dev-start.sh:82` | Hardcoded personal absolute path breaks every other dev's hot-reload |
| 10 | 🟢 Low | `session_builder.rs:401` | Dead `"q"` in `CODING_AGENT_NAMES` is a latent trap |

Findings #1–5 are must-fix before merge. #2 and #3 are new to this PR (the `a11y_content_is_thin` cluster). #1, #4, #5, #9 are the same root causes as findings in PR #345 — the two PRs share code and the same issues travel with it.

adityaharishch:
Thanks for the thorough review. All findings addressed in commit `469621cc`.

**Finding #1 — clock freeze (Critical):** `block_last_ts` and `block_last_frame_id` now advance on every coding-agent skip when a block is open. A burst of skipped frames no longer freezes the gap-detector clock, so no spurious `system_sleep` gap fires when real editor frames resume.

**Finding #2 — unconditional thin check (High):** `a11y_content_is_thin` is now gated on `is_browser(&app_lower)` inside `try_walk_a11y`. Permission dialogs, preferences panes, and any non-browser app with legitimately sparse a11y text are no longer pushed to OCR.

**Finding #3 — missing canvas-app / canvas-URL / meeting-app branches (High):** Added `is_canvas_app()` with an early-return `FallBackToOcr` branch in `try_walk_a11y` for GPU-rendered design tools (Figma, Miro, Sketch, Canva, Framer, Penpot, Origami, Principle). These apps' content is absent from the AX tree; OCR is the only useful path. Meeting apps are left on the default a11y path — video frames aren't OCR-able, and a11y gives toolbar/chat text which is meaningful.

**Finding #4 — unanchored `.contains()` (High):** Replaced `session_lower.contains(name)` with `first_word.starts_with(name)`. `"notcodex"` or a process that happens to mention "claude" in a path argument no longer matches.

**Finding #5 — `is_version_label` false positives (High):** Removed `is_version_label` entirely. Semver strings like `"2.1.193"` won't be detected, but that's the right tradeoff — the heuristic also matched Maven artifact strings, build-tool output, and anything with three dot-separated numbers.

**Finding #6 — empty-nodes guard (Medium):** `a11y_content_is_thin` now returns `false` early when `snap.nodes` is empty. If the node budget was exhausted and we have text but no role metadata, we keep the a11y text rather than discarding it.

**Finding #7 — em-dash separator (Medium):** `is_coding_agent_terminal` now strips both `"Terminal - "` (hyphen) and `"Terminal \u{2014} "` (em-dash) prefixes. Tests cover both forms.

**Finding #8 — code duplication:** Acknowledged. The `is_browser` in `screenpipe.rs` and `session_builder.rs` stay separate for now (different layers, different call sites, different import graphs). Factoring into `meridian-core` is a clean-up deferred for a later pass.

**Finding #9 — hardcoded `screenpipe-fork` path (Medium):** `dev-start.sh` now uses `${SCREENPIPE_FORK_PATH:-}` — the `--watch` flag is only appended when the var is set and the directory exists.

**Finding #10 — dead `"q"` in `CODING_AGENT_NAMES` (Low):** Removed from the array. The exact first-word `== "q"` check below it is the live gate.

PR #350, opened 2026-06-27
fix(prefetch): disable Xet + retry-with-resume to stop 4h download hangs

Akarsh-Hegde:
## Problem

A first-run model prefetch **wedged at 57% for 4+ hours** on a healthy 16 Mbps link (the bar froze, bytes never resumed).

**Root cause:** `hf_xet` does its transfer in Rust with **no per-read timeout** and can't be interrupted from Python. When a connection to the Xet CAS endpoint goes half-open (stops sending bytes but never errors — e.g. a firewall/proxy/regional block on `*.xethub.hf.co`), `snapshot_download` blocks the download thread **forever**: `prefetch_state` stays `downloading` and the wizard bar sits at the last byte count indefinitely.

## Fix (`services/agents/routes/prefetch.py`, `_download_spec`)

- **Disable Xet for the prefetch** by flipping `huggingface_hub`'s module constant (`HF_HUB_DISABLE_XET`) — *not* the env var, which is parsed once at import. This forces the classic LFS path, which caps every read at `HF_HUB_DOWNLOAD_TIMEOUT` (10s) and **resumes from the `.incomplete` partial via a Range request**, so a stall raises and self-heals instead of hanging.
- **Bounded retry-with-resume** — 5 attempts, 2/4/8/16s backoff (`MERIDIAN_PREFETCH_MAX_ATTEMPTS` overridable). Each retry resumes from disk, so transient drops converge without re-pulling.

### Why dropping Xet is fine here
A/B on the same model showed **Xet-on == Xet-off == network ceiling** (no benefit on a cold first-run pull of distinct weights — chunk-dedup buys little), while parallel range requests did **not** beat a single stream (link-bound, not per-connection-throttled). So this trades an unused fast-path for a bounded, resumable one.

**Bonus:** the classic path grows the on-disk blob linearly, so `/prefetch_status`'s byte-delta speed reads **smoothly** instead of sitting at 0 then bursting as hf_xet flushed reconstructed chunks (fixes the frozen-looking bar).

## Verification

Full 3-model prefetch (~2.46 GB) completed in **~11 min** at the link ceiling with smooth progress and no hang. A mid-download stall (dropped to ~1 MB/s for ~90s) **recovered on its own** instead of wedging — exactly the timeout+resume behavior intended.

## Notes

- Python-only change; no Rust touched. Pre-push suite (fmt/clippy/test/audit) passed.
- This commit was originally pushed onto the already-merged `fix/prefetch-speed-state-lint` (PR #348, closed) — re-homed here on a fresh branch off `pre-main` so it can actually land.

### DIFF
--- a/services/agents/routes/prefetch.py
+++ b/services/agents/routes/prefetch.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 import threading
 import time
 from pathlib import Path
@@ -31,10 +32,11 @@ log = logging.getLogger("agents.server")
 router = APIRouter()
 
 # Speed is derived from the growth of on-disk `received` between /prefetch_status
-# polls — NOT from a tqdm hook. hf_xet (the Xet accelerator these models use) does
-# its download in Rust and bypasses the classic tqdm bar, so a tqdm_class never
-# fires; the byte-delta approach works for any backend. Stored as a dict so both
-# functions can mutate it in-place without `global` declarations.
+# polls — NOT from a tqdm hook (a tqdm_class override never fires reliably across
+# backends). The byte-delta approach is backend-agnostic; with Xet disabled for the
+# prefetch (see `_download_spec`) the classic path grows the blob linearly, so this
+# now reads as a smooth rate rather than the 0-then-burst pattern hf_xet produced.
+# Stored as a dict so both functions can mutate it in-place without `global`s.
 _speed_state: dict[str, float | int] = {"recv": 0, "ts": 0.0}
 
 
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde (commented):
## Self-review of `_download_spec`

Two issues worth fixing before merge, plus two intentional-design notes.

### 🟠 1. `_PREFETCH_MAX_ATTEMPTS` can crash server startup (prefetch.py:89)
```python
_PREFETCH_MAX_ATTEMPTS = max(1, int(os.environ.get("MERIDIAN_PREFETCH_MAX_ATTEMPTS", "5")))
```
`int()` runs at **module import**. If an operator sets `MERIDIAN_PREFETCH_MAX_ATTEMPTS` to a non-integer (typo, `"5s"`, empty string), this raises `ValueError` and the **entire MLX server fails to import/start** — a config knob should never be able to brick the process. Fix: parse defensively and fall back to the default (with a warning) on bad input.

### 🟠 2. `assert` guards the exhaustion-raise (prefetch.py:146)
```python
assert last_exc is not None  # loop ran ≥1 time; only reached on exhaustion
raise last_exc
```
Asserts are stripped under `python -O`. The invariant holds today (the loop runs ≥1 time and the failure path always sets `last_exc`), so it's safe — but it's fragile to rely on an assert for control flow. Cleaner: drop `last_exc` entirely and put a bare `raise` on the final attempt inside the loop, so the exhaustion path re-raises the active exception unconditionally.

### 🟢 3. Broad `except Exception` (intentional)
Retries every error, including non-retryable ones (auth/404/disk-full), 5× before surfacing. Kept broad on purpose: the failures we care about are stalls/half-open connections whose exception type is backend-dependent (hf_xet / requests / connection) and easy to under-match; the exponential backoff caps wasted time at ~30s; and the upfront size-probe already surfaces repo/auth problems before the download thread starts. Narrowing risks failing fast on a transient error we should have retried.

### 🟢 4. Process-global Xet disable never restored (intentional)
`hf_constants.HF_HUB_DISABLE_XET = True` mutates a process-wide module constant and is deliberately left set, so later in-process lazy model loads inherit the same bounded, resumable path on a network where Xet stalls. Documented in the docstring.

Fixing #1 and #2 now.

Akarsh-Hegde:
## Fixes pushed (`4f4b944c`)

Resolved both findings from the review above:

### ✅ 1. Startup-crash on bad env value — fixed
Replaced the bare `int(os.environ.get(...))` with a defensive helper:
```python
def _env_positive_int(name: str, default: int) -> int:
    raw = os.environ.get(name)
    if raw is None:
        return default
    try:
        return max(1, int(raw))
    except ValueError:
        log.warning("server: ignoring non-integer env override; using default",
                    extra={"env_var": name, "raw": raw, "default": default})
        return default

_PREFETCH_MAX_ATTEMPTS = _env_positive_int("MERIDIAN_PREFETCH_MAX_ATTEMPTS", 5)
```
Verified: `MERIDIAN_PREFETCH_MAX_ATTEMPTS=oops` now logs a warning and falls back to `5` instead of raising at import; a valid `=3` is still respected.

### ✅ 2. `assert`-guarded raise — fixed
Dropped `last_exc` and the `assert` entirely; the loop re-raises the active exception with a bare `raise` on the final attempt:
```python
    for attempt in range(1, _PREFETCH_MAX_ATTEMPTS + 1):
        try:
            snapshot_download(spec.model_id, allow_patterns=spec.allow_patterns)
            return
        except Exception as exc:  # noqa: BLE001
            log.warning(..., extra={"attempt": attempt, ...})
            if attempt >= _PREFETCH_MAX_ATTEMPTS:
                raise  # exhausted — surface to _run_prefetch
            time.sleep(min(2**attempt, 30))
```
Now `-O`-safe (no assert) and the exhaustion path is unconditional.

### Not changed (intentional, per review notes #3 & #4)
- Broad `except Exception` retry — kept for resilience to backend-specific stall exceptions; backoff caps waste at ~30s.
- Process-global Xet disable not restored — deliberate so later lazy loads inherit the bounded path.

No behavioural change to the happy path or the hang fix. Pre-commit/pre-push checks pass.

PR #352, opened 2026-06-27
fix: hide disconnected-provider tickets everywhere + Meridian process name

Akarsh-Hegde:
## Summary

- **Provider filter** — tasks/tickets from disconnected integrations (e.g. GitHub when not connected) were leaking into Dev Tasks provider tabs, the Cleanup board, the Board Score, and the must-fix banner. All four surfaces now call `filterByConnectedProviders()` so only connected-provider tasks are shown.
- **Must-fix banner** — was not suppressing on the `/cleanup/` page because Next.js static export with `trailingSlash: true` produces `/cleanup/` not `/cleanup`. Fixed with `pathname.startsWith('/cleanup')`.
- **Meridian process name** — "meridian-tray" (Cargo binary name) was showing in the dashboard's active-session card and the macOS dock tooltip during dev. Fixed with two changes:
  1. `is_self_app()` in the capture engine skips frames where the focused app is Meridian itself (both a11y and OCR paths).
  2. `set_process_display_name("Meridian")` via `NSProcessInfo` in the setup hook makes dev mode match the production `.app` bundle name.
- **CI** — added `bun test` step to the UI job (was missing).
- **Regression tests** — 24 `bun:test` unit tests covering all the above fixes.

## Test plan

- [ ] Connect only Jira in Settings → Integrations; verify Dev Tasks shows no GitHub tab, Cleanup board shows no GitHub tickets, must-fix banner count excludes GitHub tickets
- [ ] Navigate to `/cleanup` — verify must-fix banner is hidden on that page
- [ ] Run `bun test` in `ui/` — 24 tests pass
- [ ] Run `tauri dev` — active-session card shows "Meridian" not "meridian-tray"; dock tooltip shows "Meridian"
- [ ] Verify Meridian's own dashboard window no longer appears as a captured work session

## Summary by CodeRabbit

- **New Features**
  - UI task boards and cleanup views now show only tasks from connected providers, reducing irrelevant items and improving counts and health scores.
  - The must-fix banner now follows the same connected-provider filtering and is hidden on cleanup pages.
  - The tray app now avoids capturing its own dashboard windows, and macOS app naming is more consistent.

- **Bug Fixes**
  - Improved handling of invalid environment values and tightened retry flow for downloading specs.
  - Updated CI test execution for the UI to use Bun.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,9 @@ jobs:
           cache: 'npm'
           cache-dependency-path: ui/package-lock.json
 
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@v2
+
       - name: Cache Next.js build
         uses: actions/cache@v4
         with:
@@ -86,6 +89,10 @@ jobs:
 
       - name: UI install
         run: npm ci
+        working-directory: ui
+
+      - name: UI tests
+        run: bun test
         working-directory: ui
 
       - name: UI build
--- a/services/agents/routes/prefetch.py
+++ b/services/agents/routes/prefetch.py
@@ -87,13 +87,21 @@ def _env_positive_int(name: str, default: int) -> int:
     """Read a positive-int env override, falling back to `default` on bad input.
 
     Parsed at import, so a non-integer value (operator typo) must degrade to the
-    default rather than raise ValueError and brick MLX-server startup.
+    default rather than raise ValueError and brick MLX-server startup. Non-positive
+    integers are also rejected for consistency.
     """
     raw = os.environ.get(name)
     if raw is None:
         return default
     try:
-        return max(1, int(raw))
+        val = int(raw)
+        if val <= 0:
+            log.warning(
+                "server: ignoring non-positive env override; using default",
+                extra={"env_var": name, "raw": raw, "default": default},
+            )
+            return default
+        return val
     except ValueError:
         log.warning(
             "server: ignoring non-integer env override; using default",
@@ -147,7 +155,6 @@ def _download_spec(spec: model_registry.ModelSpec) -> None:
     for attempt in range(1, _PREFETCH_MAX_ATTEMPTS + 1):
         try:
             snapshot_download(spec.model_id, allow_patterns=spec.allow_patterns)
-            return
         except Exception as exc:  # noqa: BLE001 — retry-with-resume, re-raise if exhausted
             log.warning(
                 "server: prefetch download attempt failed; will resume from partial",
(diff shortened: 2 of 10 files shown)


Akarsh-Hegde (commented):
## Code review findings — addressed in 8d5752e

### Correctness bugs fixed

**`ui/lib/integrations.ts:159` — Unknown providers silently dropped** (CONFIRMED)
Tasks from any provider string not in the static `TRACKER_IDS` registry were silently filtered out once integrations loaded. A future integration, migration artifact, or backend typo would make tasks invisible with no error. Fixed by checking `integrations as Record<string, boolean>` directly — absent keys pass through instead of being dropped.

**`ui/components/MustFixBanner.tsx:27` — Shared `Promise.all` catch suppresses must-fix count** (CONFIRMED)
`Promise.all([get_tasks, get_integrations]).catch(()=>{})` meant any `get_integrations` failure (DB contention, Tauri timeout) silently zeroed the banner count and kept it at 0 on every subsequent poll. Fixed by catching each fetch independently — a `get_integrations` failure now only means integrations are treated as `null` (pass-through), not that the count disappears.

**`ui/components/views/CleanupView.tsx:101` — Dismissed tasks inflate Board Score** (CONFIRMED)
`total` counted all connected-provider tasks including those in the `dismissed` set, while `groups` excluded them via `!dismissed.has(t.key)`. Clicking Keep or Later bumped the Board Score as if the ticket was fixed. Fixed: `total` now applies the same `dismissed` filter as `groups`.

**`ui/components/views/TasksView.tsx:134` — Blank detail pane when all providers disconnected** (CONFIRMED)
The `data.tasks.length === 0` guard checked the unfiltered list — if tasks exist but every provider is disconnected, `activeTasks` is empty and `sel` resolves to `undefined`, leaving the detail pane blank with no explanation. Fixed by adding an explicit empty-state branch after `filterByConnectedProviders` that shows `ConnectTrackers` when `activeTasks.length === 0 && integrations !== null`.

### Safety fix

**`tray/src-tauri/src/lib.rs:788` — NSProcessInfo pointer not null-checked** (PLAUSIBLE)
`ns_name` was null-guarded but `info` was dereferenced unconditionally. Added `if info.is_null() { return; }` before the dereference.

### Maintainability fix

**`tray/src-tauri/src/capture/screenpipe.rs:176` — Magic strings drift from `set_process_display_name`** (CONFIRMED)
`is_self_app` hard-coded `"meridian-tray"` and `"meridian"` independently from the `"Meridian"` literal passed to `set_process_display_name` in `lib.rs`. A productName rename in one place would silently break self-capture filtering. Fixed by extracting `pub(crate) SELF_PRODUCT_NAME_LOWER` and `SELF_BINARY_NAME` constants in `lib.rs` and referencing them from `screenpipe.rs`.

### Not addressed (by design)

- **Redundant `get_integrations` IPC fetches** — MustFixBanner, TasksView, and CleanupView each fetch independently. A shared context would halve IPC traffic but requires a larger refactor (React context or prop-drilling from DashboardShell). Deferred.
- **Missing `useMemo` in TasksView** — `filterByConnectedProviders` and downstream derivations run on every render. A real concern at scale, but no performance regression vs. pre-PR; deferred.
- **`prefetch.py` `MERIDIAN_PREFETCH_MAX_ATTEMPTS=0`** — The rejection-and-default behaviour is intentional and logged. An operator setting `0` now gets the default (5) with a warning, which is safer than silently clamping to 1. No change needed.

PR #354, opened 2026-06-28
fix(integrations): surface real Azure DevOps error instead of generic fallback

Akarsh-Hegde:
## Summary

Three layered fixes to the Azure DevOps connection flow:

**1. Surface real Tauri errors in the UI (original fix)**
- `Tauri` command rejections with `Err(String)` produce a plain JS string, not an `Error` object
- The three `catch` blocks in `AzureDevOpsSetup` checked `e instanceof Error`, which was always `false`, so the real backend message was silently dropped
- Fix: prefix each check with `typeof e === 'string' ? e :` — applied to `lookupOrgs`, `lookupProjects`, and `connect`

**2. Surface network errors from reqwest (HTTP 0)**
- When `reqwest` can't establish a TCP connection (DNS failure, TLS error, etc.) it returns status `0`; the actual error detail string was bound to `_detail` and discarded
- Fix: rename to `detail` and format as `"Could not reach Azure DevOps: {detail}"` — applied to both the profile API call and the accounts API call

**3. Handle org-scoped PATs gracefully**
- PATs scoped to a specific org cannot call `app.vssps.visualstudio.com/_apis/profile/profiles/me` — they return 401/403
- The error message now says "PAT is invalid or org-scoped — enter your org name manually below, or use an 'All accessible organizations' PAT"
- Instruction text updated to say "All accessible organizations" scope is required
- A manual org name input appears after any failed org discovery so org-scoped PAT users can still connect without recreating their token

## Root cause (original)

```ts
// Before — string rejections always hit the fallback:
catch (e) { setError(e instanceof Error ? e.message : 'Failed to fetch organisations') }

// After — Tauri string rejections are surfaced directly:
catch (e) { setError(typeof e === 'string' ? e : e instanceof Error ? e.message : 'Failed to fetch organisations') }
```

## Test plan

- [ ] Paste a PAT scoped to "All accessible organizations" → org dropdown appears normally
- [ ] Paste an org-scoped PAT → error shows the org-scoped message + manual org input appears; entering the org name and clicking "Look up projects" works
- [ ] Network unreachable (e.g. wrong URL) → shows the actual reqwest error string instead of "HTTP 0"
- [ ] Invalid/expired PAT → shows the backend error message directly instead of generic fallback

### DIFF
--- a/tray/src-tauri/src/commands/integrations.rs
+++ b/tray/src-tauri/src/commands/integrations.rs
@@ -536,9 +536,11 @@ pub async fn discover_azure_devops(
     let profile = azure_get(profile_url, &body.pat)
         .instrument(tracing::debug_span!("integrations.azure.profile"))
         .await
-        .map_err(|(status, _detail)| {
+        .map_err(|(status, detail)| {
             if status == 401 || status == 403 {
-                "PAT is invalid or expired — check it and try again".to_string()
+                "PAT is invalid or org-scoped — enter your org name manually below, or use an 'All accessible organizations' PAT".to_string()
+            } else if status == 0 {
+                format!("Could not reach Azure DevOps: {detail}")
             } else {
                 format!("Azure DevOps profile API returned HTTP {status}")
             }
@@ -558,7 +560,13 @@ pub async fn discover_azure_devops(
     let accounts = azure_get(accounts_url, &body.pat)
         .instrument(tracing::debug_span!("integrations.azure.accounts"))
         .await
-        .map_err(|(status, _detail)| format!("Could not list organizations (HTTP {status})"))?;
+        .map_err(|(status, detail)| {
+            if status == 0 {
+                format!("Could not list organizations: {detail}")
+            } else {
+                format!("Could not list organizations (HTTP {status})")
+            }
+        })?;
     let orgs = sorted_names(&accounts, "accountName");
     tracing::info!(count = orgs.len(), "azure orgs discovered");
     Ok(AzureDiscoverResponse {
--- a/tray/src-tauri/src/commands/integrations.rs
+++ b/tray/src-tauri/src/commands/integrations.rs
@@ -513,9 +513,11 @@ pub async fn discover_azure_devops(
         let body_json = azure_get(url, &body.pat)
             .instrument(tracing::debug_span!("integrations.azure.projects"))
             .await
-            .map_err(|(status, _detail)| {
+            .map_err(|(status, detail)| {
                 if status == 401 || status == 403 {
                     "PAT is invalid or lacks Work Items → Read & write scope".to_string()
+                } else if status == 0 {
+                    format!("Could not reach Azure DevOps: {detail}")
                 } else {
                     format!("Azure DevOps returned HTTP {status}")
                 }
--- a/ui/components/IntegrationConnect.tsx
+++ b/ui/components/IntegrationConnect.tsx
@@ -370,6 +370,11 @@ function AzureDevOpsSetup({ tracker, onSuccess }: { tracker: Tracker; onSuccess?
 
   const handleOrgChange = (org: string) => { setSelectedOrg(org); if (org) lookupProjects(org) }
 
+  const submitManualOrg = () => {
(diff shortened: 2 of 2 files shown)


Akarsh-Hegde (commented):
## Review — fix(integrations): surface real Azure DevOps error

Solid, well-scoped fix. The `instanceof Error` → `typeof e === 'string'` correction is right (Tauri rejects with a bare `String`, never an `Error`), and surfacing the reqwest `detail` instead of dropping it into `_detail` is a clear improvement. The org-scoped-PAT recovery path is a nice touch. A few points:

### 1. (main) `status == 0` handling is missing on the *projects* branch — the new manual-org path can still show "HTTP 0"
`tray/src-tauri/src/commands/integrations.rs:516-522`

The profile (`:539`) and accounts (`:561`) calls now special-case `status == 0`, but the projects branch was left untouched:

```rust
.map_err(|(status, _detail)| {
    if status == 401 || status == 403 {
        "PAT is invalid or lacks Work Items → Read & write scope".to_string()
    } else {
        format!("Azure DevOps returned HTTP {status}")   // ← network failure ⇒ "HTTP 0"
    }
})?;
```

This is the very branch the **new manual-org input** routes through (`lookupProjects` → `discover_azure_devops` with `org` set). So an org-scoped-PAT user who hits the manual flow and then loses connectivity gets exactly the generic "HTTP 0" message this PR set out to eliminate. Worth adding the same `status == 0 ⇒ "Could not reach Azure DevOps: {detail}"` arm here for consistency, and it also still drops `detail` (`_detail`).

### 2. (minor) `status == 0` also means "JSON parse failed", not only "couldn't reach"
`azure_get` returns `(0, e.to_string())` for **both** the connection error (`:462`) and a successful-response-but-malformed-JSON case (`:468`). So `"Could not reach Azure DevOps: …"` can be shown for a reachable-but-garbage response. Edge case, but the message would mislead. If you want to be precise, the connection vs. parse failures could carry distinct sentinels; otherwise fine to leave as a known limitation.

### 3. (minor) the 401/403 message now conflates "invalid" with "org-scoped"
`integrations.rs:540-541` — a genuinely expired/revoked PAT now reads "PAT is invalid or org-scoped — enter your org name manually below…", nudging the user toward the manual path with a dead token. They'll bottom out at the projects-branch 401 eventually, so it self-corrects, but the message is less direct than before for the plain-invalid case. Acceptable trade-off; flagging it as a deliberate choice.

### 4. (nit) duplicated submit logic in the manual-org input
`ui/components/IntegrationConnect.tsx:451-470` — the Enter `onKeyDown` and the button `onClick` both do `const org = manualOrg.trim(); setSelectedOrg(org); lookupProjects(org)`. Minor; a small `submitManualOrg()` helper would dedupe and keep the two in sync.

Nothing blocking except #1, which I'd fix before merge since it directly undercuts the new feature path. The test plan looks good — consider adding a "manual org + network drop" case to cover #1.

Akarsh-Hegde:
Addressed in a2e09d8a:

**#1 (blocking) — projects branch HTTP 0** — fixed. Renamed `_detail` → `detail` and added the `status == 0 ⇒ "Could not reach Azure DevOps: {detail}"` arm to the projects `map_err`, consistent with the profile and accounts branches. This was the path the new manual-org input routes through so it was the most important one to catch.

**#4 (nit) — duplicated submit logic** — fixed. Extracted `submitManualOrg()` helper; both the `onKeyDown` and `onClick` now call it directly.

**#2 (minor) — status 0 conflates network error and JSON parse failure** — acknowledged as a known limitation. The two cases (`reqwest::Error` from the transport layer vs. a `serde_json` decode failure on a 200 response) would need a distinct sentinel in `azure_get`'s return type to distinguish. Happy to add that in a follow-up if it becomes a real pain point, but leaving it as-is for now since the JSON case is extremely unlikely on the VSSPS/ADO APIs.

**#3 (minor) — 401/403 message conflates invalid vs. org-scoped** — deliberate trade-off. A dead token will self-correct when the user hits the projects 401, and the manual-org path is the more common case for this 401. Left as-is.

PR #355, opened 2026-06-28
fix(integrations): save API tokens on a fresh install (no .env yet)

Akarsh-Hegde:
## Bug

On a fresh `.app`/DMG install, connecting **any** tracker by API token (Jira, Linear, GitHub, Azure DevOps) failed with **"Could not save credentials."**

## Root cause

`~/.meridian/.env` does not exist on a fresh `.app` install (nothing pre-creates it — only the npm `install-from-bundle.sh` path does). So `install::detect_install_mode()` returns `Bare`, whose `env_path()` is `None`, and `save_integration_token` bailed with `"could not resolve .env path"`. The UI's `catch` only read `e.message` for `Error` instances — but Tauri rejects a `Result<_, String>` command with a **plain string** — so the real cause was discarded and the generic *"Could not save credentials"* shown instead.

It's a chicken-and-egg: the canonical `.env` is created **by** saving credentials, but saving required it to already exist.

## Fix

- **`install.rs`** — new `canonical_env_path()` → `~/.meridian/.env`, the write target.
- **`commands/integrations.rs`** — `save_integration_token` falls back to the canonical `.env` when none exists; `upsert_env` already creates the parent dir + file. `Dev`/`Canonical` modes are unchanged.
- **`components/IntegrationConnect.tsx`** — surface the real error (`typeof e === 'string' ? e : …`), so future failures show the actual message instead of the generic fallback.

## Validation

- Tray crate type-checks clean with these changes.
- Manual workaround that confirms the diagnosis (and unblocks affected machines now): `mkdir -p ~/.meridian && touch ~/.meridian/.env` flips detection to `Canonical` and saving works.

## Notes

- This is an **app** fix (ships in the tray binary via the app release), not a runtime change.
- Targeted at `pre-main`; needs to reach `main`/production to fix production installs.

### DIFF
--- a/tray/src-tauri/src/commands/integrations.rs
+++ b/tray/src-tauri/src/commands/integrations.rs
@@ -401,10 +401,15 @@ pub async fn save_integration_token(body: SaveTokenBody) -> Result<serde_json::V
     let key_count = updates.len();
     {
         let mode = crate::install::detect_install_mode();
-        let env_path = mode
-            .env_path()
-            .ok_or("could not resolve .env path")?
-            .to_owned();
+        // On a fresh `.app` install no `.env` exists yet, so `detect_install_mode`
+        // returns `Bare` (no path). Credentials must be saveable before any `.env`
+        // exists — default to the canonical `~/.meridian/.env`, which `upsert_env`
+        // creates (parent dir + file). Dev/Canonical modes keep their resolved path.
+        let env_path = match mode.env_path() {
+            Some(p) => p.to_owned(),
+            None => crate::install::canonical_env_path()
+                .ok_or("could not resolve home directory for .env")?,
+        };
         tokio::task::spawn_blocking(move || upsert_env(&env_path, &updates))
             .await
             .map_err(|e| format!("spawn_blocking panicked: {e}"))?
--- a/tray/src-tauri/src/install.rs
+++ b/tray/src-tauri/src/install.rs
@@ -65,6 +65,17 @@ pub(crate) fn detect_install_mode() -> InstallMode {
     InstallMode::Bare
 }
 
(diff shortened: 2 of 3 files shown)


Akarsh-Hegde (commented):
## Review — fix(integrations): save API tokens on a fresh install (no .env yet)

Correct fix for the chicken-and-egg. A few observations:

### What checks out

- **`upsert_env` creates the parent dir** (lines 255-257) — so `canonical_env_path()` pointing at `~/.meridian/.env` on a fresh install where `~/.meridian/` doesn't exist yet will succeed; no additional mkdir needed.
- **Consistent HOME resolution** — `canonical_env_path()` uses the same `std::env::var("HOME")` pattern as the existing `detect_install_mode()`. Correct to keep in sync.
- **`Dev`/`Canonical` modes untouched** — the fallback only fires when `env_path()` returns `None` (`Bare`). Subsequent saves after the first will hit `Canonical` (the file now exists). Correct.
- **`TokenSetup` error surfacing** (line 276) is the companion UI fix that makes the backend message visible — needed and correct.

### Merge ordering note

The change to `AzureDevOpsSetup`'s `connect` catch (line 377) is identical to what PR #354 makes to the same line. If #354 lands first, this hunk will produce a trivial conflict on rebase — same change, same result. Suggest rebasing this branch onto `pre-main` after #354 merges; everything else in this PR is disjoint.

No bugs found. This is ready to merge (rebase after #354 if that lands first).

Akarsh-Hegde:
Rebased onto `pre-main` — the duplicate `AzureDevOpsSetup` catch hunk was automatically dropped (already landed via #354). Only the `TokenSetup` fix remains. All pre-push checks pass. Ready to merge.

PR #356, opened 2026-06-28
feat(setup): two-step provision checklist + self-check causes on error

Akarsh-Hegde:
## What

Improves the first-run wizard's **Local intelligence** step so provisioning reads as clear, sequential steps and a failed download is self-diagnosable.

### 1. Two-step checklist
Two rows that tick **○ pending → ◌ spinner → ✓ done** as each phase completes:
- **Install on-device engine** (the ~214 MB runtime tarball)
- **Download private models** (the ~2.9 GB model set)

This makes the engine vs. model phases distinct (previously one ambiguous bar), and on Apple-Intelligence Macs that skip the model download, both rows tick to ✓ quickly — which answers "is it actually done?" instead of leaving the user guessing.

### 2. Self-check causes on a network error
When provisioning hits a recoverable error (e.g. `manifest request failed: error sending request…`), the wizard now lists the most-likely, user-fixable causes above **Try again**:

1. **Internet connection** — incl. café/hotel "sign in to connect" portals
2. **VPN or firewall**
3. **Network proxy** (the app doesn't pick up the macOS system proxy)
4. **Security software** (TLS interception)

Hidden for the `unavailable` hardware verdict (network help wouldn't apply there).

## Notes

- UI-only; composed entirely from existing wizard state (`downloading` / `runtimeInstalled` / `prefetching` / `modelReady`) — no backend changes.
- Targeted at `pre-main`.

### DIFF
--- a/ui/app/setup/steps.tsx
+++ b/ui/app/setup/steps.tsx
@@ -83,6 +83,41 @@ function PermissionsBody({ wiz }: { wiz: Wiz }) {
 }
 
 // ── STEP 2 — Local intelligence ──────────────────────────────────────────────
+
+/** One provisioning sub-step (engine install / model download). The leading
+ *  glyph updates live as the phase advances: hollow ○ (pending) → spinner
+ *  (active) → ✓ (done), so the two steps visibly tick off in sequence. */
+type StepState = 'pending' | 'active' | 'done'
+function ProvisionStep({ state, label }: { state: StepState; label: string }) {
+  return (
+    <div className="flex items-center" style={{ gap: 10 }}>
+      <span className="flex items-center justify-center shrink-0" style={{ width: 20, height: 20 }}>
+        {state === 'done'
+          ? <Check size={16} color="var(--success)" w={2.2} />
+          : state === 'active'
+            ? <Spinner size={15} width={2} />
+            : <span style={{ width: 12, height: 12, borderRadius: 99, border: '1.5px solid var(--rule-2)' }} />}
+      </span>
+      <span style={{
+        fontSize: 12.5,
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde (commented):
## Review — feat(setup): two-step provision checklist + self-check causes on error

Clean, UI-only change. Notes:

### What checks out

- **Smart quotes** — the `'` in `'Confirm you're online…'` is U+2019 (bytes `e2 80 99`), not ASCII U+0027. Valid inside a single-quoted TS string literal; no syntax issue.
- **State logic** — runtime step: `runtimeInstalled ? 'done' : wiz.downloading ? 'active' : 'pending'`; model step: `wiz.modelReady ? 'done' : wiz.prefetching ? 'active' : 'pending'`. Both correctly reflect the live phase without additional state.
- **Apple Intelligence path** — `modelReady` goes from false → true without `prefetching` ever being true, so the model row jumps from pending → done. This is the right experience (fast-tick).
- **Error state** — checklist is not gated on `!showErr`, so a mid-download failure shows the correct partial state (e.g. runtime ✓, model pending) alongside the error card. Informative, not misleading.
- **`Check`/`Spinner` props** — `w={2.2}` on Check and `width={2}` on Spinner match the existing call-sites in this file.
- **`SETUP_NET_HINTS` at module scope** — correct, avoids re-allocation on every render.

### Minor

- The checklist is visible even when `modelReady` (both rows ✓). This is intentional and good — gives a clear completion confirmation. Just noting it reads fine.
- `ProvisionStep` is not exported — correct, it's internal to this file.

No bugs found. Ready to merge.

PR #362, opened 2026-06-29
feat: proposed ticket approval, local-time worklog pipeline, OO dashboards, observability improvements

adityaharishch:
## Summary

This branch bundles several interconnected features and fixes that were built together across multiple sessions. The changes fall into five areas:

### 1. Proposed Ticket Approval (`feat/proposed-ticket-approval`)

When the worklog pipeline processes an hour with no matching task, it now drafts a **proposed new ticket** instead of silently dropping the hour. The user can review, edit, and approve/dismiss proposals inline in the WorklogsView.

- **Migration 050**: `pm_proposed_tasks` table (`title`, `drafted_worklog`, `status`: pending/approved/dismissed, `approved_task_id` FK once real ticket is created)
- **`meridian-core/src/readers/proposed.rs`**: `get_proposed_tasks(pool, day_local)` reader for the tray command
- **`src/pm_worklog/create.rs`**: `create_proposed_task()` upsert + `mark_proposed_approved/dismissed()` + `get_pending_proposed()`
- **Tray commands** (`worklogs.rs`): `edit_proposed_title`, `edit_proposed_worklog`, `proposed_action` (approve/dismiss)
- **UI** (`WorklogsView.tsx`): proposed items render inline in the hour timeline with editable title + comment fields and approve/dismiss buttons
- **Python pipeline** (`worklog_pipeline/workflow.py`): Tier-3 path now calls `propose_ticket()` instead of dropping; `propose_ticket.py` prompt generates a candidate title + one-paragraph worklog

### 2. Worklog Pipeline: Local-Time Hours + Today-Only (`fix/worklog-pipeline-local-time`)

The Rust hour driver and Python distiller were using UTC hours, which misbucketed sessions under sub-hour timezone offsets (e.g. IST +05:30 — a session at 10:14 IST landed in bucket "04" UTC, not "10").

- **`src/worklog_pipeline.rs`**: hour labels are now `YYYY-MM-DDTHH` in local time; midnight walk uses `Local` not `Utc`
- **`services/agents/time_utils.py`**: `local_hour_utc_bounds(label)` converts local-hour labels to UTC bounds at query time
- **`services/agents/session_distiller.py`**: distil_hour uses `local_hour_utc_bounds` instead of raw UTC arithmetic
- **Dropped yesterday backfill**: the periodic pass now always uses `day=None` (local today) — previously it re-processed yesterday on every tick, wasting LLM budget

### 3. OpenObserve Dashboards + Observability

Python telemetry no longer depends on the Rust daemon being alive to deliver spans to OpenObserve.

- **`services/agents/observability.py`**: when `otlp_enabled=true` and `oo_email`/`oo_password` are set in `settings.json`, Python uses `BatchSpanProcessor + OTLPSpanExporter` to ship directly. Spool path remains as fallback when credentials are absent.
- **`services/agents/routes/activity.py`**: `thinking_budget=8192` added to `apply_chat_template` so the Qwen3.5-2B think chain caps at 8192 tokens; `max_tokens` reduced from 65536 → 16384. Previously the model would generate indefinitely (all 99 prior runs hit exactly the old 8192 token limit inside the think chain with `think_tok=0`, meaning `</think>` was never reached).
- **Three new OO dashboards** (`services/observability/dashboards/`):
  - `activity-summary.json` — per-trace table with one-click drilldown to the 4-span waterfall
  - `pm-worklog-hour.json` — worklog pipeline run table (distil_nsess, in/out/think tokens, match outcome)
  - `session-distiller.json` — distil_hour call table
- **Scripts**: `run-activity-report.sh`, `run-distill.sh` (manual test scripts; fixed macOS `mktemp` where X's must be at end of template), `push-dashboards.sh`

### 4. Coding-Agent Indexer: Hour-Boundary Sealing + 2-Min Poll

- **Hour-boundary sealing** (`indexer.rs`, `segment.rs`): sessions are now sealed at the top of each local clock hour rather than waiting for the 1h idle backstop. The indexer wakes at each boundary (`next_tick_delay` calculates seconds-to-next-top-of-hour + 2s buffer) and runs `seal_stale_open_rows_before()`.
- **2-minute poll** (was 600s): mtime checks are cheap; faster polling gives ≤2-min process-gone detection lag.
- **`db.rs`**: `seal_stale_open_rows_before(pool, cutoff_iso)` DB helper for hour-boundary sweeps.

### 5. ETL: Filter Coding-Agent Terminal Titles from window_titles

`get_window_titles` queries by frame-id range and returns ALL frames — including coding-agent terminal frames the ETL loop already skips via `continue`. These frames were leaking spinner symbols and bare version strings (e.g. `2.1.193`) into `app_sessions.window_titles`.

- **`src/etl/extractor.rs`**: post-filters `window_titles_res` using the same `is_vscode_like + is_coding_agent_terminal` predicates the ETL loop uses
- **`src/etl/session_builder.rs`**: adds `is_version_label()` so bare semver tab titles (Claude Code's idle label) are caught
- **`tests/etl_coding_agent_skip.rs`**: integration tests covering the filter + version-label cases

### 6. Python Deps

- `transformers>=5.0.0` — `mlx-lm` requires it; Qwen3.5 architecture not recognised by 4.x (was silently installed at 4.57.6)
- `sentence-transformers>=3` — needed by `session_distiller.py`'s lazy `SentenceTransformer` import

## Test plan

- [ ] Confirm `cargo test` passes (all 336 tests green, tray_assets included)
- [ ] Run `bash scripts/run-activity-report.sh YYYY-MM-DDTHH` and confirm it completes in <60s with non-zero `think_tok`
- [ ] Verify worklogs view shows proposed items for hours with no task match
- [ ] Confirm worklog pipeline logs show local-time labels (e.g. `2026-06-29T10` for IST morning)
- [ ] Check OO dashboards load and drilldown links resolve to correct trace


tray/src-tauri/Cargo.toml
```diff
@@ -57,13 +57,13 @@ meridian = { path = "../..", optional = true }
 # screen capture + Apple Vision OCR. Optional → only pulled by the `capture`
```
Akarsh-Hegde, line 57: **[CRITICAL — Build-breaking] Hardcoded absolute path for screenpipe crates**

The `path = "/Users/adityaharish/..."` dependency replaces the previously pinned `git` revision. This breaks every build on any machine that isn't Aditya's local box — including CI and Akarsh's machine — with a Cargo path-not-found error.

Revert to a pinned `git` rev (the one this replaced) or a published crates.io version so the build is reproducible everywhere.

src/pm_worklog/post.rs
```diff
@@ -313,6 +313,98 @@ fn azure_devops_cfg(config: &Config) -> Option<&AzureDevOpsConfig> {
...
+        let cycle_index: i64 = p
+            .source_hour
+            .get(11..13)
+            .and_then(|h| h.parse().ok())
+            .unwrap_or(0);
+
+        // Stamp the real key into the drafted payload (it was minted with a
+        // placeholder task_key) so the posted comment is grounded correctly.
+        let payload_json = {
+            let mut v: serde_json::Value = p
```
Akarsh-Hegde, line 397: **[CRITICAL — Data corruption] source_hour UTC fallback stores wrong timestamps for non-UTC users**

The fallback `format!("{}:00:00+00:00", p.source_hour)` appends a UTC offset to what is actually a **local** hour label (e.g. `2026-06-28T10` for 10am IST). For a user in UTC+5:30 this writes `10:00:00+00:00` into `pm_worklogs.window_start` and posts that timestamp to Jira/Linear — placing the worklog 5.5 hours in the future relative to the user's actual work time.

Every user outside UTC who had any proposed tickets written before migration 050 (i.e. `window_start IS NULL`) will get corrupted Jira/Linear worklogs on upgrade.

Fix: treat `source_hour` as local time and attach the correct local UTC offset, or use `CURRENT_TIMESTAMP`-derived bounds that are already timezone-correct.

src/worklog_pipeline.rs
```diff
@@ -202,18 +209,14 @@ pub async fn run_loop(
         "worklog-pipeline driver starting (hour-level /worklog_hour; drafts only)"
     );
 
-    let run_pass = |scope_today: bool| {
+    let run_pass = |_scope_today: bool| {
```
Akarsh-Hegde, line 212: **[HIGH — Silent data loss] Yesterday backfill removed; last-hour sessions never drafted**

The periodic fallback pass previously included `yesterday` in its day list so that sessions completing in the 23:00–00:00 slot were caught after midnight. That guard is gone here. When the daemon restarts or the timer fires after midnight, only `today` is passed, and the previous day's last hour is permanently skipped — those sessions are never drafted into worklogs.

The old comment in this file explicitly cites a real observed case ("row 32019") that motivated the yesterday-inclusion. Restore it, or add a comment explaining why it's no longer needed.

src/etl/session_builder.rs
```diff
@@ -236,6 +261,13 @@ pub(super) fn is_coding_agent_terminal(window_name: &str) -> bool {
         _ => return false,
```
Akarsh-Hegde, line 242: **[HIGH — Silent regression] `is_coding_agent_terminal` now misclassifies Node/Python REPL versions as coding-agent terminals**

The expanded pattern now returns `true` for any bare `X.Y.Z` version label (e.g. `Terminal - 20.11.0`, `Terminal - 3.11.5`). The three assertions that Node and Python version strings must NOT be suppressed were deleted and the logic was inverted.

Consequence: VS Code integrated-terminal sessions running a Node.js REPL or Python interpreter are classified as coding-agent terminals. Their frames are skipped by the ETL loop and their window titles are stripped from `app_sessions`, so all REPL work in VS Code is silently excluded from focus-time and task-matching.

Restore the Node/Python exclusion tests or tighten the pattern so it only matches known coding-agent binary names, not arbitrary version strings.

src/coding_agent_session_ingest/summariser/config.rs
```diff
@@ -14,7 +14,6 @@ pub struct SummariserConfig {
     pub batch_per_tick: i64,
```
Akarsh-Hegde, line 14: **[HIGH — Silent config regression] `SUMMARISER_SKILL` env-var override silently dropped**

The `skill_name` config field and `SUMMARISER_SKILL` env-var are removed here. `run_claude` now embeds `SUMMARY_RULES` inline rather than invoking a configurable slash-skill.

Any deployment that set `SUMMARISER_SKILL` to a custom skill name (e.g. to route summaries through a project-specific prompt) will silently lose that customization after upgrading — the env var is ignored, the skill is never invoked, and there's no warning or error.

If the intent is to retire skill-based routing, add a `WARN` log on startup when `SUMMARISER_SKILL` is set, so operators know their config is being ignored rather than discovering it through degraded output.

ui/components/views/WorklogsView.tsx
```diff
@@ -150,11 +160,19 @@ export default function WorklogsView() {
       ) : (
         <div className="space-y-3">
           {items.map(w => (
```
Akarsh-Hegde, line 165: **[MEDIUM — UX bug] Shared `busy` state causes ID collisions between worklog and proposal cards**

The single `busy` number is compared against `w.id` for both `WorklogCard` (`pm_worklogs.id`) and `ProposedCard` (`pm_proposed_tasks.id`). Both tables use independent SQLite autoincrement sequences, so `pm_worklogs.id=5` and `pm_proposed_tasks.id=5` can coexist.

When the user approves proposal id=5, `run(5, ...)` sets `busy=5`, and any worklog card with id=5 also renders `busy={true}` — disabling its action buttons until the proposal request completes. This becomes increasingly common as both tables grow.

Fix: use a namespaced key (e.g. ``busy: { kind: 'worklog' | 'proposal'; id: number } | null``) so the two namespaces can't collide.

services/pyproject.toml
```diff
@@ -43,6 +43,8 @@ mlx = [
     # Xet downloads. Tuned harder with HF_XET_HIGH_PERFORMANCE=1 (MLX server plist).
     "hf-xet>=1.5.1,<2",
     "outlines[mlxlm]>=1.3,<2",
+    "transformers>=5.0.0",
+    "sentence-transformers>=3",
```
Akarsh-Hegde, line 47: **[HIGH — Services changes don't reach users] `pyproject.toml` version not bumped**

New deps (`sentence-transformers`, `transformers`) were added here, and this PR also modifies `session_distiller.py`, `observability.py`, `time_utils.py` (new file), and `worklog_pipeline/*.py` — but the `version` in `pyproject.toml` is unchanged at `1.63.0`.

Per `CLAUDE.md`: the MLX runtime auto-upgrade is an equality check — if `installed_version == manifest.version`, the upgrade is skipped silently. Every machine already on `1.63.0` will keep the old runtime; the new pipeline logic and new dependencies never reach them.

Bump the version and cut a runtime release after merge per the CLAUDE.md "Ship a `services/` Python change" instructions.

services/agents/worklog_pipeline/pipeline.py
```diff
@@ -366,21 +427,44 @@ def stage_propose(ctx: HourContext) -> None:
         if parsed:
```
Akarsh-Hegde, line 427: **[MEDIUM — Performance] `stage_propose` now runs two sequential blocking LLM calls**

This block adds a second `make_schema_agent` call (`generate_worklog`) immediately after `pr_agent.run()`. Both calls go to the single-threaded MLX inference server sequentially. Each takes ~30–120s, so every "no match" pipeline hour now blocks the server for 2× the previous wall-clock time and delays any concurrent request for the full combined duration.

Consider deferring the worklog draft to a follow-up pass after the proposal is confirmed, or running both agents against a local model concurrently (two separate `asyncio` tasks if the server gains async support). At minimum, add a trace attribute so the dual-call latency is visible in OO and can be measured before deciding to optimize.

src/coding_agent_session_ingest/indexer.rs
```diff
@@ -222,36 +232,80 @@ pub async fn register_records(
...
+/// hour and is eligible for force-sealing.
+fn local_hour_start_utc(now: DateTime<Utc>) -> DateTime<Utc> {
+    let local = now.with_timezone(&Local);
+    let hour_start = local
+        .date_naive()
+        .and_time(NaiveTime::from_hms_opt(local.hour(), 0, 0).unwrap_or_default());
+    Local
+        .from_local_datetime(&hour_start)
+        .single()
+        .unwrap_or(local)
```
Akarsh-Hegde: **[LOW — Maintenance] `local_hour_start_utc` re-derives hour-boundary math already in `meridian_core::date`**

This function re-implements local-time → UTC boundary math (Local + NaiveTime + DST fold) inline. `meridian_core::date::local_day_bounds` already encapsulates equivalent timezone arithmetic and is the single source of truth for this pattern across the daemon.

If the per-hour variant isn't there yet, it belongs in `meridian_core::date` alongside `local_day_bounds` — not as a private fn in the indexer. A timezone-edge fix (e.g. DST rollback ambiguous-time handling) then propagates everywhere automatically rather than requiring the same fix in two places.

services/agents/observability.py
```diff
@@ -55,10 +46,10 @@
...
 from opentelemetry.sdk._logs import LoggerProvider, LoggingHandler
-from opentelemetry.sdk._logs.export import BatchLogRecordProcessor, LogExportResult
+from opentelemetry.sdk._logs.export import SimpleLogRecordProcessor, LogExportResult
 from opentelemetry.sdk.resources import Resource
 from opentelemetry.sdk.trace import TracerProvider
-from opentelemetry.sdk.trace.export import BatchSpanProcessor, SpanExportResult
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor, SpanExportResult
 from opentelemetry.trace.propagation.tracecontext import (
     TraceContextTextMapPropagator,
 )
```
Akarsh-Hegde, line 55: **[MEDIUM — Performance] Spool exporter downgraded from `BatchSpanProcessor` to `SimpleSpanProcessor` when OO credentials are absent**

The no-credentials fallback path now uses `SimpleSpanProcessor(SpoolSpanExporter())` (and `SimpleLogRecordProcessor` for logs). `SimpleSpanProcessor` calls the exporter synchronously on every span end, blocking the inference thread while it writes to the spool file.

For the MLX server this is consequential: a busy inference request generates many child spans (distill → match → propose), and each span end now serialises to disk in the calling thread. On a slow disk or NFS home directory this can visibly inflate inference latency; the original `BatchSpanProcessor` flushed asynchronously in the background and avoided this entirely.

If the motivation is simplicity, consider keeping `BatchSpanProcessor` for the spool path and only using `BatchSpanProcessor(OTLPSpanExporter)` for the direct-OO path — the batch processor is the right choice for any high-frequency exporter regardless of destination.

services/agents/session_distiller.py
```diff
@@ -26,13 +26,15 @@
 import time
 from collections import defaultdict
 from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
```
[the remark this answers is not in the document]
adityaharishch: Fixed — removed the duplicate `datetime` import block and the unused `timedelta`/`timezone` imports. Commit `fix(review): address PR #362 review comments`.

services/agents/observability.py
```diff
@@ -471,11 +491,120 @@ def instrument_agno() -> None:
...
+
+    path = db_file or os.environ.get("AGNO_TRACE_DB") or str(
+        Path("~/.meridian/agno_traces.db").expanduser()
+    )
+    path = str(Path(path).expanduser())
+    exporter = _build_agno_db_exporter(SqliteDb(db_file=path), workflow_id, agent_id)
+    provider = _TP()
+    provider.add_span_processor(SimpleSpanProcessor(exporter))
+    AgnoInstrumentor().instrument(tracer_provider=provider)
+    _AGNO_TRACER_PROVIDER = provider
```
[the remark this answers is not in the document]
adityaharishch: Not fixing — this is a false positive from the static analyser. `_AGNO_TRACER_PROVIDER` is read on every call to `get_agno_tracer()` as a global guard (`if _AGNO_TRACER_PROVIDER is not None`) and written on initialisation. Because it is accessed only via `global _AGNO_TRACER_PROVIDER` inside functions, some linters treat it as unused at module scope. The function would throw `UnboundLocalError` on second call without it.

tray/src-tauri/Cargo.toml
adityaharishch: Fixed — reverted both `screenpipe-screen` and `screenpipe-a11y` to `git = ..., rev = "d6d2b318cfe1fffbc9e923e627e8659df1b2ff7e"`. Commit `fix(review): address PR #362 review comments`.

src/pm_worklog/post.rs
adityaharishch: Fixed — added `local_hour_to_utc(label, mm, ss)` helper in `post.rs` that parses the local-time label and converts it to UTC via `chrono::Local`. The `window_start`/`window_end` fallbacks now call this instead of naively appending `+00:00`. Commit `fix(review): address PR #362 review comments`.

src/worklog_pipeline.rs
adityaharishch: Not fixing — this is a deliberate product decision made during development of this feature. The yesterday backfill was creating duplicate drafts on morning restarts when a session crossed midnight. Today-only is the intentional behaviour for now; the backfill can be re-introduced as a separate feature when the deduplication story is solid.

src/etl/session_builder.rs
adityaharishch: Fixed — removed `is_version_label()` entirely (both the Tier-0 check in `is_coding_agent_terminal` and the function itself). The bare-semver heuristic was too broad: Node 20.11.0 and Python 3.11.5 in VS Code terminal tab titles were being filtered. The two integration test cases that exercised a version label were updated to use named-agent labels (`claude`/`codex`) which are still correctly filtered. Commits `fix(review): address PR #362 review comments` and `fix(etl): remove is_version_label and update tests`.

src/coding_agent_session_ingest/summariser/config.rs
adityaharishch: Fixed — restored `skill_name: String` field and `SUMMARISER_SKILL` env-var parsing in `SummariserConfig`. Added a `tracing::warn!` in `claude.rs` that fires when `skill_name != "session-summary"` so operators learn the var is inert and can clean it up. Commit `fix(review): address PR #362 review comments`.

ui/components/views/WorklogsView.tsx
adityaharishch: Fixed — changed `busy` state from `number | null` to `string | null` with namespaced keys: `wl:${id}` for worklog cards, `prop:${id}` for proposal cards, and `'all'` for the approve-all action. The two autoincrement sequences are independent so the same integer could refer to a worklog and a proposal simultaneously. Commit `fix(review): address PR #362 review comments`.

services/pyproject.toml
adityaharishch: Fixed — bumped to `1.64.0`. Commit `fix(review): address PR #362 review comments`.

services/agents/worklog_pipeline/pipeline.py
adityaharishch: Not fixing now — the two calls are inherently sequential: `make_score_prompt` ranks sessions against the task, and `make_propose_prompt` only runs on the winner from that ranking. Parallelising them would require firing propose for all candidates upfront, multiplying LLM cost by candidate count. Acceptable trade-off for now; could be optimised later with a single combined prompt if latency becomes a problem.

src/coding_agent_session_ingest/indexer.rs
adityaharishch: Acknowledged (LOW). The `local_hour_start_utc` function overlaps with `meridian_core::date` but it lives in the indexer's own module and targets a slightly different range (per-hour vs per-day). Consolidating it is a clean-up task for a future PR; leaving as-is to keep this PR's diff minimal.

services/agents/observability.py
adityaharishch: Fixed — switched the spool path back to `BatchSpanProcessor(SpoolSpanExporter())` and `BatchLogRecordProcessor` for both traces and logs. The `SimpleSpanProcessor` was blocking inference threads on every span flush. Commit `fix(review): address PR #362 review comments`.

src/coding_agent_session_ingest/indexer.rs
adityaharishch: Fixed — promoted `local_hour_start_utc` to `meridian_core::date::local_hour_start_utc` with DST-ambiguity handling consistent with the existing `local_day_bounds` helper. The private copy in the indexer has been removed. Commit `refactor(indexer): move local_hour_start_utc into meridian_core::date`.

Akarsh-Hegde:
## Review complete — ready to merge ✓

All 10 findings from the automated review have been addressed. Summary:

| # | Finding | Severity | Status |
|---|---------|----------|--------|
| 1 | Hardcoded `Cargo.toml` absolute path for screenpipe crates | CRITICAL | ✅ Fixed — reverted to `git` rev |
| 2 | `source_hour` UTC fallback corrupts timestamps for non-UTC users | CRITICAL | ✅ Fixed — `local_hour_to_utc()` helper added in `post.rs` |
| 3 | Yesterday backfill removed; last-hour sessions never drafted | HIGH | ✅ Accepted — intentional product decision (today-only avoids duplicate morning drafts) |
| 4 | `is_coding_agent_terminal` misclassifies Node/Python REPL versions | HIGH | ✅ Fixed — `is_version_label()` removed entirely, tests updated |
| 5 | `SUMMARISER_SKILL` env-var silently dropped | HIGH | ✅ Fixed — `skill_name` field restored with `tracing::warn!` deprecation notice |
| 6 | `pyproject.toml` version not bumped (services changes don't ship) | HIGH | ✅ Fixed — bumped to `1.64.0` |
| 7 | `busy` state ID collision between worklogs and proposals | MEDIUM | ✅ Fixed — namespaced to `wl:${id}` / `prop:${id}` |
| 8 | `stage_propose` runs two sequential blocking LLM calls | MEDIUM | ✅ Accepted — calls are inherently ordered (rank → propose-on-winner); deferred |
| 9 | Spool exporter downgraded to synchronous `SimpleSpanProcessor` | MEDIUM | ✅ Fixed — reverted to `BatchSpanProcessor` |
| 10 | `local_hour_start_utc` duplicates `meridian_core::date` math | LOW | ✅ Fixed — promoted to `meridian_core::date::local_hour_start_utc` |

No open blockers. Good to merge.

PR #363, opened 2026-06-29
feat(trello): API key prerequisite for OAuth + inline prompt UI

Akarsh-Hegde:
## Summary

- Adds `trello` to `TOKEN_FIELD_MAP` in `integrations.rs` so `save_integration_token` can persist `TRELLO_APP_KEY` to `.env` before the OAuth flow reads it
- `OAuthSetup` in `IntegrationConnect.tsx` now catches the "Power-Up app key" error from `start_oauth` and surfaces an inline input field — the user can paste their key from `trello.com/app-key` and retry without touching `.env` manually
- Button is disabled until the key is non-empty; Enter key submits

This also includes the commits already merged to `pre-main` via PRs #358–360 that this branch branched from after the last merge.

## Test plan

- [ ] Connect Trello in a dev build (no baked-in `DEFAULT_APP_KEY`) — should show the API key prompt on first attempt
- [ ] Paste a valid key, click Connect — OAuth flow proceeds and writes `~/.meridian/oauth/trello.json`
- [ ] Production build with baked-in key — prompt never appears, flow proceeds directly
- [ ] Verify `TRELLO_APP_KEY` written to `.env` after supplying key

https://claude.ai/code/session_01LXzRoAakhnwnrWtDGJHnRb

### DIFF
--- a/tray/src-tauri/src/commands/integrations.rs
+++ b/tray/src-tauri/src/commands/integrations.rs
@@ -76,6 +76,10 @@ const TOKEN_KEYS: &[(&str, &[&str])] = &[
             "AZURE_DEVOPS_ORG_URL",
         ],
     ),
+    // Trello uses OAuth for its primary auth but also stores a user-supplied app
+    // key in .env (prod has it baked in; dev users paste their own). Disconnect
+    // must strip it so a rotated key isn't silently reused on the next connect.
+    ("trello", &["TRELLO_APP_KEY"]),
 ];
 
 /// Token-based connect map: `provider → [(ui_field, env_key)]`. This is the
@@ -266,7 +270,7 @@ fn upsert_env(path: &std::path::Path, updates: &BTreeMap<String, String>) -> std
 /// Disconnect a tracker (the ported /api/integrations DELETE). Removes the
 /// OAuth token store (`~/.meridian/oauth/<p>.json`) AND strips the provider's
 /// `.env` keys — Jira can be connected either way, so both run. (trello = json
-/// only; linear/github/azure = env keys only; jira = both.) After credentials
+/// + env key; linear/github/azure = env keys only; jira = both.) After credentials
 /// are removed, clears the provider's tasks from the DB (best-effort — warns on
 /// failure but does not block the disconnect). Returns `{ ok: true }`; an unknown
 /// provider is the route's 400.
--- a/tray/src-tauri/src/commands/integrations.rs
+++ b/tray/src-tauri/src/commands/integrations.rs
@@ -112,6 +112,10 @@ const TOKEN_FIELD_MAP: &[(&str, &[(&str, &str)])] = &[
         "azure_devops",
         &[("url", "AZURE_DEVOPS_URL"), ("pat", "AZURE_DEVOPS_PAT")],
     ),
+    // Trello's app key is baked in for production builds; dev users supply their
+    // own from https://trello.com/app-key via the UI, which saves it here before
+    // the browser OAuth flow starts.
+    ("trello", &[("api_key", "TRELLO_APP_KEY")]),
 ];
 
 /// Env keys that MUST be present for a provider to count as connected. Optional
@@ -332,8 +336,9 @@ pub struct SaveTokenBody {
 /// Write a token-based tracker's credentials to the active `.env` and reload the
 /// daemon — the in-app replacement for "run `meridian config edit`" (ports the
 /// deleted `/api/auth/token` route). Covers jira (API-token / self-hosted),
-/// linear, github (PAT), and azure_devops. Browser-OAuth providers (Jira Cloud
-/// OAuth, Trello) connect via [`start_oauth`] instead.
+/// linear, github (PAT), azure_devops, and trello (API key prerequisite for the
+/// browser OAuth flow; saving `TRELLO_APP_KEY` here lets [`start_oauth`] find the
+/// key and proceed — the OAuth step still runs to write `~/.meridian/oauth/trello.json`).
 ///
 /// Validation mirrors the route: required keys must be non-empty; CR/LF are
 /// stripped from each value (an env file is line-oriented). For jira, any stored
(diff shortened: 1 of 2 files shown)


tray/src-tauri/src/commands/integrations.rs
```diff
@@ -112,6 +112,10 @@ const TOKEN_FIELD_MAP: &[(&str, &[(&str, &str)])] = &[
         "azure_devops",
         &[("url", "AZURE_DEVOPS_URL"), ("pat", "AZURE_DEVOPS_PAT")],
     ),
+    // Trello's app key is baked in for production builds; dev users supply their
+    // own from https://trello.com/app-key via the UI, which saves it here before
+    // the browser OAuth flow starts.
+    ("trello", &[("api_key", "TRELLO_APP_KEY")]),
```
Akarsh-Hegde, line 122: **Bug: `TRELLO_APP_KEY` is never cleaned up on disconnect.**

`TOKEN_FIELD_MAP` now includes Trello, but `TOKEN_KEYS` (the disconnect cleanup map) does not. Disconnecting Trello removes the OAuth JSON file but leaves `TRELLO_APP_KEY` in `.env`. Next time the user reconnects, the old key pre-fills silently — a problem if they rotated it.

Add to `TOKEN_KEYS`:
```rust
("trello", &["TRELLO_APP_KEY"]),
```

The comment in `disconnect_integration` at line 264 that says "trello = json only" would also need updating.

ui/components/IntegrationConnect.tsx
```diff
@@ -236,7 +250,14 @@ function OAuthSetup({ tracker, onSuccess }: { tracker: Tracker; onSuccess?: () =
       }, 2_000)
       pollRef.current = id
     } catch (e) {
-      setError(e instanceof Error ? e.message : String(e)); setStatus('error')
+      const msg = e instanceof Error ? e.message : String(e)
+      // Trello without a baked-in or user-supplied app key: show the API key
+      // input so the user can unblock themselves without editing .env manually.
```
Akarsh-Hegde, line 259: **Brittle string matching.** `msg.includes('Power-Up app key')` couples this branch to the exact text of a Rust error string. If the error message changes (typo fix, reword), the prompt silently stops appearing and the user sees a raw error instead.

This is the same pattern as Azure's `message.includes('enter your org name manually below')` so it's consistent with the codebase, but worth noting. Consider either a structured error type with an error code from Rust, or at minimum extracting the sentinel into a named constant on both sides so a rename is a single-site change.

Akarsh-Hegde (commented):
Good feature overall — the two-step flow (fail → prompt → retry) works cleanly for dev builds, and the immediacy of the save-before-start ordering is sound. Three issues below, one of which is a functional bug (disconnect leaves  in ).

tray/src-tauri/src/commands/integrations.rs
Akarsh-Hegde, line 122: **Bug: `TRELLO_APP_KEY` is never cleaned up on disconnect.**

`TOKEN_FIELD_MAP` now includes Trello, but `TOKEN_KEYS` (the disconnect cleanup map) does not. Disconnecting Trello removes the OAuth JSON file but leaves `TRELLO_APP_KEY` in `.env`. Next time the user reconnects, the old key pre-fills silently — a problem if they rotated it.

Add to `TOKEN_KEYS`:
```rust
("trello", &["TRELLO_APP_KEY"]),
```

The comment in `disconnect_integration` at line 264 that says "trello = json only" would also need updating to reflect that there's now an env key too.

ui/components/IntegrationConnect.tsx
Akarsh-Hegde, line 259: **Brittle string matching.** `msg.includes('Power-Up app key')` couples this branch to the exact wording of a Rust error string. If the message is ever reworded (typo fix, clarity pass), the prompt silently stops appearing and the user sees a raw error instead of the key input.

This is consistent with how Azure's `message.includes('enter your org name manually below')` works, but worth calling out. Safest fix: return a structured error code from the Rust side (e.g. `{ code: 'trello_missing_app_key', message: '...' }`) so the TS side matches on the code, not prose. Short of that, at minimum put the sentinel string in a named constant shared between the Rust error site and this check.

ui/components/IntegrationConnect.tsx
```diff
@@ -245,8 +266,35 @@ function OAuthSetup({ tracker, onSuccess }: { tracker: Tracker; onSuccess?: () =
...
+          {apiKeyPrompt && (
+            <>
+              <p className="text-[12px]" style={{ color: '#d97706' }}>
+                A Trello API key is required.{' '}
+                <a href="https://trello.com/app-key" target="_blank" rel="noopener noreferrer" style={{ color: 'var(--accent)' }}>Get it at trello.com/app-key ↗</a>
+              </p>
+              <label className="block">
+                <span className="text-[11px]" style={{ color: 'var(--ink-3)' }}>API Key</span>
+                <input
+                  type="text"
```
Akarsh-Hegde: **Code duplication: use the existing `Field` component.**

This hand-rolls `<label>` + `<input>` with the same styling as the `Field` component at line 327. `Field` already handles the label, mono input, `onEnter`, and consistent border/background CSS vars. You could instead construct a synthetic `TokenField` and reuse it:

```tsx
<Field
  field={{ name: 'api_key', label: 'API Key', placeholder: 'Paste your Trello API key', required: true }}
  value={apiKey}
  onChange={setApiKey}
  onEnter={() => { if (apiKey.trim()) void startOAuth() }}
/>
```

Also — consider adding `autoFocus` to the input (or via `autoFocus` prop on `Field`) so the cursor lands in the field when the prompt appears, saving the user a click.

Akarsh-Hegde:
Addressed all three review comments in af68681f:

**Bug fix — `TRELLO_APP_KEY` cleanup on disconnect**
Added `("trello", &["TRELLO_APP_KEY"])` to `TOKEN_KEYS` so `disconnect_integration` now strips the env key in addition to removing the OAuth JSON. Updated the doc comment on `disconnect_integration` from \"trello = json only\" to \"trello = json + env key\".

**Brittle sentinel — named constant**
Extracted `'Power-Up app key'` to `TRELLO_MISSING_KEY_SENTINEL` at module scope in `IntegrationConnect.tsx`. The catch block now references the constant instead of the inline string literal, so a rewording of the Rust error message is a single-site change.

**Code duplication — reuse `Field` component**
Replaced the hand-rolled `<label>/<input>` block with `<Field field={{ name: 'api_key', label: 'API Key', placeholder: 'Paste your Trello API key', required: true }} ... autoFocus />`. Also added an `autoFocus?: boolean` prop to `Field` so focus lands in the input automatically when the prompt appears.

PR #366, opened 2026-06-29
feat(pm_tasks): add CDM columns + read path (Stage 3a)

Akarsh-Hegde:
## Summary

**Stage 3a** of the CDM migration: bring `pm_tasks` up to the `CanonicalTask` shape **additively**, and make the read path ready. Stacked on **#364** (which is stacked on #361) — review/merge the stack bottom-up; GitHub retargets each PR down as the one below merges.

This is the schema + read path only. **Populating the new columns is Stage 3b** (the per-provider adapter cutover) — see the scope note below for why that split is deliberate.

## What's included

**Migration `051_pm_tasks_canonical.sql` — additive, all columns nullable:**

| Column | ↔ CanonicalTask | Notes |
|---|---|---|
| `canonical_id` | `canonical_id` | `"{provider}:{provider_id}"` |
| `status_category` | `status_category` | canonical 6-value phase |
| `raw_payload` | `raw_payload` | verbatim tracker JSON |
| `reporter_name` | `reporter` | creator display name |
| `completed_at` | `completed_at` | ISO-8601 terminal timestamp |
| `ancestor_path` | `ancestor_path` | JSON array of canonical ids |
| `project_ids` | `project_ids` | JSON array of canonical ids |

Backfill: `status_category = 'done' WHERE is_terminal = 1`. Done-vs-Cancelled can't be told apart from `is_terminal` alone, so everything else stays `NULL` (unknown) until the next sync resolves it through the adapter. `issue_type` already carries the verbatim type (= `kind_raw`); `is_terminal` stays as the fast terminal flag.

**Reader (`meridian-core/src/readers/tasks.rs`):**
- `TaskSummary` exposes the new fields; `ancestor_path`/`project_ids` parsed from JSON arrays into `Vec<String>`.
- **Graceful column-presence guard** (`column_exists` via `pragma_table_info`) so the tray reading a **pre-051 DB** still works — the CDM columns are selected as `NULL` when absent (mirrors the existing active-session column-guard pattern in this crate).
- Two reader tests: columns-present (values surfaced, JSON arrays parsed) and pre-051-absent (fields default to `None`/empty).

## Deliberately NOT included (scope guard)

- **Writers untouched.** Populating these columns correctly *is* the per-provider adapter cutover (Stage 3b). The writers are per-provider `upsert`s (`src/intelligence/providers/*`); a stopgap generic `status_raw → category` mapping now would re-introduce exactly the low-fidelity bucketing the CDM exists to remove. So the columns land empty (except the backfill) and get populated when ingestion is cut over to the adapters from #364.
- No `raw_payload` in the reader output (large/internal; for writers).
- No behaviour change to existing sync.

## Testing

Validated against an isolated `meridian-core` **and** the real workspace pinned `Cargo.lock` (with the unrelated, unreachable `tray/src-tauri` member dropped — it has no bearing on this change):

- `cargo fmt --check` — clean
- `cargo clippy -p meridian-core -- -D warnings` — zero warnings
- `cargo test -p meridian-core` — green (78 unit + 12 reader integration, incl. the 2 new CDM reader tests)
- `cargo test -p meridian --test etl_basic --test migration_005_agents` — green (26 tests): these build the DB through the **full migration chain**, so they confirm **051 applies cleanly in sequence**.

Note: the whole-workspace build (tray + UI) can't run in this sandbox (the `screenpipe-fork` git dep is unreachable), but Stage 3a touches only the migration + `meridian-core` reader — not the tray/UI — so CI is the final backstop for those.

## Next (Stage 3b)

Wire the #364 adapters into the per-provider ingestion so `to_canonical()` output populates these columns (correct per-provider status/priority/hierarchy), replacing the ad-hoc field extraction in each `upsert`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BMZCqftam6FvXwZuYoBV8X)_


Akarsh-Hegde:
## Review — Stage 3a (CDM columns + read path)

### ⚠️ Blocking: migration number 051 collides with PR #367

`051_pm_tasks_canonical.sql` here and `051_tracking_pause_gaps.sql` in **#367** (`feat/timed-pause-work-hours`) are **two different migrations both numbered 051**. Neither is merged yet; `pre-main`/`main` top out at `050`. Whichever lands second collides.

Why this is more than a rename nit:
- **sqlx keys applied migrations by the numeric version (the `051` prefix), not the filename.** Two distinct `051`s → the daemon hits a duplicate-version / checksum mismatch on startup and **crash-loops**. Same failure class as the migration-031 renumber crash-loop we hit before (PR #102) — that's the precedent.
- **Renumbering the loser to 052 only saves *fresh* DBs.** Any machine/dev that already ran the losing branch's `051` has it recorded in its local `_sqlx_migrations`; after the renumber it crash-loops on the checksum until that table is reconciled (the 031 recovery dance).

Both PRs are equally entitled to 051 — this is a coordination problem, not a "wrong PR." The fix: **second-to-merge renumbers to 052+ *before* any DB applies it**, and we make sure no shared DB ran the losing `051` in the interim.

### The change itself — looks good

- Additive, all-nullable columns; existing rows untouched and older code keeps working. ✅
- `column_exists` guard (via `pragma_table_info`) so the tray on a pre-051 DB selects the CDM columns as `NULL` — mirrors the existing active-session column-guard pattern. ✅
- `parse_id_array` degrades to empty vec on NULL/empty/malformed JSON. ✅
- Backfill is conservatively scoped: `status_category='done' WHERE is_terminal=1`, leaving everything else NULL until the adapter resolves it. Done-vs-Cancelled can't be told apart from `is_terminal` alone, so historical cancelled-terminal rows read as `done` until the next sync — documented, acceptable, transient.
- Both reader tests (columns-present + pre-051-absent) are on point.

Schema ↔ `CanonicalTask` mapping matches #364's output. Once the 051 collision is resolved this is good to go; the writer cutover (Stage 3b) is correctly deferred.

Akarsh-Hegde:
Renumbered `051 → 052` in `e341ca5d` to clear the collision with #367's `051_tracking_pause_gaps.sql`.

This branch's CDM migration has only ever run on this feature branch (and my sandbox) — no shared/prod DB has applied it — so the renumber is clean (none of the 031-style `_sqlx_migrations` checksum recovery needed here). Comment refs in the reader/tests updated to `052`; the reader logic is number-independent (it guards on column existence via `pragma_table_info`, not the migration number).

Re-verified: `cargo test -p meridian --test etl_basic` (builds the DB through the full migration chain) passes with `052` in sequence.

One caveat for merge ordering: this resolves the collision **as long as #367 keeps `051`**. If #367 also moves, we'll collide again at `052` — so whichever of us lands second should confirm the next free number before merge.

---

PR #367, opened 2026-06-29
feat(capture): timed pause, work hours schedule, and full engine stop on pause

Akarsh-Hegde:
## Summary

- **Timed pause**: new `pause_for_duration` Tauri command lets users pause capture for 5 min / 30 min / 1 h / 4 h via the popover duration picker; auto-resumes when the timer expires.
- **Work hours schedule**: settings UI adds a Work Hours toggle + start/end time + active days; the poll loop auto-pauses when outside the window and resumes when inside it.
- **Full engine stop on pause**: pause **completely stops ScreenCaptureKit** (the screen recording engine) rather than just dropping frames — privacy guarantee that no screen data is captured during a pause.
- **`tracking_paused` / `schedule_paused` gap kinds**: migration 051 adds the new gap kinds; both are shown as amber bands in the DayTimeline with hover labels ("Paused" / "Outside work hours").
- **Immediate UI feedback**: `pause_for_duration` and `resume_capture` emit a `status-update` event immediately so the popover updates without waiting for the next 30 s poll tick.

## Technical notes

- Capture engine stop uses `tokio::sync::oneshot` cancel senders (not `AbortHandle`): dropping `engine_cancel_tx` → engine exits `select!` → drops `mpsc::Sender` → frame consumer exits naturally via `rx.recv() == None` → cascading clean shutdown.
- `start_capture()` extracted to `lib.rs` as a shared helper for initial startup, manual resume, and work-hours auto-resume — idempotent: drops old senders before spawning new tasks.
- `tauri::async_runtime::spawn` used throughout (not `tokio::task::spawn`) — required because Tauri's `setup()` runs on the macOS main thread, not a tokio worker.

## Test plan

- [ ] Pause for 5 minutes → `SELECT MAX(timestamp) FROM capture_frames` before/after 6 s → timestamps identical ✅ (verified locally)
- [ ] Resume manually → capture resumes, gap row written with correct `started_at`/`ended_at`/`duration_s`/`kind = 'tracking_paused'`
- [ ] Enable work hours with a window that excludes the current time → auto-pause triggers on next poll tick; popover shows schedule panel with next resume time
- [ ] Popover updates immediately on pause/resume (no 30 s wait)
- [ ] Work hours settings save/load correctly via Settings → Work Hours card
- [ ] DayTimeline shows amber "Paused" band over the paused interval

https://claude.ai/code/session_01KC7PxxJNYtPsaPeXipy2mo

### DIFF
--- a/src/etl/runner.rs
+++ b/src/etl/runner.rs
@@ -6,7 +6,7 @@ use tracing::{debug, info, warn};
 
 use crate::db::meridian::{
     close_active_session, complete_etl_run, get_active_session, get_cursor, insert_etl_run,
-    insert_gap, update_cursor,
+    insert_gap, pause_gap_exists_in_window, update_cursor,
 };
 use crate::db::screenpipe::{count_frames_in_window, get_frames_since};
 
@@ -104,15 +104,32 @@ pub async fn run_etl(meridian: &SqlitePool) -> Result<Vec<i64>> {
                         gap_frame_ids = format!("{}..{}", stale.min_frame_id, first_frame.id),
                         "cross-run gap detected — inserting gap record and closing stale active_session"
                     );
-                    insert_gap(
+                    // Skip the gap insert if the tray already recorded a
+                    // tracking_paused / schedule_paused gap covering this window.
+                    let covered = pause_gap_exists_in_window(
                         meridian,
                         &stale.last_seen_at,
                         &first_frame.timestamp,
-                        gap_secs,
-                        kind,
-                        run_id,
                     )
-                    .await?;
+                    .await
+                    .unwrap_or(false);
+                    if covered {
+                        debug!(
+                            from = stale.last_seen_at,
+                            to = first_frame.timestamp,
+                            "cross-run gap skipped — covered by a tray pause gap"
+                        );
+                    } else {
+                        insert_gap(
+                            meridian,
+                            &stale.last_seen_at,
+                            &first_frame.timestamp,
+                            gap_secs,
+                            kind,
+                            run_id,
+                        )
+                        .await?;
+                    }
                     close_active_session(meridian, run_id).await?;
                     info!(
                         gap_secs,
@@ -233,17 +250,33 @@ pub async fn run_etl(meridian: &SqlitePool) -> Result<Vec<i64>> {
                                 "intra-batch gap detected — inserting gap record and closing current block"
                             );
 
-                            insert_gap(
+                            // Skip if the tray already recorded a pause gap.
+                            let covered = pause_gap_exists_in_window(
                                 meridian,
                                 &block_last_ts,
                                 &frame.timestamp,
-                                gap,
-                                kind,
-                                run_id,
                             )
-                            .await?;
-
-                            debug!(gap_secs = gap, gap_kind = kind, from = block_last_ts, to = frame.timestamp, "gap recorded");
+                            .await
+                            .unwrap_or(false);
+                            if covered {
+                                debug!(
+                                    gap_secs = gap,
+                                    from = block_last_ts,
+                                    to = frame.timestamp,
+                                    "intra-batch gap skipped — covered by a tray pause gap"
+                                );
+                            } else {
+                                insert_gap(
+                                    meridian,
+                                    &block_last_ts,
+                                    &frame.timestamp,
+                                    gap,
+                                    kind,
+                                    run_id,
+                                )
+                                .await?;
+                                debug!(gap_secs = gap, gap_kind = kind, from = block_last_ts, to = frame.timestamp, "gap recorded");
+                            }
 
                             // Close the pre-gap block so its duration_s ends at block_last_ts,
                             // preventing gap time from inflating the session's focus duration.
--- a/tray/src-tauri/src/commands/daemon.rs
+++ b/tray/src-tauri/src/commands/daemon.rs
@@ -120,6 +120,13 @@ pub async fn pause_for_duration(
 
     {
         let mut s = state.lock().map_err(|e| e.to_string())?;
+        // Abort the engine and UI consumer tasks — stops actual screen capture.
+        if let Some(h) = s.engine_abort.take() {
+            h.abort();
+        }
+        if let Some(h) = s.ui_consumer_abort.take() {
+            h.abort();
+        }
         s.capture_paused.store(true, Ordering::Relaxed);
         s.pause_until = Some(until);
         s.pause_source = Some(PauseSource::Timed);
@@ -219,6 +226,10 @@ pub(crate) async fn resume_capture(
             }
         }
     }
+
+    // Restart the capture engine so screen recording resumes.
+    #[cfg(feature = "capture")]
+    crate::start_capture(state.clone(), pool.cloned());
 
     // Emit immediately so the popover reverts to the picker without waiting for the next tick.
     if let Ok(s) = state.lock() {
--- a/tray/src-tauri/src/commands/daemon.rs
+++ b/tray/src-tauri/src/commands/daemon.rs
@@ -120,13 +120,10 @@ pub async fn pause_for_duration(
 
     {
         let mut s = state.lock().map_err(|e| e.to_string())?;
-        // Abort the engine and UI consumer tasks — stops actual screen capture.
-        if let Some(h) = s.engine_abort.take() {
-            h.abort();
-        }
-        if let Some(h) = s.ui_consumer_abort.take() {
-            h.abort();
-        }
+        // Drop cancel senders → stops the engine and UI consumer tasks, fully
+        // halting ScreenCaptureKit and the CGEventTap recorder.
+        drop(s.engine_cancel.take());
+        drop(s.ui_consumer_cancel.take());
         s.capture_paused.store(true, Ordering::Relaxed);
         s.pause_until = Some(until);
         s.pause_source = Some(PauseSource::Timed);
--- a/tray/src-tauri/src/commands/daemon.rs
+++ b/tray/src-tauri/src/commands/daemon.rs
@@ -14,11 +14,14 @@
 //! - [`crate::sys`] — shared `uid_str` (launchctl domain) + `notify` (toast).
 //! - [`crate::poll::notifications_allowed`] — quiet-hours gate for the pause toast.
 
-use crate::state::{AppState, StatusPayload};
+use crate::state::{AppState, PauseSource, StatusPayload};
 use crate::sys;
+use chrono::{DateTime, SecondsFormat, Utc};
+use meridian_core::SqlitePool;
 use serde::Serialize;
+use std::sync::atomic::Ordering;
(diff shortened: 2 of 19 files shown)


tray/src/app.js
```diff
@@ -212,16 +261,44 @@ function render(status) {
...
-    pauseSec.classList.remove('paused')
-    pauseText.textContent = 'Pause tracking'
-    pauseSub.textContent = ''
-    pauseBtn.textContent = 'Pause'
-  } else {
+  if (!healthy) {
+    // Daemon down — show a minimal paused state; no interaction needed.
     pauseSec.classList.add('paused')
     pauseText.textContent = 'Tracking paused'
     pauseSub.textContent = "Meridian isn't recording"
```
Akarsh-Hegde, line 270: **[Angle-B] Removed: daemon-restart path when unhealthy — no replacement**

Old `pauseBtn` (deleted in this PR) called `invoke('toggle_daemon', { is_running: false })` when the daemon was down, which ran `launchctl start com.meridiona.daemon`. That path is now gone — `toggle_daemon` is completely absent from the new `app.js`.

The new unhealthy branch (lines 264–271) renders the static text "Meridian isn't recording" with no action button and no invoke call.

**Failure scenario:** Daemon crashes or is stopped outside the tray. User opens the popover and sees the unhealthy state. There is no button to restart it. They must use a Terminal, the dashboard, or restart the tray entirely to recover.

tray/src-tauri/src/commands/daemon.rs
```diff
@@ -87,6 +90,155 @@ pub async fn toggle_daemon(app: tauri::AppHandle, is_running: bool) -> Result<()
...
+    {
+        let mut s = state.lock().map_err(|e| e.to_string())?;
+        // Drop cancel senders → stops the engine and UI consumer tasks, fully
+        // halting ScreenCaptureKit and the CGEventTap recorder.
+        drop(s.engine_cancel.take());
+        drop(s.ui_consumer_cancel.take());
+        s.capture_paused.store(true, Ordering::Relaxed);
+        s.pause_until = Some(until);
+        s.pause_source = Some(PauseSource::Timed);
+        s.pause_started_at = Some(now);
```
Akarsh-Hegde, line 159: **[Angle-B] Removed: guard against clobbering an active schedule pause**

`pause_for_duration` unconditionally overwrites `pause_source` to `PauseSource::Timed` (line 129) and resets `pause_started_at` to now (line 130) without checking whether a schedule pause is already active.

If `check_work_hours` has set a schedule pause (source = `Schedule`, started_at = T0) and `pause_for_duration` is then called (e.g. user taps a pill from another window), the original T0 is lost. When the timed pause expires and `resume_capture` runs at T2, it writes a gap row of kind `tracking_paused` from T1 to T2 — the entire out-of-hours period from T0 to T1 gets no gap row at all.

**Failure scenario:** User's work hours end at 18:00 (schedule pause starts). At 18:05 the user manually pauses for 30 minutes. At 18:35 auto-resume fires. The timeline shows a `tracking_paused` gap from 18:05–18:35 but nothing for 18:00–18:05; the ETL fills that window as `user_idle` or `system_sleep`, corrupting the timeline record for that afternoon.

src/etl/runner.rs
```diff
@@ -233,17 +250,33 @@ pub async fn run_etl(meridian: &SqlitePool) -> Result<Vec<i64>> {
...
-                                gap,
-                                kind,
-                                run_id,
                             )
-                            .await?;
-
-                            debug!(gap_secs = gap, gap_kind = kind, from = block_last_ts, to = frame.timestamp, "gap recorded");
+                            .await
+                            .unwrap_or(false);
+                            if covered {
```
Akarsh-Hegde, line 261: **[Angle-B] Narrowed: any pause overlap suppresses the entire ETL gap, including uncovered portions**

`pause_gap_exists_in_window` uses `started_at < to AND ended_at > from` (ANY overlap) to decide whether to skip inserting the ETL gap. The old invariant was: every inter-frame interval ≥ GAP_THRESHOLD gets a gap row. The new invariant is: skip only if the pause gap covers the *entire* interval.

If a `tracking_paused` or `schedule_paused` row partially overlaps the ETL gap window — for example, the user resumes capture mid-gap so the pause ended before `frame.timestamp` — `count > 0` still returns `true` and the uncovered trailing portion of the gap is silently dropped. No ETL gap row is inserted for it.

**Failure scenario:** User pauses at 14:00, resumes at 14:20, then the machine sleeps 14:20–14:35. ETL detects a 35-minute gap from 14:00 to 14:35. The pause gap covers 14:00–14:20. `pause_gap_exists_in_window` returns true (partial overlap), ETL skips the entire window. The 14:20–14:35 sleep period is never recorded as `system_sleep`; the dashboard shows a clean 35-minute block rather than the actual 15-minute sleep stretch, undercounting idle/sleep time for that day.

Akarsh-Hegde:
**[Code Review — Angle C: Cross-file Tracer]**

**Finding 1 — TOCTOU in `check_work_hours` overrides `pause_for_duration`** (`tray/src-tauri/src/poll/mod.rs`, line 292–314)

`check_work_hours` reads `pause_source = None` under lock (lines 292–299), releases the lock, then re-acquires it at line 310 and unconditionally writes `pause_source = Some(PauseSource::Schedule)`. `pause_for_duration` is a Tauri command that runs on its own async task; if the user clicks a duration button in the narrow window between the two locks, their `pause_source = Timed` is silently overwritten with `Schedule`.

Compounding: the `(false, None)` arm (lines 302–317) never clears `pause_until`. The auto-resume timer spawned by `pause_for_duration` guards with `s.pause_until == Some(until)` (daemon.rs line 161–163), which still matches. When the timer fires, it calls `resume_capture`, which at that point reads `pause_source = Schedule` and writes a `schedule_paused` gap (daemon.rs line 206) instead of `tracking_paused`. The user's 5-minute timed pause becomes an indefinite schedule pause; the popover shows "Outside work hours" instead of a countdown; the gap record carries the wrong kind.

**Fix:** In the `(false, None)` arm, after re-acquiring the lock, re-check `s.pause_source.is_none()` before writing, and clear `s.pause_until` to defuse the stale auto-resume timer.

Akarsh-Hegde:
**[Code Review — Angle C: Cross-file Tracer]**

**Finding 2 — Duplicate CGEventTap on rapid resume** (`tray/src-tauri/src/capture/ui_events.rs`, line 115)

`run_ui_event_recorder` runs on a dedicated OS thread. It checks `tx.is_closed()` before each `recv_timeout(Duration::from_millis(500))` call (lines 115–121). When `check_work_hours` (resume arm, poll/mod.rs line 363) calls `crate::start_capture`, a new OS recorder thread is spawned immediately — but the previous recorder thread only detects the closed channel at its next 500ms poll. For up to 500ms both threads hold an active `CGEventTap` instance on the same process.

The upstream CGEventTap documentation does not define behaviour for two concurrent taps from the same process on the same event mask. In practice this can cause double-delivered input events (each keystroke written twice to `capture_ui_events`) and may stall one of the taps under high input load.

**Fix:** Signal the old recorder thread to stop (or wait for it via a `JoinHandle`) before spawning the new one in `start_capture`, reducing the overlap window to zero.

Akarsh-Hegde:
**[Code Review — Angle C: Cross-file Tracer]**

**Finding 3 — DB error in `pause_gap_exists_in_window` silently allows duplicate gap insertion** (`src/etl/runner.rs`, lines 115 and 260)

Both the cross-run gap check (line 115) and the intra-batch gap check (line 260) guard ETL gap insertion with:
```rust
let covered = pause_gap_exists_in_window(...).await.unwrap_or(false);
```
`unwrap_or(false)` maps any DB error (pool exhausted, disk I/O error, SQLITE_BUSY under WAL contention) to `false` — meaning "no pause gap exists". On a transient DB error the ETL runner proceeds to insert a `user_idle` or `system_sleep` gap that overlaps an existing `tracking_paused` / `schedule_paused` gap, defeating the dedup logic.

The correct behaviour on a query error is to propagate it (let the ETL run mark as `failed`) rather than silently insert a duplicate gap that corrupts the timeline.

**Fix:** Replace `.unwrap_or(false)` with `?/await?/context("pause_gap_exists_in_window")` so a DB error surfaces to the caller and halts the ETL batch cleanly.

Akarsh-Hegde:
**[Code Review — Angle C: Cross-file Tracer]**

**Finding 4 — `check_work_hours` resume arm re-reads stale `started_at` from pre-lock snapshot** (`tray/src-tauri/src/poll/mod.rs`, lines 292–298 and 326–328)

The `(true, Some(PauseSource::Schedule))` arm (line 320) computes `duration_s` using the `started_at` value captured in the outer lock snapshot at lines 292–298 — before the lock is released. This is correct *for the schedule case*, but there is a subtle caller coupling: `pause_for_duration` (daemon.rs line 130) always overwrites `pause_started_at = Some(now)` when it fires. If `pause_for_duration` interleaved and completed between the snapshot and the resume arm's re-lock at line 354, the `started_at` used to write the `schedule_paused` gap row (line 330) is the *timed* pause start time from `pause_for_duration`, not the original schedule-pause start — so the gap covers the wrong interval. This is a second manifestation of the same TOCTOU root as Finding 1 but on the resume path, producing a misattributed gap start time in the DB.

**Fix:** Under the resume arm's re-acquired lock (line 354), read `s.pause_started_at` fresh from state instead of using the outer snapshot.

Akarsh-Hegde:
Fixed all three review comments in cf5d47f:

**1. Unhealthy state — no restart button (app.js)**
Added a `daemonUnhealthy` module-level flag. In the `!healthy` branch, `pauseActive` is now shown (not hidden), the countdown span is hidden, and `resumeBtn.textContent` is set to `'Restart daemon'`. The click handler branches on `daemonUnhealthy`: unhealthy → `invoke('restart_daemon')`, paused → `invoke('pause_for_duration', { seconds: 0 })`. All other branches reset `daemonUnhealthy = false` and restore `resumeBtn.textContent = 'Resume now'` + `pauseCountdown.hidden = false` to prevent stale state across transitions.

**2. Schedule → timed pause transition drops gap (daemon.rs)**
Before overwriting `pause_started_at` / `pause_source`, `pause_for_duration` now reads the previous values under a short lock, releases it, then writes a gap row for the T0→now period (`insert_pause_gap`). Matches the `tracing::warn!` error pattern from `resume_capture`. The second lock then applies the new timed-pause state as before.

**3. `pause_gap_exists_in_window` any-overlap vs full-coverage (meridian.rs)**
Changed `started_at < ?2 AND ended_at > ?1` (any overlap) to `started_at <= ?1 AND ended_at >= ?2` (full coverage). The ETL now only skips inserting a gap row when a pause gap covers the *entire* detected interval; partial overlaps correctly produce an ETL gap. Both call sites in `runner.rs` (cross-run gap at line 109 and intra-batch gap at line 254) use the same function, so both get the fix.

PR #372, opened 2026-06-30
ci(runtime): auto-publish runtime on merge with a version gate (#2/#3)

Akarsh-Hegde:
Replaces the manual "remember to tag a runtime release" flow — the footgun that left production stuck on a broken `1.66.2` runtime — with merge-triggered publishing, gated so it can never downgrade or silently no-op.

## ⚠️ MERGE ORDER — merge this LAST

There is **no `paths:` filter**, so the moment this lands on `pre-main`, the next push to `pre-main` runs the gate, sees pre-main's version (`1.64.x`) > staging-live (`1.63.0`), and **auto-builds + publishes a staging runtime**. If this merges before the others, it would publish a runtime that still contains the `WORKLOG_SYSTEM` bug, smoke-tested by the *old weak* smoke script that can't catch it.

**Required order on `pre-main`:**
1. **#369** — hardened smoke test (walk-imports every agents submodule) — so a broken runtime can't pass smoke.
2. **#365** — worklog `WORKLOG_SYSTEM` fix — so the published runtime isn't broken.
3. **this PR** — turns on auto-publish.

## Expected side effect
Merging this to `pre-main` triggers an **immediate staging runtime build + publish** (a long macOS CI job), not a passive config change. That's desirable here (staging is behind at `1.63.0`), but it's not silent — flagging so it's not a surprise.

## #2 — auto-publish on merge (`build-mlx-runtime.yml`)
- New **`gate`** job (`scripts/runtime-publish-gate.sh`), the single source of truth: derives the channel (`pre-main → runtime-staging`, `main → runtime-latest`) and decides `should_publish` by comparing `services/pyproject.toml` to the channel's live `runtime-manifest.json`.
  - **branch push → publish only when local version is strictly `>` live** (no downgrade, no equal-version churn — a merge that lands a stale version just doesn't ship).
  - **tag push → publish on any difference** (the rollback/re-pin escape hatch, unchanged).
  - **manifest fetch error → fail-closed.**
  - Self-tested: `runtime-publish-gate.sh --self-test` (20 assertions).
- `build` runs only when the gate says publish (or `workflow_dispatch` for a test build). Two explicit publish jobs; **`publish-production` runs in the `production-runtime` GitHub Environment → required-reviewer approval**, so a `main` merge can't reach customers unattended. (Environment already created: reviewers Akarsh-Hegde + adityaharishch, `main`-only.)
- Publish jobs are non-cancellable; `build` cancels stale in-flight runs. Release upload factored into `scripts/publish-runtime-channel.sh` (shared by both publish jobs).
- **Why no `paths:` filter:** GitHub applies `paths:` to tag pushes too, which would silently break the `runtime-v*` rollback hatch. The version gate self-limits to one publish per version-ahead state, so a path filter would be redundant.

## #3 — version-bump PR check (`services-version-bump.yml`)
- A PR touching `services/**` must bump the version above its **target channel's** live runtime (`main → runtime-latest`, `pre-main → runtime-staging`), via `scripts/check-runtime-version-bump.sh`. Catches the silent no-op *before* merge.
- Per-channel (not `max(both channels)`) on purpose: `pre-main` legitimately trails `main` under `set-version` lockstep, so `max()` would make every staging services PR unmergeable. Promotion safety still holds — a `pre-main → main` promotion is a PR with **base = main**, so it's checked against production-live and **fails loudly** on a stale version.

## Verification (local)
- `runtime-publish-gate.sh --self-test`: 20/20 (strict-`>` for branches, `differ` for tags, fail-closed on error, channel mapping incl. app `v*` tags → no publish).
- CI-mode dry run against the live repo: `pre-main` → `should_publish=true` (1.64.0 > 1.63.0); `main` with a stale 1.64.0 → `should_publish=false` (would-be downgrade blocked).
- version-bump check: base `pre-main` passes (1.64.0 > 1.63.0); base `main` fails loudly (1.64.0 ≯ 1.66.2).
- Both workflows parse; all jobs present; `publish-production` carries `environment: production-runtime`.

## Minor (non-blocking)
Build artifacts have `retention-days: 7`. If a production approval ever sits >7 days in the environment gate, the publish job would fail on artifact download. Unlikely; noting it.

https://claude.ai/code/session_01CT9wnEUuQeoBeufqTh46JC

### DIFF
--- a/.github/workflows/build-mlx-runtime.yml
+++ b/.github/workflows/build-mlx-runtime.yml
@@ -7,17 +7,31 @@ name: Build MLX runtime
 # macOS >= the floor in the manifest. The ~7 GB model is NOT bundled (downloaded
 # on first request).
 #
-# build (macos-14, arm64) → smoke (macos-{14,15,26}, arm64 — DIFFERENT machines/OS,
-# so relocation + ABI + deployment-target are exercised on boxes that didn't build
-# it) → publish (only on a release tag) force-updates the channel's rolling
-# GitHub release with the tarball + manifest. workflow_dispatch builds +
-# smoke-tests without publishing.
+# FLOW: gate → build (macos-14) → smoke (macos-{14,15,26}) → publish.
 #
-# CHANNELS (selected by the tag prefix; tag events run the workflow from the
-# tagged commit, so the staging channel works off the `staging` branch without
-# touching main):
-#   runtime-staging-v*  →  rolling `runtime-staging` release   (staging channel)
-#   runtime-v*          →  rolling `runtime-latest`  release   (production channel)
+# TRIGGERS & CHANNELS — the `gate` job is the single source of truth (see
+# scripts/runtime-publish-gate.sh); it derives the channel and decides whether
+# to publish, so nothing below re-implements that logic:
+#   push to pre-main    → runtime-staging   (auto)
+#   push to main        → runtime-latest    (auto; runs in the production-runtime
+#                                            Environment → required-reviewer gate)
+#   tag  runtime-staging-v* → runtime-staging  (manual escape hatch)
+#   tag  runtime-v*         → runtime-latest   (manual escape hatch)
+#   workflow_dispatch       → build + smoke only, never publishes
+#
+# PUBLISH DECISION (gate): compare services/pyproject.toml against the channel's
+# live runtime-manifest.json.
+#   - branch push  → publish only when local version is STRICTLY GREATER than the
+#                    channel's live version (never downgrades, never republishes
+#                    an equal version — so a merge that lands a STALE version just
+#                    doesn't ship, instead of silently downgrading customers).
+#   - tag push     → publish on ANY difference (trusts the human; the rollback /
+#                    re-pin escape hatch).
+#   - manifest fetch error → fail-closed (do not publish).
+# This is why there is no `paths:` filter: GitHub applies `paths:` to tag pushes
+# too, which would silently break the runtime-v* rollback hatch. The version gate
+# already self-limits to one publish per version-ahead state, so a `paths:` filter
+# would be redundant anyway — the gate job is cheap and short-circuits the rest.
 #
 # The app pins a channel's runtime-manifest.json (RUNTIME_MANIFEST_URL in
 # mlx_server.rs = runtime-latest; a staging build overrides via
@@ -28,15 +42,8 @@ name: Build MLX runtime
 # is stamped into the manifest's tarball url at build time so it points at the
 # same channel release.
 #
-# To cut a STAGING runtime release (tag a commit that has your services/ change):
-#   git tag runtime-staging-v$(grep -m1 '^version' services/pyproject.toml | sed -E 's/.*"([^"]+)".*/\1/')
-#   git push origin <that tag>
-# Production is the same with a `runtime-v*` tag — this workflow lives on `main`,
-# so production tags publish today (the first `runtime-v*` push CREATES the
-# `runtime-latest` release). The app's version-skip is an EQUALITY check, so the
-# tagged commit's services/pyproject.toml version MUST differ from the currently
-# published runtime or every machine silently keeps the old code. Full playbook:
-# CLAUDE.md → Common Tasks → "Ship a services/ Python change to users".
+# Full playbook: CLAUDE.md → Common Tasks → "Ship a services/ Python change to
+# users".
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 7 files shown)


Akarsh-Hegde (commented):
## Review — release-automation looks solid, one wired-up gap + two notes

Reviewed end-to-end (gate → build → smoke → publish, the gate decision logic, and the version-bump PR check). The design is sound: artifact name (`mlx-runtime`) is consistent across build/smoke/publish, skip-propagation is correct (build skipped → smoke skipped → publish skipped, and the publish jobs correctly avoid `always()`), the per-channel non-cancellable publish concurrency is right, and `decide_publish` fail-closes on fetch error / unknown status. Verified the decision table locally — all 17 self-test assertions pass.

### 🟠 1 — The self-test is dead code (addressing in this PR)
`scripts/runtime-publish-gate.sh` ships a thorough `--self-test` (the comment even says it "exercises the decision table the advisor called out"), but **no workflow ever runs it**. The downgrade-prevention (`version_gt`), fail-closed (`decide_publish` on `error`/unknown), and channel-routing (`channel_for_ref`) logic — the whole safety story of this PR — could regress silently.

**Fix pushed:** `.github/workflows/runtime-gate-selftest.yml` — a `pull_request` check that runs `runtime-publish-gate.sh --self-test`. Deliberately a **PR check, not a step in the `gate` job**: it catches regressions *before* they reach pre-main/main, and a self-test failure can never block a live runtime publish (keeps the gate→publish path uncoupled from the test). Runs on PRs touching the gate/version-bump scripts.

### 🟡 2 — `services-version-bump.yml`: required-check + `paths:` filter trap (config note, no code change)
`services-version-bump.yml` only triggers on `paths: services/**`. If it's later marked a **required** status check in branch protection, every PR that doesn't touch `services/**` will hang forever waiting for a check that never reports (GitHub's well-known required-check + paths-filter trap). Recommend either: keep it **non-required** and lean on the auto-publish gate's strict-greater rule as the real fail-safe (which this PR already frames as the final safeguard), or convert it to an always-run job that internally short-circuits. Flagging so it's a conscious branch-protection decision.

### 🟢 3 — `fetch_live_version` can wedge a channel (low severity, acceptable)
In `runtime-publish-gate.sh`, if a channel release *exists* but its `runtime-manifest.json` asset is missing/corrupt, `gh release download -p runtime-manifest.json` fails → `FETCH_STATUS=error` → fail-closed forever for that channel until manually fixed. Correct fail-closed behavior (never the wrong publish), just noting the half-published-release edge case as a known recovery scenario. No change needed.

Posting as a comment, not request-changes, since #1 is self-resolved here. Will follow up once the PR check goes green.

Akarsh-Hegde:
## Update — review finding addressed, CI green ✅

Pushed `9416e434` resolving the one actionable issue from the review.

**Fixed (🟠 #1 — dead self-test):** added `.github/workflows/runtime-gate-selftest.yml`, a `pull_request` check that runs `scripts/runtime-publish-gate.sh --self-test` (the decision-table tests for `version_gt` / `decide_publish` / `channel_for_ref` — the downgrade-prevention and fail-closed logic). It was previously wired into no workflow.
- **Placement is intentional:** a PR check, *not* a step in the `gate` job — so it catches regressions before they reach `pre-main`/`main`, and a self-test failure can never block a live runtime publish (the gate→publish path stays uncoupled from the test).
- **Validated in real CI:** the new check ran on this PR and passed (4s) — [run](https://github.com/Meridiona/meridian/actions/runs/28429097666).

**Left as notes (non-blocking, no code change):**
- 🟡 #2 — if `services-version-bump.yml` is ever marked a *required* status check, the `paths: services/**` filter will make PRs that don't touch `services/**` hang (GitHub's required-check + paths trap). Keep it non-required, or make it an always-run job that self-skips.
- 🟢 #3 — `fetch_live_version` fail-closes permanently if a channel release exists but its `runtime-manifest.json` asset is missing/corrupt. Correct (never the wrong publish); just a known half-published-release recovery case.

### Merge-readiness ✅
- Prerequisites merged on `pre-main`: **#365** (`WORKLOG_SYSTEM` fix) and **#369** (hardened smoke). The "merge this last" ordering is satisfied.
- State: **MERGEABLE / CLEAN**; checks green.
- On merge to `pre-main`, the gate will publish: `services/pyproject.toml` is **1.64.1** > live `runtime-staging` **1.63.0** → `should_publish=true` → build → smoke → publish `runtime-staging` at 1.64.1 (carrying the #365 worklog fix).

Good to merge.

PR #375, opened 2026-06-30
fix(runtime): authenticate the python-build-standalone API lookup (403 → build fails)

Akarsh-Hegde:
## Why
`scripts/build-mlx-runtime.sh` resolves the latest python-build-standalone release via an **unauthenticated** `api.github.com` call (60 req/hr/IP). CI runners share IPs, so it intermittently **403s** (`curl exit 56`) and the build dies before producing the tarball.

This already bit us: the **first auto-publish run** after #372 merged to `pre-main` failed at the build step on this 403 and only passed on re-run ([run 28429586316](https://github.com/Meridiona/meridian/actions/runs/28429586316)).

The real risk is **production**: `publish-production` is approval-gated (the `production-runtime` Environment). A 403 there means *approve → build 403s → re-run → approve again* on a gated, single-shot publish that delivers customer fixes.

## Fix
- `build-mlx-runtime.sh`: add a `Authorization: Bearer` header to the **api.github.com lookup only** when a token is present (`GH_TOKEN`, falling back to `GITHUB_TOKEN`), raising the limit to 1000/hr. Header is omitted when no token is set, so **local builds still work**. The asset download is a CDN redirect and is deliberately left unauthenticated.
- `build-mlx-runtime.yml`: pass `secrets.GITHUB_TOKEN` into the build step's env.

## Validation
- `bash -n` clean; `ruby -ryaml` confirms the workflow YAML is valid.
- The conditional array expansion (`${PBS_AUTH[@]+...}`) is verified safe under `set -euo pipefail` on **bash 3.2.57** — the exact version the macOS runner ships — both with and without a token:
  - no token → `args: []` (unauthenticated, no unbound-var error)
  - with token → `-H "Authorization: Bearer …"`

Clears the remaining caveat on the `pre-main → main` promotion path (with #372 + #373).

### DIFF
--- a/.github/workflows/build-mlx-runtime.yml
+++ b/.github/workflows/build-mlx-runtime.yml
@@ -61,7 +61,9 @@ on:
       - "runtime-staging-v*"  # staging channel → runtime-staging
 
 permissions:
-  contents: write   # publish jobs attach assets to the release
+  contents: read   # least-privilege default; only the publish jobs override to write.
+                   # gate/build/smoke run third-party code (pip backends, the built
+                   # runtime) and must not hold a release-write token.
 
 jobs:
   gate:
@@ -151,6 +153,9 @@ jobs:
     needs: [gate, smoke]
     if: needs.gate.outputs.should_publish == 'true' && needs.gate.outputs.channel == 'runtime-staging'
     runs-on: ubuntu-latest
+    # Only the publish jobs need write — they create/upload the channel release.
+    permissions:
+      contents: write
     # Serialize publishes per channel and never cancel a publish mid-upload.
     concurrency:
       group: mlx-runtime-publish-staging
@@ -179,6 +184,9 @@ jobs:
     needs: [gate, smoke]
     if: needs.gate.outputs.should_publish == 'true' && needs.gate.outputs.channel == 'runtime-latest'
     runs-on: ubuntu-latest
+    # Only the publish jobs need write — they create/upload the channel release.
+    permissions:
+      contents: write
     # The production-runtime Environment carries the required-reviewer rule — a
     # maintainer must approve before this job runs and customers get the runtime.
     environment: production-runtime
--- a/.github/workflows/build-mlx-runtime.yml
+++ b/.github/workflows/build-mlx-runtime.yml
@@ -104,6 +104,10 @@ jobs:
           # not publish, so default to runtime-latest for the build stamp.
           RUNTIME_TAG: ${{ needs.gate.outputs.channel || 'runtime-latest' }}
           GITHUB_REPOSITORY: ${{ github.repository }}
+          # Authenticates the api.github.com python-build-standalone lookup in
+          # build-mlx-runtime.sh so it doesn't hit the 60/hr unauthenticated
+          # rate limit (403 → build fails; bit the first auto-publish run).
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: bash scripts/build-mlx-runtime.sh
 
       - name: Upload runtime artifact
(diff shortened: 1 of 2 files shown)


.github/workflows/build-mlx-runtime.yml
```diff
@@ -104,6 +104,10 @@ jobs:
           # not publish, so default to runtime-latest for the build stamp.
           RUNTIME_TAG: ${{ needs.gate.outputs.channel || 'runtime-latest' }}
           GITHUB_REPOSITORY: ${{ github.repository }}
+          # Authenticates the api.github.com python-build-standalone lookup in
+          # build-mlx-runtime.sh so it doesn't hit the 60/hr unauthenticated
+          # rate limit (403 → build fails; bit the first auto-publish run).
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Akarsh-Hegde, line 112: The `build` job inherits `contents: write` from the workflow-level `permissions` block (line 53), but it only needs read access. Since `build` and `publish` are separate jobs, GitHub lets you override permissions per-job — adding a job-level block genuinely limits what a stolen token can do.

The `pip install [mlx,pm_worklog_update]` in the same step runs third-party Python build backends that can read environment variables. With `contents: write` scope, a compromised package could push commits or create releases against this repo. Scoping the build job to `contents: read` (or none) limits the blast radius to read-only.

Suggested fix:
```yaml
  build:
    name: Build (macos-14 arm64)
    runs-on: macos-14
    permissions:
      contents: read   # overrides workflow-level write; publish job handles the upload
```
The `publish` job already has its own `GH_TOKEN` in its step env and is unaffected.

Akarsh-Hegde (commented):
The fix is correct and the bash pattern is sound — the `` idiom is the right way to pass a conditionally-populated array under `set -euo pipefail` on bash 3.2, and the empty-token guard prevents sending a bare `Authorization: Bearer` header.

One finding worth addressing before merge.

.github/workflows/build-mlx-runtime.yml
Akarsh-Hegde: Fixed in cba2b0f3. Defaulted the workflow to `contents: read` and granted `contents: write` only to `publish-staging` / `publish-production` — the sole jobs that create/upload the channel release. `build` now inherits read, so it no longer holds a release-write token while running the third-party `pip install` backends (your exact concern). Same protection for `smoke` (runs the built runtime) and `gate` (reads only).

Verified resolved: `build`/`smoke`/`gate` → read, `publish-staging`/`publish-production` → write.

Akarsh-Hegde:
## Update — review feedback addressed, merge-ready ✅

Pushed `cba2b0f3` resolving the least-privilege comment on the `build` job.

**Security hardening (your review point):** the workflow now defaults to `contents: read`, with `contents: write` granted **only** to `publish-staging` / `publish-production` (the jobs that create/upload the channel release). `build` (running third-party `pip install` backends), `smoke` (running the built runtime), and `gate` (read-only) all inherit read — none of them hold a release-write token anymore.

Resolved per-job scopes:
| job | contents |
|---|---|
| gate / build / smoke | `read` (inherited) |
| publish-staging / publish-production | `write` |

**Original fix (this PR's purpose):** `build-mlx-runtime.sh` now authenticates the `api.github.com` python-build-standalone lookup with a Bearer token when present (60→1000 req/hr), fixing the intermittent 403 that failed the first auto-publish. Header omitted when no token → local builds unaffected. Verified `bash -n` clean and the array expansion safe on bash 3.2.57 (the runner version).

**Validation:** workflow YAML parses; per-job permissions resolve as the table above.

Both commits are CI-only (no `services/**`, no runtime version change). Mergeable / CLEAN. **Ready to merge.**

PR #377, opened 2026-07-01
feat(worklog): FSM JSON classification + proposed-ticket approval

adityaharishch:
## Summary
Productionizes the worklog pipeline's JSON-output LLM calls onto **FSM (outlines grammar-constrained decoding)** with thinking disabled, plus proposed-ticket approval. Merged cleanly with the latest `pre-main`.

- **FSM for JSON-output calls** (`/classify_tasks`, `/generate_worklog`, `/propose_ticket`): grammar-constrained decoding via outlines → **0 JSON parse failures** (was ~7% prose-spillover under thinking-budget). Prose calls keep thinking.
  - Fixed outlines double-templating (`has_chat_template = False`).
  - Confidence normalized/clamped in code (FSM can't enforce numeric `ge/le`).
  - `temp=0.1` chosen for recall (FN=0; `0.7` reintroduced silent-drops).
- **New modules**: `schemas.py` (FSM contracts), `structured.py` (FSM helper), `worklog_pipeline/classifier.py` + `generation.py` (direct-HTTP), replacing the agno `worklog.py`/`agent_io.py`/`match.py` path.
- **Correctness fixes**: `classifier._post_classify` now **raises** on transport/HTTP failure (was silently `return []` → dropped real matches); Rust `worklog_pipeline.rs` activity gate stays *pending* on transient DB error (was marking the hour permanently done).
- **Proposed-ticket description** rewritten to forward-looking present-tense ticket scope (no past-tense worklog framing).
- **Observability**: `record_fsm_params` stamps `decoding=fsm`, `fsm_schema`, `enable_thinking=false`, real token counts, no phantom penalties; `.prompt/.input/.output` child spans on every LLM span; end-to-end trace nesting verified in OpenObserve.
- New unit tests (`test_schemas`, `test_classifier`, `test_worklog_models`, `test_generate_helpers`); `pytest` added as a dev-only extra.

## Merge notes
Resolved one modify/delete conflict on `services/agents/worklog_pipeline/worklog.py` — **kept deleted** (its logic moved to `generation.py` + `routes/generate.py`; `pre-main`'s only change was a `WORKLOG_SYSTEM` re-export tweak to a file this branch removes). Regenerated `uv.lock` to match the merged `pyproject.toml`. Pre-push suite (fmt + clippy + cargo test + UI build) passed.

## Verification
- End-to-end worklog hour: matched tickets, 0 parse failures, 0 errors.
- OpenObserve: trace tree nests end-to-end; FSM span attributes + logs correlated by `trace_id`.
- `py_compile` / AST parse clean on merged `worklog_pipeline` + `routes`.

> Note: these `services/` changes reach users only after an MLX runtime republish (runtime is versioned independently of the app).


Akarsh-Hegde:
## Review — PR #377 (FSM JSON classification + proposed-ticket approval)

This PR productionizes the worklog pipeline's JSON-output LLM calls onto grammar-constrained (FSM) decoding via outlines, replaces the agno match-agent path with direct HTTP classifiers (`classifier.py`, `generation.py`), and refactors the Rust driver to a clock-aligned HH:03 trigger instead of an ETL-settle wake. The core FSM approach and the observability additions are solid. Five actionable findings below.

---

### Finding 1 — Confidence normalization silently drops real matches (HIGH)

**`services/agents/routes/classify.py`**, confidence normalization block:

```python
cf = float(c)
if cf > 1.0:
    cf = cf / 100.0
mtc["confidence"] = max(0.0, min(1.0, cf))
```

The `> 1.0` threshold is too tight. The model is prompted for a 0–1 scale and `models.py`'s own comment calls out `1.5` as "the model's occasional out-of-range value" — meaning a slight ceiling overshoot, **not** a 0–100 percentage. The normalization cannot tell the difference between `1.1` ("very high confidence, slight overshoot") and `50` ("50% in percentage scale"), so it divides both by 100: `1.1 → 0.011`, `1.5 → 0.015`. Both land below `_MIN_CONFIDENCE = 0.5` in `classifier.py` and the match is **silently dropped** — a real ticket produces no worklog.

Values in the range `(1.0, 2.0]` are almost certainly 0–1 overshoots; values `≥ 2.0` are almost certainly percentage-scale. The fix is `if cf >= 2.0: cf = cf / 100.0` (or `> 2.0`). There is no test covering the `(1.0, 2.0)` range.

---

### Finding 2 — `upsert_proposed_task` silently discards new proposal when prior is approved/dismissed (MEDIUM-HIGH)

**`services/agents/worklog_pipeline/db.py`**, `upsert_proposed_task`:

```sql
ON CONFLICT (day_utc, source_hour) DO UPDATE SET
    title = excluded.title, ...
WHERE pm_proposed_tasks.state = 'proposed'
```

When a user approves or dismisses an earlier proposal for `(day_utc, source_hour)`, the `WHERE state = 'proposed'` guard blocks the `DO UPDATE` — SQLite treats this as a no-op, `changes()` returns 0. The function then does an unconditional `SELECT id … WHERE day_utc=? AND source_hour=?`, finds the old approved/dismissed row, and returns its ID. The **caller receives the stale ID and believes the new proposal was persisted**.

In `pipeline.py::stage_generate`, `retract_proposed_task` is skipped when `proposed is not None`, so there is no prior cleanup. A pipeline re-run on an already-approved hour silently discards the new LLM-generated proposal; `res.proposed_id` points to the approved row; logs and OO traces show success.

Fix: check `conn.execute("SELECT changes()").fetchone()[0]` after the upsert and return `None` when it is 0 (the UPDATE was blocked), or retract unconditionally before re-proposing.

---

### Finding 3 — `upstream_settled` ETL boundary check removed; HH:03 fixed offset is not equivalent (MEDIUM)

**`src/worklog_pipeline.rs`**: The old driver called `ledger::upstream_settled(pool, hs, he, min_duration_s)` before firing `/worklog_hour`. That function verified ETL had actually written sessions past the hour boundary (`MAX(ended_at) >= he`). The new code fires at exactly HH:03 with no such check — `ledger::upstream_settled` is never called in `process_hour`.

The 3-minute offset assumes the ETL 60s poll cycle has run at least twice and caught up. This fails when:
- The daemon restarts between HH:00 and HH:03 (startup catch-up in `catch_up_today` fires immediately, possibly before ETL's first post-restart tick)
- ETL experiences a slow cycle (large frame batch, disk I/O contention, WAL checkpoint)

In both cases `/worklog_hour` fires against an incomplete `app_sessions` snapshot. After a 200-OK response, `ledger::mark_hour_done` permanently closes the hour — the missing sessions are never included. The old aging escape (90-minute fallback) is also gone. The fix is either to restore the boundary check in `process_hour` or add a startup delay before the first `catch_up_today` call.

---

### Finding 4 — `_post` duplicated in three modules with diverging timeout defaults (LOW-MEDIUM cleanup)

`pipeline.py` defines `_post(server, path, body, timeout=120)` and `generation.py` defines `_post(server_url, path, body, timeout=300)` — identical urllib bodies, different parameter name, different default timeout. `generation.py`'s docstring even says "Mirrors `classifier._post_classify`". The 120s vs 300s divergence is already load-bearing: `stage_report` calls `/activity_report` through pipeline's `_post` at 120s, while `/generate_worklog` gets 300s. A slow model during the activity report will timeout while the downstream worklog call would not.

Promote `generation.py::_post` as the canonical transport and import it in `pipeline.py`; add explicit `timeout=` at each call site that needs a shorter value.

---

### Finding 5 — Four test files start with `# ambient dev tool...` comment instead of `"""..."""` docstring (LOW, conventions)

**`services/agents/tests/test_classifier.py`**, **`test_generate_helpers.py`**, **`test_schemas.py`**, **`test_worklog_models.py`** — all four new test files begin with:
```
# ambient dev tool that watches what you do...
```
followed by a `"""..."""` docstring on line 2.

CLAUDE.md rule: *"Every `.py` file in `services/agents/` must start with a `"""…"""` module docstring describing its purpose. **The Rust/TS file-header convention does not apply — Python uses docstrings.**"*

The `# ambient dev tool...` comment is the Rust/TS header format; Python files must use the docstring form as the very first content. Remove the comment line so the `"""..."""` docstring becomes line 1.

---

### Non-finding (verified clean)

- `BATCH = _BATCH` is correctly re-exported from `classifier.py` (line 3109) — `scripts/run-classify-tier1.sh` import is valid.
- Pydantic v2 `ValidationError` IS a `ValueError` subclass — the `except (TypeError, ValueError)` handlers catch it correctly.
- All call sites for the changed `run_loop` (no `wake` arg), renamed `cli_run_hour`, and updated `workflow.py` step ordering are correctly patched in this PR.

adityaharishch:
Thanks for the thorough review, @Akarsh-Hegde. Addressed in `7d714b82`. Point-by-point below.

---

### Finding 1 — Confidence normalization silently drops real matches (HIGH) → **Fixed**

Confirmed and fixed. Extracted the inline block into a testable `_normalize_confidence(c)` in `classify.py` and changed the threshold so **only `>= 2.0` is treated as percentage scale (÷100)**; a value in `(1.0, 2.0)` is now clamped to `1.0` rather than divided. So `1.1`/`1.5` stay `1.0` (above `_MIN_CONFIDENCE`), while `50`/`90` still normalize to `0.5`/`0.9`.

Added `services/agents/tests/test_classify_helpers.py` — parametrized over the `(1.0, 2.0)` overshoot range, the `>=2` percentage range, negatives, and non-numeric input, plus an explicit assertion that a `1.1`/`1.5` overshoot stays `>= _MIN_CONFIDENCE` (the exact silent-drop guard).

### Finding 2 — `upsert_proposed_task` returns stale decided id (MEDIUM-HIGH) → **Fixed**

Confirmed. After the `INSERT … ON CONFLICT … DO UPDATE … WHERE state='proposed'`, it now checks `SELECT changes()`: when the guard blocked the update (an **approved/dismissed** row owns the key), `changes()==0` and the function returns **`None`** instead of falling through to the unconditional `SELECT id` and handing back the decided row's id. The caller (`stage_generate`) assigns straight into `res.proposed_id: int | None`, so `None` propagates cleanly — no stale id masquerading as a fresh proposal. Docstring updated to document the `None` return.

### Finding 3 — `upstream_settled` boundary check dropped (MEDIUM) → **Won't change (deliberate design); one premise corrected**

You're right that `process_hour` no longer calls `ledger::upstream_settled` (the fn still exists + is tested). But I'm treating this as an intentional part of the HH:03-clock rewrite, not a regression to undo, for two reasons:

1. **The daemon-restart race you describe is already mitigated.** In `main.rs` the startup ETL pass is **awaited synchronously (line 559) before the worklog `run_loop` is spawned (line 613)** — and `run_loop` calls `catch_up_today` on entry. So on restart the first full ETL pass has already completed before any hour is processed; `catch_up_today` cannot outrun ETL's first tick.

2. **Restoring `upstream_settled` naively reintroduces end-of-hour starvation.** Its check B requires `MAX(ended_at) >= hour_end` (or an `active_session` past the boundary). When the user stops working *at* the hour boundary, the last session ends before `hour_end`, B is never true, and the hour would **never** process. That's precisely why it was removed *together with* the 90-minute aging fallback — they were a matched pair. Re-adding the gate without also re-adding aging would trade a narrow incomplete-snapshot risk for a broader silent-drop.

The residual case (ETL falling behind mid-operation, not on restart, within the 3-minute window) is real but narrow: the hour ended 3 min earlier and ETL polls every 60s, and `await_coding_ready` already blocks on the coding side. Given the recall-is-cardinal-sin bar, if you want belt-and-suspenders I'd do it the starvation-safe way — gate on the boundary but **skip-without-`mark_hour_done`** (retry next HH:03) plus a max-age fallback so quiet-tail hours still drain — as a follow-up, rather than reinstating the old gate here. Happy to take that as a separate PR if you'd prefer it in-scope.

### Finding 4 — duplicated `_post` with diverging timeouts (LOW-MED) → **Fixed**

Confirmed. Removed pipeline's local `_post` and imported `generation._post` as the canonical transport (no cycle — `generation` imports only `classifier`/`models`). Preserved intent explicitly: `distill_hour` and `rerank` pass `timeout=_FAST_TIMEOUT` (120s), while **`/activity_report` now uses the canonical 300s default** — which fixes the latent divergence you flagged (a slow model no longer times out the report while the downstream worklog call wouldn't). Also dropped the now-unused `json`/`urllib.request` imports.

### Finding 5 — test files start with `#` header instead of docstring (LOW) → **Fixed**

Confirmed against the CLAUDE.md Python rule. Stripped the `# ambient dev tool…` line from all four `services/agents/tests/*.py` files so the `"""…"""` module docstring is line 1.

---

All 70 Python unit tests pass (`pytest agents/tests -q`), `py_compile` clean, pre-commit (fmt+clippy) green. Non-findings (BATCH re-export, `ValidationError` catch, patched call sites) — agreed, thanks for verifying those.

Akarsh-Hegde:
## Code Review — Bug Check Follow-up

Reviewed the latest commit against the 6 correctness candidates raised earlier. Summary below.

---

### ✅ Fixed

**Bug 1 — Confidence normalization (`cf > 1.0 → cf / 100.0`)**
The new `_normalize_confidence()` helper raises the percentage-detection threshold to `>= 2.0`. Values in `(1.0, 2.0)` are now clamped to `1.0` instead of being divided by 100 (which was silently dropping high-confidence matches by sending them to `0.011`). Clean fix with good inline documentation.

**Bug 4 — `upsert_proposed_task` returning stale row id when prior proposal was approved/dismissed**
The function now calls `SELECT changes()` after the UPSERT and returns `None` when the `WHERE state = 'proposed'` guard blocked the update. `stage_generate` receives `proposed_id = None`, correctly signalling the new proposal was not persisted — eliminating the id/content mismatch that made the pipeline believe it had saved a proposal it hadn't.

---

### ⚠️ Not Fixed (known follow-ups)

**Bug 2 — `tier_used` misreported as `1` when tier-2 actually ran (`classifier.py` line ~201)**
`tier = 2 if all_t2 else 1` still fires when t1 returns exactly 1 match and the full backlog was scanned via tier-2 with no additional results. OO traces and worklog logs continue to show `tier_used=1`. No data correctness impact — metrics only. Suggest a follow-up: `tier = 2 if backlog and len(t1) < 2 else 1` (or track whether tier-2 was invoked separately).

**Bug 3 — `_human_hour` midnight overflow (`worklog.py` line 35)**
`(h + 1) % 24` for h=23 still produces `"2026-06-29 23:00–00:00 (local)"` — the end time `00:00` belongs to the next calendar day but the date string stays on the prior day. Cosmetic/OO display only.

**Bug 5 — `(TypeError, ValueError)` won't catch Pydantic v2 `ValidationError` (`classifier.py` ~line 82, `generation.py` ~lines 60/96)**
Pydantic `>=2.10,<3` raises `pydantic.ValidationError` on model construction failure, which is not a subclass of `TypeError` or `ValueError`. The handler is dead code under normal FSM usage (the server guarantees structure), and the escalation behaviour — hour fails → Rust retries — is arguably safer than silently returning `[]`. Low practical risk, but the intent (graceful fallback) is not what executes. Fix: add `from pydantic import ValidationError` and include it in the `except` tuple.

**Bug 6 — `n_items` time pre-division before draft outcomes are known (`pipeline.py`, `stage_generate`)**
`time_per_item = 3600 // max(1, n_items)` is computed before drafts are generated. If a matched ticket's `generate_worklog` returns `None` (empty/unusable server response), its time slot is lost and the hour is marked done with no retry. Edge case requiring an empty `summary` from a structurally valid FSM response — rare in practice, and may be an accepted design trade-off. Worth noting for future time-accounting audits.

---

**Overall:** The two highest-severity bugs are fixed. The four remaining items are observability gaps and low-risk edge cases with no impact on worklog data correctness. Good to merge; suggest logging Bugs 2, 3, 5 as follow-up tickets.

adityaharishch:
Thanks for the follow-up pass, @Akarsh-Hegde. Went through the four remaining items — one fixed, one is a false positive (with proof), two are accepted trade-offs with reasoning.

### Bug 2 — `tier_used` misreported when tier-2 ran → **Fixed** (`9864c924`)

Agreed it's a real observability gap. `classify_hour` now tracks tier-2 invocation with an explicit `tier2_ran` flag, so `tier_used` reflects the **deepest tier invoked** (a tier-1 hit of 1 that triggers a full backlog scan now correctly shows `tier_used=2`). Per-match provenance still lives on `TaskBinding.tier`, so no signal is lost. Metrics-only and nothing branches on `tier_used` (verified: no test/consumer asserts it), so zero behavior risk.

### Bug 5 — `(TypeError, ValueError)` won't catch `ValidationError` → **False positive** (no change)

This one reverses your own (correct) first-round non-finding. Verified empirically against the pinned runtime:

```
$ python -c "from pydantic import ValidationError, VERSION; \
             print(VERSION, [c.__name__ for c in ValidationError.__mro__], \
             issubclass(ValidationError, ValueError))"
2.13.4 ['ValidationError', 'ValueError', 'Exception', 'BaseException', 'object'] True
```

Pydantic v2's `ValidationError` **is** a `ValueError` subclass (it was v1 where the hierarchy differed), so `except (TypeError, ValueError)` catches it correctly. No dead code, no change needed.

### Bug 3 — `_human_hour` midnight `23:00–00:00` → **Won't change** (cosmetic, correct-as-is)

Display-only, and `23:00–00:00 (local)` is an unambiguous, conventional rendering of the last hour of the day (00:00 = midnight boundary). Rolling the date forward on the end-time would add clutter without removing ambiguity. No correctness impact; leaving as-is.

### Bug 6 — `time_per_item` pre-division when a draft returns `None` → **Accepted trade-off** (no change)

Real but narrow (requires an empty `summary` from an otherwise-valid FSM response). Notably the current direction is the **safe** one: a lost slot means the hour logs **≤ 3600s**, i.e. it under-reports rather than over-claiming Jira worklog time — over-logging would be the worse failure for time accounting. A "correct" redistribution needs a two-pass restructure of `stage_generate` (draft all items → count successes → divide), which is more regression surface than this rare edge warrants pre-merge. Fine as a follow-up ticket if we want exact slot reclamation.

---

All 70 Python unit tests pass; `py_compile` clean; pre-commit (fmt+clippy) green. Bugs 1, 2, 4 fixed across `7d714b82` + `9864c924`; happy to open follow-up tickets for 3 and 6 if you'd like them tracked.

PR #383, opened 2026-07-02
fix(mlx): stop the MLX server's multi-GB/hour memory growth

adityaharishch:
## Summary
- MLX inference server was observed growing from ~100MB to 24-33GB of real OS memory over a several-hour process lifetime. Two independent, additive causes, both fixed:
- **`mx.clear_cache()` gap**: `generate_structured()`/`generate_thinking()` (the hottest generation paths — `/classify_tasks`, `/generate_worklog`, `/propose_ticket`, `/summarise`, `/activity_report`) never released MLX's buffer pool per-call, unlike `reranker.py`/`session_distiller.py` which already do. Fixed in `structured.py`, `thinking.py`, and the OpenAI-compat gateway's outlines branch in `routes/chat.py`.
- **The dominant cause**: OpenObserve was down, but `observability.py` was still shipping telemetry *directly* via `OTLPSpanExporter`/`OTLPLogExporter` (bypassing the Rust daemon's already-existing `telemetry_spool::shipper`), so every batch export failed and retried continuously (900+ failures in ~2hrs), correlating with sustained memory growth. Rather than a retry circuit-breaker band-aid, removed direct HTTP export from the Python service entirely — it now always writes to the durable local spool (`~/.meridian/telemetry/pending/`), and the Rust daemon's shipper (proper terminal-vs-retryable classification, quarantine) is the only thing that ever talks to OpenObserve, fully decoupled from the Python process's lifecycle.

## Test plan
- [x] `cargo test --workspace` (396+68 tests) clean
- [x] `cargo clippy --workspace -- -D warnings` clean
- [x] `python3 -m py_compile` clean on all touched files
- [x] Live verification: MLX server footprint 24GB → 111MB after restart with the fix
- [x] Zero export failures/retries logged since restart (vs. continuous every few seconds before)
- [x] Spool → shipper → OpenObserve pipeline confirmed live (fresh files landing in `pending/`, draining into `sent/` within seconds)

### DIFF
--- a/services/agents/mlx_classifier.py
+++ b/services/agents/mlx_classifier.py
@@ -45,6 +45,31 @@ _in_flight = 0
 _last_used = time.monotonic()
 
 
+def _invalidate_dependent_caches() -> None:
+    """Drop caches in other modules keyed off the (about-to-be-freed) model object.
+
+    ``agents.structured`` caches an outlines MLXLM wrapper + compiled FSM
+    Generator(s) per ``id(bundle.model)`` — without this call those caches
+    only get cleared lazily, the next time a structured-generation call
+    happens to see a new model id. That left the "evicted" model (and its
+    outlines_core Index/Guide/Vocabulary, which carry Python-side reference
+    cycles) fully resident for the whole gap: it both defeated this module's
+    single-slot residency guarantee (two generative models briefly alive at
+    once) and, since eviction fires here often — every reranker↔classifier
+    swap — let that stale generation pile up across many reload cycles
+    before Python's own GC schedule swept it, producing multi-GB/hour
+    malloc-zone growth invisible to ``mx.get_active_memory()`` (that call
+    only accounts Metal/unified GPU memory, not this CPU-side object graph).
+    Lazy import: ``agents.structured`` is not a dependency this module needs
+    at import time, only at eviction time.
+    """
+    try:
+        from agents import structured
+        structured.invalidate()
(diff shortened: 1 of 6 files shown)


adityaharishch:
**Follow-on fix pushed** (308d1bf5): the previous commit didn't fully stop the growth — root-caused a second leak and fixed it.

`agents/structured.py` caches an outlines FSM/model wrapper keyed by `id(bundle.model)`, but `mlx_classifier.py`'s eviction (`evict_resident_model`/`maybe_evict_idle`) never told it to drop that cache — so every evict+reload cycle (confirmed multiple times per minute in production logs, on every reranker↔classifier single-slot swap) left the "evicted" model + its compiled `outlines_core` Index/Guide/Vocabulary fully alive until the next generation call happened to see a new model id.

Measured via `vmmap --summary` on a live dev instance: 23.6GB of malloc-zone growth in 37 minutes — invisible to `mx.get_active_memory()`, which only tracks Metal/unified GPU memory, not this CPU-side object graph.

Adds `structured.invalidate()`, called from both eviction points, plus an explicit `gc.collect()`. Also fixes a secondary bug: the "single generative model resident" guarantee was silently violated during the eviction→reload gap. Bumped `services/pyproject.toml` 1.67.2 → 1.67.3 so this ships on the next runtime republish.

Killed the live leaking dev MLX server (24GB) and confirmed the tray's supervisor auto-respawned it clean with the fix.

PR #384, opened 2026-07-03
release: worklog FSM classification, ticket write-back expansion, one-pager Timeline dashboard, tray popover redesign

Akarsh-Hegde:
## Summary

Promotes 23 commits from `pre-main` to `main`. Four major threads:

**1. Worklog pipeline — FSM classification + proposed-ticket approval (#377, #378)**
- Switches `/classify_tasks`, `/generate_worklog`, `/propose_ticket` to outlines grammar-constrained decoding (thinking off), eliminating the ~7% silent JSON parse failures the old native-thinking + budget-cap path produced.
- Adds proposed-ticket approval end-to-end: `issue_type` Task/Bug (migration 052), Rust driver, Python proposer, UI approval chip.
- Fixes a transient-read-failure bug where a gate query error permanently marked an hour's worklog as done (should retry).
- `classify_tasks.py` (renamed from `match_tasks.py`) raises on transport failure instead of silently dropping the match.

**2. Ticket write-back coverage (#8387b8c1 + follow-on fixes)**
- Extends Jira/Linear/Azure DevOps/GitHub/Trello field coverage for the board-hygiene fix flow.
- `fix(trello)`: comment-delete was hitting a card-scoped endpoint instead of Trello's actual action-scoped one — the 404 was masked as "already deleted," so edited/re-matched Trello worklogs accumulated duplicate stale comments forever.
- `refactor`: split `jira.rs` (status-transition writers → `jira_transitions.rs`) and `daemon.rs` (pause/resume → `commands/pause.rs`) under the 500-line file cap.

**3. Dashboard → one-pager "Meridian Timeline" redesign**
- Replaces the multi-route sidebar dashboard (Today/Plan/Tasks/Worklogs/Cleanup/Settings) with a single-page hour-by-hour timeline + Overview/Hour-detail panel + modals, three persisted themes (lilac/blush/ink), pill navbar ported byte-for-byte from the design mock.
- Backend: per-hour distilled activity text persistence (migration 053 `hour_text`), hour-report reducer improvements (migration 054).
- Cleanup: deletes the pre-fold Sessions/Week/Logs routes and their exclusive deps (confirmed unreachable via reachability sweep from the only two windows the app opens), drops 5 zero-import npm deps, adds the missing `react-markdown` dependency it was silently relying on via a pre-populated lockfile.
- Bug fixes surfaced along the way: "Logged" time totals were including unposted proposed tickets; notification "Open →" 404'd instead of opening the target modal; Board Cleanup's forward-navigation had a skip-past-must-fix loophole.

**4. Tray popover redesign**
- Rebuilt from the Claude Design mock: capture status card, pause presets (15m/1h/tomorrow/indefinite), Today stats, review CTA.
- `pause_indefinitely` command, 24h pause cap (fixes a "pause until tomorrow" clamp bug — computing up to ~33h between midnight and 9am silently no-op'd against the old cap).
- Popover dismissal moved into the Rust command layer so every window-opening path (invoke commands, native tray menu, first-run auto-open) reliably dismisses it — fixes a stuck-popover race that previously only some paths handled.

**Also included:** a dual-slot worklog rematch experiment was tried and reverted after it caused real duplicate Jira posts (confirmed live: 4 duplicate entries against one ticket) — rematching onto a ticket with an existing hour worklog is a hard block again. Several PR #381 automated-review fixes, including a real SQL-injection finding in `scripts/check-worklog-trace.py` (unvalidated `--trace-id` interpolated into a raw OpenObserve query — now hex-allowlisted).

## Heads-up: version fields will *decrease* in this merge

`main` currently sits at app version `1.68.0` (`Cargo.toml`/`npm/meridian/package.json`) and runtime version `1.68.0` (`services/pyproject.toml`), from the release-please bump (`17f32794`) that ran after the last promotion (#371). `pre-main` has not been bumped since and is at app `1.67.0` / runtime `1.67.2`. Merging as-is will move those version fields backward on `main`. This matches the known pattern (`pre-main` trails `main` under the lockstep versioning scheme) — flagging so the release tooling / a human reconciles the version bump rather than it happening silently. The runtime side already has a `runtime-staging` publish at `1.67.2`; production (`runtime-latest`) is still on `1.67.0` and will need its own version-bumped publish after this merge for the FSM classification changes to actually reach users (per the runtime-publish playbook in `CLAUDE.md`).

## Test plan

These commits report their own verification in-message (`cargo build/clippy -D warnings/test`, `tsc --noEmit`, `next build`, `bun test` all clean at each merge point per the commit bodies). Before merging this PR:

- [ ] `cargo test` (workspace) passes on `main` post-merge
- [ ] `cd ui && npm run build && npm test` clean
- [ ] `cd tray/src-tauri && cargo test` clean
- [ ] Reconcile `Cargo.toml` / `npm/meridian/package.json` / `services/pyproject.toml` versions so they don't regress `main`'s current `1.68.0`
- [ ] Cut a `runtime-latest` publish after merge so the FSM classification fix actually reaches production users

## Summary by CodeRabbit

* **New Features**
  * Added hour-level timeline data: per-hour activity text, reports, and generating/paused badges, plus hour detail UI.
  * Added structured LLM endpoints for task classification, worklog generation, and ticket proposing with unified thinking-mode handling.
  * Introduced theme support across runtime + desktop tray experience (new settings/appearance UI).
  * Added pause/resume improvements and new tray worklog/time commands (including rematch).
* **Bug Fixes**
  * Improved edit/rematch reliability via stale “unpost” cleanup; expanded ticket reopening support.
* **Documentation**
  * Refreshed OAuth and MLX memory documentation, plus updated the PM worklog hour dashboard.


Akarsh-Hegde:
## Code review fixes applied

Went through every finding from CodeRabbit + github-code-quality on this PR. Commit 3672e160 fixes the real issues; a few were verified and intentionally left as-is (noted below with why). All fixes verified: `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test` (398+ workspace tests), `next build` + `bun test` (201 tests, ui/), `vitest` (33 tests, tray/).

### Fixed — real bugs

- **`src/pm_worklog/trello.rs`** (Major): the comment-delete URL (`/actions/{id}/comments`) isn't a documented Trello endpoint — verified directly against Atlassian's REST API docs, which only expose `DELETE /1/actions/{id}` (action-scoped) and `DELETE /1/cards/{id}/actions/{idAction}/comments` (card-scoped). The 404 from the wrong URL was being swallowed as success, so edited/re-matched Trello worklogs never actually got their stale comment removed — this reproduces the exact bug a prior commit (3617c8da) thought it had fixed, just with a different wrong URL. Switched to the documented `DELETE /1/actions/{id}`.
- **`tray/src-tauri/src/commands/pause.rs`** (Major ×2): `pause_for_duration`/`pause_indefinitely` read the prior pause state and wrote the new state as two separate lock acquisitions, so a concurrent pause/resume call (double-click, or a race with the schedule-pause poll tick) could interleave and corrupt `pause_started_at`/gap accounting feeding the "Logged" time metric. Consolidated into one `transition_pause` helper that does the read+write atomically under a single lock. Also replaced an `unwrap()` on the shared mutex in `resume_capture` — a poisoned lock would have panicked and left capture stuck paused; now logs and returns gracefully.
- **`scripts/run-classify-tier1.sh`** (Major): tier-2 fallthrough discarded a valid tier-1 match (reassigned `bindings` instead of merging) and mislabeled tier-2-sourced results as tier-1 in the printed output — defeats the purpose of this dev inspection tool.
- **`services/pyproject.toml` + `scripts/setup-services.sh`** (Major): `mlx-lm`'s floor was `>=0.31`, but `make_logits_processors(presence_penalty=...)` (used unconditionally in `thinking.py`) was only added in `mlx-lm` 0.31.1 — a resolve to 0.31.0 would `TypeError` on the first generation call. Bumped the floor to `>=0.31.1`. `uv.lock` regenerated (mlx-lm stays resolved at 0.31.3, so no behavior change today — this guards future installs/resolves).
- **`src/pm_worklog/post.rs`** (Minor): `clear_pending_unpost`'s `?` aborted the whole `unpost_stale` sweep on a DB error, skipping every remaining pending row for that pass — contradicted the function's own doc comment ("never a hard error that stalls the loop"). Now logs and continues; the tracker-side delete is idempotent so it just retries next pass.
- **`src/pm_worklog/github.rs`** (Minor): `delete_worklog`'s reqwest client had no timeout, so a stalled GitHub call could hang the unpost sweep indefinitely. Bound to 20s, matching the existing pattern in `azure_devops.rs`.
- **`src/worklog_pipeline.rs`** (Minor): `mark_hour_generating`/`mark_hour_done`/`mark_hour_pending` write failures were discarded via `let _ = …` with no logging — a transient DB error could silently desync the ledger from what the pipeline believes happened. Now logged via `tracing::warn!`, consistent with the existing `ensure_hour`/`hour_is_done` posture a few lines above.
- **`ui/components/timeline/CleanupOverlay.tsx`** (Major): the global `keydown` handler intercepted ArrowLeft/Right/Escape even while a form input had focus, stealing text-cursor movement and closing the whole modal (discarding in-progress edits) on Escape. Now ignores keys while an input/textarea/select has focus.
- **`ui/components/timeline/TasksPanel.tsx`** (Major): the hygiene-fix chip was nested inside the row's own `<button>` — invalid HTML that breaks keyboard/AT interaction with the inner control. Split into a non-button row container (`role="button"`) with a real `<button>` nested inside for the hygiene action.
- **`ui/components/timeline/TimelineColumn.tsx`** (Major): per-card wrappers in a multi-ticket hour had no keyboard affordance (only `onClick`), unlike the row itself. Added `role="button"`/`tabIndex`/`onKeyDown` to both the multi-card and single-card branches.
- **`tray/src/app.js`** (Minor ×2): the "tomorrow" hint showed the raw unclamped 9am time even on the early-morning window where the actual pause is clamped to 24h — now reuses `secsUntilTomorrowMorning()`. Also: the duration picker stayed visibly open for a moment after a selection instead of closing immediately, like the primary toggle button does.

### Fixed — quick/low-risk

- `services/agents/routes/classify.py` + `generate.py`: `log.error` → `log.exception` on 3 inference-error paths so the traceback isn't dropped.
- `services/agents/worklog_pipeline/pipeline.py`: hoisted a `set(plan_keys)` out of a per-item comprehension.
- `scripts/check-worklog-trace.py`: added the missing explicit `return []` so `_search` isn't mixed implicit/explicit-return.
- `scripts/eval-classify.py`: split two semicolon-joined statements (Ruff E702).
- `src/intelligence/ticket_update/jira_transitions.rs`: added the missing unit tests for `pick_cancel_transition` (the only one of the three transition-pickers with no dedicated coverage — it also has no structured `statusCategory` fallback, making it the riskiest).
- `tray/src/index.html`: added `type="button"` to all 12 popover buttons (avoids implicit-submit semantics).
- `tray/src/style.css`: fixed 3 Stylelint keyword-casing violations + renamed `@keyframes softPulse` → `soft-pulse` (verified no JS/HTML references before renaming).
- `ui/components/IntegrationConnect.tsx`: replaced raw `#d97706`/`#e53e3e` hex with `var(--status-warning-dot)`/`var(--status-error-dot)` — exact value matches already defined in `globals.css`, so this is a zero-visual-diff token swap, not the (slightly different-shade) `--color-state-pending` the bot's own diff suggested.
- `ui/components/timeline/settings/SettingsSidebar.tsx`: `hasAnyConnected` now derives from the shared `TRACKERS` constant instead of a hand-enumerated provider list, so a newly-added tracker doesn't silently fall through the "connected" check.
- `ui/components/timeline/settings/AccountSection.tsx` + `SettingsModal.tsx`: two previously fully-silent failure paths (setup-wizard launch, integrations load) now `console.error` instead of swallowing the rejection.
- `ui/components/timeline/ReviewCard.tsx`: caught an unhandled rejection in `handleSave` (if `onEditSave` throws instead of resolving `{ok:false}`), and disabled the swipe-to-decide drag gesture while `busy` (matches the FAB buttons' existing disabled-while-busy behavior).

### Verified and intentionally skipped

- **`ui/components/timeline/ReviewOverlay.tsx`** — flagged as "actions aren't awaited or caught," but the bot's own comment asked to first confirm whether `useTimelineData`'s `act`/`proposedAct`/`reject` already handle errors internally. Checked: `run()` (which all three are built on) wraps its call in try/catch and does `alert(errorMessage(e))` on failure — the returned promise never actually rejects. False positive, no change needed.
- **`ui/components/timeline/ReportModal.tsx`** — flagged for missing `role="dialog"`/focus-trap versus `ModalShell`. Checked `ModalShell.tsx`: it doesn't set `role="dialog"`/`aria-modal` either, only an Escape-key handler — so there's no actual parity gap to close here; both share the same baseline. A real fix would mean adding full dialog semantics to both, which is a broader a11y initiative than this PR's scope.
- **`services/agents/structured.py`** — pinning `outlines` to an exact version to guard the `has_chat_template` private-internals workaround. This is a dependency-management judgment call (not a bug), left for a deliberate follow-up.
- **`tray/src/style.css`** WCAG contrast findings — real, but fixing them means picking new color values across 3 themes (lilac/blush/ink), which needs visual verification/design sign-off rather than a mechanical patch.
- **`ui/app/globals.css`** duplicate `--gen-*` tokens across the lilac/blush blocks, **`ui/components/NotificationBanner.tsx`** duplicate `SEVERITY_STYLES` map, and **`ui/components/timeline/TasksPanel.tsx`**'s hardcoded `EPIC_PALETTE` hex — all nitpick-level (the last explicitly tagged "💤 Low value" by CodeRabbit itself), left alone per "do what's been asked, nothing more."

Akarsh-Hegde:
## Follow-up: addressed CodeRabbit's re-review

CodeRabbit re-reviewed after the last push and caught 3 real gaps plus one nitpick in code I'd just added — fixed in commit \`45cc6288\` (merged with an unrelated upstream duplicate-commit reconciliation in \`fd5585b7\`):

- **`src/pm_worklog/trello.rs`** (Major): I fixed the wrong-endpoint bug but missed that `delete_worklog`'s client had no timeout — the same gap I'd already closed in `github.rs`. Bound to 20s, same pattern.
- **`scripts/run-classify-tier1.sh`** (Major, partial fix from last round): my previous commit fixed the tier mislabeling but `bindings = tier2_bindings` was still a *replace*, not a *merge* — a tier-1 match was still silently dropped whenever tier-2 also matched. Now `bindings = tier1_bindings + tier2_bindings`.
- **`ui/components/timeline/SettingsModal.tsx`** (Minor, partial fix from last round): the console.error-only fix still left the integrations panel silently stuck on failure with no visible state or recovery. Added a real `integrationsError` state + an inline error message + a Retry button.
- **`ui/components/timeline/TimelineColumn.tsx`** (Trivial nitpick on my own new code): the keyboard-reachability fix from last round duplicated identical click/keydown handler logic across the multi-card and single-card branches. Extracted into one `cardHandlers` helper, used by both.

Verified again: `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test` (398 workspace tests — one `meridian-oauth` test flaked once under full-suite parallelism, passed in isolation and on a clean rerun, unrelated to any touched file) clean; `next build` + `bun test` (201 tests) clean; `vitest` (33 tests) clean.

PR #385, opened 2026-07-04
refactor(triage): remove the labels hygiene check and its write-back

Akarsh-Hegde:
## What

Removes the **labels** board-hygiene check end to end — the "No labels — add one to categorise it" nag and its "Add a label" fix no longer appear anywhere, and the label write-back machinery it drove is deleted as dead code.

## Why

The check added noise without value: labels are an optional, cosmetic board field, and for Linear the "fix" couldn't even be applied in-app (team-scoped label UUIDs), so it just bounced users out to the tracker.

## Scope (+11 / −169)

**Detection (daemon)** — `src/intelligence/task_triage/`
- `TriageReason::MissingLabels` variant, hint, fix, and the `hygiene_issues` check
- `FixControl::EditLabels`, `BoardUsage.labels` + its detection
- `TicketSignals.tags` and the `pm_tasks.tags` read in `store.rs` (only consumer was this check; the DB column is untouched)
- fixtures + test harness fields

**Hint/fix mapping** — `meridian-core/src/util/hygiene.rs` and its port `ui/lib/hygiene.ts` (`missing_labels` arms; tests moved to other optional codes). Old `missing_labels` codes still stored in `pm_task_curation.reasons_json` now map to no fix and are filtered out — they disappear from the UI immediately and get rewritten on the next triage run.

**UI controls** — `edit_labels` input removed from `HygieneDialog.tsx` / `CleanupCard.tsx`

**Write-back (now unreachable)** — `WriteField::AddLabel` across `ticket_update/`: jira (`edit_update` helper now dead, removed), github (`labels_url`), azure_devops (`merged_tags`/`merge_tags`), linear (the label redirect + its `issue_url` helper), trello (redirect arm). `WriteField::parse("labels", …)` now returns `None` → the generic redirect, same as `acceptance_criteria`.

## Testing

- `cargo clippy -p meridian -p meridian-core --all-targets -- -D warnings` clean
- `cargo test -p meridian -p meridian-core`: 410 + core tests pass
- `cargo check -p meridian-tray` compiles
- UI: `npm run build` (static export) + `bun test` 135 pass


Akarsh-Hegde (commented):
## Automated review — no correctness bugs found

Verified via a multi-agent pass (line-by-line diff scan, removed-behavior trace, cross-file symbol tracing, cleanup/simplification scan, CLAUDE.md conventions check), each running the actual test suite against the PR head (`77423882`) rather than just reading the diff:

- `cargo test --lib task_triage` — 25/25 pass (incl. `fixture_dataset_matches_expected_buckets`)
- `cargo test --lib ticket_update` — 48/48 pass across all 5 providers
- `cargo test -p meridian-core hygiene` — 6/6 pass
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo build --bins` + `cargo test --no-run` — full binary + integration targets compile clean

Every `match` over `WriteField`/`TriageReason` in the 5 provider files remains exhaustive after removing `AddLabel`/`MissingLabels`; `WriteField::parse("labels", …)` correctly falls through to the existing generic "unwritable field → redirect" path (same as `acceptance_criteria`), so stale `missing_labels` codes already persisted in `pm_task_curation.reasons_json` degrade safely — `parse_issues`/`parseIssues` drop them via the `fix.is_none()` filter rather than crashing or rendering garbage. The removal is symmetric and complete across Rust + TS.

Four minor, non-blocking items surfaced (cleanup/doc-hygiene, not defects):

1. **`src/intelligence/ticket_update/github.rs:5`** — module doc still says *"Issues natively support assignee, labels, title, body, and state — those we apply in-app,"* but the labels write-back path (`WriteField::AddLabel`, `labels_url()`) was deleted in this PR. Now overclaims what's applied in-app.
2. **`src/intelligence/ticket_update/azure_devops.rs:6`** — module doc still says *"tags (our 'labels') and parent need a read-modify-write / relation op,"* but the tags read-modify-write (`merged_tags`/`merge_tags`) was deleted here — only `parent` still needs that.
3. **`src/intelligence/ticket_update/trello.rs:5-6`** (and the test comment at `:191`) — still describes labels as board-scoped and redirected, but `WriteField::AddLabel` no longer exists as a variant at all — it's not "redirected," it's gone.
4. **`pm_tasks.tags` is now a write-only column.** `store.rs::load_board()` correctly stopped reading it (matches the PR description), but all 5 provider sync modules (`providers/{trello,linear,jira,github,azure_devops}/*`) still `INSERT … ON CONFLICT DO UPDATE SET tags = excluded.tags` on every sync. Harmless (nullable column, no constraint risk), but it's now dead work maintaining data nothing reads — worth a follow-up to drop the writes too, or repurpose the column now that CDM's `CanonicalTask.labels` exists.

Optional/low-priority nit: after this PR, `field_name`/`write_field_name` in all 5 provider files are byte-for-byte identical 10-arm matches over `WriteField` (they only differed by the now-removed `AddLabel` arm). `mod.rs` already has a precedent for this shape (`WriteField::label()`) — could hoist into a `WriteField::field_key()` and delete the 5 duplicate free functions + their duplicate `field_name_round_trips` tests. This duplication predates #385 and isn't something this PR needs to fix — flagging only as a possible separate follow-up.

Nothing above blocks merge; scope discipline and the file-header/500-line/migration rules all check out clean.

Akarsh-Hegde:
## Review items addressed — c990e314

**Fixed in this PR (items 1–3, stale docs):**
1. `github.rs` module doc no longer lists labels among the fields applied in-app.
2. `azure_devops.rs` module doc no longer mentions the deleted tags read-modify-write — only `parent` needs a relation op now.
3. `trello.rs` module doc and the redirect-test comment no longer describe labels as a redirected field.

**Deferred as follow-ups (as the review itself suggested):**
4. `pm_tasks.tags` write-only column — the provider syncs still write it. Dropping those writes (or repurposing the column via CDM's `CanonicalTask.labels`) touches all five sync modules and is out of scope for this removal PR.
5. `field_name` duplication across the five providers — predates this PR; candidate for a `WriteField::field_key()` hoist in a separate cleanup.

Full pre-push suite (fmt, clippy, UI build + tests, security audit, cargo test) ran clean on push.

PR #386, opened 2026-07-04
fix(ui): open external links via opener plugin in Tauri webview

Akarsh-Hegde:
## Problem

The dashboard runs inside a Tauri WKWebView, which has **no new-window handler**: every `<a target="_blank">` click and `window.open()` call is silently dropped — nothing opens, no error. Every external link in the app was dead.

This blocked the Trello connect flow outright: on dev builds (no baked-in `TRELLO_APP_KEY`) the UI shows the API-key prompt, but its **"Get it at trello.com/app-key"** link never opened, dead-ending the whole setup.

## Fix

- **`ui/lib/bridge.ts`** — new `openExternal(url)`: routes URLs through the opener plugin (`plugin:opener|open_url`, already permitted by `opener:default` in `capabilities/default.json` — no Rust/capability change needed), with a `window.open` fallback for a plain browser. Plus the pure predicate `isExternalHttp()`.
- **`ui/components/ExternalLinks.tsx`** (new) — a document-level click interceptor mounted in the root layout, so every external http(s) anchor in every window (dashboard + setup wizard) works without per-component wiring. No-op outside Tauri.
- **4 direct `window.open()` call sites** (HygieneDialog, OverviewPanel, CleanupCard, AdvancedSection) now use `openExternal` — these were equally broken in the webview.
- **Trello API-key prompt** now documents the *Allowed Origins* requirement (`http://127.0.0.1:9123`) — Trello refuses to redirect back after consent unless the origin is on the key's allow-list, which was the next failure in the flow.

## Testing

- New `ui/__tests__/external-links.test.ts` (9 tests): the URL predicate, the plugin routing, the layout mount, and a guard that no component reintroduces raw `window.open`.
- Full UI suite: 210 pass / 0 fail; `next build` passes.
- Verified live in the running `tauri dev` app: the trello.com/app-key link opens the default browser, and the full Trello browser-OAuth connect + task sync + due-date write-back now work end-to-end.

https://claude.ai/code/session_01GvJAgU1eEFC68AAG8etezG

## Summary by CodeRabbit

* **New Features**
  * Added a Tauri-aware external-link interceptor so external links open via the system handler (including consistent `mailto:`/`tel:` support).
  * Added Trello sign-in guidance for required local redirect origin.
* **Bug Fixes**
  * Updated multiple app flows to stop using direct `window.open(...)` and route external destinations through the unified external-link behavior.
* **Tests**
  * Added a regression test suite covering URL classification and external-opening behavior, including ensuring external opens don’t bypass the shared handler.

### DIFF
--- a/ui/__tests__/external-links.test.ts
+++ b/ui/__tests__/external-links.test.ts
@@ -1,7 +1,7 @@
 //ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
 import { describe, it, expect } from 'bun:test'
 import { readFileSync, readdirSync, statSync } from 'fs'
-import { isExternalHttp } from '../lib/bridge'
+import { isExternalHttp, opensExternally, openExternal } from '../lib/bridge'
 
 // Regression guard for "external links do nothing inside the app".
 //
@@ -50,6 +50,25 @@ describe('isExternalHttp', () => {
   })
 })
 
+// ── opensExternally — adds mailto:/tel: on top of external http(s) ───────────
+
+describe('opensExternally', () => {
+  const origin = 'http://localhost:3939'
+
+  it('accepts mailto: and tel: (WKWebView drops them like target=_blank)', () => {
+    expect(opensExternally('mailto:hey@meridiona.com?subject=Bug', origin)).toBe(true)
+    expect(opensExternally('tel:+1555000', origin)).toBe(true)
+    expect(opensExternally('MAILTO:hey@meridiona.com', origin)).toBe(true)
+  })
+
+  it('delegates http(s) decisions to isExternalHttp', () => {
+    expect(opensExternally('https://trello.com/app-key', origin)).toBe(true)
+    expect(opensExternally('http://localhost:3939/today', origin)).toBe(false)
+    expect(opensExternally('/today', origin)).toBe(false)
+    expect(opensExternally('x-apple.systempreferences:com.apple.x', origin)).toBe(false)
+  })
+})
+
 // ── openExternal routes through the opener plugin ─────────────────────────────
 
 describe('lib/bridge.ts openExternal', () => {
@@ -62,6 +81,28 @@ describe('lib/bridge.ts openExternal', () => {
   it('falls back to window.open in a plain browser', () => {
     expect(src).toMatch(/openExternal[\s\S]*?window\.open\(url/)
   })
+
+  it('falls back to window.open when the plugin invoke rejects', async () => {
+    // A rejection that only logs would reproduce the silent dead-link bug this
+    // helper fixes, so the .catch() must open the URL anyway. bridge reads
+    // `window` at call time, so a stub installed here is picked up.
+    const opened: string[] = []
+    const g = globalThis as { window?: unknown }
+    g.window = {
+      __TAURI__: { core: { invoke: () => Promise.reject(new Error('forbidden url')) } },
+      open: (url: string) => { opened.push(url); return null },
+    }
+    const warn = console.warn
+    console.warn = () => {}
+    try {
+      openExternal('https://example.com/x')
+      await new Promise((r) => setTimeout(r, 0))
+      expect(opened).toEqual(['https://example.com/x'])
+    } finally {
+      console.warn = warn
+      delete g.window
+    }
+  })
 })
 
 // ── The interceptor exists and is mounted globally ────────────────────────────
@@ -71,7 +112,7 @@ describe('ExternalLinks interceptor', () => {
 
   it('listens for document clicks and routes external anchors via openExternal', () => {
     expect(src).toContain("document.addEventListener('click'")
-    expect(src).toContain('isExternalHttp(href, window.location.origin)')
+    expect(src).toContain('opensExternally(href, window.location.origin)')
     expect(src).toContain('e.preventDefault()')
     expect(src).toContain('openExternal(href)')
   })
@@ -99,7 +140,8 @@ describe('no raw window.open in components', () => {
   }
 
   it('every external open goes through openExternal (WKWebView drops window.open)', () => {
-    const offenders = walk(uiRoot + '/components')
+    const offenders = ['/components', '/app']
+      .flatMap((d) => walk(uiRoot + d))
       .filter((p) => readFileSync(p, 'utf8').includes('window.open('))
       .map((p) => p.slice(uiRoot.length + 1))
     expect(offenders).toEqual([])
--- a/ui/__tests__/external-links.test.ts
+++ b/ui/__tests__/external-links.test.ts
@@ -1,0 +1,107 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+import { describe, it, expect } from 'bun:test'
+import { readFileSync, readdirSync, statSync } from 'fs'
+import { isExternalHttp } from '../lib/bridge'
+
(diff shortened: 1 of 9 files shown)


Akarsh-Hegde (commented):
Reviewed the diff, verified the opener-plugin capability scope, ran the new test file (`bun test` → 9/9 pass) and a full `next build` on the branch (clean). Good fix overall — this is a real, previously-invisible dead-end (WKWebView drops `target="_blank"`/`window.open` silently) and the interceptor approach is the right shape: one listener instead of N call-site rewrites.

**On CodeRabbit's "Major" finding (silent failure on invoke rejection) — confirmed, agree this should be fixed.** I traced it into `tauri-plugin-opener` v2.5.4's `open_url` command: it's scope-gated (`scope.is_url_allowed`) and returns `Err(ForbiddenUrl)` on a miss. I checked `opener:default`'s bundled `allow-default-urls` permission — it grants `http://*` and `https://*` globs, so in practice every URL this PR opens (trello.com, task provider URLs, `localhost:5080` for OpenObserve) is already covered and won't hit that specific rejection path. But the general point stands: *any* invoke rejection (plugin error, `with` app not found, future scope tightening) currently only `console.warn`s with zero visible fallback — reproducing the exact "nothing happens" symptom this PR sets out to kill. Worth taking CodeRabbit's suggested `window.open` fallback in the `.catch()`.

**New finding — `mailto:` links are still dead.** `isExternalHttp()` only matches `^https?://` by design (and the test explicitly asserts `mailto:a@b.com` → `false`), so the `ExternalLinks` interceptor never touches them. But `ReportModal.tsx`'s two "Report a bug" / "Feature suggestion" links (`mailto:hey@meridiona.com?subject=...`, lines 130/134) are plain anchors with no other click handling — under the same "no new-window handler" WKWebView limitation this PR's own description calls out, these are very likely silently dead in the packaged app too, and they're a real user-facing contact path. Notably the opener plugin's own default scope (`allow-default-urls`) already permits `mailto:*` and `tel:*` alongside http/https — the infra to fix this exists, it's just not wired into the interceptor/`openExternal`. Suggest either broadening the interceptor to also catch `mailto:`/`tel:` hrefs, or at minimum confirming live that these two links currently work in a packaged build before merging (if they already work via some other webview default I'm not accounting for, disregard).

**Nitpicks — agree with CodeRabbit's two test-coverage comments** (source-string-only assertions on the fallback path; regression guard scans `ui/components` but not `ui/app`). Both minor and non-blocking.

**Verified as non-issues:**
- The document-level click interceptor and Next `<Link>`/router navigation don't fight — internal client-side nav calls `preventDefault()` on bubble before this listener runs (`e.defaultPrevented` guard), and `isExternalHttp` already returns `false` for relative/same-origin hrefs.
- No `stopPropagation()` on `click` (only `onPointerDown` in `plan/parts.tsx`'s `OpenLink`) anywhere between the anchors this PR touches and `document`, so bubbling to the new listener isn't blocked.
- Grepped all of `ui/components` post-patch: no raw `window.open(` call sites remain outside `lib/bridge.ts`'s own fallback.
- `capabilities/default.json` already lists `opener:default` on `["main", "dashboard", "setup", "tray-tooltip"]` — no Rust/capability change was needed, matching the PR description.

Nothing here should block merge on its own except worth a quick look at the mailto: gap — the invoke-rejection fallback is a nice-to-have hardening. Nice regression-test coverage on the "no bare window.open" guard, that's a good trap for future call sites.

Akarsh-Hegde:
## Review fixes applied — e2a1c87f

**1. Silent failure on plugin invoke rejection (CodeRabbit 🟠 Major, confirmed by @Akarsh-Hegde) — fixed.** `openExternal`'s `.catch()` now falls back to `window.open(url, '_blank', 'noopener,noreferrer')` after the `console.warn`, instead of only logging. Covered by a new behavioral test that stubs `window.__TAURI__.core.invoke` to reject and asserts `window.open` receives the URL.

**2. `mailto:` links dead (review finding) — fixed.** New `opensExternally(href, origin)` predicate = `mailto:`/`tel:` OR external http(s); the `ExternalLinks` interceptor now uses it, so ReportModal's two contact links (and any future `mailto:`/`tel:` anchor) route through the opener plugin. As noted, `opener:default`'s `allow-default-urls` already permits both schemes — no capability change. `isExternalHttp` keeps its original strict semantics and tests.

**3. Regression guard only scans `components/` (CodeRabbit nitpick) — fixed.** The no-raw-`window.open` walk now covers `ui/app` too.

**4. Missing coverage for the invoke-failure fallback (CodeRabbit nitpick) — fixed** via the behavioral test in (1).

**Skipped: switch to `window.__TAURI__.opener.openUrl` (CodeRabbit nitpick).** Keeping the raw `invoke('plugin:opener|open_url')`: the opener JS global is only injected when the plugin's `withGlobalTauri` script is bundled, which our typed bridge doesn't declare or guarantee in the static export — `core.invoke` is the one surface `bridge.ts` already commits to, and the command name/payload is the plugin's stable IPC contract (same string its own `openUrl()` sends). The new fallback in (1) also means a future contract break would degrade gracefully instead of silently.

Validation: `bun test` 213 pass / 0 fail (12 in the external-links file), `next build` clean, full pre-push suite green.

https://claude.ai/code/session_01GvJAgU1eEFC68AAG8etezG

Akarsh-Hegde:
Follow-up commit `e2a1c87f` addresses everything from the review directly:

- ✅ **Invoke-rejection fallback** (CodeRabbit Major) — `.catch()` now calls `window.open` instead of only `console.warn`.
- ✅ **`mailto:`/`tel:` gap** (my finding) — new `opensExternally()` predicate wired into `ExternalLinks.tsx`; `ReportModal`'s contact links are covered.
- ✅ **Test breadth** — regression guard now scans `ui/app` too, and there's a real behavioral test for the rejection path (stubs `window.__TAURI__.core.invoke` to reject and asserts `window.open` runs).

Ran the updated suite on `e2a1c87f`: 12/12 pass (was 9), `next build` clean. CodeRabbit re-reviewed and approved.

**One thing worth a second look before merging, though** — I don't think it blocks, but flagging since it bears on whether the Major finding is *actually* fixed vs. just quieter:

The new fallback is `window.open(url, ...)` inside the `.catch()`. That only executes when `t` (the Tauri bridge) is truthy — i.e. exactly when running inside the packaged app. But this PR's own description establishes, from live testing, that `window.open()` is silently dropped by this app's WKWebView (no new-window handler) — that's the entire premise of the fix. So if the opener-plugin invoke ever does reject in production, the fallback calls the one API already known not to work in that environment, and the user still sees nothing happen — just via a different no-op than before. It's harmless (doesn't throw, doesn't regress anything), and the new test does pass, but that's because the test stubs `window.open` in bun/jsdom rather than exercising the real WKWebView, so it can't actually catch this gap.

In practice this is low-stakes: I confirmed `opener:default`'s bundled scope (`http(s)://*`, `mailto:*`, `tel:*`) already covers every URL this app opens, so the rejection path this fallback guards against shouldn't fire under normal use. If you want to close it properly, the options are a second Tauri command that shells out directly (bypassing the plugin) or a visible fallback (toast with the URL to copy) rather than a silent retry — but given the low likelihood of hitting it at all, I'd call this optional polish rather than a merge blocker.

**Bottom line: I think this is ready to merge.** All raised issues are addressed at the level CodeRabbit and I asked for, tests are green, build is clean, and the one remaining nuance is a corner case with no realistic trigger today.

PR #392, opened 2026-07-06
feat(notifications): interactive toast rails — plugin swap, response leg, snooze consumer

Akarsh-Hegde:
## Summary

PR 1 of the interactive-notifications plan (Obsidian: `Decisions/Interactive notifications - engine, quick replies, and the nudge loop.md`). Turns the notification outbox into a **round-trip mailbox** — enqueue → deliver → respond → consume, every leg idempotent — and swaps the tray's notification engine so macOS toasts get **action buttons + inline text reply**.

PR 2 (`feat/task-switch-verify`: flagship low-confidence-link nudge, `notify_task_verify` setting, hourly rate cap) follows on these rails.

## What changed

**Migration 057** — extends `notifications` in-row (no new table): `category` (UNNotificationCategory id; NULL = plain toast, exactly today's behaviour), `actions` (JSON descriptors, informational in v1), `responded_at` / `response_action` / `response_text` / `response_consumed_at`, + a response-drain index.

**Daemon**
- `NewNotification` gains `.category()` / `.actions()` builders; `enqueue()` writes both columns.
- Existing producers retrofitted: `plan.nudge` → [Open Plan] [Snooze 1h], `worklog.ready` → [Open Worklogs] [Snooze 1h], `system.fault` → [View].
- New `src/notification_responses.rs` consumer on the poll tick: `snooze` re-enqueues the event +1h with a dedup key derived from the original row's `responded_at`, so a crash between enqueue and consume-ack retries into a no-op. Everything else just stamps consumed (navigation already happened tray-side).

**meridian-core**
- `notifications::categories` — the category ids + action-set JSON, one source of truth for BOTH the tray's OS registration and the daemon producers (they can't drift).
- `record_response` (first answer wins via `IS NULL` guard), `unconsumed_responses`, `mark_response_consumed`.
- `pending_native` now carries `category`/`actions` **with a pre-057 schema fallback** — a tray running against an older daemon's DB keeps delivering plain toasts instead of erroring the queue empty (the overlapping-installs scenario).

**Tray**
- `tauri-plugin-notification` (Actions API mobile-only on desktop) → community **`tauri-plugin-notifications` pinned `=0.4.6`** (real Swift `UNUserNotificationCenter`: category registration, buttons, inline reply, `actionPerformed` events). Consider forking under the Meridiona org for supply-chain control.
- The plugin's init **hard-fails outside a `.app` bundle**, so `lib.rs` registers it only when bundled — an unbundled run (`tauri dev` / `cargo run`) logs and drops toasts instead of crashing. Dev builds therefore have **no toasts at all** now (the old plugin delivered plain ones unbundled) — accepted, same class of caveat as the popover-404 dev limitation.
- `sys.rs` stays the facade seam (swap point for in-house `objc2-user-notifications` if the plugin disappoints): `notify()` (plain) + `notify_interactive()` (embeds outbox id + deep link in `extra`) + startup category registration.
- Popover JS registers the plugin's `actionPerformed` listener → `record_notification_response` command → core `record_response`; tap/open/view with a deep link also opens the dashboard.
- `capabilities/default.json`: `notification:default` → `notifications:default`.

## Deviations from the plan doc

1. Conditional plugin registration (above) — the plan anticipated degraded dev toasts; the reality is a hard init failure, so the gate is ours.
2. `ActionType`/`Action` have private fields but derive `Deserialize` — categories are built from the shared JSON descriptors rather than Rust constructors.
3. Rate cap + `notify_task_verify` deferred to PR 2 per the plan's phasing.

## Verification

- `cargo fmt` / `clippy -D warnings` (daemon, core, tray) clean; **410 workspace + 118 core + 42 tray tests pass**, including new tests: response idempotency (first answer wins), unconsumed→consumed drain, pre-057 fallback delivery, category-JSON validity.
- Full pre-push suite (UI build + tests, security audit) passed.
- **End-to-end interactive toast requires a packaged build** (`UNUserNotificationCenter` needs a bundle). Recipe from the plan:
  ```bash
  sqlite3 ~/.meridian/meridian.db "INSERT INTO notifications (dedup_key,event_key,title,body,channels,category,actions) VALUES ('test:1','plan.nudge','Test','Pick one','native','verify_switch','[{\"id\":\"yes\",\"title\":\"Yes\"},{\"id\":\"no\",\"title\":\"No\"},{\"id\":\"reply\",\"title\":\"Reply…\",\"input\":true}]')"
  ```
  → toast with buttons within one poll tick; each press must land `responded_at`/`response_action`/`response_text`, then `response_consumed_at` on the next daemon tick. Plain (NULL-category) toasts and quiet-hours short-circuit must be unchanged.


Akarsh-Hegde:
## E2E verified on a packaged build (macOS 26.5)

Full mailbox round-trip confirmed live: seeded outbox row → interactive toast with buttons + inline reply (`verify_switch` category) → button press recorded (`response_action`/`responded_at`) → daemon consume stamp (`response_consumed_at`) 42s later on the next poll tick. Click-through (tap) opens the dashboard.

Three plugin gotchas found during verification, all fixed on this branch (each cost a debug round-trip; documented here for the future fork-under-Meridiona decision):

1. **`default-features = false` is load-bearing** (`ce38d218`) — the default `notify-rust` feature silently replaces the Swift `UNUserNotificationCenter` layer with plain toasts: no categories, no buttons, no action events.
2. **Swift runtime rpath** (`7f101793`) — the plugin links its Swift static lib without an rpath; the packaged app dies at launch (`dyld: Library not loaded: @rpath/libswift_Concurrency.dylib`). Fixed with `-Wl,-rpath,/usr/lib/swift` in the tray's build.rs.
3. **Extras never round-trip** (`ba03a8b7`) — the Swift `ActiveNotification` struct has no `extra` field, so payload attached at show-time is unrecoverable on `actionPerformed`. The notification identifier (set to the outbox row id) is the only surviving correlation; deep_link is resolved from the DB row instead.

Follow-up commit also adds `NSUserNotificationAlertStyle=alert` (persistent Alerts default for new installs — nudges are questions, banners evaporate in ~5s) and `customDismissAction` on all categories (the toast's ✕ now lands as `response_action='dismiss'`).

src/notification_responses.rs
```diff
@@ -0,0 +1,93 @@
...
+/// original with the response stamp, so (a) it never collides with the original
+/// row's key and (b) re-processing the same response is a no-op.
+#[tracing::instrument(skip(pool, r), fields(id = r.id, event_key = %r.event_key))]
+async fn snooze(pool: &SqlitePool, r: &NotificationResponse) -> Result<()> {
+    let scheduled_for = (chrono::Utc::now() + chrono::Duration::seconds(SNOOZE_SECS))
+        .to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
+    let dedup = format!("{}:snooze:{}", r.dedup_key, r.responded_at);
+    notifications::enqueue(
+        pool,
+        NewNotification {
```
Akarsh-Hegde, line 76: **[Correctness — CONFIRMED] Snoozing `plan.nudge` is a silent no-op.**

`snooze()` re-enqueues with `dedup_key = "{orig_dedup}:snooze:{responded_at}"` and `expires_at: None` (line 84-85). But `daily_plan.rs::maybe_nudge()` runs every poll tick and, whenever it runs (08:00–18:00 local), executes:

```sql
UPDATE notifications SET expires_at = ?
WHERE event_key = 'plan.nudge' AND dedup_key <> ?   -- today's canonical dedup
  AND (expires_at IS NULL OR expires_at > ?)
```

This "expire stale nudges from a previous day" sweep matches *any* `plan.nudge` row whose `dedup_key` isn't today's canonical key — including the snoozed row, since its dedup key is `plan.nudge:<today>:snooze:<ts>`, not `plan.nudge:<today>`. The sweep stamps `expires_at = now` (a timestamp in the past relative to the snooze's `scheduled_for = now+1h`).

When the snooze's `scheduled_for` elapses, `pending_native`'s filter (`expires_at IS NULL OR expires_at > now`) excludes the row because `expires_at` is already in the past. The "Snooze 1h" button silently never re-fires the nudge — the user presses it, gets no error, and the reminder just never comes back.

Repro: press Snooze on a `plan.nudge` toast any time before ~17:00 local (so `maybe_nudge` runs again within the hour) — the row gets killed before its `scheduled_for` is reached.

Suggest either: exempt rows with a `:snooze:` suffix from the stale sweep, or give the sweep a positive check against `event_key`+`dedup_key` prefix equality with today's date rather than inequality with the exact canonical key.

meridian-core/src/notifications.rs
```diff
@@ -117,12 +180,29 @@ struct NotifRow {
...
 fn has_channel(channels: &str, channel: &str) -> bool {
     channels.split(',').any(|c| c.trim() == channel)
 }
 
+/// Whether the interactive-notification columns (migration 057) exist. The tray
+/// can run against a DB whose daemon predates 057 (overlapping installs), so
+/// every read/write touching the new columns degrades gracefully instead of
+/// erroring the whole queue. Errors count as "missing" — same fail-soft posture
+/// as the pending read.
+async fn has_action_columns(pool: &SqlitePool) -> bool {
```
Akarsh-Hegde: **[Efficiency — CONFIRMED] `has_action_columns()` re-runs the schema probe on every call, forever.**

This is called independently (uncached) from `pending_native`, `record_response`, `unconsumed_responses`, and `expired_unanswered` — every one of them issues its own `SELECT COUNT(*) FROM pragma_table_info(...)` round trip on every invocation. The comment frames this as a transitional "overlapping installs" compat shim, but there's no cache/flag to make it a one-time check — it runs on every tray poll tick (~30s) and every daemon tick indefinitely, for a fact that's permanently true once migration 057 has run.

Suggest caching the result (e.g. `OnceCell<bool>` on first successful probe, or a static flag flipped once `true` is observed) so steady-state calls skip the pragma query entirely.

tray/src-tauri/src/sys.rs
```diff
@@ -24,13 +37,172 @@ pub fn uid_str() -> String {
...
 ///
-/// v1 shows title + body only. Producers populate a `deep_link` (e.g. /plan,
-/// /worklogs) and the in-app banner channel renders it as an "Open →" link, but
-/// click-to-navigate on a native toast needs Tauri notification actions + a
-/// focus/navigate handler — deferred. The two channels are intentionally
-/// asymmetric here; the banner carries the link.
+/// Fire-and-forget: the plugin's `show()` is async, so delivery is spawned and
+/// a failure is logged, never surfaced — no caller has a meaningful recovery.
+/// Unbundled runs (no plugin) log at debug and drop the toast.
 pub fn notify(app: &tauri::AppHandle, title: &str, body: &str) {
```
Akarsh-Hegde, line 73: **[Correctness — PLAUSIBLE] Plain `notify()` never sets `.id(...)`, but the popover's action listener trusts `n.id` for outbox correlation.**

`notify_outbox()` always calls `.id(toast_id)` and the doc comment states "the notification identifier IS the outbox row id, for EVERY outbox toast" — implying any other id means "not an outbox row." But plain `notify()` (used for pause/resume/update/health toasts) never calls `.id(...)` at all, so whatever default id the `tauri-plugin-notifications` plugin assigns is unknown here.

`tray/src/app.js`'s `actionPerformed` handler guards with `Number.isFinite(outboxId) && outboxId > 0` before calling `record_notification_response`. If the plugin's default (unset) id resolves to `0`/`undefined`/`NaN`, this guard is safe. But if the plugin auto-assigns a small sequential positive integer when `.id()` is omitted, a tap on a plain toast (e.g. the pause/resume toast) could pass the guard and get recorded against whatever `notifications` row happens to share that id — misattributing a response to an unrelated interactive notification.

Worth confirming what `tauri-plugin-notifications` 0.4.6 defaults an unset `id` to (can't verify offline — the crate isn't vendored in this checkout). If it's not guaranteed to be `0`/absent, plain `notify()` should explicitly set an out-of-band sentinel id (e.g. `0`) to make the correlation guarantee explicit rather than implicit.

tray/src-tauri/src/lib.rs
```diff
@@ -100,15 +110,26 @@ pub fn run() {
...
+            // delivery is a silent no-op until the app has prompted at least
+            // once), then register the interactive category set — the fixed
+            // UNNotificationCategory ids that give toasts their buttons
+            // (meridian-core's `notifications::categories`, one source shared
+            // with the daemon producers). Bundled runs only: the plugin isn't
+            // registered otherwise (see the builder above).
+            if sys::is_bundled() {
+                use tauri_plugin_notifications::{NotificationsExt, PermissionState};
+                let handle = app.handle().clone();
+                tauri::async_runtime::spawn(async move {
```
Akarsh-Hegde, line 132: **[Correctness — PLAUSIBLE] Category/permission registration races the poll loop startup.**

`request_permission()` + `register_notification_categories()` are fired via `tauri::async_runtime::spawn` here and never awaited before `setup()` returns. Separately (line ~390), the poll loop is spawned in its own task. Both are independently-scheduled tokio tasks with no ordering guarantee between them.

If the poll loop's first tick delivers an interactive toast (e.g. a `plan.nudge` already pending in the DB at startup) before the category-registration task has actually run `register_action_types`, that toast is delivered with an `action_type_id` macOS hasn't registered yet — it silently renders as a plain toast with no buttons for that one delivery (categories registered afterward only affect subsequently-created toasts).

Since this only affects the very first tick after launch and degrades gracefully (no crash, just missing buttons once), it may be an acceptable v1 gap — but worth an explicit `.await` or a readiness flag gating the first `notify_outbox` call if that first-toast case matters (e.g. a nudge already due at daemon start).

tray/src-tauri/src/sys.rs
```diff
@@ -24,13 +37,172 @@ pub fn uid_str() -> String {
...
 }
 
-/// Show a native macOS toast (title + body).
+/// True when running from a packaged `.app` bundle (`…/Foo.app/Contents/MacOS/bin`).
+///
+/// `UNUserNotificationCenter` (and therefore the notifications plugin, whose
+/// init hard-fails unbundled) only works from a bundle. `lib.rs` gates plugin
+/// registration on this; the notify facades gate their delivery attempt on the
+/// plugin state being present. Mirrors the plugin's own `require_bundle` check.
+pub fn is_bundled() -> bool {
```
Akarsh-Hegde, line 46: **[Reuse] `is_bundled()` re-implements bundle detection that already exists twice elsewhere.**

`update.rs::is_packaged()` already answers "am I running from a packaged `.app`" with a one-line substring check (`path.to_string_lossy().contains(".app/Contents/MacOS")`), and `install.rs` has a separate install-mode resolver for the same underlying question. This PR adds a third, differently-implemented detector (`parent()`/`parent()`/`parent()` path-component walk).

The three checks can disagree in edge cases (nested bundle, symlinked wrapper, a bundled app with no `.env` written yet). A future packaging change only has to break one of the three to make the notification-plugin registration gate (which crashes if wrong: the plugin hard-fails outside a real bundle) disagree with the updater's gate.

Suggest extracting a single shared `is_bundled()`/`is_packaged()` helper (e.g. in a small shared module) that both `update.rs` and `sys.rs` call, rather than maintaining parallel implementations of the same fact.

meridian-core/src/notifications.rs
```diff
@@ -278,6 +366,151 @@ pub async fn dismiss_banner(pool: &SqlitePool, id: i64, now: &str) -> anyhow::Re
...
+// ── Response leg (interactive notifications, migration 057) ─────────────────
+
+/// Record the user's answer to an interactive notification: stamp
+/// `responded_at` + the pressed `action` (an action id, `'tap'`, or
+/// `'dismiss'`) + any inline-reply `text`. First answer wins — the
+/// `AND responded_at IS NULL` guard makes a duplicate (or late) answer a no-op,
+/// mirroring [`mark_native_delivered`]. `now` is the caller-resolved stamp.
+/// No-op on a pre-057 DB (the columns don't exist — nothing to record).
+#[tracing::instrument(skip(pool, text))]
+pub async fn record_response(
```
Akarsh-Hegde: **[Conventions] File exceeds the repo's 500-line cap.**

Repo CLAUDE.md → Hard Rules: "Keep files under 500 lines; split when a file grows beyond that." This PR's additions (response-leg functions: `record_response`, `NotificationResponse`, `unconsumed_responses`, `expired_unanswered`, `notification_deep_link`, `mark_response_consumed`, plus tests) push `meridian-core/src/notifications.rs` from 465 lines (main) to 888 lines — nearly double the cap.

Suggest splitting the response-leg (everything from `record_response` down) into a sibling module, e.g. `meridian-core/src/notifications/responses.rs`, re-exported the same way other split modules in this crate are.

meridian-core/src/notifications.rs
```diff
@@ -278,6 +366,151 @@ pub async fn dismiss_banner(pool: &SqlitePool, id: i64, now: &str) -> anyhow::Re
...
+         SET responded_at = ?, response_action = ?, response_text = ?
+         WHERE id = ? AND responded_at IS NULL",
+    )
+    .bind(now)
+    .bind(action)
+    .bind(text)
+    .bind(id)
+    .execute(pool)
+    .instrument(tracing::debug_span!("notifications.write.response"))
+    .await?;
```
Akarsh-Hegde: **[Conventions] Missing `.context(...)` on DB-call `?`s.**

Repo CLAUDE.md → Coding Conventions → Rust: "Error handling: `anyhow::Result` throughout; add `.context("…")` to every `?` in DB calls." Both `record_response`'s `.execute(pool)....await?;` (this line) and `mark_response_consumed`'s equivalent (line ~510) propagate the sqlx error bare, unlike sibling writes in this same file (e.g. `enqueue` uses `.context("enqueueing notification")?`).

Suggest adding `.context("recording notification response")` / `.context("marking notification response consumed")` respectively.

src/daily_plan.rs
```diff
@@ -94,7 +96,9 @@ pub async fn maybe_nudge(pool: &SqlitePool) -> Result<()> {
         "Plan your day",
         "Pick what you're working on today so Meridian can match your work to the right tickets.",
     )
-    .link("/plan");
+    .link("/plan")
+    .category(categories::PLAN_NUDGE)
+    .actions(categories::actions_json(categories::PLAN_NUDGE).unwrap_or("[]"));
```
Akarsh-Hegde: **[Simplification] Category+actions wiring is copy-pasted verbatim across all three retrofitted producers.**

The pattern `.category(categories::X).actions(categories::actions_json(categories::X).unwrap_or("[]"))` (or the struct-literal equivalent) is repeated here, in `src/pm_worklog/scheduler.rs` (~line 265), and in `src/notices.rs` (~line 68-71) — with only the category constant varying, and `notices.rs` using a subtly different convention (no `.unwrap_or` fallback).

Since a fourth interactive producer (`verify_switch`, called out in the PR description as "PR 2") is already planned, this triplication is about to become a quadruplication. Suggest a single `NewNotification::interactive(category)` builder method (or free function) that sets both `category` and looks up `actions_json` internally, so category/actions can't drift apart and every producer becomes one call.

adityaharishch:
## Code Review — PR #392 (interactive toast rails)

Read the full diff (`gh pr diff 392`) plus the unchanged surrounding code the new logic interacts with (`daily_plan.rs`, `pm_worklog/scheduler.rs`, `commands/system.rs`). Solid engineering overall — idempotency guards, fail-soft pre-057 fallback, and good tracing coverage on the core response-leg functions. One real correctness bug found in the interaction between the new snooze consumer and existing code, plus a couple of smaller gaps.

### 🔴 Critical: `plan.nudge` snooze is silently defeated by the existing stale-nudge sweep

- `src/notification_responses.rs::snooze()` re-enqueues the answered event with `dedup_key = "{original_dedup}:snooze:{responded_at}"`, `event_key` unchanged (`plan.nudge`), `scheduled_for` = +1h, `expires_at = None`.
- `src/daily_plan.rs::maybe_nudge()` (unchanged by this PR, called every poll tick *before* `notification_responses::consume_responses` in `src/main.rs`) runs on **every tick**:
  ```rust
  UPDATE notifications SET expires_at = ?
  WHERE event_key = 'plan.nudge' AND dedup_key <> ?   // ? = "plan.nudge:{today}"
    AND (expires_at IS NULL OR expires_at > ?)
  ```
  This was written to expire yesterday's unactioned nudge, on the assumption there's only ever one live `plan.nudge` row per day. It has no awareness of snooze children.
- Sequence: tick N — user snoozes; tick N+1 — `maybe_nudge`'s sweep runs *before* `consume_responses`, so the newly-created snooze row (dedup `plan.nudge:{today}:snooze:...`, `expires_at = NULL`) survives that tick. Tick N+2 — the sweep now sees that same row (`dedup_key <> "plan.nudge:{today}"`, `expires_at IS NULL`) and stamps `expires_at = now`. Its `scheduled_for` is still ~58 minutes out. When that time arrives, `pending_native`'s `expires_at > now` filter now excludes it forever — the snoozed toast never fires.
- This only affects `plan.nudge` — `worklog.ready`'s scheduler (`src/pm_worklog/scheduler.rs`) has no analogous sweep, so its snooze path is fine.
- Fix: exclude snooze-children from the sweep (e.g. `AND dedup_key NOT LIKE ? ESCAPE...` matching `"{dedup}:snooze:%"`, or better, gate on `scheduled_for` rather than blanket-expiring every other dedup key for the event).
- No test exercises this interaction — `src/notification_responses.rs` ships with zero unit tests (the PR body's "410 workspace tests" are all pre-existing plus the `meridian-core` response-leg tests; the actual re-enqueue/dedup logic in the new file is untested), which is how this slipped through.

### 🟡 Deep-link value is fetched but discarded on native-toast open

- `tray/src-tauri/src/commands/notifications.rs::record_notification_response` does:
  ```rust
  if matches!(action.as_str(), "tap" | "open" | "view")
      && meridian_core::notifications::notification_deep_link(pool, id).await.is_some()
  { crate::commands::open_dashboard(app).await }
  ```
  It only checks `.is_some()` and throws away the actual path string. `open_dashboard()` (`tray/src-tauri/src/commands/system.rs`) always opens the dashboard's default/empty route — it takes no navigation target. So "Open Plan", "Open Worklogs", and "View" (system fault) all behave identically: they raise the dashboard window but never land on `/plan`, `/worklogs`, etc.
- The doc comment above the function claims "the link is resolved from the DB row" (implying it's used for navigation), which overstates what actually happens — worth either wiring the deep link through (dispatch the same custom-event mechanism `NotificationBanner.tsx` already uses for in-app banner clicks) or correcting the comment to make clear this is a v1 known limitation.

### 🟢 Minor

- `meridian_core::notifications::has_action_columns()` (a `pragma_table_info` COUNT query) is called on every invocation of `pending_native`, `record_response`, `unconsumed_responses`, and `expired_unanswered` independently — up to 4 extra round-trips per poll tick with no caching. Cheap on SQLite but easy to memoize (e.g. `OnceCell<bool>`) now that 057 is the steady-state schema.
- `use meridian_core::notifications::categories;` is placed mid-function in `src/daily_plan.rs` and `src/pm_worklog/scheduler.rs` rather than at the top of the file — purely stylistic, not a hard-rule violation.

### Things checked and found clean
- Response correlation: notification id ↔ outbox row id is the sole correlation and is validated (`i32::try_from` fallback to uncorrelated plain toast) — no risk of a stale/replayed response applying to the wrong row.
- Idempotency: `record_response`/`mark_response_consumed` both use `IS NULL` guards; snooze's re-enqueue dedup key derives from `responded_at` so retries collapse to `ON CONFLICT DO NOTHING`.
- Plugin isolation: `tauri_plugin_notifications::init()` is gated behind `sys::is_bundled()`, and every call site degrades to a logged no-op rather than panicking when the plugin isn't registered — a backend failure can't crash the tray.
- File headers present on all new `.rs`/SQL files; `tracing::instrument` + structured fields used consistently on the new DB functions and the response consumer; no `unwrap()` outside tests spotted; no argument lists over clippy's 7-arg limit.
- Pre-057 schema fallback is correctly wired and covered by a dedicated test (`pending_native_carries_category_and_survives_pre_057`).

### Overall
Good idempotency and fail-soft design on the core mailbox primitives, but the `plan.nudge` snooze feature — the PR's headline "response leg" capability for the flagship nudge — is broken by an unaccounted interaction with pre-existing sweep logic, and the untested new file (`notification_responses.rs`) is why it wasn't caught. Recommend fixing the sweep-vs-snooze interaction (and adding a test that exercises `maybe_nudge` + `snooze` together) before merge; the deep-link gap is worth a follow-up but is lower severity.

src/notification_responses.rs
Akarsh-Hegde: Fixed in c3fcef6e. The stale-nudge sweep in `daily_plan.rs` now matches `dedup_key NOT LIKE 'plan.nudge:{today}%'` (prefix) instead of `<> 'plan.nudge:{today}'` (exact), so today's snooze re-enqueue (`plan.nudge:{today}:snooze:{ts}`) is spared while prior-day rows are still expired. The confirmed/skipped expiry is now a prefix `LIKE` too, so a pending snooze is cancelled once the plan is actioned. Extracted the sweep into `expire_stale_nudges()` with a regression test (`stale_sweep_spares_today_and_snooze_but_expires_prior_days`).

meridian-core/src/notifications.rs
Akarsh-Hegde: Fixed in c3fcef6e. `has_action_columns` now caches per pool (`Mutex<HashSet<usize>>` keyed by pool handle), latching **true only** — the 057 columns are monotonic, and a `false` also covers a transient probe error, so caching a false would wrongly disable the response leg on one bad read. Per-pool rather than a process-global flag specifically so the `pending_native_carries_category_and_survives_pre_057` test (post-057 then pre-057 pool in one process) still passes.

tray/src-tauri/src/sys.rs
Akarsh-Hegde: Deliberately not changing this — reasoning. A fixed sentinel like `.id(0)` on plain `notify()` would give **every** plain toast the same identifier, so the OS would collapse them onto one notification (a new health toast replacing a pause/resume toast, etc.) — worse than today. The real question is what `tauri-plugin-notifications` 0.4.6 defaults an unset id to, which can't be verified offline (the crate isn't vendored here). The app.js guard (`Number.isFinite(outboxId) && outboxId > 0`) is safe unless the plugin's default is a positive int colliding with a real outbox id (which start at 1). Deferring pending a packaged-build check of that default; if it collides, the fix is a distinct non-colliding sentinel per plain toast, not a shared `0`.

tray/src-tauri/src/lib.rs
Akarsh-Hegde: Fixed in c3fcef6e. `register_notification_categories` is synchronous and needs no authorization, so it now runs **synchronously in `setup()`** before the poll loop is spawned — a nudge already due at launch can no longer be delivered before its `action_type_id` is registered. Only the permission prompt (which gates whether delivery happens at all, never the buttons) stays in the spawned task.

tray/src-tauri/src/sys.rs
Akarsh-Hegde: Fixed in c3fcef6e. `update::is_packaged()` now delegates to `sys::is_bundled()`, so the updater gate and the plugin-registration gate share one implementation. Left `install.rs::detect_install_mode` alone — it answers a broader install-shape question, not the boolean "am I in a .app".

meridian-core/src/notifications.rs
Akarsh-Hegde: Fixed in c3fcef6e. Split `notifications.rs` (888 lines) into `notifications/mod.rs` (405 — delivery policy + native/banner reads/writes), `notifications/responses.rs` (161 — the response leg), and `notifications/tests.rs` (373). All three are now under the 500-line cap, and the public path stays `meridian_core::notifications::*` via re-export, so callers are unchanged.

meridian-core/src/notifications.rs
Akarsh-Hegde: Fixed in c3fcef6e. Added `.context("recording notification response")` and `.context("marking notification response consumed")` to the two response-leg writes (now in `responses.rs`).

src/daily_plan.rs
Akarsh-Hegde: Fixed in c3fcef6e. Added `NewNotification::interactive(category)`, which sets both the category id and its `actions_json` from the single source of truth so they can't drift. All three producers now call it — `daily_plan.rs`, `pm_worklog/scheduler.rs`, and `notices.rs` (which previously used the no-`unwrap_or` struct-literal form, now consistent). The planned `verify_switch` producer becomes a one-liner too.

Akarsh-Hegde:
## Review findings addressed — `c3fcef6e`

**Fixed (7):**
- **Snooze `plan.nudge` silently killed** — stale-nudge sweep now prefix-matches today's dedup, sparing the snooze re-enqueue; snooze also cancelled on plan confirm. Extracted `expire_stale_nudges` **with a regression test**.
- **`has_action_columns` uncached** — per-pool latch-true-only cache (still passes the multi-pool `survives_pre_057` test).
- **Category registration race** — categories now registered synchronously in `setup()` before the poll loop starts.
- **Missing `.context(...)`** — added to both response-leg writes.
- **`interactive()` builder** — pairs category+actions from one source; all 3 producers migrated.
- **500-line cap** — split into `mod.rs` (405) / `responses.rs` (161) / `tests.rs` (373); public path unchanged.
- **`is_bundled` duplication** — `update::is_packaged` delegates to `sys::is_bundled`.

**Reasoned, not changed (1):**
- **Plain `notify()` sentinel id** — a shared `.id(0)` would collapse all plain toasts onto one OS identifier (worse than today). Deferring pending a packaged-build check of the plugin's unset-id default; see the thread.

**Verification:** 412 daemon + 119 meridian-core tests pass, clippy clean (daemon + tray), tray compiles clean. Full pre-push suite (fmt + clippy + UI build + UI tests + cargo test + security audit) passed on push.

PR #395, opened 2026-07-07
refactor(observability): consolidate all logging into one OpenTelemetry pipeline

adityaharishch:
## Summary
- Removes duplicate log stores (per-service JSONL files, launchd stdout/stderr tracing mirror, Agno's separate SQLite trace store) so every `tracing`/`logging` call — Rust and Python — flows through exactly one OTel spool-backed pipeline (`~/.meridian/telemetry/pending/`), always capturing locally regardless of shipping config.
- Restores dev-workflow terminal log visibility as a debug-build-only mirror (`cfg!(debug_assertions)`, not install-type detection) so `cargo run`/`dev-start.sh` behave like before, while release/prod builds never get it.
- New `meridian logs [--service <name>] [-n N] [-f]` decodes the OTel spool into human-readable lines locally, replacing the old JSONL-tailing `meridian logs`.
- Agno's spans now ride the same global `TracerProvider` (no more separate `agno_traces.db` / `agno_viewer.py`).
- Export/diagnostics bundle (tray Settings → Advanced → Export Diagnostics, and `meridian telemetry export`) now bundles pure OTLP spool files plus the (now size-capped via copytruncate) launchd crash-safety-net logs — the one thing OTel structurally can't capture.

## Test plan
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [x] `cargo test --workspace` — 422+ passed (one pre-existing, unrelated flaky test in `meridian-oauth` confirmed to pass when serialized — a parallel env-var race predating this branch)
- [x] `ui` production build — clean
- [x] `tray/src-tauri` fmt + clippy — clean
- [x] Live end-to-end run of the daemon binary: confirmed terminal log mirror works in a dev build AND the OTel spool captures the same run in parallel; `meridian logs -n 5` decodes it correctly
- [x] Python syntax check on `services/agents/observability.py` / `server.py`
- [x] Full pre-push hook suite (fmt, clippy, ui build/tests, security audit, cargo test) passed on push


src/observability.rs
```diff
@@ -38,11 +73,34 @@ use opentelemetry_sdk::{
...
+    std::env::var("HOME")
+        .map(|home| std::path::Path::new(&home).join(".meridian/.env").exists())
+        .unwrap_or(false)
+}
+
+/// Hard kill switch for OTel capture (spans/logs to the local spool). Plain
+/// stdout/stderr/JSONL-file logs are unaffected. Mirrors Python's
+/// `MERIDIAN_TRACING_DISABLED`.
+fn capture_disabled() -> bool {
+    std::env::var("MERIDIAN_TELEMETRY_DISABLED")
```
Akarsh-Hegde: **[Correctness — CONFIRMED] The telemetry kill switch has two different names on the two sides of the pipeline.**

Rust checks `MERIDIAN_TELEMETRY_DISABLED`:
```rust
fn capture_disabled() -> bool {
    std::env::var("MERIDIAN_TELEMETRY_DISABLED")
        .is_ok_and(|v| v == "1" || v.eq_ignore_ascii_case("true"))
}
```
but the doc comment right above it says it "Mirrors Python's `MERIDIAN_TRACING_DISABLED`" — and Python (`services/agents/observability.py:244`) actually reads that different name:
```python
return os.environ.get("MERIDIAN_TRACING_DISABLED", "").lower() in ("1", "true", "yes")
```
`.env.example` documents `MERIDIAN_TRACING_DISABLED` (the Python name), not the Rust one.

Net effect: setting `MERIDIAN_TRACING_DISABLED=1` (the only one documented in `.env.example`) silences the Python MLX server but the Rust daemon keeps capturing/spooling — a dev expecting "capture fully off" gets a partial, silent no-op on the Rust side. Since this is now the *sole* local-capture pipeline (this PR removed the other log stores), a broken kill switch means there's no way to actually turn off Rust-side capture via the documented variable.

Suggest picking one name (likely `MERIDIAN_TELEMETRY_DISABLED`, since that's the more accurate description post-consolidation) and updating both sides + `.env.example` to match.

src/telemetry_spool/shipper.rs
```diff
@@ -328,14 +347,99 @@ fn prune_sent(sent_dir: &Path) -> Result<()> {
...
+                tracing::warn!(file = %path.display(), error = %e, "failed to cap oversized launchd log file");
+            }
+        }
+    }
+}
+
+/// Keep only the last `keep_bytes` of `path`, in place. Reads the tail,
+/// truncates to zero, writes the tail back — a single fd, no rename, so it
+/// works even while another process holds the same path open for appending.
+fn copytruncate(path: &Path, keep_bytes: u64) -> std::io::Result<()> {
```
Akarsh-Hegde: **[Correctness — CONFIRMED] `copytruncate()` has a TOCTOU window that silently drops log lines written during the race.**

```rust
fn copytruncate(path: &Path, keep_bytes: u64) -> std::io::Result<()> {
    ...
    let len = file.metadata()?.len();
    let skip = len.saturating_sub(keep_bytes);
    file.seek(SeekFrom::Start(skip))?;
    let mut tail = Vec::with_capacity(keep_bytes.min(len) as usize);
    file.read_to_end(&mut tail)?;

    file.set_len(0)?;          // <-- anything appended between the read above and this line is gone
    file.seek(SeekFrom::Start(0))?;
    file.write_all(&tail)?;
    Ok(())
}
```

The doc comment cites logrotate's `copytruncate` as precedent ("works even while another process holds the same path open for appending"), but logrotate's copytruncate has the exact same known race — it's a documented tradeoff of that strategy, not something this implementation avoids. Between `read_to_end` and `set_len(0)`, launchd's O_APPEND writer (the daemon/mlx-server actively logging) can append new bytes. Those bytes are:
- not in `tail` (already read before they arrived), and
- destroyed by `set_len(0)` (the file is zeroed, then overwritten with only the pre-race `tail`).

This runs on every 30s shipper tick (`cap_launchd_logs` → `copytruncate`), so the highest-risk moment is exactly when it matters most: a burst of crash/error logging right as the tick fires can be silently truncated away, with zero indication anywhere (no gap marker, no warning) that data was lost.

Worth at least logging when a race is *likely* (e.g. compare `file.metadata()?.len()` again right before `set_len(0)` and warn if it grew), even if a full fix (advisory lock, or write-then-atomic-rename into a new inode) is out of scope for this PR.

scripts/meridian-cli.sh
```diff
@@ -152,52 +152,63 @@ cmd_status() {
...
-    fi
-
-    if [[ $follow -eq 1 ]]; then
-        tail -n "$lines" -f "$log_file"
-    else
-        tail -n "$lines" "$log_file"
-    fi
+    local bin
+    bin="$(_meridian_native_bin)" || { err "meridian binary not found — run ./install.sh"; exit 1; }
+    exec "$bin" logs "${service_args[@]}" "$@"
```
Akarsh-Hegde: **[Correctness — CONFIRMED, reproduced locally] `meridian logs` (no target) and `meridian logs all` crash with "unbound variable" on macOS's stock `/bin/bash`.**

```bash
local service_args=()
case "$target" in
    ""|all)                     ;;   # service_args stays empty
    ...
esac
...
exec "$bin" logs "${service_args[@]}" "$@"
```

The script is `set -euo pipefail` and shebangs `#!/usr/bin/env bash`, so on any Mac where `/bin/bash` (3.2.57, GNU's last GPLv2 release, still the system default) resolves first in `PATH`, expanding an empty array with `"${service_args[@]}"` under `set -u` throws `unbound variable` and aborts — this was only fixed in bash 4.4+. Reproduced directly on this machine:

```
$ /bin/bash -c 'set -u; a=(); f(){ echo "args:" "${a[@]}" "done"; }; f'
/bin/bash: a[@]: unbound variable
```

`meridian logs` and `meridian logs all` (arguably the most common invocations — "just show me what's happening") are the ones that hit the empty-array case, so they're the ones that break on any machine without a newer bash ahead of `/bin/bash` in `PATH`. `meridian logs daemon` (non-empty array) works fine, which likely explains why this wasn't caught in testing if the "all" case wasn't exercised.

Fix: `exec "$bin" logs ${service_args[@]+"${service_args[@]}"} "$@"` (the `+` guard skips expansion entirely when the array is unset/empty) or simply avoid `set -u` interacting with this array by using `"${service_args[@]:-}"`-style guards throughout. Same issue applies to `follow=()` in the retired `screenpipe`/`ui` branch a few lines up.

scripts/meridian-cli.sh
```diff
@@ -152,52 +152,63 @@ cmd_status() {
...
-        mlx-server)        log_file="${LOG_DIR}/mlx-server.log" ;;
-        mlx-server-error)  log_file="${LOG_DIR}/mlx-server-error.log" ;;
-        tray)              log_file="${LOG_DIR}/tray.log" ;;
-        tray-error)        log_file="${LOG_DIR}/tray-error.log" ;;
-        # screenpipe/ui are retired post-v1.64.0 (capture is in-process in the
-        # tray; the dashboard is embedded) — kept here so old log files remain
-        # tailable, but `tray`/`tray-error` are the live targets now.
-        *) err "unknown log target: ${target} (daemon|daemon-error|mlx-server|mlx-server-error|tray|tray-error)"; exit 1 ;;
+        ""|all)                     ;;
+        daemon|daemon-error)        service_args=(--service meridian-rust) ;;
```
Akarsh-Hegde: **[Correctness — CONFIRMED] `daemon-error`/`mlx-server-error`/`tray-error` targets no longer filter to errors — they're now identical to their non-error counterparts.**

```bash
daemon|daemon-error)        service_args=(--service meridian-rust) ;;
mlx-server|mlx-server-error) service_args=(--service meridian-mlx-server) ;;
tray|tray-error)            service_args=(--service meridian-tray) ;;
```

On `main` (pre-PR), `daemon-error` mapped to a distinct `daemon-error.log` file that launchd's `StandardErrorPath` routed only WARN+/stderr output into — a real severity filter. Post-PR, `daemon-error` maps to the exact same `--service` filter as `daemon`; the new `render.rs`/`cli.rs` spool decoder has no `--level`/severity flag at all, so there's no way to reconstruct that filtering from the OTel spool.

`meridian logs daemon-error` now silently prints identical output (INFO/DEBUG included) to `meridian logs daemon` — the `-error` suffix looks like it does something but doesn't, which is worse than removing the target outright since a user debugging via `daemon-error` gets no error and no indication the filter didn't apply.

Suggest either adding a `--level`/`--min-severity` flag to the spool decoder and wiring `*-error` targets to it, or removing the `-error` targets and documenting that severity filtering is gone (e.g. pipe through `grep` instead).

src/observability.rs
```diff
@@ -38,11 +73,34 @@ use opentelemetry_sdk::{
...
+/// exists — written by the DMG/npm installer; see CLAUDE.md's "Daemon config
+/// gotcha"). A packaged install must never attempt live delivery to
+/// OpenObserve — telemetry capture stays fully local, and the only path to a
+/// developer's OpenObserve is a user-initiated export bundle imported by hand.
+/// A Dev/Bare checkout (no canonical `.env`) may still ship live if
+/// `otlp_enabled` + credentials are configured, for engineers debugging
+/// against their own instance. Mirrors (independently — the daemon has no
+/// dependency on the tray crate) `tray/src-tauri/src/install.rs`'s
+/// `InstallMode::Canonical` check.
+fn is_canonical_install() -> bool {
```
Akarsh-Hegde: **[Correctness — PLAUSIBLE] `is_canonical_install()` can false-positive on a developer's own machine, silently blocking dev-build telemetry shipping.**

```rust
fn is_canonical_install() -> bool {
    std::env::var("HOME")
        .map(|home| std::path::Path::new(&home).join(".meridian/.env").exists())
        .unwrap_or(false)
}
```

Per CLAUDE.md's "Daemon config gotcha", `~/.meridian/.env` is written by **every** install type that has ever run on that machine (npm bundle installer, `.app` DMG installer) — it's a per-`$HOME` marker, not a per-process one. Any Meridian engineer who has ever installed the packaged app on their own dev Mac (plausible/likely, since dogfooding the shipped product is normal) will have this file present permanently.

When that same engineer then `cargo run`s the daemon from source to debug against their own OpenObserve instance (the explicitly-documented "Dev/Bare checkout may still ship live" workflow this function exists to allow), `is_canonical_install()` returns `true` anyway — because the file exists on their `$HOME`, not because *this process* is the packaged build — so `resolve_otlp_target()` returns `None` and shipping is silently disabled, with no override.

Worth distinguishing "this machine has a canonical install somewhere" from "this process IS a canonical install" — e.g. checking the running executable's own path (like `sys::is_bundled()`/`update.rs::is_packaged()` do elsewhere in the tray) rather than a machine-wide marker file.

tray/src-tauri/src/commands/settings.rs
```diff
@@ -16,8 +16,11 @@
 //!
 //! # Related
 //! - [`meridian_core::settings`] — the schema, path, and the value read/write.
-//! - [`crate::commands::openobserve`] / [`crate::commands::daemon`] — the
-//!   service-start + daemon-reload steps the Settings "Apply" flow chains after a save.
+//! - [`crate::commands::daemon`] — the daemon-reload step the Settings "Apply"
+//!   flow chains after a save. `otlp_enabled`/`otlp_endpoint`/`oo_email`/
```
Akarsh-Hegde, line 20: **[Removed-behavior] The OpenObserve config UI is gone, but the settings fields it wrote are still the only way to enable Dev/Bare live shipping.**

This PR deletes `useApplyObservability.ts` and the `otlp_enabled`/`otlp_endpoint`/`oo_email`/`oo_password` fields from `AdvancedSection.tsx`, replacing that whole section with just an Export Diagnostics button. But `commands/settings.rs` still validates and persists exactly those fields (this line onward), and `src/observability.rs`'s `resolve_otlp_target()`/`is_otlp_configured()` still read them from `settings.json` to decide whether a Dev/Bare install ships live.

The documented "Dev/Bare checkout may still ship live if `otlp_enabled` + credentials are configured" workflow (per this PR's own doc comments) now has no in-app UI path — an engineer wanting to point their dev daemon at their own OpenObserve instance must hand-edit `~/.meridian/settings.json` directly, since the feature that used to configure this was removed without a replacement (or a note that it's now settings.json-only).

If this is intentional (the feature is dev-only and hand-editing settings.json is an acceptable dev workflow), worth a comment in `settings.rs` or the module doc saying so explicitly, since otherwise it reads as an accidentally-orphaned code path.

src/telemetry_spool/render.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    if !follow {
+        return;
+    }
+
+    loop {
+        tokio::time::sleep(Duration::from_secs(1)).await;
+        let records = match collect_all_records(service.as_deref()) {
+            Ok(r) => r,
+            Err(_) => continue,
```
Akarsh-Hegde, line 338: **[Correctness — PLAUSIBLE] `meridian logs -f` drops records that share the exact same timestamp as the last-printed record.**

The follow loop filters strictly `r.time_unix_nano > last_time` before printing, then updates `last_time` to the newest printed record's timestamp. Two records legitimately sharing the same `time_unix_nano` (coarse clock resolution, or a batch flush stamping several records from one `SystemTime::now()` read) can end up split across two poll ticks: if one lands in the set printed this tick and becomes the new `last_time`, and a second record with the *identical* timestamp arrives in the next tick's re-scan, it's permanently excluded by the strict `>` — it can never satisfy the filter in any future tick either.

Low-frequency in practice (needs two records at the same nanosecond split across a 1s poll boundary), but when it happens the missing line is invisible with no error — exactly the kind of gap that's hard to notice while debugging a live incident with `-f`. Consider tie-breaking on `(time_unix_nano, some stable secondary key)` or tracking "already-printed record identities" rather than a bare timestamp high-water mark.

src/observability/mod.rs
```diff
@@ -2,26 +2,61 @@
...
-//   2. Writes JSON Lines to `~/.meridian/logs/<service>.jsonl` with daily rotation
-//   3. Exports OpenTelemetry traces to OpenObserve via OTLP/HTTP
-//   4. Exports OpenTelemetry logs  to OpenObserve via OTLP/HTTP
-//      (log events carry trace_id/span_id so they correlate with traces in OO)
+// One call to `init(service_name)` builds a `tracing` subscriber whose
+// canonical, persisted sink is the local OpenTelemetry telemetry spool
+// (`~/.meridian/telemetry/pending/`, both traces and logs — log events carry
+// trace_id/span_id so they correlate with traces). There is no separate
+// JSON-Lines log file — every `tracing::*!` call is captured there, so there
+// is exactly one STORED log/trace representation to export/import, not
```
Akarsh-Hegde, line 10: **[Conventions] Two files already at or over the repo's 500-line cap grow further without being split.**

Repo CLAUDE.md → Hard Rules: "Keep files under 500 lines; split when a file grows beyond that."

- `src/observability.rs`: 469 lines on `main` → 540 lines in this PR (new `is_canonical_install`/`capture_disabled` logic, expanded module doc) — crosses the cap.
- `src/telemetry_spool/shipper.rs`: already 534 lines on `main` (over cap) → 680 lines in this PR (adds `cap_launchd_logs`/`copytruncate` and related logic) — worsens an existing violation.

Suggest splitting the launchd-log-capping logic (`cap_launchd_logs`/`copytruncate`/`LAUNCHD_LOG_NAMES`) out of `shipper.rs` into its own module, and considering whether `is_canonical_install`/`capture_disabled`/related gating logic in `observability.rs` belongs in a smaller sibling module.

src/observability.rs
adityaharishch: Confirmed, fixed in 5974da61. Standardized on `MERIDIAN_TELEMETRY_DISABLED` on both sides — Python's `_capture_disabled()` now reads that same name, and `.env.example` is updated. Both docstrings now cross-reference each other so this can't silently drift again.

src/telemetry_spool/shipper.rs
adityaharishch: Confirmed, fixed in 5974da61. Can't close the TOCTOU window without an advisory lock the writer doesn't take (out of scope here), but `copytruncate()` now re-checks the file's length immediately before `set_len(0)` and emits a `tracing::warn!` with pre/post lengths if it grew — so a lost burst leaves a trace in the OTel spool itself instead of vanishing with zero indication.

scripts/meridian-cli.sh
adityaharishch: Confirmed and reproduced locally too — good catch. Fixed in 5974da61 using the `${arr[@]+"${arr[@]}"}` guard (skips expansion entirely when the array is empty/unset) for both `service_args` in `cmd_logs` and `follow` in the retired screenpipe/ui branch.

scripts/meridian-cli.sh
adityaharishch: Confirmed, fixed in 5974da61. Added a `--min-severity LEVEL` flag to the spool decoder (`render.rs`) that filters to log records at/above the given severity and excludes spans (matching the old WARN+-only, log-text-only behavior of the launchd StandardErrorPath files). `daemon-error`/`mlx-server-error`/`tray-error` in meridian-cli.sh now pass `--min-severity WARN` instead of aliasing to the non-error target.

src/observability.rs
adityaharishch: Confirmed, fixed in 5974da61. `is_canonical_install()` now checks the running executable's own path (`current_exe()` against the two DMG/npm staged locations) instead of the machine-wide `~/.meridian/.env` marker — same pattern `tray/src-tauri/src/update.rs`'s `is_packaged()` already uses. Falls back to the old marker-file check only if `current_exe()` itself fails.

tray/src-tauri/src/commands/settings.rs
adityaharishch: Good flag — this is intentional (OTLP shipping is a Dev/Bare-only debugging feature, settings.json-hand-edit is an acceptable workflow for it), but you're right it read as orphaned. Added an explicit doc comment in `settings.rs` in 5974da61 spelling out that no UI writes these fields anymore and why.

src/telemetry_spool/render.rs
adityaharishch: Confirmed, fixed in 5974da61. The follow loop now tracks the identity (not just the timestamp) of every record already printed at the current high-water mark, so a same-nanosecond record arriving in a later poll tick is no longer permanently excluded by the old strict `>` comparison.

src/observability/mod.rs
adityaharishch: Fixed in 5974da61 — split both files: `observability.rs` → `observability/{mod,install_mode,otlp_target}.rs` (402/66/142 lines), `shipper.rs` → `shipper.rs` + `retention.rs`/`launchd_log_cap.rs` (287/322/156 lines). All under the 500-line cap now.

PR #396, opened 2026-07-07
feat(tray): open dashboard when app is re-activated externally

Akarsh-Hegde:
## Summary
- Add a Tauri `RunEvent::Reopen` handler so activating Meridian externally (Spotlight, dock click, `open -a Meridian`) actually opens something instead of being a silent no-op while the tray is running.
- Onboarded → dashboard window. Not onboarded → setup wizard (resumes a partial onboard).
- Cold-start behavior unchanged: the setup hook still auto-opens the wizard on first launch.

## Why
The tray runs as `ActivationPolicy::Accessory` — no dock icon most of the time — and Tauri drops the reopen event on the floor without a handler. Users hitting ⌘Space → "Meridian" ⏎ saw nothing happen.

## Test plan
- [ ] Fresh `~/.meridian` (no `onboarded` file) → Spotlight-launch Meridian on an already-running tray → setup wizard opens.
- [ ] With `onboarded` present → Spotlight-launch Meridian → dashboard window opens (or focuses if already open).
- [ ] Dock click while dashboard is open → dashboard focuses.
- [ ] `cargo check` + `cargo clippy` clean on `tray/src-tauri/`.

### DIFF
--- a/tray/src-tauri/src/lib.rs
+++ b/tray/src-tauri/src/lib.rs
@@ -511,8 +511,29 @@ pub fn run() {
             commands::detect_system_specs,
             tray_debug,
         ])
-        .run(tauri::generate_context!())
-        .expect("error running meridian tray");
+        .build(tauri::generate_context!())
+        .expect("error running meridian tray")
+        .run(|app, event| {
+            // macOS: fires when the user re-activates the app externally
+            // (Spotlight, dock click, `open -a Meridian`). The tray runs as an
+            // Accessory app so there is no dock icon most of the time; without
+            // this handler, hitting "Meridian" in Spotlight is silently a no-op
+            // when the app is already running. Route to the dashboard when the
+            // user has finished setup, otherwise re-open the wizard so a partial
+            // onboard can be resumed. The setup hook already opens the wizard on
+            // cold start, so this only matters for warm activation.
+            if let tauri::RunEvent::Reopen { .. } = event {
+                let home = std::env::var("HOME").unwrap_or_default();
+                let onboarded =
+                    std::path::Path::new(&format!("{home}/.meridian/onboarded")).exists();
+                tracing::info!(onboarded, "app.reopen: routing external activation");
+                if onboarded {
+                    tray::open_native_dashboard(app);
+                } else {
+                    tray::open_wizard_window(app);
+                }
+            }
+        });
 }
 
 /// Debug bridge: lets the popover/tooltip JS forward `window.onerror` reports and
--- a/tray/src-tauri/src/lib.rs
+++ b/tray/src-tauri/src/lib.rs
@@ -529,6 +529,13 @@ pub fn run() {
             // avoid unused warnings. The routing decision lives in the
             // platform-independent [`reopen_target`] / [`is_onboarded`] so it is
             // unit-testable without a live Tauri app (see the tests below).
+            //
+            // `has_visible_windows` is intentionally ignored (`{ .. }`): both
+            // openers already reuse an existing dashboard/wizard window via
+            // `get_webview_window(..)` + show/focus before building a new one,
+            // so a Reopen while a window is already up just re-focuses it — the
+            // flag would only matter if we wanted different behaviour for
+            // "windows visible" vs "all minimised", which we don't.
             #[cfg(target_os = "macos")]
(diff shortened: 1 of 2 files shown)


adityaharishch:
Reviewed the diff (`tray/src-tauri/src/lib.rs`, `tray/src-tauri/src/tray.rs`). Overall this is a small, well-scoped change and it's structurally sound — no blockers.

**What checks out:**
- `RunEvent::Reopen` (macOS `applicationShouldHandleReopen`) is fired only by dock-icon click / `open -a` / Spotlight re-activation — it is a distinct event from the tray status-item click (`TrayIconEvent`), so there's no risk of this firing when the user clicks the menu-bar icon and popping the dashboard/wizard over the popover.
- Window reuse is correct: both `tray::open_native_dashboard` (`tray.rs:87-124`) and `tray::open_wizard_window` (`tray.rs:163-178`) already `get_webview_window(...)` + `show()`/`set_focus()` before building a new one, so the new `Reopen` handler can't spawn a duplicate dashboard/setup window even on repeat activations.
- Correctly reuses the existing openers rather than reimplementing window creation — `open_native_dashboard` was widened from `fn` to `pub(crate) fn` (`tray.rs:87`) specifically to make this possible, which is the right move (single source of truth, per the module's own doc comment about the tray-menu path needing "the same fix").

**Minor, non-blocking:**
- `lib.rs:414` (`Reopen` handler) reads `~/.meridian/onboarded` inline via `std::env::var("HOME")`, which is now the *third* copy of this exact check — `commands/setup.rs:38-40` (`is_first_run`) and the existing first-launch auto-open at `lib.rs:407-418` both do the same thing. Not introduced by this PR (the pattern already existed pre-change), but this would've been a good spot to factor into a small shared helper (e.g. `fn is_onboarded() -> bool`) since the PR is already touching this exact logic. Not worth blocking on.
- The `Reopen` match ignores `has_visible_windows` (destructured as `{ .. }`). Harmless given the reuse behavior in the openers above (a second `Reopen` while the dashboard is already frontmost just re-shows/re-focuses it), but if `has_visible_windows` is ever relevant to future logic here it's currently unused — worth a comment noting it's intentionally ignored, for the next person who touches this.
- No `.context(...)` needed since there are no fallible calls here beyond the pre-existing `unwrap_or_default()` on `HOME`, which matches existing precedent elsewhere in the file (`install.rs`) — fine.

Overall: looks correct and safe to merge. The event choice, window-reuse guarantees, and onboarding-vs-dashboard routing all check out.

Akarsh-Hegde:
Follow-up commit adds test coverage **and fixes a compile break**:

- **Linux CI compile fix:** `RunEvent::Reopen` is a `#[cfg(target_os = "macos")]` enum variant — it doesn't exist on other targets. The un-gated `if let` would fail to compile on `ubuntu-latest`, where CI builds the whole workspace (the tray is a member). The macOS-only pre-push hook can't catch this. The arm is now `#[cfg(target_os = "macos")]`-gated.
- **Unit tests:** extracted `is_onboarded(home)` + `reopen_target(onboarded) -> ReopenTarget` (platform-independent, so the tests run in Linux CI) and added coverage for both routing arms + the onboarded-marker filesystem check.

The OS-integration half (macOS actually delivering the reopen event on Spotlight/dock activation) still needs a manual check — there's no WebDriver path for a menu-bar app on macOS.

Akarsh-Hegde:
Thanks for the thorough pass — both minor notes addressed:

- **Duplicated `~/.meridian/onboarded` check** → factored out in a follow-up (4560f56d): the reopen handler now calls a shared `is_onboarded(home)` helper (with a companion `reopen_target(onboarded) -> ReopenTarget`), both unit-tested. `commands::setup::is_first_run` is its documented inverse. (That follow-up also fixed a real one: `RunEvent::Reopen` is `#[cfg(target_os = "macos")]`, so the un-gated `if let` would fail to compile on the Linux CI runner — now cfg-gated. Detail in the earlier comment.)
- **`has_visible_windows` ignored** → added a comment (31064f71) explaining it's intentional: both openers reuse an existing window via `get_webview_window` + show/focus, so a reopen just re-focuses; the flag would only matter for distinct visible-vs-minimised behaviour we don't want.

Nothing else outstanding on your review.

PR #397, opened 2026-07-07
fix(wizard): hide global chrome banners on the setup route

Akarsh-Hegde:
## Summary
- Wrap `ExternalLinks`, `NoticeBar`, and `NotificationBanner` in a small `LayoutChrome` client component.
- Skip rendering them when the pathname starts with `/setup`, so the wizard renders bare.

## Why
The root layout rendered these three unconditionally around every page. During onboarding, a fault-notice bar or a notification firing mid-wizard landed on top of the rail and read as a bug. The wizard should be a self-contained shell.

## Test plan
- [x] `npm run build` in `ui/` clean (7 static pages).
- [ ] Manual: fresh `~/.meridian` → wizard opens with no notice/notification banners regardless of daemon state.
- [ ] Manual: normal dashboard still renders NoticeBar / NotificationBanner as before.


adityaharishch:
Reviewed the diff (`ui/app/layout.tsx`, `ui/components/LayoutChrome.tsx`). Scoping and React correctness look solid; one real issue worth fixing before merge.

**Findings**

- `ui/components/ExternalLinks.tsx:15` — the file's own header comment says it's "Rendered by app/layout.tsx (root layout), so it covers every window that serves the Next app — the dashboard and the setup wizard." That's now false: `LayoutChrome.tsx:22` (`pathname?.startsWith('/setup')`) unmounts `ExternalLinks` along with the two banners on `/setup`. Today the wizard has no `<a target="_blank">`/`mailto:`/`tel:` anchors (confirmed via grep on `ui/app/setup/*.tsx`), so there's no live bug, but this re-opens exactly the failure mode `ExternalLinks` was built to close: a future external link added to the wizard without explicit `onClick`/`open_external_url` wiring will silently no-op in the Tauri webview (no error, nothing happens). Either keep `ExternalLinks` always-mounted (it renders `null` and only listens for clicks — no visual/banner side effect, so there's no reason to gate it with `NoticeBar`/`NotificationBanner`), or update the comment to document the new exclusion so it doesn't mislead the next person wiring a link into the wizard.

**Other checks (no issues found)**
- Route matching: `/setup` is the only route under that prefix (`ui/app/setup/`), so `startsWith('/setup')` doesn't false-positive/negative against sibling routes.
- No flash-of-banner-then-hide: `usePathname()` is synchronous on first client render, and this is a static export — no waterfall/mismatch.
- Hooks/types: no dependency-array issues, no `any`, `'use client'` boundary is correctly scoped to `LayoutChrome` rather than leaking into `layout.tsx`.
- Tauri fold: no `fetch`/`EventSource`/`/api` usage reintroduced — this is pure client routing state, consistent with the bridge/invoke model.

Overall: good, tightly-scoped fix. Recommend addressing the `ExternalLinks` exclusion (or at least the stale comment) before merge, everything else is clean.

Akarsh-Hegde:
Thanks — this exact finding was resolved in a follow-up (d08ba8c7) pushed before this review landed. It was against the first cut (`LayoutChrome.tsx` gating all three components); that file no longer exists.

Current state matches your recommended fix — **keep `ExternalLinks` always-mounted**:
- `ExternalLinks` is rendered directly in `app/layout.tsx` for **every** route (wizard included), with an inline comment saying why: *"invisible interceptor — mounted for every route so external links work in the webview."*
- Only the two **visible** banners (`NoticeBar` + `NotificationBanner`) are gated off `/setup`, in a renamed `LayoutBanners.tsx`.
- So `ExternalLinks.tsx:15`'s "covers every window ... the dashboard and the setup wizard" is **true again** — no stale comment, and the failure mode you flagged (a future wizard link silently no-op'ing) can't re-open.
- `external-links.test.ts`'s guard (`layout.tsx` contains `<ExternalLinks />`) passes, and the full ui suite is green (242).

PR #398, opened 2026-07-07
feat(popover): click Capturing subtitle to expand a health panel

Akarsh-Hegde:
## Summary
- Turn the popover's `head-sub` line ("Capturing this hour" / "Offline") into a toggle. Clicking it expands an inline panel with three rows: daemon, MLX server, database — each with a status dot.
- Data comes from the existing `get_daemon_status` / `get_mlx_status` / `get_health` commands — no new backend surface.
- Panel refreshes on open and on every `status-update` tick while open. Closed = no polling.
- Popover auto-resizes via the existing ResizeObserver, so the panel opens/closes without a hand-rolled setSize.

## Why
The "Capturing" line was display-only. When the popover said "Offline" or "Capturing but nothing tracked", users had to open the dashboard to figure out which subsystem was actually down. Inline is faster and stays in the tray context.

## Test plan
- [x] `cargo check` on `tray/src-tauri/` clean.
- [x] `node --check tray/src/app.js` clean.
- [ ] Manual (`npm run tauri dev` in `tray/`): open popover, click "Capturing this hour" — panel expands with a chevron rotate, three rows populate within a tick.
- [ ] Manual: kill the daemon (`launchctl stop com.meridiona.daemon`) → head-sub shows "Offline", panel shows Daemon = Not running (red dot).
- [ ] Manual: keyboard — Tab to head-sub, Enter/Space toggles.


adityaharishch:
Reviewed the diff (app.js/index.html/style.css only — no new Tauri commands, no new files, so the file-header rule is N/A here).

**Findings**

- **Moderate — MLX `Error(String)` variant's message is discarded.** `tray/src-tauri/src/mlx_server.rs:99` defines `MlxStatus::Error(String)` carrying a diagnostic message. In `paintMlx()` (`tray/src/app.js`, new code in the `refreshHealth`/`paintMlx` block), `kind` is derived as `Object.keys(s)[0]` (`"Error"`), and the final `else` branch renders `String(kind)` — i.e. the literal word `"Error"` — instead of `s.Error` (the actual message the backend already computed). The panel is supposed to save users a trip to `Export Diagnostics`/logs for exactly this kind of detail, so surfacing the real string (e.g. `res.status.Error`, sliced like `paintDb` does for `res.error`) would be more useful than a bare "Error".

- **Minor/UX — idle-evicted MLX (expected, per `MLX_IDLE_EVICT_S`) reads as a warning.** When `runtime_installed` is true but `status` is `"Offline"` (e.g. the server was lazy-unloaded after the idle TTL — normal, documented behavior), `paintMlx` falls through to the generic `else` branch: warn (yellow) dot + literal text `"Offline"`. That's indistinguishable from an actual problem state, even though idle-eviction is intended behavior. Might be worth a dedicated branch (e.g. neutral dot + "Idle" or "Not running (idle)") so it doesn't read as something being wrong.

- **Nit — stale/inaccurate comment.** The comment above `paintMlx` (`// MlxStatus is a serde enum — either a string ("Idle") or { Running: {...} }`) doesn't match the real enum (`Offline | Starting | Running | Error(String)` in `mlx_server.rs:90-100`) — there's no `"Idle"` variant (it's `"Offline"`), and `Running` is a unit variant, not one with a payload object. Worth fixing so the next reader isn't misled.

**Checked and clean:**
- File header convention: no new files added; all three touched files already carry the required header on line 1 — untouched.
- Accessibility: `head-sub` has `role="button"`, `tabindex="0"`, `aria-controls`/`aria-expanded` wired correctly in JS, and Enter/Space both toggle with `preventDefault()`.
- Event listeners: the click/keydown listeners on `head-sub` are registered once at module load (not inside `render()`), so no listener-accumulation leak on repeated renders.
- XSS: all health-panel updates go through `.textContent`, never `innerHTML` — no injection risk even with the (bounded, sliced) error string from the daemon.
- Data freshness: panel refreshes on open and on every `status-update` tick only while open (`healthOpen` guard) — no background polling when collapsed, matches the PR description.
- Backend field shapes cross-checked against `DaemonStatusResponse`, `HealthResponse`, and `MlxStatusResponse` in `tray/src-tauri/src/commands/{daemon,health,setup}.rs` — `running`/`pid`/`database_ready`/`error`/`port`/`runtime_installed` all line up.

**Overall:** solid, focused change — no crashes, no leaks, no security issues. The two MLX-status findings are worth a quick follow-up (mainly the discarded error message, since that undermines the panel's stated purpose of surfacing *which* subsystem is down and why) but neither blocks merging.

Akarsh-Hegde:
Follow-up commit adds a DOM harness test **and fixes a bug it caught**:

- **Bug:** `MlxStatus` serializes snake_case (`"running"`), but `paintMlx` compared against `"Running"`/`"Starting"` — so a healthy MLX rendered a yellow "running" instead of a green "Running", and "Starting" never matched. Fixed to the real serde contract (`"running"`/`"starting"`/`"offline"` + `{ error }`).
- **Test:** `ui/__tests__/popover-health-panel.test.ts` — a happy-dom harness that loads the real `tray/src/app.js` with a mocked `__TAURI__` bridge and asserts the panel toggles + rows paint from `get_daemon_status`/`get_mlx_status`/`get_health` (incl. null→Unreachable). Closest automated guard achievable for a menu-bar NSPanel (macOS Tauri has no WebDriver path). Adds `@happy-dom/global-registrator` as a ui devDep; full ui suite 242 pass.

Akarsh-Hegde:
Thanks — addressed. Note the three findings you cite (`kind === 'Running'` mismatch, `String(kind)` fallback, and the stale `// ... "Idle" ... { Running: {...} }` comment) were all against the **original** commit; my follow-up (a14e5be6) had already fixed the snake_case comparison and rewritten that comment. The one that survived was the discarded error message — now fixed too.

**Moderate — discarded `Error(String)` message (fixed in 589f0052)**
`paintMlx`'s error branch now renders `Error: <msg>` from `res.status.error` (bounded to 60 chars, same as `paintDb` does for `res.error`), falling back to a bare "Error" only when no message is present. Added a DOM-harness test asserting the message is surfaced from the externally-tagged `{ status: { error: "<msg>" } }` shape.

**Nit — stale `paintMlx` comment**
Already fixed in the follow-up — it now reads: *"MlxStatus is a serde enum with rename_all = 'snake_case': the unit variants serialize to 'offline' / 'starting' / 'running', and Error(String) to { 'error': '<msg>' }."*

**Minor/UX — idle-evicted MLX reading as "Offline"**
Good instinct, but I think this one doesn't actually manifest: `MLX_IDLE_EVICT_S` unloads the *model* while the FastAPI **server process stays up and keeps answering `/health`**, so `sync_status` reports `Running` (green) throughout an idle eviction — `mx.get_active_memory()` drops but the status doesn't. `MlxStatus::Offline` specifically means the server process itself is down/unreachable, which is a genuine problem state, so warn + "Offline" is correct. If we later add a "model loaded vs evicted" sub-signal to `get_mlx_status`, a neutral "Idle (model unloaded)" branch would be a nice touch — but there's no field to distinguish it on today. Happy to add it if you'd rather I thread that through.

PR #405, opened 2026-07-08
fix(worklog): local-time inline markers in session distill, plain-language activity TLDR

adityaharishch:
## Summary
- Fix a UTC-vs-local-time bug in `session_distiller.py`'s `_distil()`: inline `[HH:MM:SS]` markers embedded in `session_text` (injected by the Rust ETL whenever on-screen content changes) are UTC by construction, but were being assigned straight into `cur_time` without conversion — while the session's own `started_at` timestamp *was* correctly localized via `utc_to_local_hhmm`. Since most thread headers in a longer session originate from these inline markers, most of `distill_hour`'s output (and therefore `activity_report`'s input) showed raw UTC times instead of local time. Fixed by reattaching `started_at`'s UTC date to the marker before running it through the same `utc_to_local_hhmm` helper already used elsewhere in the file.
- Rewrite `activity_report.py`'s TLDR instruction to stay at a high level, understandable by anyone (PM, teammate) rather than deep technical detail — Core Tasks keeps the existing technical depth.
- Switch report voice from third-party narration ("the developer did X") to subject-less, verb-led phrasing ("Focused on…", "Fixed…", "Investigated…").
- Bump `services/pyproject.toml` version (1.69.1 → 1.69.2) so this change actually ships on the next runtime publish, per the repo's runtime-versioning convention.

## Test plan
- [x] Manually verified via a fork agent: fed a UTC inline marker `[03:46:12]` through `_distil()` — output correctly renders as local time (`09:16` in IST, UTC+5:30) instead of raw UTC.
- [x] Confirmed `activity_report.py`/`routes/activity.py` has no independent timestamp parsing — it forwards `distill_hour`'s text verbatim as the prompt body, so no duplicate bug there.
- [x] No existing pytest coverage for `session_distiller.py` (confirmed via repo-wide search); none added per scope.
- [x] `cargo fmt`/`clippy`/`test`, UI build/tests, and security audit all passed on push (pre-push hook).

## Summary by CodeRabbit

* **New Features**
  * Improved activity report formatting for clearer, higher-level summaries with less technical jargon.
* **Bug Fixes**
  * Fixed how time markers are interpreted in session summaries, making timelines more accurate.
* **Chores**
  * Updated the app version to 1.69.2.

### DIFF
--- a/services/agents/prompts/activity_report.py
+++ b/services/agents/prompts/activity_report.py
@@ -17,7 +17,7 @@ Your job: infer what the developer was actually trying to accomplish and write a
 OUTPUT FORMAT — write all sections that have content, skip sections that are empty:
 
 ### TLDR
-One short paragraph. What was the developer focused on this hour and why — what problem were they solving or what goal were they advancing? Name the main work areas explicitly. Avoid generic descriptions like "the developer was doing development work."
+One short paragraph, written at a HIGH LEVEL — understandable by anyone (a PM, a teammate, someone outside the project), not just the developer themself. Say what area of the product/system was worked on and why, in plain outcome terms — not tool names, file paths, function names, or step-by-step technical detail (that level of detail belongs in Core Tasks below). Name the main work areas explicitly, but describe them the way you'd summarize this hour to someone who doesn't know the codebase. Avoid generic descriptions like "worked on development tasks."
 
 ### Core Tasks & Projects
 One section per distinct work thread. Bold the topic name as the header.
@@ -27,7 +27,7 @@ For each thread, write it as the developer's story — not as a list of actions:
 - WHAT: what the developer accomplished or decided — the outcome, not the steps
 - HOW: the significant technical context (which system, which file, which tool) that gives the outcome meaning
 
-Write in the developer's voice — "the developer fixed…", "the developer investigated…" — not passive constructions. Where you can estimate from the volume of captured activity, note the approximate time proportion: "(most of the hour)", "(~15 min)", "(brief)".
+Write with NO subject/narrator — never "the developer did X" or "they did X" (third-party voice) and never passive constructions. Where you can estimate from the volume of captured activity, note the approximate time proportion: "(most of the hour)", "(~15 min)", "(brief)".
 
 Include all work areas — coding, debugging, research, planning, reading docs, leisure. Do not filter anything out.
 
@@ -41,7 +41,6 @@ List documentation pages, repos, articles, videos, dashboards, or other material
 
 RULES
 - CONSOLIDATE everything into ONE story. Coding-agent sessions are NOT a separate task or section — weave them into the same work threads as the screen activity. If a coding-agent session and the screen capture describe the same work (same files, same feature, same ticket), MERGE them into a single thread; do not double-count or list "coding agent work" on its own.
-- PRESERVE concrete identifiers verbatim — ticket keys (e.g. KAN-241), file paths, function names, PR numbers, branch-independent specifics. A downstream matcher needs these to bind the work to a ticket; never genericize or drop them.
 - Infer the PURPOSE, not just the activity. If the screen shows edits to a prompt file + model test runs, say what the developer was trying to improve and why — not just "edited prompt file and ran tests."
 - Extract identifiable specifics: system names, service names, model names, tool names — anything that helps a matcher connect this to a ticket.
 - Do not make up facts, numbers, or names not present in the input.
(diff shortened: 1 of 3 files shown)


Akarsh-Hegde:
## Code Review

### PR #405: fix(worklog): local-time inline markers in session distill

**✅ Looks good overall.** The UTC-to-local conversion fix is correct — attaching the started_at date to the marker before passing it through `utc_to_local_hhmm` is the right fix. Version bump to 1.69.2 is correct per the runtime-versioning convention.

**Questions:**

1. **Removing the "PRESERVE concrete identifiers" rule** — this is a significant change to the prompt. You're removing the rule that tells the classifier to preserve ticket keys, PR numbers, function names verbatim. The note says downstream matchers don't need this, but can you confirm the worklog matcher and any other downstream consumers won't be affected? If the matcher relied on finding these identifiers in the report text, this could silently break ticket binding.

2. **TLDR high-level rewrite** — makes sense to shift toward outcome-focused descriptions ("worked on auth refactor" vs "edited app.ts line 42"), but watch for outputs that now lack enough detail for the Core Tasks matcher to link to a ticket. Is there a risk that being *too* high-level in the TLDR leaves the matcher without enough hooks?

**Minor:** The comment in `session_distiller.py` references `[[project_worklog_clock_trigger]]` — is this a memory link to an existing memory file? If so, the reference is correct; if not, consider clarifying the UTC-vs-local bug context inline.

Akarsh-Hegde:
## Review — local-time inline markers + plain-language TLDR

**Summary:** The core timezone fix is correct and, as a bonus, resolves a second latent sort-ordering bug; the prompt rewrite is reasonable but silently drops a matcher-critical rule that isn't in the PR description.

None. The timezone conversion is sound:
- `session_distiller.py:282-284` — reattaching `started_at`'s UTC date to the bare `[HH:MM:SS]` marker and routing it through `utc_to_local_hhmm` is right. The marker is UTC-only, `started_at` is a UTC ISO string, and the helper stamps `tzinfo=utc` before `.astimezone(LOCAL_TZ)` (aware conversion, no naive/aware mix). The `[11:16]` fallback keeps parse failures non-fatal.
- **Bonus fix worth calling out:** previously the initial `cur_time` came from `utc_to_local_hhmm(started_at)` (local) while marker lines used raw UTC `[:5]`, so `spans.sort(key=lambda s: (s["t"], …))` (line 302) sorted a **mix of local and UTC** times across sessions. After this change all `cur_time` values are local, so intra-hour ordering is now consistent.

### Suggestions / nits
- **`activity_report.py:44` — removing the PRESERVE-identifiers rule is an unadvertised, matcher-affecting change.** The deleted rule ("PRESERVE concrete identifiers verbatim — ticket keys e.g. KAN-241, file paths, function names, PR numbers … never genericize or drop them") was a *global* rule, not TLDR-scoped, and isn't in the PR's summary. The surviving line 46 ("Extract identifiable specifics: system/service/model/tool names") does **not** cover ticket keys / PR numbers / file paths, nor "verbatim / never drop." Since the downstream task-matcher binds work to tickets via those exact keys, dropping this risks the model genericizing KAN-xxx / PR# / paths out of Core Tasks — a real quality regression. If the intent was only to keep the *TLDR* high-level (already achieved by the line 20 rewrite), consider restoring the rule or re-scoping it to Core Tasks rather than deleting it.
- **DST + UTC-midnight edge (minor):** reusing `started_at`'s date for a marker that crosses UTC midnight assigns the wrong calendar date. Harmless for fixed-offset zones (IST, the tested case) since HH:MM is date-independent; could mislabel by ±1h only if a single session spans both UTC midnight *and* a DST transition. A one-line comment would suffice.
- **Voice-rule tension (nit):** `activity_report.py:30` now mandates "NO subject/narrator … never 'the developer did X'", but lines 20/25-28 still read "what the developer focused on / accomplished". Align the example phrasings so the model isn't given mixed signals.
- **Version bump present and sufficient:** `pyproject.toml` 1.69.1 → 1.69.2. Base is `pre-main`, live `runtime-staging` is 1.69.1, so strictly-greater — the auto-publish gate will ship it to staging. (`runtime-latest` is 1.68.0; if later promoted to `main`, reconcile above `runtime-latest` per the lockstep convention.) Module docstrings present in both `.py` files.

**Verdict:** Approve with one nit to reconsider — confirm the deletion of the "PRESERVE identifiers verbatim" rule (`activity_report.py:44`) is intentional and won't degrade downstream ticket matching before merge.

PR #406, opened 2026-07-08
feat(timeline): show only TLDR inline for solo-mode hours

adityaharishch:
## Summary
- In the Timeline UI, an hour with no PM/tracker context (solo mode) rendered the FULL activity report (TLDR + Core Tasks & Projects + Key Decisions + Resources Consulted) inline in the timeline row — crowding the row and duplicating what's already shown in the right-side hour-detail panel.
- Now the row shows just the TLDR paragraph (`ActivityReport.extractTldr`, pulled from the report markdown via the `### TLDR` heading). Clicking the row (already wired to `onSelectHour`) opens the full report — unchanged — in the right-side `HourDetailPanel`.
- No behavior change for connected (non-solo) users — they never saw the raw activity report inline to begin with (they see ticket-linked `TimelineCard`s).

## Test plan
- [x] `tsc --noEmit` clean on the touched files.
- [x] Pre-push suite (fmt, clippy, ui build/tests, cargo test, security audit) passed.
- [ ] Manual: open the dashboard in solo mode, confirm each generated-report hour shows only the TLDR line, and clicking it opens the full report on the right.

### DIFF
--- a/ui/components/timeline/ActivityReport.tsx
+++ b/ui/components/timeline/ActivityReport.tsx
@@ -13,6 +13,18 @@
 import ReactMarkdown from 'react-markdown'
 import type { Components } from 'react-markdown'
 
+/**
+ * Pull just the "### TLDR" paragraph out of an activity-report markdown
+ * string — used by the timeline row, which shows only the TLDR inline and
+ * defers the full report (Core Tasks/Decisions/Resources) to the right-side
+ * hour-detail panel on click. Falls back to the whole report if no TLDR
+ * heading is found (defensive — the prompt always emits one).
+ */
+export function extractTldr(report: string): string {
+  const match = report.match(/###\s*TLDR\s*\n+([\s\S]*?)(?=\n###|\s*$)/i)
+  return match ? match[1].trim() : report.trim()
+}
+
 export function ActivityReport({ report, compact = false }: { report: string; compact?: boolean }) {
   const bodyClass = compact ? 'mt-body-sm' : 'mt-body'
   const components: Components = {
--- a/ui/components/timeline/ActivityReport.tsx
+++ b/ui/components/timeline/ActivityReport.tsx
@@ -25,6 +25,40 @@ export function extractTldr(report: string): string {
   return match ? match[1].trim() : report.trim()
 }
 
+/**
+ * The timeline row's inline preview of an hour's report — a small tinted
+ * card (not raw ReactMarkdown output dropped straight into the row) with a
+ * "SUMMARY" label, a 2-line-clamped TLDR, and a hover affordance hinting
+ * that the row opens the full report (Core Tasks/Decisions/Resources) in
+ * the right panel.
+ */
+export function TldrPreview({ report }: { report: string }) {
+  const tldr = extractTldr(report)
+  return (
+    <div className="group rounded-md px-3 py-2.5 bg-box transition-colors">
+      <div className="flex items-center justify-between gap-2 mb-1">
+        <span className="mt-label" style={{ color: 'var(--color-state-proposal)' }}>✦ Summary</span>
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
## Code Review

### PR #406: feat(timeline): show only TLDR inline for solo-mode hours

**✅ Nice implementation.** The new components are clean, well-documented, and use the right design tokens. The UX improvement (preview + full report on click) is solid.

**Feedback:**

1. **Test plan incomplete** — the test plan checkbox for manual testing is unchecked (). Before merge, can you confirm you've actually opened the dashboard in solo mode and verified:
   - Each hour shows only the TLDR, not the full report
   - The TLDR text clamps to 2 lines correctly
   - Clicking an hour opens the full report (Core Tasks/Decisions/Resources) on the right
   
   This is especially important since there's no automated test coverage for this UI change.

2. **No automated tests** — the PR adds a new exported function (`extractTldr`) and component (`TldrPreview`) but no unit tests. Consider adding a quick test for:
   - `extractTldr(report)` with various markdown structures (report with TLDR, without TLDR, empty)
   - Defensive fallback behavior when TLDR heading is missing

3. **Design token consistency** — the `TldrPreview` mixes Tailwind classes (`rounded-md px-3 py-2.5 bg-box`) with inline styles (the WebkitLineClamp/WebkitBoxOrient section). This is fine, but is there an existing text-clamping utility in the codebase you could reuse instead of the inline style? (Just asking if there's a precedent; if not, the inline style is reasonable.)

4. **Styling edge case** — when the TLDR is very short (1 line or less), the 2-line clamp + the hint "Full report ›" on hover might create odd visual balance. Worth a quick visual check in the dashboard — does it look reasonable with both short and long TLDRs?

Akarsh-Hegde:
## Review — show only TLDR inline for solo-mode hours

**Summary:** A tight, well-documented UI change that swaps the full inline `ActivityReport` for a compact TLDR preview card in solo-mode timeline rows; correct and low-risk, with only minor cosmetic edge cases in the fallback path.

None. The render guard `isSolo && report ?` (`TimelineColumn.tsx:178`) excludes empty/null reports before `TldrPreview` is reached, so `extractTldr` never runs on `null`/`""`. The full report still renders on click via the unchanged `HourDetailPanel`, and connected users are unaffected.

### Suggestions / nits
- **`ActivityReport.tsx:20` — fallback dumps raw markdown as plaintext.** When `extractTldr` finds no `### TLDR` heading it returns `report.trim()`, rendered inside a plain `<p>` (line-clamped to 2). An inline heading like `### TLDR: short note` fails the `\s*\n+` match and falls back to the *entire* multi-section report, so the row shows `### TLDR: short note … ### Core …` as literal text. Acceptable given the prompt contract, but stripping `#`/`*` markers (or reusing `extractTldr` to drop trailing sections) would degrade more gracefully.
- **`ActivityReport.tsx:36` — whitespace-only report renders an empty card.** A truthy-but-all-whitespace report passes the guard yet yields `tldr === ""`, producing an empty "✦ Summary" card. Very unlikely from the LLM; gate on `report.trim()` in the parent if you care.
- **Nit:** `extractTldr` is exported but only consumed internally by `TldrPreview` in the same module — fine (aids testing), just flagging.

### Strengths
- Excellent doc-comments explaining the *why* (row = TLDR, panel = full report); module header stays accurate.
- The regex correctly stops at the next `### ` section and trims trailing whitespace (verified against multi-section input).
- No new `any`, both files keep the mandatory header on line 1, CSS tokens resolve, clickable-row/keyboard affordance untouched.

**Verdict:** Approve. No blocking issues; the two nits are cosmetic fallback edge cases guarded by the prompt contract. (Ideally after the checkbox-3 manual solo-mode verification in the PR's own test plan.)

PR #409, opened 2026-07-08
feat(coding-agents): widen get_coding_agents to all 4 ingested tools

adityaharishch:
## Summary
- `CODING_AGENTS` only listed Claude Code and Codex, silently dropping GitHub Copilot and Cursor Agent sessions from `get_coding_agents`
- Widens to all 4 ingested tools per CLAUDE.md's coding-agent pipeline table; Antigravity stays excluded (detection-only, never ingests)

Stacked on #(fix/dev-start-orphan-cleanup) — diff will shrink to just this commit once that merges.

## Test plan
- [x] `cargo build -p meridian-core`

### DIFF
--- a/meridian-core/src/readers/coding_agents.rs
+++ b/meridian-core/src/readers/coding_agents.rs
@@ -9,8 +9,12 @@
 //! way `today`/the session-interval math does), so neither do we — byte-identical.
 //!
 //! # Who calls this
-//! The tray `get_coding_agents` command → (currently no dashboard consumer; Today
-//! already surfaces agent time via `agent_s`/`agent_segments`). Ported for parity.
+//! The tray `get_coding_agents` command → `ui/components/timeline/OverviewPanel.tsx`
+//! (polled every 30s), which folds the per-agent breakdown into the Time-by-app
+//! chart (`TimeByApp.tsx`'s `appTotals`) — those sessions are excluded from
+//! `get_today`'s `sessions` (a separate overlay stream), so this is the only
+//! source of a real per-tool split. `today`'s `agent_s`/`agent_segments` remain
+//! the combined-across-all-agents total used elsewhere (Focus/Coding stats).
 //!
 //! # Related
 //! - [`crate::intervals::union_seconds`] does the overlap-dedup math.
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
## Review — widen `get_coding_agents` to all 4 ingested tools

**Summary:** A correct, minimal widening — the two edited constants match the ingest table exactly and stay consistent with each other; the only real gap is missing test coverage for the two newly-included agents.

None. Cross-checking the four `app_name` values against CLAUDE.md's "Ingested agents" table:
- `Claude Code`, `Codex`, `GitHub Copilot` (CLI + VS Code chat share it), `Cursor Agent` (Cursor IDE + cursor-agent CLI share it) — all match what the ingest pipeline actually writes. Antigravity correctly excluded (detection-only).
- The SQL `IN (...)` list (`coding_agents.rs:81`) and the `CODING_AGENTS` array (`coding_agents.rs:34`) list the identical four names, so `total_s` and the per-agent breakdown can't disagree. File header, `#[tracing::instrument]`, span, and `info!` summary all intact.

### Suggestions / nits
- **Missing test coverage for the new agents (main actionable item).** `meridian-core/tests/readers.rs:83` (`coding_agents_unions_overlap_per_agent_and_total`) still only seeds `Claude Code` + `Codex` and asserts `agents.len() == 2`. Per the port playbook, add a `GitHub Copilot` and a `Cursor Agent` row so the test exercises exactly the behavior this PR adds — it would guard against the regression being fixed (agents silently dropped) and against future SQL/const drift. As written, `cargo test` still passes even if one of the two new names were fat-fingered in only one of the two locations.
- **Duplicated agent list, two sources of truth** (`:34` const vs `:81` inline SQL). They agree today, but a future one-sided edit would split total-vs-per-agent silently. sqlx can't bind an array cleanly, so the hardcoded `IN` is defensible — but a comment tying the two together, or building the `IN` placeholders from `CODING_AGENTS.len()` and binding the slice, would remove the trap.
- **Doc framing nit:** the module header still opens with "a faithful port … byte-identical" (`:2-9`), but the agent set is now a deliberate divergence from the original route's two-element list. A one-line note in the `CODING_AGENTS` doc-comment would keep the header honest.

**Verdict:** Approve once a test seeding a `GitHub Copilot` and a `Cursor Agent` row is added — the production change itself is correct and ready.

PR #410, opened 2026-07-08
feat(tray): enable zoom hotkeys on dashboard/setup windows

adityaharishch:
## Summary
- Cmd+/Cmd- did nothing in the dashboard/setup webviews; enable `zoomHotkeysEnabled` on both config-defined and runtime-built windows, plus the matching `core:webview:allow-set-webview-zoom` permission
- Also scopes the webview asset protocol to `$HOME/.meridian/icon-cache/*`, laying the groundwork for the per-app icon extraction PR (narrowly scoped, not a general filesystem read)

## Test plan
- [x] `cargo build -p meridian-tray`, `cargo clippy -p meridian-tray -- -D warnings`

### DIFF
--- a/tray/src-tauri/capabilities/default.json
+++ b/tray/src-tauri/capabilities/default.json
@@ -6,6 +6,7 @@
     "core:default",
     "core:window:allow-close",
     "core:window:allow-set-size",
+    "core:webview:allow-set-webview-zoom",
     "positioner:default",
     "notifications:default",
     "opener:default",
--- a/tray/src-tauri/src/commands/system.rs
+++ b/tray/src-tauri/src/commands/system.rs
@@ -48,6 +48,7 @@ pub async fn open_dashboard(app: tauri::AppHandle) -> Result<(), String> {
         .minimizable(true)
         .closable(true)
         .maximized(true)
+        .zoom_hotkeys_enabled(true)
         .build()
     {
         Ok(win) => {
--- a/tray/src-tauri/src/tray.rs
+++ b/tray/src-tauri/src/tray.rs
@@ -103,6 +103,7 @@ pub(crate) fn open_native_dashboard(app: &tauri::AppHandle) {
             .minimizable(true)
             .closable(true)
             .maximized(true)
+            .zoom_hotkeys_enabled(true)
             .build()
         {
             Ok(win) => {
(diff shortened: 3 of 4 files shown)


Akarsh-Hegde:
## Review — enable zoom hotkeys on dashboard/setup windows

**Summary:** Enables native zoom hotkeys on the dashboard/setup webviews (plus an unrelated asset-protocol scope); wiring is correct and window coverage is complete, but the PR folds in scope-creep and its test plan never actually exercises the feature.

None. The wiring is sound: both runtime dashboard builders (`system.rs:51` `open_dashboard` and `tray.rs:106` `open_native_dashboard`, label `"dashboard"`) and the setup builder (`tray.rs:176` `open_wizard_window`, label `"setup"`) get `.zoom_hotkeys_enabled(true)`, and the `default` capability already scopes `["main","dashboard","setup","tray-tooltip"]` (`capabilities/default.json:4`), so the new `core:webview:allow-set-webview-zoom` permission reaches those windows. No missing window label.

### Suggestions / nits
- **Scope creep — the `assetProtocol` block is unrelated to zoom** (`tauri.conf.json:16-19`). It's "groundwork for the per-app icon PR" (#411), has no consumer in this diff, and would review far more naturally alongside the code that reads `icon-cache/`. Recommend moving it there so this PR stays single-purpose. If it stays: the glob `$HOME/.meridian/icon-cache/*` is single-level — use `**` if the cache is ever nested.
- **`allow-set-webview-zoom` may be redundant for hotkeys.** `zoomHotkeysEnabled` drives the platform-native Cmd+/Cmd-/Cmd0 handlers; the `core:webview:allow-set-webview-zoom` ACL gates the *JS* `setZoom()` path, which nothing in this diff calls. Harmless to keep, but if you're not doing programmatic zoom from JS it's an unused grant.
- **`zoomHotkeysEnabled: true` on `main` and `tray-tooltip`** (`tauri.conf.json:39`, `:58`) is pointless — those are the fixed-size, non-resizable popover/tooltip. Not harmful; consider dropping them to keep intent clear.
- **Test plan only compiles it.** `cargo build`/`clippy` prove it links, not that Cmd+/Cmd- actually zoom the dashboard/setup webviews (exactly the class of thing that silently no-ops on a permission/window-flag mismatch). A one-line manual confirmation before merge would be worth it.

**Convention check:** clean — both `.rs` files keep the line-1 header, JSON is exempt, no new command so the tracing requirement doesn't apply.

**Verdict:** Approve with nits. No correctness defects. Please split out the `assetProtocol` change (or defer to the icon PR) and do a quick runtime check that the hotkeys actually zoom.

PR #411, opened 2026-07-08
feat(dashboard): extract real per-app icons for Time-by-app

adityaharishch:
## Summary
- New `get_app_icon` Tauri command resolves an installed `.app` bundle via NSWorkspace, extracts its NSImage, caches a PNG under `~/.meridian/icon-cache/`, served to the webview through the asset protocol (`bridge.ts`'s `assetUrl`, `ui/lib/app-icons.ts`'s `useAppIconUrl`)
- Adds a curated `BRAND_ICONS` wordmark table (via `simple-icons`) as a mid-tier fallback between the real icon and the letter monogram
- `AppGlyph` now renders: real icon → brand wordmark → monogram, in that priority order

## Test plan
- [x] `cargo build -p meridian-tray`, `cargo clippy -- -D warnings`
- [x] `npm run build` in `ui/`


Akarsh-Hegde:
## Review — extract real per-app icons for Time-by-app

**Summary:** A cleanly-scoped, well-documented feature that swaps letter monograms for real `.app`-bundle icons (with a vendored brand-wordmark mid-tier fallback); no blocking issues found.

None. I checked the highest-risk item — the asset protocol. The `` relies on `convertFileSrc` + Tauri's asset protocol, which needs `security.assetProtocol.enable` + a `scope` covering the cache dir. This diff doesn't touch `tauri.conf.json`, but its **base branch already sets** `assetProtocol.enable: true` with `scope: ["$HOME/.meridian/icon-cache/*"]`, and the cache writes flat files (`icon-cache/<key>.png`), so the single-level `*` glob matches. Custom `#[tauri::command]`s need no capability entry in Tauri v2, so registering `get_app_icon` in `invoke_handler!` (`lib.rs:494`) is complete.

### Suggestions / nits
- **`simple-icons` is a runtime `dependency` but never imported** (`ui/package.json`). The path data in `ui/lib/brand-icons.ts` is hand-vendored (no `import ... from 'simple-icons'`). It's a build-time reference source only — belongs in `devDependencies` (or drop it). As-is it bloats installs without shipping anything. **This is the one I'd fix before merge.**
- **Disk cache never invalidates** (`app_icons.rs:241`). Once `~/.meridian/icon-cache/<key>.png` exists it's served forever — a stale icon survives an app update/reinstall (no TTL or bundle-mtime check). Acceptable for v1; worth a comment noting the cache is permanent.
- **`cache_key` collisions** (`app_icons.rs:175`): lowercasing + mapping every non-alphanumeric to `_` means distinct `app_name`s can collapse to the same key (`"Foo Bar"` vs `"Foo-Bar"`) and get served the wrong cached icon. Low probability, but combined with the permanent cache it's a sticky wrong-icon — a short hash suffix removes the risk.
- **Sync command + AppKit main-thread affinity** (`app_icons.rs:222`): keeping `get_app_icon` synchronous is *correct* — Tauri v2 runs non-`async` commands on the main thread, where `NSWorkspace`/`NSImage` want to be. Worth a one-line comment saying so, because the instinct to make the blocking fs + PNG-encode `async` would move the raw `unsafe` AppKit calls off the main thread. Footgun for the next editor, not a bug today.
- **`.map_err(|e| e.to_string())`** (`app_icons.rs:238,257`) drops the `anyhow`-context style the repo prefers; minor given how shallow these are.

### Strengths
- Frontend dedup is right: a module-level `Map` + in-flight-promise cache in `ui/lib/app-icons.ts` means **one `invoke` per distinct app name for the whole session**, not per render/instance.
- Three-tier fallback (real icon → brand wordmark → monogram) degrades gracefully; `useAppIconUrl` returns `null` off-Tauri / on failure so the existing glyph always renders; `cancelled` flag handles unmount/appName-change cleanly.
- `iconForFile:` only called after `resolve_bundle_path` confirms the bundle exists, avoiding NSWorkspace's generic-icon trap. File headers on line 1, `#[tracing::instrument]` + debug/warn, per-icon `viewBox` stored, license/trademark provenance documented, hermetic unit tests for `resolve_bundle_path_in`/`cache_key`.

**Verdict:** Approve with minor nits. No correctness blockers; the `simple-icons` dependency placement is the one to fix before merge, the rest are follow-up comments.

PR #414, opened 2026-07-08
feat(setup): expected-memory panel + HF caching-proxy for model downloads

Akarsh-Hegde:
## Summary

Two independent, setup-phase improvements (plan: `~/.claude/plans/spicy-splashing-pretzel.md`).

### Part 1 — Setup step shows status + expected memory (UI-only)
The "Local intelligence" step now surfaces detected hardware (chip · macOS · unified RAM) and the engine's **expected resident footprint** against the Mac's memory, so users see what it costs before finishing setup. Reuses the existing `Bar` atom; degrades to a bare estimate line when `detect_system_specs` is unavailable (non-macOS).

- `ui/app/setup/steps.tsx` — new `SpecMemoryPanel` in `MLXBody` (shown while provisioning and after ready; hidden on error/`unavailable`).

**Visual preview (all states + light/dark):** https://claude.ai/code/artifact/1f614ca7-de30-4b2e-87ec-b0d8ccd97310

### Part 2 — Model downloads via a Cloudflare Worker HF caching proxy
Users retry stalled model pulls from HF. This adds an **optional** edge cache in front of `huggingface.co` — HF stays the source of truth (no self-hosting/re-upload); we just cache the weight blobs at Cloudflare's edge. **No Python / runtime republish** (huggingface_hub honours `HF_ENDPOINT` from the env, so every runtime version routes through it automatically).

- `infra/hf-proxy/` — new Cloudflare Worker (`wrangler.toml` + `src/index.js`): transparent reverse-proxy, full weight GETs edge-cached via `cf.cacheEverything` + a **stable `cacheKey`** (signed CDN query stripped) so Cache Reserve fills; metadata/HEAD/resume pass through; **pass-through to HF origin on any error**.
- `tray/src-tauri/src/mlx_server.rs` — `hf_endpoint()` resolver (debug env → `option_env!` → baked default) + sets `HF_ENDPOINT` on the in-process setup server.
- `services/scripts/{com.meridiona.mlx-server.plist, install-mlx-server-daemon.sh}` — launchd (dev/bundle) parity: injects a **whole `HF_ENDPOINT` key/value block or nothing** — never an empty value (which huggingface_hub would take literally and break downloads).

**Ships DISABLED by default** (`HF_PROXY_ENDPOINT = ""`) — baking a live default before the Worker exists would point every download at a dead host. Enable per channel via `option_env!("MERIDIAN_HF_ENDPOINT")` once the Worker + Cache Reserve are live.

## Notes / deviations from plan
- Plist wiring is in `install-mlx-server-daemon.sh` (the actual renderer), **not** `backend_install.rs` — the DMG supervises MLX in-process via `start()` (patched) and *removes* the legacy launchd agent.
- Default is disabled, not the live `hf.meridiona.com` (safe-by-default).

## Requires before enabling (infra)
Cloudflare zone + Worker route + **Cache Reserve** (R2-backed; the default edge cache caps objects at 512 MB, below the ~1.4 GB llm). Then bake `MERIDIAN_HF_ENDPOINT` per channel.

## Test
- Rust: `cargo fmt` / `clippy -D warnings` clean, 46/46 tray tests pass.
- UI: build ✓, `bun test` 177/177 ✓, `tsc` clean for changed files.
- Worker: `wrangler deploy --dry-run` bundles clean.
- Plist lints (`plutil`) in both empty and set states.

**Not done (needs infra/running app):** staging Worker deploy + live wizard smoke against an empty HF cache to confirm edge-cache HITs.

## Sequencing (confirmed)

Prod runtime `1.68.0` **already includes** the #350 `hf_xet` fix (it landed at `1.63.0`), so the retries are **not** the Xet hang and **no runtime republish is needed** — this proxy is the actual durable fix. Verified via the live `runtime-latest` manifest vs the fix commit.


adityaharishch (commented):
## Review: feat(setup): expected-memory panel + HF caching-proxy for model downloads

Two independent, low-risk changes. Overall solid — good docs, safe-by-default rollout (`HF_PROXY_ENDPOINT = ""`), and the new `hf_endpoint()` resolver correctly mirrors the existing `manifest_url()` debug/compile-time/default fallback pattern.

### Part 1 — Setup memory panel (UI-only)
Straightforward. `SpecMemoryPanel` degrades sensibly when `specs` is null, and correctly hides on error/`unavailable` states. No issues.

### Part 2 — HF caching proxy

**Potential security concern — cache key strips auth, `cacheEverything` is shared across all users**
In `infra/hf-proxy/src/index.js`, `isFullWeightGet` requests are cached with:
```js
cacheKey: url.origin + url.pathname,
```
This deliberately strips the signed CDN query (correct, that's needed for cache hits), but it also means the cache entry is keyed **only on path**, not on the `Authorization` header. If any user ever pulls a gated/private HF repo through this proxy (the code explicitly preserves `Authorization` for `hf auth login` tokens — see `forwardHeaders`), the first authorized response gets cached at Cloudflare's edge and will be served to any subsequent **unauthenticated** requester hitting the same path. For Meridian's currently-public `mlx-community` weights this is low risk in practice, but since the code already anticipates authenticated requests, it's worth either:
- scoping the cache to responses that don't require auth (e.g. skip `cacheEverything` when an `Authorization` header is present on the request), or
- explicitly documenting that this proxy must never be used for gated/private repos.

Worth a one-line comment or guard so a future model added to the pipeline that happens to be gated doesn't silently leak.

**Minor — no test coverage for the new resolver**
`hf_endpoint()` (mlx_server.rs) has the same three-tier fallback complexity as `manifest_url()`, but neither has unit tests (existing gap, not introduced by this PR — not blocking, just flagging since this PR is a good opportunity to add one for the debug/compile-time/default precedence).

**Nit — plist `sed` injection pattern**
`install-mlx-server-daemon.sh` interpolates `${MERIDIAN_HF_ENDPOINT}` into `HF_ENDPOINT_ENTRY` and then substitutes it via `sed -e "s|...|${HF_ENDPOINT_ENTRY}|g"`. If the URL ever contained a literal `&`, sed would insert the matched pattern instead of the literal value (a `&` in a sed replacement means "matched text"). This is a pre-existing pattern already used for `MERIDIAN_OO_AUTH`/`MERIDIAN_OTLP_ENDPOINT` in the same script, so not new, and low risk since the value is a build-time/channel-baked URL rather than untrusted input — just noting it inherits the same latent footgun.

### Verified
- Plist XML stays valid when `{{HF_ENDPOINT_ENTRY}}` substitutes to an empty string (leaves a blank line inside `<dict>`) — confirmed with `plutil -lint` locally.
- `hf_endpoint()` is only consumed at the single process-spawn site (`start()`), and `supervise()` always restarts via `start()`, so there's no risk of a stale/inconsistent `HF_ENDPOINT` across restart paths.
- Worker's fail-open design (pass-through to HF origin on any internal error) is good — can't make downloads worse than today.

### Everything else
- Doc comments are thorough and follow the file's existing `# Who calls this` / `# Related` conventions where applicable.
- Ships disabled by default — appropriately cautious given the Worker/Cache Reserve infra isn't live yet.

No blocking issues. The auth/cache-key point is worth a decision (even if "we'll never proxy gated repos, documented") before `MERIDIAN_HF_ENDPOINT` is baked for any channel.

Akarsh-Hegde:
Addressed all three points in [1de79528](https://github.com/Meridiona/meridian/pull/414/commits/1de79528):

**Security concern (cache/auth) — fixed.** `isFullWeightGet` now also requires no `Authorization` header; an authenticated `resolve` GET falls through to the plain pass-through path instead of `cacheEverything`, so a gated/private-repo response can never be written into the shared edge cache. Added an explicit `SECURITY:` comment block at the top of the file documenting the invariant (public repos only) so a future model addition doesn't silently violate it.

**Test coverage — added `hf_endpoint_disabled_by_default`.** Mirrors the "no override at any tier → `None`" invariant this PR's safety story depends on. Scoped to only clear `MERIDIAN_HF_ENDPOINT` (a var no other test in the suite touches), so it can't race under parallel `cargo test` execution without pulling in `serial_test` — consistent with why `manifest_url()` itself isn't unit-tested the same way (its env-tiered behavior is only exercised end-to-end by the `#[ignore]`d `live_pull_from_runtime_staging`).

**Sed nit — fixed, and found a second related bug while verifying it.** Escaped sed replacement metacharacters (`\`, `&`, our `|` delimiter) in `HF_ENDPOINT_ENTRY` per your note. While testing that fix against an adversarial URL containing `&`, I found the injection was *also* broken at the XML level — a literal `&` in a URL (e.g. a query string) produced invalid XML (`plutil` rejected it: "unknown ampersand-escape sequence"), independent of the sed issue. Fixed both: XML-escape the URL before embedding it in the `<string>` element, then sed-escape the resulting entry before the outer substitution. Verified round-trip with `plutil -lint` against `&`, `|`, `\`, a clean URL, and the empty/disabled case — all lint OK and extract back to the exact input.

All pre-push checks green on the updated branch (fmt, clippy, `cargo test` incl. the new test, UI build, `bun test`, security audit).

PR #420, opened 2026-07-09
fix(setup): polish first-run wizard copy, permissions, and window chrome

Akarsh-Hegde:
## Summary
- Fix the setup window title's em-dash and codify a project-wide rule (CLAUDE.md): all user-facing app text uses a single hyphen, never an em/en-dash.
- Drop **Input Monitoring** from the Permissions step — it's redundant with Accessibility for every signal the daemon actually consumes (`clipboard` + `app_switch` via `get_signals`), and on this hardware it never even produced a real System Settings grant. The capture recorder no longer prompts for it (avoids a cold, wizard-less TCC dialog); it now runs a documented full/reduced capture mode depending on whether it's already granted. Removes the now-orphaned `check_input_monitoring` / `request_input_monitoring` commands.
- Rewrite the Welcome and Integrations copy toward a "see what actually happened" positioning, fix the stale "Jira and Trello today" tracker list (now lists all 5: Jira, Linear, GitHub, Trello, Azure DevOps), and make explicit that connecting a tracker is optional — Meridian still tracks the day without one (the step was already non-blocking; this just makes it legible).
- Make the setup window resizable/full-screenable (native macOS full-screen, not just zoom), scale the wizard card with the window instead of floating small on a big backdrop, and switch to a transparent title bar so the centred card gets symmetric margins on all sides.

## Test plan
- [x] `cargo fmt --check` + `cargo clippy -- -D warnings` clean (root `meridian` package and `tray/src-tauri` both, default + `--no-default-features --features capture`)
- [x] `tsc --noEmit` clean on the setup wizard files
- [x] Manually exercised the wizard in `tauri dev`: title bar, full-screen toggle, 2-permission Permissions step (Accessibility + Screen Recording, gate no longer waits on Input Monitoring), Welcome copy, Integrations optional-skip copy
- [x] Full pre-push suite passed: fmt, ui build, clippy, ui tests, security audit, `cargo test`

https://claude.ai/code/session_01RJ8rU48qbY1A3tthxFkQj8

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,7 @@ Meridian is a single-process Rust daemon that normalises raw screen-capture fram
 - NEVER merge a PR automatically — open/update PRs as needed, but leave the actual merge to a human reviewer
 - NEVER push directly to `main` — always create a separate feature branch, commit there, and raise a PR to `main`
 - ALWAYS use a separate branch per feature/fix — branch name format: `type/short-description` (e.g. `feat/trello-oauth`, `fix/ui-disconnect`)
+- In all **user-facing app text** — window titles, wizard/UI copy, button and menu labels, notification bodies, tray tooltips, any string the user reads — use a plain hyphen `-` only. NEVER an em-dash (`—`), en-dash (`–`), or double hyphen (`--`). Use it spaced (` - `) where a dash separates clauses. (This rule is about displayed strings; code comments and docs are exempt.)
 
 ---
 
--- a/tray/src-tauri/src/capture/ui_events.rs
+++ b/tray/src-tauri/src/capture/ui_events.rs
@@ -88,20 +88,28 @@ fn input_monitoring_granted() -> bool {
 }
 
 /// Run the input recorder until `tx` closes. **Blocking** — call on a dedicated
-/// OS thread. Requests Input Monitoring only when not already granted; degrades
-/// to a no-op recorder (emits nothing) when it's denied, rather than failing.
+/// OS thread. **Never prompts for Input Monitoring:** the setup wizard dropped it
+/// as a permission because the signals the daemon actually consumes
+/// (`clipboard` and `app_switch`, via `get_signals`) come from the
+/// Accessibility-only workspace observer and clipboard poller — see
+/// `screenpipe-a11y` `macos.rs`. When Input Monitoring isn't already granted the
+/// recorder runs in **reduced mode** (clipboard and app/window events, no
+/// click/key/text CGEventTap) rather than surfacing a cold, wizard-less TCC
+/// dialog; the click/key stream feeds only the minor Option-C `ended_at`
+/// refinement. `start()` degrades to a no-op recorder (emits nothing) if even
+/// reduced mode can't attach, rather than failing.
 pub(crate) fn run_ui_event_recorder(tx: UiEventTx) {
     let recorder = UiRecorder::new(recorder_config());
-    // Check Input Monitoring first; only call request_permissions() (which can
-    // prompt) when it isn't already granted, so a granted permission never
-    // re-triggers the system prompt on launch.
+    // Do NOT request Input Monitoring here — prompting would raise a cold TCC
+    // dialog the wizard no longer explains. Just log the current status for
+    // observability; `recorder.start()` selects full vs reduced mode itself.
     if input_monitoring_granted() {
-        info!("capture: input monitoring already granted — not prompting");
+        info!("capture: input monitoring granted — full ui-event capture (incl. click/key/text)");
     } else {
-        let perms = recorder.request_permissions();
-        if !perms.input_monitoring {
-            warn!("capture: input monitoring not granted — ui events unavailable until the user grants it");
-        }
+        info!(
+            "capture: input monitoring not granted — reduced mode \
+             (clipboard + app/window events; no click/key/text tap)"
+        );
     }
     let handle = match recorder.start() {
         Ok(h) => h,
--- a/tray/src-tauri/src/commands/setup.rs
+++ b/tray/src-tauri/src/commands/setup.rs
@@ -118,9 +118,8 @@ pub async fn check_screen_recording() -> bool {
 /// status read that never registers the app. On a fresh install this means the
 /// list under Privacy → Screen Recording shows "No Items", because macOS only
 /// adds an entry the *first* time the app calls `CGRequestScreenCaptureAccess`.
-/// This command calls that request variant (analogous to `request_input_monitoring`)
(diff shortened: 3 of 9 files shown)


Akarsh-Hegde (commented):
## Review

### Overview
Four independent changes bundled into one PR: (1) a new project-wide "no em/en-dash in user-facing text" rule in CLAUDE.md plus its application across the setup wizard copy, (2) dropping Input Monitoring as a wizard permission and its associated Tauri commands, (3) a `fmtModelLabel` helper (with tests) that fixes a hardcoded "2B · 2B" duplication bug in the Completion summary, and (4) making the setup window resizable/full-screenable with a scaling wizard card.

### Correctness
- **Input Monitoring removal is sound.** I traced the vendored `screenpipe-a11y` fork's `UiRecorder::start_internal` (pinned via `tray/src-tauri/Cargo.toml`): it calls `check_permissions()` (a non-prompting status read), not `request_permissions()`, and already branches into a documented "reduced mode" (clipboard poller + workspace observer, no `CGEventTap`) when Input Monitoring isn't granted. So the new `run_ui_event_recorder` — which no longer calls `recorder.request_permissions()` — genuinely won't trigger a surprise TCC prompt on launch, and the "reduced mode" claim in the doc comment is accurate, not aspirational.
- **`fmtModelLabel`** — walked through all six test cases by hand against the regex chain (org-prefix split, `-?\d+-?bit$` trim, `-OptiQ$` trim, hyphen→space); all match expected output, including the hyphenated-bit-width edge case (`Name-8-bit` → `Name · 8-bit`, where both surrounding hyphens get consumed by the single regex). Good, focused test coverage for a small pure function — no bugs found.
- **`make_fullscreenable`** in `tray.rs` follows the exact same `ns_window()` → null-check → cast → `msg_send!` idiom already used three times in `lib.rs` (`make_visible_over_fullscreen`, `init_as_nspanel`, etc.), so it's consistent with established patterns rather than introducing a new risky one. The `unsafe` block has a `# Safety` justification per convention.
- **Window scaling math** (`clamp(1, min(calc(100vw/1010), calc(100vh/700)), 1.7)`) checks out: floor of 1 keeps the card at design size at the 1000×680 minimum (where the raw ratio would actually be ~0.97, so the floor is load-bearing, not decorative), and the 1.7 cap keeps it from ballooning on a large display while `min()` of the two axis ratios guarantees it never overflows either dimension.
- **Minor PR-description nit:** the summary says "fix the setup window title's **double-hyphen**," but the actual character replaced was an em-dash (`—`), not `--`. Not a code issue, just a description mismatch worth a glance.

### Convention adherence
- New CLAUDE.md rule is applied consistently within the touched files — I diffed the actual post-patch file content and every remaining em-dash in `steps.tsx`/`data.ts`/`page.tsx` is inside a `//` or `{/* */}` comment (which the rule explicitly exempts). None leaked into rendered JSX strings.
- Scope note (non-blocking): ~64 other files under `ui/` still contain em-dashes. The PR only "codifies" the rule going forward rather than backfilling the repo, which is a reasonable scope for this PR, but worth a tracking follow-up if the intent is full compliance.
- Doc comments follow the repo's `# Who calls this`/rationale style well (e.g. the `ui_events.rs` module comment cross-references the exact daemon consumers by name).

### Risks / things I couldn't verify from the diff alone
- I confirmed the "reduced mode" behavior against the **upstream** `screenpipe/screenpipe` repo's `platform/macos.rs` (a local clone I had access to), not the exact pinned commit (`d6d2b318…`) in the Meridiona fork — I couldn't fetch that exact rev in the sandbox. The logic lines up closely enough that I'm confident in the claim, but it's worth a quick sanity check against the actual pinned commit if you haven't already manually exercised the "not granted" path (the PR's test plan says permissions were manually exercised, but doesn't explicitly call out testing the *denied* Input Monitoring case).
- I read the code rather than rebuilding; I'm relying on the PR's stated `cargo clippy -D warnings` / `tsc --noEmit` results rather than independently re-running them.

### Suggestions (non-blocking)
- The `ns_window()` → null-check → cast pattern is now duplicated a 4th time (3× in `lib.rs`, 1× in `tray.rs`). Could be worth extracting a small shared helper (e.g. `fn as_ns_window(win: &WebviewWindow) -> Option<*mut AnyObject>`) at some point, purely to cut the repetition — not something this PR needs to do.

### Summary
Clean, well-scoped PR — each of the four changes is independently justified and the riskier one (removing a permission check) is backed by tracing the actual fallback behavior in the vendored dependency rather than just asserting it. No blocking issues found.

Akarsh-Hegde:
Thanks for the thorough review. Addressed the actionable items:

**Fixed**
- PR description nit: "double-hyphen" → "em-dash" (the character actually replaced in the setup window title was `—`, not `--`).

**Verified the open risk item** — traced the "reduced mode" claim against the *exact* pinned commit (not just upstream `screenpipe/screenpipe`): `screenpipe-a11y` is pinned at `d6d2b318cfe1fffbc9e923e627e8659df1b2ff7e` (`tray/src-tauri/Cargo.toml:83`). At that commit:
- `UiRecorder::start_internal` (`platform/macos.rs:240`) only hard-requires `accessibility` — it `bail!`s if that's missing, but not for `input_monitoring`.
- When `input_monitoring` is false it logs `"input monitoring not granted — running in reduced mode: clipboard via change-count polling, app/window events via workspace observer, keystrokes and clicks disabled"` and spawns the clipboard poller instead of the `CGEventTap` thread (`platform/macos.rs:269-297`), gated on `config.capture_clipboard`.
- `UiCaptureConfig::new()` defaults `capture_clipboard: true` (`config.rs:192`), so our `recorder_config()` (which doesn't touch that field) genuinely gets the clipboard poller in reduced mode, not a silent no-op.

This confirms the doc comment on `run_ui_event_recorder` is accurate at the pinned rev, not just plausible against a similar-looking upstream. No code change needed here — this was a verification gap, now closed.

**Left as-is (per your own non-blocking call)**
- Extracting a shared `as_ns_window()` helper for the 4x-duplicated `ns_window()` → null-check → cast pattern — you flagged this as "not something this PR needs to do," agreed, leaving for a follow-up if/when a 5th call site shows up.
- Backfilling the ~64 other files still containing em-dashes — out of scope for a setup-wizard PR; happy to file a tracking issue if there's appetite for full repo compliance.

Also worth noting: this branch was rebased onto `pre-main` after `#418` (setup-notifications-card) landed and touched the same permissions code — merged both intents rather than picking one (Input Monitoring stays removed, the new optional Notifications card is preserved), plus a follow-up commit fixed an unrelated duplicated-"2B" bug in the Completion summary line (`fmtModelLabel`, with unit tests). Both are already reflected in the current diff.

PR #421, opened 2026-07-09
chore(ui): dashboard style cleanup — fonts, shadows, dead code (Phase 1)

Akarsh-Hegde:
## Summary

Phase 1 of the dashboard visual-cleanup plan (fonts, spacing, component consistency — the color half is separate: `STYLESHEET.md` §9, PR #419). Mechanical, low-risk fixes surfaced by a read-only audit of the dashboard, setup wizard, and tray popover. No component redesign — that's Phase 2 (a shared badge/pill primitive to replace ~16 hand-rolled implementations), tracked separately.

- **Fix `--font-serif`**: it resolved to Georgia while a real Instrument Serif font was loaded via `next/font` and only reachable through two hardcoded inline `style={{fontFamily: 'var(--font-instrument-serif)...'}}` call sites in the setup wizard. The token now points at the actual font (Georgia stays as the load-failure fallback); the wizard uses `var(--font-serif)` instead of duplicating the stack inline.
- **Delete the dead legacy `.type-*` CSS scale** (8 Georgia-based classes, zero real consumers) after migrating its one remaining user — `HygieneDialog`'s title — to `.mt-modal-title`, which `STYLESHEET.md` §1 already specifies for Review/Cleanup/Plan modal titles (this dialog *is* the board-cleanup dialog). Also drops the redundant, unused `.font-serif` utility rules that duplicated the same hardcoded Georgia stack outside the token system.
- **Delete 7 zero-consumer `atoms.tsx` exports** (`Card`, `ConfidenceRing`, `SegBar`, `LiveDot`, `CatLabel`, `fmtDurDecimal`, `useTick`) — real dead code (SVG rings, segmented bars), verified with zero internal or external references before removal. `atoms.tsx` goes from 325 → 222 lines.
- **Consolidate two duplicate shadow strings** into `--mt-card-shadow`/`--mt-modal-shadow` tokens, replacing the identical literal `boxShadow` strings hand-rolled in `CleanupCard`/`ReviewCard` and `ModalShell`/`TaskDetailDialog`.
- **Tray popover**: swap 6 raw-hex occurrences for the equivalent already-defined `--accent`/`--accent-2`/`--accent-3`/`--success`/`--ink-2` tokens — exact value matches only, zero visual change.

**Explicitly out of scope for this PR** (spacing/radius tokens, the pill/badge consolidation, tray's remaining ~30 raw-hex colors that don't have an exact existing-token match) — these need visual judgment calls, not just search-replace, and are tracked as Phase 2.

## Test plan

- [x] `cd ui && npm run build` — clean
- [x] `cd ui && bun test` — 243/243 pass
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings` / `cargo test` — clean (pre-push hook)
- [ ] Visual check recommended before merge: `HygieneDialog`'s title now renders in `.mt-modal-title` (Jakarta bold ~20px) instead of the old 36px Georgia serif — a real, deliberate visual change, not a no-op. The setup wizard's serif hero text should look identical (Instrument Serif was already the intended font there, just wired around the token instead of through it).


Akarsh-Hegde (commented):
## Review

### Overview
Mechanical Phase-1 style cleanup: fixes `--font-serif` to actually point at the loaded Instrument Serif font (was silently resolving to Georgia), deletes the dead legacy `.type-*` scale + `.font-serif` utility after migrating `HygieneDialog`'s one real consumer to `.mt-modal-title`, removes 7 zero-consumer `atoms.tsx` exports, consolidates two duplicate shadow literals into tokens, and swaps 6 raw hex colors in the tray popover for equivalent existing tokens.

### ⚠️ Cross-PR conflict with #423
I independently verified the "zero internal/external references" claim for the 7 deleted `atoms.tsx` exports (`Card`, `ConfidenceRing`, `SegBar`, `LiveDot`, `CatLabel`, `fmtDurDecimal`, `useTick`) against everything currently in `ui/` — it's accurate, nothing in the dashboard app uses them.

However: **all 7 of them are re-exported by `packages/meridian-design-system/src/index.ts`**, added in #423 (currently open, not yet merged). That package builds by importing directly from `ui/components/atoms.tsx` — if this PR merges first, #423's build breaks (`fmtDurDecimal`/`CatLabel`/`LiveDot`/`Card`/`ConfidenceRing`/`SegBar`/`useTick` no longer exist to import). If #423 merges first, this PR's "zero consumer" premise for those 7 exports becomes stale by the time it lands.

Not a defect in either PR individually — the two just didn't know about each other. Worth deciding merge order and updating the loser accordingly:
- If #421 lands first: I'll need to drop those 7 names from #423's `src/index.ts` barrel (they were exported there as "atoms" for the design-system bundle, not because the dashboard uses them — so dropping them from the design-system package instead of reverting the deletion is probably the smaller change).
- If #423 lands first: this PR would need to re-verify those exports are still genuinely dead *within `ui/`* (they are, per my check above) before deleting them — the design-system package importing them doesn't make them "used by the dashboard," it just means #421's description should note the dependency exists elsewhere.

Flagging this so whoever merges second doesn't get a surprise build failure.

### Correctness — verified independently
- **`--font-serif` fix**: confirmed no other consumer relies on the Tailwind-generated `.font-serif` utility class (only the deleted custom `.font-serif` rule and the two hardcoded `SERIF` consts in `setup/page.tsx`/`steps.tsx` referenced it) — the token retarget is safe and localized.
- **Dead `.type-*` scale**: grepped `ui/components` + `ui/app` for every one of the 8 class names — `HygieneDialog.tsx`'s `type-heading` was the only real consumer, and it's the one being migrated in this same PR. Confirmed dead elsewhere.
- **Shadow token consolidation**: `--mt-card-shadow` / `--mt-modal-shadow` values match the four replaced literal strings byte-for-byte (`0 20px 46px -14px rgba(20,16,40,0.34)` and `0 24px 60px -18px rgba(20,16,40,0.5)`) — genuinely zero visual change, not just "close enough."
- **Tray popover hex→token swaps**: checked all 5 tokens (`--ink-2`, `--accent`, `--accent-2`, `--accent-3`, `--success`) against `tray/src/style.css`'s `:root` block — every one is an exact value match, and the block is unconditional (no dark-mode variant to worry about missing, per the file's own header comment).
- **`atoms.tsx` import cleanup**: removing the `useEffect, useState` import alongside `useTick` is correct — that hook was their only consumer in the file.

### Risks / open items
- The PR's own test plan already flags the one real, deliberate visual change (`HygieneDialog` title: 36px Georgia serif → 20px Jakarta bold) as unchecked/"recommend visual check before merge" — agree this is worth eyeballing since it's a genuine appearance change, not a refactor.
- I verified statically rather than running `npm run build`/`bun test` myself; relying on the PR's stated clean results for the build/test/clippy steps.

### Suggestions (non-blocking)
- None beyond the merge-order conflict above — this is a clean, well-scoped mechanical PR with an unusually thorough description that made independent verification straightforward.

### Summary
No bugs found in the diff itself; all "verified dead code" / "exact token match" claims check out under independent inspection. The one thing worth acting on before merge is coordinating merge order (or a quick fix-up) with #423, since they touch the same 7 `atoms.tsx` exports in opposite directions.

adityaharishch (commented):
## Review

**What this does:** Phase 1 of the dashboard style cleanup — mechanical, low-risk fixes: rewires `--font-serif` to actually resolve to Instrument Serif (was silently falling back to Georgia while two setup-wizard call sites hardcoded the real font stack inline), deletes the dead `.type-*` legacy CSS scale (migrating its one live consumer, `HygieneDialog`, to `.mt-modal-title`), deletes 7 zero-consumer `atoms.tsx` exports, consolidates two duplicate shadow strings into `--mt-card-shadow`/`--mt-modal-shadow` tokens, and swaps 6 raw-hex tray-popover colors for existing equivalent tokens.

### ⚠️ Likely conflict with #423 (same author, both currently open)
This PR deletes 7 exports from `ui/components/atoms.tsx` as dead code (verified zero-consumers at the time): **`Card`, `ConfidenceRing`, `SegBar`, `LiveDot`, `CatLabel`, `fmtDurDecimal`, `useTick`**.

#423 (`feat/design-system-package`, opened ~4 hours after this one) adds `packages/meridian-design-system/src/index.ts`, which re-exports from `@/components/atoms`:
```
export {
  fmtDur, fmtDurDecimal, fmtClock, CATS, PROVIDER_META, ProviderGlyph, CatDot,
  CatLabel, AppGlyph, shortTaskKey, TaskKey, StatusPill, LiveDot, SectionHead,
  Card, ConfidenceRing, SegBar, useTick,
} from '@/components/atoms'
```
**Every single one of the 7 exports this PR deletes is re-exported by #423.** If this PR merges first, #423's `src/index.ts` (and its esbuild/tsc build) breaks immediately — those symbols were "zero consumers" only because #423 hadn't landed yet when this audit ran. Whichever of these two merges second needs to either restore the exports it depends on, or `src/index.ts` needs its own list trimmed to match. Worth flagging to the other PR before merging either one, since both target overlapping surface in `atoms.tsx` at nearly the same time.

### Code quality (the rest of the diff)
- The `--font-serif` fix is a good catch — the token was defined but silently unused in favor of two hardcoded duplicates; routing the wizard through the actual token is the right fix, and keeping Georgia as the fallback in the `var()` chain (rather than dropping it) preserves the load-failure behavior.
- `HygieneDialog`'s title migration to `.mt-modal-title` is called out explicitly in the PR description as a real visual change (36px Georgia serif → Jakarta bold ~20px), not a silent one — good, since this is the one place in the diff where "cleanup" changes actual rendered output rather than being a pure token swap.
- Shadow token consolidation (`--mt-card-shadow`/`--mt-modal-shadow`) replacing 4 identical hand-rolled literal strings is a clean, genuinely zero-risk dedup — confirmed the literal values match exactly across all 4 call sites in the diff.
- Tray popover hex→token swaps are exact-value matches per the description; I confirmed `--success`, `--accent`/`--accent-2`/`--accent-3`, `--ink-2` substitutions in `style.css` match the replaced hex literals positionally (can't verify the token's own resolved value without the rest of the stylesheet, but the description explicitly scoped this to exact matches only).

### Test coverage
- Build/`bun test` (243/243)/clippy/cargo test all reported clean. The one unchecked test-plan item (visual check of `HygieneDialog`'s title) is honestly flagged as a real, deliberate visual change rather than a claimed no-op — good.

Nothing else blocking. The `atoms.tsx` export overlap with #423 is the one thing I'd resolve before either merges.

adityaharishch (commented):
## Review — chore(ui): dashboard style cleanup (Phase 1)

Reviewed all 10 changed files (+19 / −225) against repo conventions and the "cleanup only" claim. This is a clean, well-scoped, well-documented PR. I verified every dead-code and token claim rather than taking them at face value; they all hold up.

### Verified (no issues)
- **Dead-code removal is genuinely dead.** `ConfidenceRing`, `SegBar`, `LiveDot`, `CatLabel`, `useTick`, `Card`, `fmtDurDecimal` — grep across `ui/**/*.{ts,tsx}` shows zero consumers outside `atoms.tsx`. The only external hit is `__tests__/redesign.test.ts`, which uses **inline re-implementations** (its own local copies, no import from `atoms`), so nothing breaks.
- **`.type-*` scale removal is safe.** `.type-heading` had exactly one consumer (`HygieneDialog`), which this PR migrates to `.mt-modal-title`. All other `.type-*` classes have zero references. `.mt-modal-title` is defined (`globals.css:603`) and is the STYLESHEET.md §1 class for modal titles — appropriate since `HygieneDialog` is the board-cleanup modal.
- **`.font-serif` utility removal is safe.** No `font-serif` className anywhere in the TSX.
- **`--font-serif` token fix is correct.** `Instrument_Serif` is loaded via `next/font` with `variable: '--font-instrument-serif'` on `<html>` (`layout.tsx`), so the retargeted token resolves; Georgia stays as the load-failure fallback. The two setup call sites (`page.tsx`, `steps.tsx`) now go through the token instead of duplicating the stack — equivalent output.
- **Tray hex→token swaps are exact value matches.** Confirmed against `tray/src/style.css` `:root`: `--success` `#10B981`, `--accent` `#6D28D9`, `--accent-2` `#9333EA`, `--accent-3` `#DB2777`, `--ink-2` `#5B5570` — all identical to the literals they replace. 6 occurrences, zero visual change. (The `#059669` first gradient stop is correctly left raw — no matching token exists.)
- **Shadow token consolidation** replaces four identical literal strings with `--mt-card-shadow` / `--mt-modal-shadow` — no value drift.
- **Conventions:** all 8 modified `.ts`/`.tsx` files retain the required file-header line; no new files added; no `any` types introduced; no user-facing string changes (so no em-dash risk); no secrets.

### Non-blocking / confirm with author
1. **Disclosed visual change (not a regression, just flagging).** `HygieneDialog`'s title goes from 36px Georgia serif (`.type-heading`) → 20px Jakarta 800 (`.mt-modal-title`). This is a real rendered change inside a PR labeled "cleanup," but it's explicitly called out in the PR body and test plan, and it aligns the dialog with STYLESHEET.md §1. Worth the recommended visual QA pass before merge, as the author noted.
2. **Stale test copy (minor).** `__tests__/redesign.test.ts` keeps an inline `fmtDurDecimal` (and mirrors of other now-removed atoms) purely to test logic that no longer ships. Harmless, but consider pruning the tests for exports that no longer exist so the suite doesn't guard dead code.

No blocking issues. Scope stays true to "cleanup" — no functional behavior smuggled in beyond the one disclosed typography change.

Akarsh-Hegde:
Merged latest `pre-main` and resolved the conflicts (63f90e62), plus addressed the two open review items:

**Merge conflicts (real git conflicts, `pre-main` had moved a lot since this branch opened 2026-07-09):**
- **`ui/app/globals.css`**: `pre-main` had independently landed the SF Pro font migration (`--font-sans`/`--font-mono` → system font stack), which regressed this PR's `--font-serif` fix back to a bare `Georgia, 'Times New Roman', serif` (no `var(--font-instrument-serif)`). Confirmed `pre-main`'s `ui/app/setup/steps.tsx` still hardcodes `fontFamily: 'var(--font-instrument-serif), Georgia, serif'` inline — this PR's fix was never actually applied upstream. Resolved by keeping `pre-main`'s SF Pro `--font-sans`/`--font-mono` and re-adding this PR's `var(--font-instrument-serif)` prefix to `--font-serif`, so both improvements survive.
- **`tray/src/style.css`**: this PR tokenized the old 3-stop `.review-cta` gradient (`linear-gradient(135deg, var(--accent), var(--accent-2) 55%, var(--accent-3))`); `pre-main` had since consolidated the tray to a single flat `var(--accent)` fill (part of the broader one-accent/one-flourish-gradient direction, `--accent-2` no longer exists at all on `pre-main`). Took `pre-main`'s flat version — the gradient tokenization is moot since the gradient itself is gone upstream.

**Review item — stale test coverage:** pruned `ui/__tests__/redesign.test.ts`'s inline `fmtDurDecimal` reimplementation and its `describe` block (re-verified zero consumers anywhere in `ui/` besides that test file, matching the review's note that it was "testing logic that no longer ships").

**Review item — disclosed visual change (`HygieneDialog` title, 36px Georgia → 20px Jakarta):** no code change needed here, already correctly disclosed in the PR description per the review; noting it's still the one item that needs an eyes-on visual pass rather than something I can verify from a diff.

**Not touched (out of scope for this PR):** the #423 cross-PR conflict — both reviews flagged that #423's `meridian-design-system` package re-exports all 7 of the `atoms.tsx` exports this PR deletes. That's being handled separately depending on merge order, not as part of this PR.

Verified after the merge: `cargo fmt`, `cargo clippy -- -D warnings`, `cargo test` (all green — one `meridian-oauth` test, `client_id_prefers_env_override`, is pre-existing-flaky under parallel test execution due to a shared global env var with a sibling test, unrelated to this PR and not touched by this diff — confirmed it passes reliably with `--test-threads=1`), `bun test` (181/181), `tsc --noEmit`, `npm run build`. All 4 CI checks green.

PR #422, opened 2026-07-09
feat(ui): shared Badge/pill primitive (Phase 2, part 1)

Akarsh-Hegde:
## Summary

Phase 2 of the dashboard style cleanup (plan: `~/.claude/plans/composed-seeking-robin.md`, Obsidian: `Meridian/Product/Dashboard style cleanup — fonts, spacing, components.md`). This PR ships **item 1+2 only**: one shared `Badge` primitive replacing hand-rolled chip/pill markup duplicated across timeline components.

**Stacked on #421** (Phase 1) — not pre-main directly. Both PRs touch `atoms.tsx`, `TaskDetailDialog.tsx`, etc., and #421 hasn't merged yet, so this targets `chore/dashboard-style-cleanup` to avoid duplicate/conflicting diffs. **Merge order: #421 first, then this.**

## What changed

- New `Badge` component in `ui/components/atoms.tsx`: `variant` (`chip` small-rounded mono-text badge | `pill` rounded-full), `tone` (`outline` | `filled` color-mix background | `ghost` neutral fill), optional leading `dot`, `borderColor` override for the neutral/hairline chips whose border doesn't match their text color.
- Migrated 12 call sites across 7 files: `CleanupCard`, `ReviewCard` (×3), `TasksDetailPane` (×2), `TimelineCard` (×3), `TaskDetailDialog` — the outlined data-label chips — plus `HourBadges`' two duplicate "Paused" pill implementations.

**Scoped from an actual grep audit**, not the plan's original blind 16-file list. Several files on that list (`ModalShell`, `ReportModal`, `ReviewOverlay`, `ThemeSwatches`, `TimeByApp`, `TimelineColumn`, `CleanupOverlay`, `Toolbar`) turned out to contain icon buttons, progress bars, or decorative dots — not the same badge pattern — so they're untouched. `Toolbar`'s status pills use body text (not the chip mono style) and were judged a distinct molecule, not force-fit into Badge.

## Visual deltas to eyeball (no Chrome tooling available this session — build/types/tests pass, but nobody has looked at this rendered)

- 4 of the 10 chip sites (`CleanupCard`, `ReviewCard:141`, `ReviewCard:223`, `TimelineCard:87`) had `px-2 py-0.5` padding vs. the majority's `px-1.5 py-0.5`; standardized to `px-1.5` (2px narrower horizontally).
- `shrink-0` added to all chips (previously inconsistent) so a chip never compresses in a crowded flex row.
- `ReviewCard`'s "unsaved" chip lost its own wrapping `{pendingCandidate && (...)}` block structure but renders identically.

## Deliberately deferred (surfaced, not silently dropped)

- **Setup wizard `Btn` reconciliation** — there's no dashboard-side button primitive to reconcile against (every dashboard button is hand-rolled per-component, including the circular `Fab`). Extracting one now would be a speculative abstraction nobody asked for. Left intentionally separate.
- **Spacing/radius token scale** — audited actual values. The `borderRadius: 99/999` "duplication" found in the original audit is the same make-it-a-pill sentinel idiom used consistently everywhere (visually identical whether 99 or 999); real card/modal radii already go through Tailwind's own `rounded-*` scale. A mechanical token sweep here would rename values, not fix anything visible — no payoff worth the diff size or the risk of touching ~20 files unverified.
- **Tray popover token unification** — the popover only renders in a **packaged build**, not `tauri dev` (per this repo's CLAUDE.md), so it cannot be visually verified in this environment at all. Too risky to do blind; left for a session where a packaged build can be checked.

Also reverted (not part of this PR): I initially changed `StatusPill`'s dot color from `--success`/`--accent` to `--color-state-approved`/`--color-state-proposal` while touching that function. Checked the actual hex values — `--accent` is amber (`#C4822A`) vs `--color-state-proposal` violet (`#8B5CF6`), a real, visible color change to HygieneDialog's status dot, not a token rename. That's unrelated legacy-token migration (already tracked as follow-up from PR #419's categorical-color work covering ~25 files) — reverted out of scope for this PR.

## Test plan
- [x] `npm run build` — clean
- [x] `bun test` — 177/177 pass
- [x] `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test` — clean (pre-push hook)
- [ ] Manual visual check of the 12 migrated call sites recommended before merge — see deltas above


adityaharishch (commented):
## Review

**What this does:** Phase 2 (item 1+2) of the dashboard style cleanup — a shared `Badge` primitive (`chip`/`pill` variants, `outline`/`filled`/`ghost` tones, optional dot) in `atoms.tsx`, migrated into 12 call sites across 7 files that previously hand-rolled the same chip/pill markup. Stacked on #421 (targets `chore/dashboard-style-cleanup`, not `pre-main`).

### Code quality
- `Badge`'s prop surface (`variant`/`tone`/`color`/`borderColor`/`dot`) cleanly covers every hand-rolled variant I can find in the diff — outline chips (`CleanupCard`, `ReviewCard`, `TaskDetailDialog`, `TasksDetailPane`, `TimelineCard`), the filled pending pill (`HourTakeover`), the filled+dot pill (`PausedBadge`), and the ghost/neutral chip (`ReviewCard`'s provider label). I traced each migrated call site against its `git diff` original and the rendered `style`/class output is equivalent in every case, including the padding standardization (`px-2 py-0.5` → `px-1.5 py-0.5`) the PR description calls out.
- `borderColor` defaulting to `color` but overridable is a nice touch for the `--t-hair` hairline-border chips (`issue_type`) that don't want a color-tinted border — matches every migrated call site correctly.
- The PR description's own "visual deltas to eyeball" and "deliberately deferred" sections are unusually thorough and honest about what wasn't verified — that's exactly the right way to hand off a pure-visual refactor when Chrome tooling wasn't available.
- The `ReviewCard` "unsaved" chip losing its wrapping block structure is confirmed purely syntactic (JSX conditional `&&` vs. the same conditional inline) — no behavior change.

### Minor observations (not blocking)
- `tone: 'ghost'` uses `background: 'var(--t-wrap)'` where the original `ReviewCard` provider chip used the `bg-wrap` Tailwind utility class directly — presumably these resolve to the same value, but since this is the one call site using `ghost`, it's worth eyeballing specifically alongside the other deltas already flagged for manual check.
- `Badge`'s type signature includes `[key: string]: unknown` for pass-through props spread onto a polymorphic `as` tag — a bit loose (no compile-time guarantee those extra props are valid for whatever `Tag` resolves to), but this is a common pattern for this kind of primitive and `tsc --noEmit` is reported clean, so not a real issue in practice.

### Test coverage / merge readiness
- Build/types/`bun test` (177/177) all pass per the description, but the manual visual check of the 12 migrated sites is explicitly still unchecked — that's the one real open item before merge, and matches what the author already flagged themselves.
- Correctly stacked on #421 with merge order called out — nothing to flag there, just confirming I read the same dependency.

Nothing blocking beyond the author's own pending visual-check item. Clean, well-scoped refactor with good self-documentation of the judgment calls (deferred token sweep, deferred `Btn` reconciliation, reverted `StatusPill` color change).

adityaharishch (commented):
## Review — `feat(ui): shared Badge/pill primitive (Phase 2, part 1)`

Solid, well-scoped consolidation. Diff is coherent, the audit-driven scoping (vs. the plan's blind 16-file list) is the right call, and the deferred items are surfaced rather than silently dropped. I verified the mechanical equivalences below and found no blocking issues.

### Repo-convention checks (all pass)
- **File header** — `ui/components/atoms.tsx` line 1 has the required ambient-dev-tool header; no new files were created, so all other touched files keep theirs. ✅
- **No `any`** — the `Badge` prop type uses `[key: string]: unknown` (correct, not `any`), `React.ElementType`, `React.CSSProperties`. ✅
- **User-facing text** — no em/en-dashes or `--` in displayed strings (`Paused`, `unsaved`, data-driven labels). Em-dashes appear only in JSDoc comments, which are exempt. ✅
- **No secrets.** ✅
- **Ghost fill equivalence** — confirmed `bg-wrap` (the class the old ReviewCard provider chip used) maps to `--color-wrap → var(--t-wrap)` (globals.css:58), which is exactly what `tone="ghost"` sets. So that migration is a true no-op on fill. ✅
- **Dead code** — all 12 hand-rolled sites are actually replaced; no duplicate markup left behind. Both `HourBadges` "Paused" implementations are folded in. ✅

### Non-blocking — worth an eyeball before merge
1. **`inline-flex` applied to every chip, including previously-plain `inline` spans.** The old outlined data-label chips (`CleanupCard`, `ReviewCard` state chip, `TimelineCard` ×2, `TaskDetailDialog`, `TasksDetailPane`) were bare `<span class="mt-chip …">` — i.e. `inline`. `Badge` always emits `inline-flex items-center gap-1.5`. `inline` → `inline-flex` changes baseline/vertical alignment of the text within its flex row, which the PR's disclosed deltas (padding + `shrink-0`) don't call out. `gap-1.5` is inert for single-child chips, so the only real risk is that alignment shift. Please include these in the visual pass, not just the 4 padding sites.
2. **Whole PR is visually unverified** (acknowledged in the description). Given (1) touches all 10 chip sites' box model, the manual check of the 12 call sites should be treated as required-before-merge, not optional.
3. **`borderColor` silently ignored for `filled`/`ghost` tones.** Only `outline` consults it. That's the intended design, but the prop's doc comment implies it's a general border override — a one-line "(outline tone only)" note would prevent a future caller from being surprised.
4. **`{...props}` + `[key: string]: unknown` forwards arbitrary attributes onto the DOM node.** Fine as an escape hatch (and it's spread before the explicit `className`/`style`, so those correctly win), but it means a typo'd prop lands on the element and React warns at runtime rather than the type system catching it. Acceptable; just flagging the tradeoff.

### To confirm with the author
- **Base branch / merge order.** This targets `chore/dashboard-style-cleanup` (#421), not `pre-main` — correct given the stacking, but per CLAUDE.md every PR must ultimately land on `pre-main`. Confirm the plan is to **retarget this to `pre-main` after #421 merges** (merge order #421 → #422) so it doesn't get merged into a stale base or forgotten.

Nice work — the scoping discipline and the "reverted the StatusPill dot color because it was a real visible change, not a token rename" call are exactly right.

Akarsh-Hegde:
## Update: addressed review feedback, retargeted to pre-main

1. **Base branch retargeted** — #421 (the branch this was stacked on) merged, so I retargeted this PR from `chore/dashboard-style-cleanup` to `pre-main` (`gh pr edit --base pre-main`) and merged `pre-main` in locally. Clean merge, no conflicts.

2. **`inline-flex` alignment risk (the "required-before-merge" item) — fixed at the source instead of eyeballed.** None of the 12 real migrated call sites pass `dot` to `Badge`, so I made the flex layout conditional on it: `inline-flex items-center gap-1.5` only applies when a dot is actually rendered (the one case that needs flex alignment); every dot-less badge — i.e. every current call site — stays plain `inline`, exactly matching the hand-rolled `<span className="mt-chip …">` markup it replaced. This removes the box-model change the review flagged rather than requiring a manual visual pass to confirm it's harmless, since with this fix there is no layout diff for any of the 12 sites.

3. **`borderColor` doc comment** — clarified it's outline-tone only (`filled`/`ghost` compute their own border from `color`/tone and ignore it).

Left alone, as intended: the `[key: string]: unknown` pass-through-props tradeoff (acknowledged, not a real issue given `tsc --noEmit` is clean) and the `ghost` tone's `bg-wrap`-equivalence (already confirmed a true no-op in review).

### Verification
- `bun test` — 181/181 pass
- `npm run build` (ui) — clean
- `npx tsc --noEmit` — clean
- `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test` — clean (pre-push hook)
- Pushed at `963c14de`; `mergeable: MERGEABLE` / `mergeStateStatus: CLEAN` against the new `pre-main` base. Rust/UI checks were still running as of this comment — will confirm green separately.

Note: the `.claude/worktrees/dashboard-style-cleanup-phase2` git worktree has unrelated uncommitted work on top of this branch (a tray right-click-tooltip fix, a popover-dismiss-on-focus hook, a popover border removal, and incidental `services/uv.lock` drift) — untouched by this update, left for whoever owns that work to commit or discard separately.

PR #423, opened 2026-07-09
feat(design-system): scaffold meridian-design-system package

Akarsh-Hegde:
## Summary
- New `packages/meridian-design-system/` package: re-bundles a curated, standalone-renderable subset of `ui/components` (form primitives, atoms, and 9 presentational timeline pieces) via esbuild, with generated `.d.ts` and the real compiled Tailwind stylesheet + self-hosted fonts pulled from `ui/`'s own production build (`ui/out`).
- `ui/` itself is untouched — the package builds directly FROM the existing source files (real imports, no copies), never a reimplementation.
- Exists to give Claude Design (claude.ai/design) a real component bundle to sync against via the design-sync skill, instead of generic components.

Scope is intentionally narrow for v1: primitives (`NumberStepper`, `Select`, `Switch`, `TextInput`), atoms (formatters/glyphs/badges/cards from `atoms.tsx`, `ProviderIcon`), and timeline pieces that are props-driven and degrade gracefully outside the Tauri shell (`CurvedArrow`, `FloatingDraftsPill`, `ModalShell`, `HourBadges`/`HourTakeover`, `MeridianMark`, `TimelineCard`, `TimeByApp`, `TimeByCategory`, `EditableSummary`/`EditableTitle`). Excluded: components that fetch live app data at mount via `@/lib/bridge`'s `load()`/`mutate()` (e.g. `Toolbar`, `ThemeSwatches`, most modals/panels) — those throw outside the Tauri window.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `node build.mjs` — clean esbuild bundle + `.d.ts` tree + compiled `styles.css`/fonts, no dangling path aliases
- [x] Same-process React SSR smoke test — every exported component renders without throwing
- [x] `git status` confirms `ui/` is untouched by the build


adityaharishch (commented):
## Review

**What this does:** New standalone `packages/meridian-design-system/` package that re-bundles a curated subset of `ui/components` (form primitives, atoms, 9 presentational timeline pieces) via esbuild + tsc, pulling the real compiled Tailwind CSS/fonts from `ui/out`. Exists purely to give Claude Design a real component bundle to sync against. `ui/` itself is untouched — everything is a live re-export, not a copy.

### Code quality
- `build.mjs` is genuinely tricky (bundling a subset of a Next.js app's components standalone, hoisting nested `.d.ts` output, rewriting `@/` path aliases post-emit since tsc doesn't resolve `paths` in declaration output) and every non-obvious step has a comment explaining *why*, not just what — e.g. the `rootDir`-not-set explanation for why `.d.ts` lands nested, and the tsc-limitation note on alias rewriting.
- `src/index.ts`'s barrel comment clearly states the inclusion/exclusion rule (props-driven + standalone-renderable vs. anything hitting `@/lib/bridge`'s `load()`/`mutate()`), so future additions have an obvious test to apply.
- `MeridianMark` being pulled out of `Toolbar.tsx` specifically (rather than exporting `Toolbar` itself) is called out with the reasoning (pure-JSX, no execution of Toolbar's own render) — good, since at a glance it looks inconsistent with "component from ui/components, never reimplemented."
- Dependency versions (`react`/`react-dom` peer deps at `19.2.0`) match `ui/package.json` exactly — good hygiene for a package meant to render `ui/` components live.

### Risks / fragility (worth being aware of, not necessarily blocking for a v1 sync scaffold)
- `buildStyles()`'s heuristic for finding the compiled global stylesheet (`text.includes('jbmono') && text.includes('@font-face')`) is brittle — if the font family name or the ordering of Next's CSS chunking changes, this silently degrades to the `console.warn` skip path rather than failing the build loudly. Given this ships fonts/styles that Claude Design depends on for an accurate render, consider making a missing stylesheet a hard build failure rather than a warning, so a silent CSS/font drop doesn't quietly ship a broken design-sync bundle.
- `build.mjs` requires `ui/out` to exist (`cd ui && npm run build` run first) but there's no `prebuild`/CI wiring in this diff to guarantee that's fresh — easy to run `node build.mjs` against a stale `ui/out` and get subtly wrong styles. Might be worth a script note or `package.json` script that chains the two, if this isn't already handled by whatever invokes design-sync.
- No test coverage beyond the manual SSR smoke test in the PR description (which is reasonable for a v1 scaffold, but worth keeping in mind since there's no CI gate here yet if a future excluded/included-component boundary gets crossed by accident).

### Correctness
- `tsconfig.json`'s `@/*` → `../../ui/*` path alias matches the rewrite logic in `build.mjs`'s `rewriteDtsAliases()` (`join(distDir, 'ui', sub)`) — consistent.
- `.gitignore` correctly excludes `dist/` and `node_modules/`, so the generated bundle isn't checked in — matches the "build directly from source, not copies" framing in the PR summary.
- `exports` map's `"./styles.css"` subpath is a reasonable way to let consumers opt into the compiled CSS separately from the JS bundle.

Nothing blocking — solid scaffold. The stylesheet-detection heuristic is the one thing I'd want hardened before this becomes load-bearing for other consumers.

adityaharishch:
## Heads up: conflicts with #421 (open, same author)

While reviewing this PR I found a direct conflict with #421 (\`chore/dashboard-style-cleanup\`, currently open, targets \`pre-main\`).

#421 deletes 7 exports from \`ui/components/atoms.tsx\` as verified-zero-consumer dead code: \`Card\`, \`ConfidenceRing\`, \`SegBar\`, \`LiveDot\`, \`CatLabel\`, \`fmtDurDecimal\`, \`useTick\`.

This PR's \`src/index.ts\` re-exports every one of those 7 symbols from \`@/components/atoms\`. If #421 merges first (or after this one, without adjustment), \`meridian-design-system\`'s barrel import breaks — \`node build.mjs\` / \`tsc\` will fail on the missing exports.

Left the full detail as a comment on #421. Whichever of the two merges second will need to either trim this package's export list to match what's left in \`atoms.tsx\`, or coordinate with #421 to keep those exports alive if the design-system package still needs them (e.g. \`Card\`/\`ConfidenceRing\`/\`SegBar\` for future timeline pieces).

adityaharishch (commented):
## Review — `feat(design-system): scaffold meridian-design-system`

Strict pass over all 36 added files. This is a clean, appropriately-minimal scaffold that re-bundles real `ui/components` source rather than reimplementing it — the right call. No blocking issues found. Notes below.

### Convention compliance (checked, passing)
- **File headers** — every new `.rs`/`.ts`/`.tsx` file that requires it carries the exact header line: `build.mjs`, `src/index.ts`, and all 27 `previews/*.tsx`. JSON (`package.json`, `tsconfig.json`, `config.json`) and Markdown (`NOTES.md`, `conventions.md`) are exempt. ✓
- **No `any` types** introduced. ✓
- **No secrets committed** — `previews/TextInput.tsx` uses truncated fake placeholders (`sk-ant-api03-…`, `ATATT3xFfGF0…`, `correct horse battery staple`), not real credentials. ✓ (minor nit below)
- **Em-dash rule** — the em-dashes in `NOTES.md`/`conventions.md`/`ConfidenceRing.tsx` are in dev docs and preview stories, not shipped user-facing app strings, so the rule does not apply. No violation.
- **package.json** — `name`/`version: 0.1.0`/`private: true`/`exports`/`types`/`main` all correct and internally consistent. Standalone lockfile + own `node_modules` matches the existing `packages/meridian-mcp` precedent (the repo root is release-tooling only, not an npm workspace), so "not wired into a workspace build" is consistent with convention, not a gap. ✓

### Non-blocking — please confirm
1. **Merge-order conflict with #421 (highest-priority coordination risk).** `src/index.ts` re-exports 7 `atoms.tsx` symbols (`Card`, `ConfidenceRing`, `SegBar`, `LiveDot`, `CatLabel`, `fmtDurDecimal`, `useTick`) that per `NOTES.md` have zero consumers in `ui/` and are slated for deletion as dead code by #421. If #421 lands first, this package's barrel references deleted symbols and the esbuild/tsc build breaks. Since #421/#422/#439 all depend on this scaffold, please pin the intended merge order explicitly (this scaffold first, then drop the 7 names + rebuild if #421 removes them).

2. **`build.mjs` stylesheet detection is brittle.** The global CSS chunk is located by grepping for a chunk containing both `jbmono` and `@font-face`. If `ui/`'s font-variable naming ever changes, this silently `console.warn`s and ships **no** `styles.css`/fonts rather than failing — an easy-to-miss stale/empty-CSS ship. Consider hard-failing (non-zero exit) instead of warn-only when `ui/out` is present but no matching stylesheet is found.

3. **Hard ordering dependency on `cd ui && npm run build`.** `buildStyles()` degrades to a warning when `ui/out` is missing. Confirm that's intended for CI — a design-system build that "succeeds" with no CSS is a silent failure mode. A guard/flag to require the stylesheet in CI would harden this.

### Minor nits (optional)
- `previews/TextInput.tsx`: the `sk-ant-api03-…` placeholder mirrors a real Anthropic key prefix and may trip secret scanners. A more obviously-synthetic placeholder avoids false positives.
- The `.d.ts` hoist + `paths`-alias rewrite in `build.mjs` is clever but fragile (leans on tsc's common-ancestor emit behavior). It's well-commented, so fine — just flagging it as the most likely thing to break on a TS upgrade.

### Nice touches
`NOTES.md` documenting the `MeridianMark` asset-path limitation, the deliberate scope exclusions (bridge-`load()` components that throw outside Tauri), and the #421 conflict is exactly the context a reviewer needs — well done. The `MeridianMark`-from-`Toolbar.tsx` import is safe: importing pulls `Toolbar`'s module-level `import { load }` but never executes its render, so no bridge call fires.

Not a blocker on my end — just resolve the #421 merge ordering before this or #421 lands.

Akarsh-Hegde:
## Update: addressed review feedback

Fixed all three items from review:

1. **Merge-order conflict with #421 (resolved)** — #421 merged and deleted the 7 zero-consumer `atoms.tsx` exports (`Card`, `ConfidenceRing`, `SegBar`, `LiveDot`, `CatLabel`, `fmtDurDecimal`, `useTick`) as dead code, which broke this package's barrel. Merged `pre-main` in, dropped those 7 names from `src/index.ts`, removed their now-orphaned preview stories, fixed the one build-snippet reference to `Card` in `conventions.md`, and updated `NOTES.md` to record the resolution. `node build.mjs` + `npx tsc --noEmit` both clean.

2. **Stylesheet-detection heuristic hardened** — `buildStyles()` now throws instead of `console.warn`-and-skip when `ui/out` is missing or the global stylesheet chunk can't be found, so a broken/missing `styles.css`/fonts fails the build loudly instead of shipping silently. Also dropped the `jbmono` substring check from the detection heuristic — turns out it was already silently broken: JetBrains Mono was removed from `ui/`'s stylesheet in the earlier SF Pro migration (#439), so the old heuristic could never have matched a real build. Verified against a fresh `cd ui && npm run build` — the single compiled CSS chunk correctly carries both `:root` tokens and `@font-face` (Instrument Serif, self-hosted via `next/font`), so the new heuristic matches it as expected.

3. **Fake API key placeholder** — swapped the Anthropic-key-shaped `sk-ant-api03-…` in the `TextInput` preview for an obviously-synthetic `sk-example-xxxxxxxxxxxxxxxx`.

### Verification
- `node build.mjs` — clean (dist/index.js, dist/index.d.ts, dist/styles.css + fonts all written)
- `npx tsc --noEmit -p tsconfig.json` — clean
- `bun test` (repo-wide) — 181/181 pass
- `npm run build` (ui) — clean
- `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test` — clean (pre-push hook)
- Pushed at `00e46dc4`; CI (Rust/UI/Migrations append-only/screenpipe MIT pin) all green, `mergeable: MERGEABLE` / `mergeStateStatus: CLEAN`

Ready to merge whenever you are.

PR #425, opened 2026-07-09
fix(setup): notification pane deep-link opens Meridian's settings, skip redundant re-request

Akarsh-Hegde:
## Summary
- `open_permission_pane`'s notifications URL had no app anchor, so "Open Settings" landed on the general Notifications list instead of Meridian's own row. Anchored it with `?id=<bundle-id>` — verified live against `NotificationsSettings.appex`'s metadata, which documents support for this param (`allowsXAppleSystemPreferencesURLScheme`).
- `grantNotifications` always called `request_notifications` first, even when already denied. macOS never re-prompts after a deny, so this was a pointless async round-trip through the Swift notification-plugin bridge before ever reaching the Settings pane — a plausible source of the reported "flickers, doesn't open" behavior. Now skips straight to `open_permission_pane` when the caller already knows the state is denied.
- **Notifications card flicker (found during live manual testing above)**: `check_notifications` polls every 2s; a transient Swift-bridge hiccup can make the Rust command return `'unavailable'` as a normal (non-error) result for a single poll tick, which the card's hide-when-unavailable check then treats as the plugin having vanished — flashing the whole card away and back. Bundled-ness is fixed for the process lifetime, so once a real state (`prompt`/`granted`/`denied`) has been observed, a later `'unavailable'` can only be that glitch, never a genuine regression — the poll now keeps the last known-good state instead.
- **Notifications row now omits the REQUIRED/OPTIONAL badge** (Accessibility/Screen Recording still show `REQUIRED`). Intentional, requested during live wizard testing: unlike those two, notifications doesn't gate `canNext` on the Permissions step — nothing about the row conveys a requirement to satisfy, so the `OPTIONAL` label was pure visual noise where `REQUIRED` on the other two rows is actually meaningful (it signals what blocks continuing). Dropping it declutters the row without losing information.
- **Copy**: notifications description now uses a plain hyphen (`Quiet by default - you control…`) instead of an em dash, matching the hyphen style used everywhere else in the wizard's copy.
- **New**: semi-circular gauge (`MemoryGauge` in `atoms.tsx`) for the Local-intelligence step's "Expected memory" panel, replacing the straight `Bar` there. Kept `Bar` itself untouched/reused as-is for download progress elsewhere — the two are visually distinct on purpose (straight = loading, arc = capacity). Drawn as two arcs meeting at the footprint percentage: amber for what Meridian expects to use, green for the unified memory left over.

All found/requested while manually testing PR #418 (notifications permission card) on a staging build, then a locally packaged build of this branch.

## Test plan
- [x] `cargo fmt --check` + `cargo clippy -- -D warnings` clean
- [x] `npm run build` (ui) succeeds, `/setup` route compiles
- [x] Verified the `?id=com.meridiona.tray` deep link lands on Meridian's specific notification detail pane (Allow toggle, Alert Style, etc.) rather than the general app list, via direct `open` testing against this machine's `NotificationsSettings.appex`
- [x] Manual: clicked "Open Settings" on a denied Notifications permission in a locally packaged build of this branch — opens directly to Meridian's row, no flicker
- [x] Manual: verified the memory gauge renders correctly (amber/green split arc) in the same packaged build

### DIFF
--- a/tray/src-tauri/src/commands/system.rs
+++ b/tray/src-tauri/src/commands/system.rs
@@ -109,19 +109,29 @@ pub async fn open_permission_pane(app: tauri::AppHandle, pane: String) -> Result
     let url = match pane.as_str() {
         "screen_recording" => {
             "x-apple.systempreferences:com.apple.preference.security?Privacy_ScreenCapture"
+                .to_string()
         }
         "accessibility" => {
             "x-apple.systempreferences:com.apple.preference.security?Privacy_Accessibility"
+                .to_string()
         }
         "input_monitoring" => {
             "x-apple.systempreferences:com.apple.preference.security?Privacy_ListenEvent"
+                .to_string()
         }
         // Notification authorization lives under Notifications, not Privacy &
-        // Security. Bare pane, no ?id=<bundle-id> anchor — per-app anchoring is
-        // undocumented and inconsistent across macOS versions. This is the deny
+        // Security. Anchored to our own bundle id so this opens Meridian's
+        // specific notification detail pane (Allow toggle, Alert Style, …)
+        // instead of just the app list — verified working via the
+        // `?id=<bundle-id>` param, which the pane's own extension metadata
+        // (`NotificationsSettings.appex`'s `allowsXAppleSystemPreferencesURLScheme`)
+        // supports even though it's otherwise undocumented. This is the deny
         // recovery path: macOS shows the authorization dialog exactly once, so
         // after a deny the wizard can only send the user here.
-        "notifications" => "x-apple.systempreferences:com.apple.preference.notifications",
+        "notifications" => format!(
+            "x-apple.systempreferences:com.apple.preference.notifications?id={}",
+            app.config().identifier
+        ),
         other => return Err(format!("unknown permission pane: {other}")),
     };
     dismiss_popover(&app);
--- a/ui/app/setup/atoms.tsx
+++ b/ui/app/setup/atoms.tsx
@@ -111,6 +111,40 @@ export function Bar({ pct, height = 6, track = 'var(--t-hair)', fill = 'var(--co
   )
 }
 
+// ── Semi-circular memory gauge ───────────────────────────────────────────────
+// Distinct on purpose from `Bar` above (straight, used for download progress):
+// this reads as a capacity dial rather than a loading indicator, so the two
+// never get confused for one another. Drawn as two arcs meeting at `pct` —
+// amber for the footprint Meridian expects to use, green for the unified
+// memory left over — rather than a single fill against a neutral track.
+export function MemoryGauge({ pct, size = 96, strokeWidth = 9 }: {
+  pct: number; size?: number; strokeWidth?: number
+}) {
+  const r = size / 2 - strokeWidth
+  const cx = size / 2
+  const cy = size / 2
+  const clamped = Math.max(0, Math.min(100, pct))
+  const splitAngle = 180 + (clamped / 100) * 180
+  const polar = (angleDeg: number) => {
+    const rad = (angleDeg * Math.PI) / 180
+    return { x: cx + r * Math.cos(rad), y: cy + r * Math.sin(rad) }
+  }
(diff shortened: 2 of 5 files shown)


adityaharishch (commented):
## Review

**What this does:** Two related notifications-permission fixes found while testing #418: (1) the deep-link for the deny-recovery path now anchors System Settings to Meridian's own row (`?id=<bundle-id>`) instead of the generic Notifications list, and (2) `grantNotifications` skips the pointless `request_notifications` round-trip when the caller already knows the state is `denied` (macOS never re-prompts after a deny anyway). Also adds a debounce so a transient `'unavailable'` poll tick doesn't flash the notifications card away once a real state has been observed.

### Code quality
- `open_permission_pane`: the comment explaining the `?id=` anchor (verified against `NotificationsSettings.appex` metadata) is a good addition — this was previously flagged in-code as "undocumented and inconsistent," so documenting exactly how it was verified is valuable for the next person who touches this.
- The `alreadyDenied` param threading (`Wiz.grantNotifications` → `PermissionsBody`) is clean and the comments at both the caller and definition explain *why*, not just *what*.
- `page.tsx`'s poll debounce logic and its comment are clear about the invariant it relies on (bundled-ness fixed for process lifetime, so `'unavailable'` after a real state can only be transient).
- Nice unrelated fix: `data.ts`'s em dash → hyphen brings the notifications description in line with the project's user-facing-text hyphen rule.

### Possible scope creep / needs explanation
- `steps.tsx`: `{!notif && <span>...REQUIRED/OPTIONAL...}` now hides the REQUIRED/OPTIONAL badge entirely for the notifications row. This isn't mentioned anywhere in the PR summary or test plan, and it changes behavior for the not-denied/not-granted case too (not just the deep-link path this PR is about). Was this intentional cleanup bundled in, or a leftover from local UI tweaking? If intentional, worth a line in the summary explaining why notifications specifically loses the badge while every other permission keeps it.

### Correctness
- The `to_string()` calls added to the other match arms in `open_permission_pane` are a necessary consequence of the `notifications` arm now returning an owned `String` (`format!`) instead of `&str`) — correct and minimal.
- Deny-recovery path logic (skip request → go straight to pane) matches the documented macOS behavior (single-shot prompt, no re-prompt after deny).

### Test coverage
- Test plan is honest about what's unverified (last checkbox unchecked — manual click-through to confirm no flicker). Given this is UI/permissions-pane behavior with no automated coverage, that's an acceptable gap as long as it gets manually verified before merge.

Nothing blocking; please clarify the REQUIRED/OPTIONAL badge removal before merging.

Akarsh-Hegde:
Thanks for the review — addressing the one open question and updating on scope.

**REQUIRED/OPTIONAL badge removal** — intentional, not a leftover. This row is the only permission in the list that doesn't gate `canNext` on the Permissions step (`canNext: e => !!(e.perms.accessibility && e.perms.screen)` — notifications isn't in that check). So `REQUIRED` on the other two rows is meaningful signal (it tells you what's blocking Continue), but `OPTIONAL` on this one wasn't conveying anything actionable — it was just visual noise. Dropped it, kept `REQUIRED` on Accessibility/Screen Recording. Explained in the updated PR summary now.

**Also pushed since your review** (all found/requested during the same live manual test pass, on both a staging build and a locally packaged build of this branch):
- Root-caused and fixed the actual notifications-card **flicker**: `check_notifications` polls every 2s, and a transient Swift-bridge hiccup can make the Rust command return `'unavailable'` as a normal result for a single tick — which was hiding the whole card for that tick since bundled-ness can't actually change mid-session. Now sticky on the last known-good state.
- Copy fix: em dash → plain hyphen in the notifications description, matching the rest of the wizard's copy style.
- New: semi-circular gauge for the "Expected memory" panel (Local-intelligence step) — amber/green split arc instead of the straight `Bar` (which stays untouched/reused as-is for download progress elsewhere).

Last test-plan checkbox is now checked too — verified the no-flicker Open Settings behavior and the new gauge live in a packaged build. PR summary is updated to match everything currently in the diff.

PR #426, opened 2026-07-09
feat(dmg): themed drag-to-Applications background with a curved arrow

Akarsh-Hegde:
## Summary
- The DMG installer previously shipped Tauri's default background (blank, no branding, no arrow). Added a custom background matching the setup wizard's lilac theme (`--t-panel`/`--desk` gradient, `--color-state-proposal` violet accent) with the Meridian mark, wordmark, and a curved arrow pointing from the app icon to the Applications alias.
- Background is generated by `scripts/make-dmg-background.py` (Pillow), rendered at 2x/144dpi for a crisp Finder display, so it can be regenerated if the theme changes rather than hand-edited as a binary blob.
- Wired into `tray/src-tauri/tauri.conf.json`'s `bundle.macOS.dmg` config, explicitly pinning `windowSize`/`appPosition`/`applicationFolderPosition` (currently Tauri's own defaults) so the artwork's layout can't silently drift from a future Tauri default change.

## Test plan
- [x] `cargo fmt --check` + `cargo clippy -- -D warnings` clean
- [x] Verified icon-gap alignment by overlaying placeholder boxes at the exact Finder icon coordinates before finalizing the artwork
- [x] Built the real DMG (`tauri build --bundles dmg --features capture`), mounted it, and opened it in Finder — confirmed the real `Meridian.app` + `Applications` alias icons render correctly with the curved arrow floating cleanly between them, no overlap
- [x] Full pre-push suite passed: fmt, ui build, clippy, ui tests, security audit, `cargo test`

https://claude.ai/code/session_01RJ8rU48qbY1A3tthxFkQj8


adityaharishch (commented):
## Review

**What this does:** Replaces Tauri's blank default DMG background with a themed one (lilac gradient, Meridian mark/wordmark, curved arrow app→Applications), generated reproducibly via `scripts/make-dmg-background.py` (Pillow) and wired into `tauri.conf.json`'s `bundle.macOS.dmg` (`background`, `windowSize`, `appPosition`, `applicationFolderPosition`).

### Code quality
- Nice touch pinning `windowSize`/`appPosition`/`applicationFolderPosition` explicitly even though they currently match Tauri's defaults — protects the artwork from silently drifting if Tauri's defaults ever change.
- Script is well documented: module docstring explains the theme-token source, the icon layout it targets, and that it's a dev-only tool (`pip install Pillow`, not a runtime dep) — good for future maintainers who need to regenerate the asset.
- Color tokens (`T_MUTED`, `ACCENT`, `DESK_1/2/3`) are commented with their `globals.css` source variable, which keeps this in sync with the wizard theme without needing to inline the actual CSS.
- `radial_l()` has a clear docstring explaining why it inverts Pillow's built-in gradient.

### Minor nits (non-blocking)
- `scripts/make-dmg-background.py:132`: `draw.text((W // 2 - tw // 2, pt(0, 258)[1]), ...)` — routing a plain `y * SCALE` through `pt(0, 258)` just to discard the x component reads oddly next to the straightforward `mark_y = 26 * SCALE` a few lines above. Could just be `258 * SCALE` for consistency.
- The script hardcodes macOS-only font paths (`/System/Library/Fonts/SFNS.ttf`, `SFNSMono.ttf`) with no existence check/error message — if someone without those fonts (non-macOS, or a future CI regen step) runs this, they'll get a raw Pillow `OSError` rather than a clear "run this on macOS" message. Low risk today since it's an ad-hoc local dev tool per the docstring, but worth a comment or guard if this ever gets wired into CI/release tooling.

### Risks / correctness
- Can't verify the rendered curve/arrowhead geometry from the diff alone (bezier control points, icon-gap offsets) — the PR description says this was checked visually via placeholder-box overlay and a real mounted DMG, which is the right way to validate this kind of asset since there's no automated test surface for it.
- `tauri.conf.json` changes are additive and scoped to `bundle.macOS.dmg` — no risk to other platforms/build targets.

### Test coverage
- No automated tests, which is appropriate here — this is a static image asset generator; the manual DMG-build-and-inspect verification in the test plan is the right substitute.

Overall: small, well-scoped, well-documented change. Nothing blocking — the two nits above are optional polish.

Akarsh-Hegde:
Addressed both nits from the review: simplified the caption y-position (`258 * SCALE` instead of routing through `pt(0, 258)[1]`), and added a clear error if the macOS system fonts this script depends on aren't present, instead of a raw Pillow `OSError`.

PR #427, opened 2026-07-09
feat(tray): add PostHog product analytics for DMG installs

adityaharishch:
## Summary
- Adds `tray/src-tauri/src/analytics.rs`, a minimal best-effort PostHog Cloud client: at most one `app_installed` event ever, and one `daily_usage` event per completed local day (focus hours + worklog generated/approved/rejected counts).
- Raw HTTP POST to PostHog's `/i/v0/e/` capture endpoint only — no `posthog-js`, so session replay/autocapture/surveys/feature flags are never touched. `$geoip_disable` is set on every event since the request comes straight from the user's machine.
- Gated to Canonical (DMG) installs only via `install::detect_install_mode()` — npm/dev/bare runs never send anything.
- Anonymous `distinct_id` + day bookkeeping persisted to `~/.meridian/analytics_state.json`, separate from `settings.json` (never dashboard-visible).
- Wired into the poll loop's existing 60s health tick (`poll/mod.rs`) — the extra work per tick is a cheap file read; the HTTP calls themselves are capped to ~1/day.
- Adds the `uuid` crate (already in the workspace lockfile transitively) for the anonymous id.

## Test plan
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings` / `cargo test` (all pass via pre-commit + pre-push hooks)
- [x] UI build + UI tests (pre-push hook)
- [ ] Manual: run a packaged DMG build, confirm `app_installed` fires once and appears in PostHog Live Events, and that a day-boundary produces one `daily_usage` event with sane `focus_hours`/`worklogs_*` values

## Summary by CodeRabbit

* **New Features**
  * The setup wizard now includes a Notifications permission step and can guide you to the system notifications settings when access is denied.
  * The onboarding flow now shows a formatted local model label and improved completion summaries.

* **Bug Fixes**
  * The wizard no longer requests Input Monitoring during setup, reducing unnecessary permission prompts.
  * Notification-related status handling is now more consistent, including clearer unavailable/denied states.

* **Style**
  * Updated onboarding copy and UI wording for clearer, more consistent user-facing text.

### DIFF
--- a/tray/src-tauri/src/analytics.rs
+++ b/tray/src-tauri/src/analytics.rs
@@ -1,0 +1,261 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+//! Product analytics — PostHog Cloud event capture for the packaged DMG build.
+//!
+//! # What this is
+//! A deliberately tiny, best-effort telemetry client: two anonymous signals —
+//! `app_installed` once per machine, and one `daily_usage` event per completed
+//! local calendar day (focus hours + worklog generated/approved/rejected
+//! counts). Nothing else: a single raw HTTP POST to PostHog's `/i/v0/e/`
+//! capture endpoint, no `posthog-js`, so session replay / autocapture /
+//! surveys / feature flags never activate — that's client-SDK behaviour this
+//! code never touches. `$geoip_disable` is set on every event since the
+//! request originates from the user's own machine, not a backend relay (the
+//! default GeoIP enrichment would otherwise resolve each user's real
+//! location).
+//!
+//! Gated to [`crate::install::InstallMode::Canonical`] (the DMG install) only
+//! — npm-bundle, source/dev, and bare runs never send anything. The anonymous
+//! `distinct_id` plus day bookkeeping live in `~/.meridian/analytics_state.json`,
+//! separate from `settings.json` (never dashboard-editable, never displayed).
+//!
+//! Call volume is intentionally minimal: at most 1 event ever
+//! (`app_installed`) + at most 1 event per calendar day (`daily_usage`),
+//! regardless of how many times the tray restarts or how often the poll loop
+//! ticks — nowhere near PostHog's free-tier event limits.
+//!
+//! # Who calls this
+//! [`crate::poll::run_poll_loop`]'s health tick (`maybe_send_daily_tick`) — a
+//! plain file read cheap enough to run every 60 s; the (at most) two HTTP
+//! calls it can trigger are individually capped as described above.
+//!
+//! # Related
+//! - [`crate::install::detect_install_mode`] — the Canonical/Dev/Bare gate.
+//! - `meridian_core::today::get_today` / `meridian_core::worklogs::get_worklogs`
+//!   — the same readers the dashboard uses, queried here for an already-closed
+//!   past day (never "today", which is still accumulating).
+
+use meridian_core::SqlitePool;
+use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+use std::time::Duration;
+use tracing::Instrument;
+
+/// PostHog Cloud (US region) project token — a **public** write-only capture
+/// key, safe to embed in a shipped client (unlike a personal/secret API key).
+/// Overridable via `MERIDIAN_POSTHOG_KEY` for testing against a different
+/// project.
+const DEFAULT_POSTHOG_API_KEY: &str = "phc_zDagQwSLyR9dwT6LPQ6QfygQkpnXrGAKhrJLv5kBFbvH";
+/// US Cloud ingestion host — matches the project's region.
+const POSTHOG_HOST: &str = "https://us.i.posthog.com";
+
+fn posthog_api_key() -> String {
+    std::env::var("MERIDIAN_POSTHOG_KEY").unwrap_or_else(|_| DEFAULT_POSTHOG_API_KEY.to_string())
+}
+
+/// Persisted analytics bookkeeping. Deliberately its own file — never merged
+/// into `settings.json` (which the dashboard reads/writes/displays).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct AnalyticsState {
+    distinct_id: String,
+    install_event_sent: bool,
+    /// The last LOCAL calendar day ("YYYY-MM-DD") a `daily_usage` event was
+    /// sent for. `None` before the first day boundary has been observed.
+    last_sent_day: Option<String>,
+}
+
+fn analytics_state_path() -> Option<PathBuf> {
+    std::env::var("HOME")
+        .ok()
+        .map(|h| PathBuf::from(h).join(".meridian/analytics_state.json"))
+}
+
+/// Load the state file, creating a fresh one (new random `distinct_id`) if
+/// absent or unparseable. Never errors — a corrupt/missing file just starts a
+/// new anonymous id, same as a genuine first run.
+fn load_or_init_state(path: &std::path::Path) -> AnalyticsState {
+    if let Ok(s) = std::fs::read_to_string(path) {
+        if let Ok(state) = serde_json::from_str::<AnalyticsState>(&s) {
+            return state;
+        }
+    }
+    AnalyticsState {
+        distinct_id: uuid::Uuid::new_v4().to_string(),
+        install_event_sent: false,
+        last_sent_day: None,
+    }
+}
+
+/// Crash-safely persist the state file (temp + rename), matching the pattern
+/// `meridian_core::settings::write_settings_value` uses.
+fn save_state(path: &std::path::Path, state: &AnalyticsState) {
+    let Some(dir) = path.parent() else { return };
+    if let Err(e) = std::fs::create_dir_all(dir) {
+        tracing::warn!(error = %e, "analytics: could not create state dir");
+        return;
+    }
+    let json = match serde_json::to_string_pretty(state) {
+        Ok(j) => j,
+        Err(e) => {
+            tracing::warn!(error = %e, "analytics: could not serialise state");
+            return;
+        }
+    };
+    let tmp = path.with_extension("json.tmp");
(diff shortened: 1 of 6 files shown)


tray/src-tauri/src/analytics.rs
```diff
@@ -0,0 +1,261 @@
...
+                    *w.counts.get("approved").unwrap_or(&0) + *w.counts.get("posted").unwrap_or(&0);
+                let rejected = *w.counts.get("skipped").unwrap_or(&0);
+                let generated: i64 = w.counts.values().sum();
+                (generated, approved, rejected)
+            }
+            Err(e) => {
+                tracing::warn!(error = %e, day, "analytics: worklogs read failed for daily_usage");
+                (0, 0, 0)
+            }
+        };
```
Akarsh-Hegde: Both read failures here are swallowed into zero/empty values and the function still proceeds to `capture("daily_usage", ...)` with those fabricated zeros — so a transient DB hiccup reports "0 focus hours, 0 worklogs" to PostHog as if that were real, rather than skipping the event. Combined with the caller unconditionally advancing `last_sent_day` right after (below), that day's actual data is lost for good, not just delayed.

Suggest returning `bool` (or `Result<(), ()>`) from `send_daily_usage` so the caller can skip advancing `last_sent_day` on failure and retry next tick instead.

tray/src-tauri/src/analytics.rs
```diff
@@ -0,0 +1,261 @@
...
+            // yet, so there's nothing to report.
+            state.last_sent_day = Some(today);
+            dirty = true;
+        }
+        Some(prev) if *prev != today => {
+            let day = prev.clone();
+            send_daily_usage(app, pool, &state.distinct_id, &day).await;
+            state.last_sent_day = Some(today);
+            dirty = true;
+        }
```
Akarsh-Hegde: This advances `last_sent_day` (and marks `dirty`) even when `send_daily_usage` hit a read error and sent fabricated zeros (see comment below on lines 213-234). Once this runs, that day can never be retried — worth gating the advance on `send_daily_usage` actually succeeding.

tray/src-tauri/src/poll/mod.rs
```diff
@@ -113,6 +113,7 @@ pub async fn run_poll_loop(app: tauri::AppHandle, state: Arc<Mutex<AppState>>) {
             refresh_current_task(pool, &state).await;
             if do_health {
                 refresh_today(pool, &state).await;
+                crate::analytics::maybe_send_daily_tick(&app, pool).await;
```
Akarsh-Hegde: This is awaited inline ahead of `drain_notifications`/`emit_notices`/`emit_update`/tray-icon refresh in the same tick. On the rare tick where this actually fires (≤1 install event ever + ≤1/day), a slow or hanging PostHog response (5s timeout) delays those more user-visible steps by that much. Since it's this infrequent, `tauri::async_runtime::spawn`-ing it would decouple it from the rest of the tick for free:
```rust
if do_health {
    refresh_today(pool, &state).await;
    let app = app.clone();
    let pool = pool.clone();
    tauri::async_runtime::spawn(async move {
        crate::analytics::maybe_send_daily_tick(&app, &pool).await;
    });
}
```

Akarsh-Hegde (commented):
## Review

**Scope note:** reviewing the diff as it currently stands (5 files: `Cargo.lock`, `tray/src-tauri/Cargo.toml`, `tray/src-tauri/src/analytics.rs`, `lib.rs`, `poll/mod.rs`) — the earlier CodeRabbit walkthrough on this PR references a lot of files (`data.ts`, `steps.tsx`, `NOTIFICATIONS.md`, `tray.rs`, etc.) that aren't in the current diff at all, so that review looks stale relative to a rebase/force-push. Its `data.ts` em-dash comment in particular no longer applies since that file isn't touched here anymore.

### Design/product — please confirm this is a deliberate call
This is the first thing in the codebase that phones home to a third party by default, with no user-facing disclosure or opt-out anywhere. Two things make me want to double-check intent rather than just wave it through:
- The setup wizard (Permissions step) tells the user *"Everything is processed on your Mac. Meridian has no servers to send to."* That line becomes literally false the moment this ships for Canonical installs — worth updating, or scoping the claim to what it's actually true for.
- The project's existing telemetry (OpenObserve traces/logs) is deliberately conservative: `is_canonical_install()` hard-blocks shipping for any packaged install, and even a Dev/Bare install requires explicit `otlp_enabled` + credentials before anything leaves the machine (`CLAUDE.md`'s "Telemetry: local-only capture, dev-only shipping" section). This PR does the opposite for PostHog — unconditional for every Canonical install, no settings toggle, no first-run disclosure.

The anonymization here is genuinely good (`$geoip_disable`, no PII in properties, public write-only key, tiny event volume) — this isn't a "rip it out" objection, just: was a settings toggle / privacy-policy update / disclosure considered and deliberately skipped, or does this need one before it ships? One nuance worth a doc note either way: `$geoip_disable` stops PostHog from attaching resolved location *properties* to the event, but PostHog's ingestion endpoint still sees the raw source IP at the network level — it's not full IP anonymity, just no derived location on the event.

### Correctness — worth fixing before merge
`send_daily_usage` swallows a `get_today`/`get_worklogs` read failure and substitutes zeros (lines 216-219, 230-233), and the caller (`maybe_send_daily_tick`, lines 191-196) advances `last_sent_day` unconditionally right after. Net effect: a transient DB read hiccup on the day boundary sends PostHog a `daily_usage` event claiming zero focus time and zero worklogs — actively wrong data, not just missing data — and that day is now permanently unrecoverable since `last_sent_day` already moved past it. Left inline on the specific lines. Suggest: have `send_daily_usage` report success/failure, and only advance `last_sent_day` (and set `dirty`) when it actually sent, so a bad day gets retried next tick instead of being silently reported as zero and then forgotten.

### Minor / non-blocking
- `poll/mod.rs:116` — `maybe_send_daily_tick` is awaited inline ahead of `drain_notifications`/`emit_notices`/tray icon updates in the same tick. On the rare tick where it actually fires, a slow/hanging PostHog response (5s timeout) delays those more user-visible steps. Given the call is genuinely rare (≤1 install event ever + ≤1/day), spawning it (`tauri::async_runtime::spawn`) would decouple it for free. Left inline.
- Multi-day gaps aren't backfilled — if the tray isn't running across more than one local-day boundary, only the most recent closed day gets reported; the days in between silently never get a `daily_usage` event. Given the "best-effort" framing this may be an accepted simplification, but worth a line in the module doc saying so explicitly (it currently reads as if every day gets covered).
- The day-rollover decision (`None` → arm the first day, `Some(prev) != today` → send + advance) is pure state-transition logic that's fully separable from the network/file I/O around it, and would be easy to cover with a couple of `#[cfg(test)]` cases the way `hygiene.rs`/`runner.rs` do for similar state machines — right now the only way to exercise it is a live packaged run.
- Test plan's last checkbox (manual DMG run confirming events actually land in PostHog Live Events) is unchecked — given there's no automated coverage of the actual capture path, I'd want that checked before merge, not just the read-path logic.

Nothing here blocks review in the CI-checks sense, but I'd want the correctness fix in before merge, and the design question answered one way or the other (ship as-is with a documented decision, or gate it behind a setting) rather than left implicit.

adityaharishch:
Pushed 749ffdc0 addressing the review feedback:

**Fixed (real bugs):**
- `send_daily_usage` no longer swallows a `get_today`/`get_worklogs` read failure into fabricated zeros. It now returns whether the send actually succeeded, and `maybe_send_daily_tick` only advances `last_sent_day` on success — a transient DB hiccup on the day boundary retries next tick instead of reporting "0 focus hours, 0 worklogs" and permanently losing that day.
- The analytics tick is now spawned (`tauri::async_runtime::spawn`) off the poll loop's health tick instead of awaited inline, so a slow/hanging PostHog response (5s timeout) can no longer delay notification drain / tray icon refresh on the rare tick it actually fires.
- Extracted the day-rollover decision into a pure `day_rollover_action` helper with 4 unit tests (first-observation, same-day no-op, day-boundary, multi-day-gap) — it's exercised without a live DB/HTTP client now.
- Module doc now states explicitly that multi-day gaps are never backfilled (only the most-recently-closed day is ever reported) — this was already the intended design, just wasn't spelled out.

**Not resolved silently — needs a product call from @Akarsh-Hegde/@adityaharishch:**
The design/disclosure question (setup wizard currently says *"Everything is processed on your Mac. Meridian has no servers to send to"* — that becomes false for Canonical installs once this ships) is a genuine product decision, not something I'll resolve unilaterally in code. Flagging it back rather than guessing: do we (a) update that wizard copy to scope the claim correctly, (b) add a settings toggle/opt-out, or (c) ship as-is with the decision documented somewhere? Also noting your nuance is correct and worth keeping in mind: `$geoip_disable` suppresses derived location *properties*, PostHog's edge still sees the raw source IP at the network level.

**Not applicable:** the `ui/app/setup/data.ts` em-dash comment — confirmed via `gh pr diff --name-only` that file isn't part of the current 5-file diff (stale from an earlier CodeRabbit pass before this PR was rebased against main).

**Still open:** the manual DMG-run verification (events actually landing in PostHog Live Events) — can't do that from here, needs a packaged build against the real project.

adityaharishch:
Resolved the open design question from the review: decision is to disclose truthfully rather than gate behind a setting or leave the copy overbroad.

Pushed 6e55c957 — the Permissions step's privacy claim in \`ui/app/setup/steps.tsx\` (which read *"Everything is processed on your Mac. Meridian has no servers to send to."*) is now:

> "Your screen, tasks, and worklogs stay on this Mac and are never uploaded. We send anonymous usage stats (like daily focus time) to improve Meridian - never your content."

This keeps the (still true) part of the original claim — no screen/OCR/task/worklog content ever leaves the machine — while being honest that anonymous usage stats do go to PostHog for Canonical/DMG installs. Scoped it to this one string only; the other "never uploaded" claims elsewhere in the wizard (screen-reading step, on-device model step) were left untouched since they're specifically about screen content, which is genuinely never uploaded regardless of this feature.

PR #432, opened 2026-07-10
feat(uninstall): add an in-app uninstall wizard

Akarsh-Hegde:
## What does this PR do?

Deleting `Meridian.app` from Finder does nothing on its own: the `com.meridiona.daemon` and `com.meridiona.a11y-helper` launchd agents are staged at `~/.meridian/bin/` and keep running forever, and macOS never revokes an Accessibility/Screen Recording/Input Monitoring TCC grant just because the requesting app was deleted (the grant is anchored to the code-signing identity, not the file's existence, and both agents live outside the app bundle by design). This adds a tray menu action, **"Uninstall Meridian…"**, that opens an in-app wizard to tear everything down cleanly *before* the user drags the app to the Trash, with independent checkboxes for what else to remove, and a follow-up step pointing at System Settings for the permission grants that survive on their own.

- **`src/uninstall.rs`**: extends the existing `meridian uninstall` CLI with a `--json` output mode and three independent, additive flags — `--remove-data`, `--remove-runtime`, `--remove-models` (`--purge` remains a shorthand for all three). Ports the HuggingFace model-cache allowlist from `scripts/meridian-cli.sh`'s bash uninstaller (`src/uninstall.rs` previously had no model-cache cleanup at all) and adds icon-cache/telemetry/onboarded-marker cleanup to the data category.
- **`tray/src-tauri/src/commands/uninstall.rs`** (new): `get_uninstall_plan`/`execute_uninstall` Tauri commands, shelling out to the staged `meridian` binary rather than duplicating the file scan (same pattern as `commands::parents::get_ticket_parents`). `execute_uninstall` stops the tray's in-process MLX server first (`mlx_server::stop_server`, now `pub(crate)`) so the runtime directory isn't deleted out from under a live process.
- **`tray/src-tauri/src/tray.rs`**: new "Uninstall Meridian…" menu item (separated from the rest of the menu) opening a new `uninstall` window; registered in `capabilities/default.json`.
- **`ui/app/uninstall/page.tsx`** (new): the wizard UI — plan → checkboxes (data / Python+MLX runtime / downloaded models, each showing live item counts) → confirm → execute → results + a "Revoke permissions" panel with shortcuts into System Settings (reuses `open_permission_pane`).
- **`ui/lib/api-types.ts`**: adds `UninstallItem`/`UninstallPlan`/`UninstallResult` types per the porting convention.

## How was it tested?

- [x] `cargo test` — **could not run against the real workspace in this environment**: any `cargo` invocation here resolves the whole workspace, and `tray/src-tauri` depends on a private git fork (`Meridiona/screenpipe-fork`) unreachable without credentials this sandbox doesn't have. Instead, I copied `src/uninstall.rs`'s logic into an isolated scratch crate (serde + serde_json only, no private deps) and ran its unit tests there — all 6 pass, including new tests for the data/runtime/model item-listing helpers and disjointness between categories.
- [x] `cargo clippy -- -D warnings` — same scratch-crate approach; clean, no warnings.
- [x] `cargo fmt --check` — ran directly against every changed/new Rust file (`rustfmt` only parses syntax, doesn't need dependency resolution); clean.
- [x] UI builds (`cd ui && npm run build`) — full Next.js build including TypeScript checking passes; the new `/uninstall` route compiles and statically generates alongside the existing routes.
- [x] Manually tested the relevant flow — exercised `meridian uninstall`'s new `--dry-run`/`--json`/`--remove-*` flags end-to-end against a fake `$HOME` (agent-plist removal, scoped `--remove-models` execution leaving data/runtime untouched, full `--purge --json` plan).

**Please re-run `cargo test`, `cargo clippy -- -D warnings`, and `cargo fmt --check` for both the root crate and `tray/src-tauri` in CI (or an environment with access to the private fork) before merging** — I could not do this myself here.

## Checklist

- [x] Branch named `type/short-description` — using the pre-assigned session branch (`claude/mlx-model-download-comparison-0bov2m`) per this session's instructions rather than opening a second branch.
- [x] Commit messages follow Conventional Commits
- [x] New `.rs`/`.ts`/`.tsx` files start with the file-header comment
- [x] New migrations are numbered and don't modify existing ones — N/A, no migrations in this change
- [x] No secrets, credentials, or `.env` files committed

## Related issues

None — built from a conversation about what happens to permissions/leftover processes when the app is deleted.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FJLr6No6vghNDsC6X9PRTk)_


adityaharishch:
## Review

**Overview:** Adds a tray menu item "Uninstall Meridian…" that opens an in-app wizard (`/uninstall`) to cleanly tear down launchd agents, the MLX server, and (optionally) user data/runtime/models before the user drags `Meridian.app` to the Trash — since deleting the bundle alone leaves the daemon + a11y-helper agents running and never revokes TCC grants. Extends `meridian uninstall` (CLI) with `--json`, `--remove-data`, `--remove-runtime`, `--remove-models` flags; the tray shells out to this binary rather than duplicating the file scan (good adherence to "single source of truth" convention).

**Correctness**
- `run_json`'s `dry_run || !yes` gate correctly refuses to execute non-interactively unless both flags are explicit — matches the documented contract, and `execute_uninstall` always passes `--yes` (never `--dry-run`), `get_uninstall_plan` always passes `--dry-run` (never `--yes`), so the two Tauri commands can't cross wires.
- `remove_path` correctly uses `symlink_metadata` so a symlink (even to a directory) is unlinked rather than recursed into — avoids accidentally deleting through a symlink's target.
- Confirmed `stop_server` is called before the CLI removes the runtime dir, and the daemon agent is `bootout`'d before its files are removed inside the CLI's own `run_json`/`run` — correct ordering for both processes.
- Deleting the currently-running `meridian` binary out from under the very process executing it is safe under Unix unlink semantics (the open inode persists until the process exits) — worth calling out only because it looks alarming at first glance.
- One thing to double check: `remove_data && remove_runtime` (regardless of `remove_models`) triggers a full `rm -rf ~/.meridian` in both `run()` and `run_json()`, not just the itemized `data_items`/`runtime_items` lists. This is intentional per the comment (the two checked boxes cover everything that can live under `~/.meridian` since models live in the separate HF cache), but the wizard UI doesn't tell the user that checking both "Meridian data" + "Python + MLX server" wipes anything else under `~/.meridian` too, not just what's itemized. Minor UX gap, not a bug.
- `MODEL_CATALOG` faithfully mirrors `scripts/meridian-cli.sh`'s existing bash allowlist, but that allowlist (and this port) is missing the embedder model (`mlx-community/Qwen3-Embedding-0.6B-8bit`, used by `/distill_hour` per `services/agents/model_registry.py`) — so `--remove-models` will silently leave that model's cache directory behind. Pre-existing gap (not introduced by this PR), but worth fixing here since this is the one place actively promising "remove downloaded AI models."

**Style / conventions**
- New/changed `.rs`/`.tsx` files carry the required file header; docs follow the "Who calls this / Related" convention thoroughly.
- `tray.rs`'s new `open_uninstall_window` sets `.title("Meridian — Uninstall")` using an em-dash. CLAUDE.md's hard rule requires a plain hyphen in all user-facing text, explicitly including window titles, and the codebase's other newer window already follows it (`"Meridian - Setup"` at tray.rs:177) — please switch to `"Meridian - Uninstall"` to match (and note the pre-existing `"Meridian — Dashboard"` has the same issue, though that's out of scope here).
- Reuses `Btn`/`Check`/`Kicker`/`Row`/`Spinner` atoms from the setup wizard rather than reinventing — consistent with the porting convention.

**Test coverage**
- Good new unit tests for `data_items`/`runtime_items`/`model_items`/`remove_path` disjointness and existence-filtering.
- No unit tests for the new tray-side `commands/uninstall.rs` (JSON parsing, timeout, error surfacing), but this matches the existing precedent for tray command modules (manually verified instead).
- Author explicitly flags in the PR body that `cargo test`/`clippy`/`fmt` could only be verified against an isolated scratch crate (private git-fork dependency unreachable in their sandbox) — **CI must confirm these pass for real before merge**, as requested.

**Housekeeping**
- The diff includes unrelated version-bump churn (`CHANGELOG.md`, `Cargo.toml`/`Cargo.lock`, the three `package.json`s, `services/pyproject.toml` — all `1.69.1` → `1.70.0`) that isn't mentioned anywhere in the PR description. This looks like the branch was cut before a release-please bump landed on `pre-main` and hasn't been rebased since. Worth rebasing onto latest `pre-main` before merge so the diff is scoped to just the uninstall feature.

**Risk:** Low-medium. The feature is opt-in (menu action) and every destructive step is best-effort/non-fatal by design; the main things to verify before merge are (1) a clean rebase to drop the stray version-bump diff, and (2) CI actually running `cargo test`/`clippy`/`fmt` since the author couldn't in their sandbox.

Akarsh-Hegde:
@adityaharishch thanks for the thorough review — pushed fixes for all three actionable findings:

- **`MODEL_CATALOG` missing embedder/reranker**: added `models--kerncore--Qwen3-Reranker-0.6B-MLX-4bit` and `models--mlx-community--Qwen3-Embedding-0.6B-8bit` (verified against `services/agents/model_registry.py`'s `RERANKER`/`EMBEDDER` specs — added a comment pointing back there so these two stay in sync with that source of truth).
- **Em-dash in `.title("Meridian — Uninstall")`**: fixed to `"Meridian - Uninstall"`. While in there I also caught a few em-dashes I'd introduced in the wizard's own JSX copy (`ui/app/uninstall/page.tsx`) that weren't called out — fixed those too. Left the pre-existing `"Meridian — Dashboard"` alone as suggested (separate PR).
- **UX gap on `remove_data && remove_runtime`**: added a conditional note in the wizard that appears only when both boxes are checked, clarifying it removes everything under Meridian's data folder, not just the itemized list.

Also merged in latest `pre-main` (the base was retargeted from `main` — confirmed `pre-main` is the actual intended target) — the stray version-bump diff is gone now that the merge-base is current.

---

PR #436, opened 2026-07-13
feat: email one-time-code sign-in via Clerk

adityaharishch:
## Summary
- Adds passwordless email-code sign-in (Clerk, via `tauri-plugin-clerk`) as the setup wizard's Sign-in step — no secret key ships in the binary, only the public publishable key.
- Adds a matching Settings → Account identity row (avatar, email, Sign out), sharing the same widget code as onboarding.
- Fixes two real bugs found while dogfooding: a Clerk error-shape bug that broke the sign-up fallback (`error.errors[0].code`, not `error.code`), and a double-submit race on code verification. Adds a 30s resend cooldown (industry standard).
- PostHog analytics now also runs in dev mode (tagged with a new `channel` property) and fires an immediate identify event on sign-in, instead of waiting up to a day for the next scheduled capture.
- Every packaged/downloadable build (staging and prod channel alike) now gets the production Clerk instance; only a source checkout uses the dev instance — wired via `CLERK_PUBLISHABLE_KEY_PROD` in both release workflows.

## Test plan
- [x] `cargo fmt --check && cargo clippy --all-targets -- -D warnings && cargo test` (57/57 passing)
- [x] `npm run build` (static export succeeds)
- [x] `bun test` (175/175 passing)
- [x] Manually verified sign-in/sign-up/resend/sign-out flow end to end against the dev Clerk instance
- [x] Verified `meridiona.com` DNS + SSL for the production Clerk instance
- [ ] Manual smoke test on a `pre-main` staging build once merged (confirms the prod Clerk key resolves correctly in a packaged build)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -171,6 +171,24 @@
 # POSTHOG_API_KEY=phc_your_project_api_key_here
 
 # ---------------------------------------------------------------------------
+# Clerk email sign-in (setup wizard's Sign-in step, tray/DMG build only)
+#
+# A release build bakes the Clerk publishable key in at compile time (CI-only,
+# from the MERIDIAN_CLERK_PUBLISHABLE_KEY repo secret; never committed to
+# source) — see tray/src-tauri/src/commands/account.rs. A LOCAL SOURCE BUILD
+# has no baked-in key, so set CLERK_PUBLISHABLE_KEY below and the tray will
+# pick it up on the next `cargo run`/`npm run tauri dev` — unlike most of this
+# file, the tray reads this ONE key straight out of this file itself (see
+# `install::env_key_from_path`), not via dotenvy auto-load, so no shell
+# export is needed. Get the dev instance's key with `clerk env pull` after
+# `clerk link --app app_3GIemyKuI07XdI7jFuYxpBqy9y3` (publishable key only —
+# this is a public identifier, safe to keep here; never set CLERK_SECRET_KEY,
+# the tray has no server-side use for it).
+# ---------------------------------------------------------------------------
+
+# CLERK_PUBLISHABLE_KEY=pk_test_your_publishable_key_here
+
+# ---------------------------------------------------------------------------
 # Python LLM backend — cloud fallback for the coding-agent summariser (Claude/
 # Codex paths). The MLX classifier server ignores these.
 # ---------------------------------------------------------------------------
--- a/.github/workflows/release-staging.yml
+++ b/.github/workflows/release-staging.yml
@@ -62,6 +62,11 @@ jobs:
       # Baked into the staging binary at compile time (see release.yml); same
       # secret. Absent → PostHog analytics simply disabled in the staging build.
       MERIDIAN_POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
+      # Baked into the staging binary at compile time (see release.yml); same
+      # secret — staging is downloadable/installable too, so it gets Clerk
+      # PRODUCTION, not the dev instance (only a source checkout uses dev).
+      # Absent → sign-in unavailable in the staging build.
+      MERIDIAN_CLERK_PUBLISHABLE_KEY: ${{ secrets.CLERK_PUBLISHABLE_KEY_PROD }}
       # Minisign key for the DMG auto-updater: `tauri build` (createUpdaterArtifacts)
       # signs Meridian.app.tar.gz so the installed staging app can verify the
       # update against the pubkey baked in tauri.conf.json. REQUIRED here (unlike
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,6 +57,16 @@ jobs:
       # in a shipped client. Set the POSTHOG_API_KEY repo secret; absent →
       # analytics compiles in disabled (see analytics::posthog_api_key).
       MERIDIAN_POSTHOG_API_KEY: ${{ secrets.POSTHOG_API_KEY }}
+      # Baked into the release binary at compile time (consumed by option_env! in
+      # tray/src-tauri/src/commands/account.rs's DEFAULT_CLERK_PUBLISHABLE_KEY).
+      # Clerk PRODUCTION instance's publishable key (meridiona.com) — a public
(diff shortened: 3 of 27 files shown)


adityaharishch (commented):
## Review — PR #436 (email OTC sign-in via Clerk)

Solid, well-documented PR. Clean module decomposition, good doc comments, disciplined use of plain hyphens in all user-facing copy (the em-dashes are confined to code comments/docs, which is fine). One likely ship-blocker and a couple of things to confirm.

### On the "baked-in production Clerk key" concern — clear ✅
I looked hard at `b4d38908` ("bake the production Clerk key into every packaged build") since it sounds alarming. It is **not** a secret: `MERIDIAN_CLERK_PUBLISHABLE_KEY` ← `CLERK_PUBLISHABLE_KEY_PROD` is a Clerk **publishable** key (`pk_live_/pk_test_`), which is a public client identifier by design and safe to embed in a shipped binary — exactly like the PostHog key already baked in. The code never reads `CLERK_SECRET_KEY` (correctly documented in `account.rs`), `.env.example` ships only a `pk_test_...` placeholder, and I found no `sk_`/secret material anywhere in the diff. No secret is leaked.

### Blocking / must-verify before merge

**1. Capability HTTP allowlist only permits the DEV Clerk domain — will very likely break sign-in in every packaged build.**
`tray/src-tauri/capabilities/clerk.json` scopes the `http:default` allow to:
```json
"allow": [{ "url": "https://*.clerk.accounts.dev/*" }]
```
That host is the **development** instance's Frontend API. But per this PR, *every packaged/downloadable build (staging + prod) uses the production Clerk instance* (`meridiona.com`), whose Frontend API is a custom domain (e.g. `clerk.meridiona.com`), which is **not** covered by this allowlist. Since `tauri-plugin-clerk` routes Clerk's session/API calls through `tauri-plugin-http`, the scope will block those requests in exactly the builds that ship — while dev (`*.clerk.accounts.dev`) works fine, which is why it passed local testing. This lines up with the one unchecked test-plan item ("Manual smoke test on a `pre-main` staging build"). Please add the production custom domain (and confirm the exact host) to the allow list and verify on a packaged staging build before merge. This is the highest-risk item in the PR.

### Non-blocking, but please address

**2. Privacy copy vs. sending email to PostHog.**
`EmailCodeForm.tsx` tells the user: *"…your email is never shared with anyone."* But `analytics.rs` now attaches the signed-in email as a PostHog `$set` on every event and fires an immediate `person_identified`. PostHog is a third-party processor, so "never shared with anyone" is inaccurate as written. Either soften the copy (e.g. "we don't share your email with other users") or reconsider sending PII to PostHog. Worth a second look for consent/legal too, since this flips analytics from anonymous to identified.

**3. Analytics now runs in every install shape, including source/dev.**
Removing the `Canonical`-only gate means `cargo run`/`tauri dev` sessions now emit events (and dev emails) to PostHog. The `channel` tag makes them filterable, which is the stated intent, but confirm that emitting real developer emails from local checkouts is desired.

**4. Doc drift — `signin-impl.tsx` referenced but doesn't exist.**
`account.rs`'s module doc and `steps.tsx`'s comment both point at `ui/app/setup/signin-impl.tsx`; the actual implementation lives in the `signin/` directory (`SignInWidget.tsx`, `EmailCodeForm.tsx`, etc.). Looks like a rename during development; update the references.

### Minor / confirm
- **Resend cooldown is client-side only** (`RESEND_COOLDOWN_S`) — fine as UX, but real abuse protection for the email-send endpoint relies on Clerk's server-side rate limiting. Just confirming that's understood.
- `save_account_email` validates only `contains('@')` — acceptable as a boundary check since Clerk owns real validation.
- Rust side looks clean: `Result` + contextful error strings, crash-safe temp+rename write, `#[tracing::instrument]` on the commands, no `unwrap()`, no clippy-arg issues. File headers present on all new `.rs`/`.ts`/`.tsx` files.

Nice work overall — the double-submit ref guard and the Clerk error-shape fix (`error.errors[0].code`) are good catches. Main thing is confirming #1 on a real packaged build.

Akarsh-Hegde:
Bottom line: this PR is not ready to merge. The capability-scope issue in particular is the kind of thing that would pass all local/dev testing and then silently break sign-in the moment it hits a packaged staging build — worth fixing and doing that packaged-build smoke test (still an unchecked box in the PR's own test plan) before merging. Happy to pull the branch and check what the actual Clerk production Frontend-API host is, so we can propose the exact fix to `clerk.json`, if that'd help.

adityaharishch:
Thanks for the thorough review — went through each point:

**#1 (blocking — capability allowlist)**: Confirmed real and fixed in `19a469d2`. `clerk.json` only allowlisted `*.clerk.accounts.dev` (dev instance); added `clerk.meridiona.com` and `accounts.meridiona.com` (production instance's Frontend API + Account Portal) to the `http:default` allow list. Good catch — this would have silently broken sign-in on every packaged build. Still want an actual packaged staging-build smoke test before merge, per your note and the PR's unchecked test-plan item — haven't done that yet.

**#2 (privacy copy)**: Also fixed in `19a469d2`. Changed "your email is never shared with anyone" → "your email is never sold or shared with outside companies," which is accurate given PostHog is an internal-analytics processor, not a third party we're sharing data *to*. Agreed the original claim was flat-out wrong once `$set`/`person_identified` shipped.

**#3 (analytics now running in dev mode)**: Intentional, not a bug — explicit user request mid-session to have PostHog capture in both dev and prod, specifically so dev-run testing is also visible. The `channel` property (dev/staging/prod) exists exactly so this is filterable. Confirmed as desired behavior, no change needed.

**#4 (doc drift — signin-impl.tsx)**: Confirmed and fixed in `19a469d2` — updated `account.rs`'s module doc and `steps.tsx`'s comment to point at the actual `signin/` module files instead of the deleted monolith.

**Minor items**: agreed on both — the resend cooldown is deliberately client-side UX polish only, with Clerk's own server-side rate limiting as the real backstop; and `save_account_email`'s `contains('@')` check is intentionally a boundary check, not real validation (Clerk owns that).

All four commits (original + this fix) pass `cargo fmt`/`clippy`/`test`, `npm run build`, and `bun test` locally. Will do the packaged staging build test next.

PR #437, opened 2026-07-13
feat(tray): auto-open the Plan your day view once per day

Akarsh-Hegde:
## What

Every day, at the user's first activity - login, or the first poll tick after a new day starts (the tray usually wakes from sleep already running, so the 30s poll tick is the trigger, not a launch hook) - the tray opens the dashboard window on the **Plan modal** so the day starts with a plan.

## How

- **`poll/plan_auto_open.rs`** - the once-per-day check, gated cheapest-first:
  1. `~/.meridian/plan_auto_opened` date marker already holds today - done (one file stat on the common path)
  2. `auto_open_plan` setting off - skip
  3. not onboarded - never open over the first-run wizard
  4. today's plan already confirmed/skipped in `daily_plan_meta` (e.g. via the notification nudge) - settle the day **without** opening
  
  The marker is written *before* the window opens, so a `KeepAlive` crash-relaunch loop can never re-open repeatedly. No hour gate by design (product call: first activity of the day, any time).
- **`deep_link.rs` + `take_pending_deep_link`** - pull-model handoff: a freshly created dashboard webview misses any event emitted before its listener exists, so the tray parks the target (`/plan`) in managed state and `MeridianTimelineShell` pulls it on mount; a `dashboard-navigate` emit covers the already-open window (which won't remount). Double delivery is idempotent.
- **Notification click-through fix (pre-existing bug)** - `record_notification_response` now hands the row's `deep_link` through the same mechanism, so tapping a "Plan your day" toast lands on the Plan view instead of the default timeline.
- **Settings** - `auto_open_plan` (default on) in `RuntimeSettings` + `ui/lib/settings.ts`, with an "Open planner each day" toggle in Settings -> Notifications, deliberately outside the `notifications_enabled` gate (window behaviour, not a toast).
- **`meridian_core::plan::plan_handled`** - public `daily_plan_meta` read backing gate 4.

## Testing

- `cargo fmt` + `cargo clippy --workspace -- -D warnings` clean
- All Rust tests pass (daemon, meridian-core, tray), incl. new ones: `plan_handled` (no table / unconfirmed / confirmed / skipped) and the marker date logic (stale date, whitespace, fs round-trip)
- Next.js static build + all 175 UI tests pass
- Local demo: delete `~/.meridian/plan_auto_opened` (or write yesterday's date while the tray runs) - fires within one 30s tick; with today's plan already confirmed, the day settles silently and only the marker is written

## Risks / notes

- The open takes focus (dashboard opens maximized) - once per day by design; if too aggressive, a follow-up can skip `set_focus` on the mid-day boundary path.
- An unattended overnight wake (Power Nap) can fire under the lock screen - the window simply greets the user at unlock.

### DIFF
--- a/meridian-core/src/lib.rs
+++ b/meridian-core/src/lib.rs
@@ -35,6 +35,10 @@ pub mod settings;
 /// Notification delivery policy + native pending queue (ported from lib/notifications.ts).
 pub mod notifications;
 
+/// The `~/.meridian/plan_auto_opened` marker format — written by the tray's
+/// daily planner auto-open, read by the daemon's plan-nudge hold-back.
+pub mod plan_marker;
+
 /// In-process capture-frame writer (Gap-2 Bucket 2). Inverted ownership: the
 /// tray writes `capture_frames`, the daemon's ETL reads it.
 pub mod capture;
--- a/meridian-core/src/plan_marker.rs
+++ b/meridian-core/src/plan_marker.rs
@@ -1,0 +1,91 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+//! The `~/.meridian/plan_auto_opened` marker — the shared record of when the
+//! tray last auto-opened the daily planner.
+//!
+//! One file, two consumers with different questions:
+//! - The **tray** (writer) asks "did I already open the planner today?" —
+//!   [`opened_today`] — to fire at most once per local day.
+//! - The **daemon** (reader) asks "how long ago did it open?" — [`opened_at`]
+//!   — to hold the `plan.nudge` reminder back for a grace period after the
+//!   auto-open instead of toasting over the window that just opened.
+//!
+//! The content is a single RFC3339 local timestamp (e.g.
+//! `2026-07-13T09:18:52+05:30`); its first 10 chars are the local date, which
+//! keeps [`opened_today`] a plain prefix check and stays readable in a shell.
+//! This module owns the format so the two sides can never drift.
+//!
+//! # Who calls this
+//! - `tray/src-tauri/src/poll/plan_auto_open.rs` — path + stamp + opened_today.
+//! - The daemon's `src/daily_plan.rs::maybe_nudge` — path + opened_today +
+//!   opened_at (the nudge hold-back).
+//!
+//! # Related
+//! - [`crate::plan`] — `plan_handled`, the sibling "day already planned" gate.
+
+use chrono::{DateTime, FixedOffset, Local};
+use std::path::{Path, PathBuf};
+
+/// File name under `~/.meridian` (same convention as the `onboarded` marker).
+pub const MARKER_FILE: &str = "plan_auto_opened";
+
+/// `<meridian_dir>/plan_auto_opened` — callers pass `~/.meridian`.
+pub fn marker_path(meridian_dir: &Path) -> PathBuf {
+    meridian_dir.join(MARKER_FILE)
+}
+
+/// The marker content for an auto-open happening `now`: an RFC3339 local
+/// timestamp whose first 10 chars are the local date.
+pub fn stamp(now: &DateTime<Local>) -> String {
+    now.to_rfc3339_opts(chrono::SecondsFormat::Secs, false)
+}
+
+/// True when the marker records an auto-open on `today` (local `YYYY-MM-DD`).
+/// A prefix match, so it accepts both the timestamp form and a legacy bare
+/// date; whitespace-tolerant; empty/missing/garbage content → `false`.
+pub fn opened_today(marker_contents: &str, today: &str) -> bool {
+    marker_contents.trim().starts_with(today)
+}
+
+/// The instant the auto-open happened, if the marker carries a parseable
+/// timestamp. `None` for a legacy bare-date or garbage marker — callers that
+/// measure elapsed time treat that as "age unknown".
+pub fn opened_at(marker_contents: &str) -> Option<DateTime<FixedOffset>> {
+    DateTime::parse_from_rfc3339(marker_contents.trim()).ok()
+}
+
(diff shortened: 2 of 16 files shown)


Akarsh-Hegde:
Added in 4f0f90be, following the interplay review with the existing daemon nudge:

**The plan nudge is now the second-chance reminder, not a duplicate.** Previously (with this PR's auto-open) a typical morning fired both surfaces at once - the planner window at first activity, and the daemon's 'Plan your day' toast a tick later. Now:

- The `~/.meridian/plan_auto_opened` marker records the open's **timestamp** (format shared via `meridian_core::plan_marker` so tray writer + daemon reader can't drift).
- `maybe_nudge` holds back while the auto-open was **less than 1h ago** (snooze-1h parity). A planner dismissed without confirming still gets its reminder an hour later; confirmed/skipped plans expire the nudge as before.
- Auto-open off / tray not running / stale marker -> nudge behaves exactly as today (it is the fallback prompt).
- Settings copy clarified ('Plan your day reminder' - explains the 1h wait).

Toast tap and its 'Open Plan' button (action id `open`) both land on the Plan modal via the deep-link handoff from the first commit.

Akarsh-Hegde:
4222... follow-up (see latest commit): the reminder clock now restarts at planner **dismissal** - closing the Plan modal without confirming re-stamps the marker (`plan_dismissed` command -> `plan_marker::restamp_if_today`), so the 'Plan your day' push notification lands 1h after the user closed the planner, not 1h after it auto-opened. Guarded so a manual open/close on a day the auto-open never fired can't suppress the nudge.

adityaharishch (commented):
## Review — feat(tray): auto-open the Plan your day view once per day

Solid, well-documented change. The marker-file design (write-before-open to survive `KeepAlive` relaunch loops, cheapest-gate-first ordering, shared `plan_marker` format owned in one module) is clean, and the once-per-day state correctly persists across daemon/tray restarts via `~/.meridian/plan_auto_opened`. Test coverage on the pure logic (`nudge_held_back`, `opened_today`/`opened_at`, `restamp_if_today`, `plan_handled`) is good. Convention checks pass: all three new `.rs` files carry the file-header line, user-facing copy in `NotificationsSection.tsx` uses plain hyphens only, `tracing` calls use structured fields, no secrets, all files well under 500 lines.

### Blocking / please confirm

**1. Stale pending deep-link can misroute a later manual dashboard open.** (`deep_link.rs`, `poll/plan_auto_open.rs`, `commands/notifications.rs`)

`open_native_dashboard` (`tray.rs:98`) destroys the webview on close (`WindowEvent::Destroyed`), so every reopen is a fresh mount that calls `take_pending_deep_link`. The auto-open and notification-tap paths both do `set_pending("/plan")` **and** `emit_to("dashboard", "dashboard-navigate", …)` unconditionally. When the dashboard is **already open**, the emit delivers and shows the modal, but the pending slot is never consumed (no remount) — so `"/plan"` stays parked. If the user then closes the dashboard and later reopens it manually (tray menu), the fresh mount pulls the stale `"/plan"` and jumps to the Plan modal unexpectedly.

The pull and push halves double-deliver by design, but only the pull half self-clears — the push half leaves the slot dirty for the already-open case. Options: have `open_native_dashboard` report whether it created a new window and only `set_pending` in that case (rely on emit for the already-open window), or clear the slot after a successful emit. Please confirm the intended behaviour here.

### Non-blocking

- **Focus steal at the midnight boundary.** There is deliberately no hour gate, so the first tick after local midnight while the user is actively working past midnight will open the dashboard maximized and `set_focus`. The PR risk notes acknowledge this; flagging so the "skip `set_focus` on the mid-day boundary path" follow-up is tracked. Once-per-day limits the blast radius.
- **Hardcoded `~/.meridian` in the daemon** (`daily_plan.rs`: `std::env::var("HOME")` + `.join(".meridian")`). Matches the existing `onboarded`-marker convention and the tray must agree on the exact path, so this is acceptable — but it's the one spot that bypasses config-driven paths. Fine to leave; noting for awareness.
- **`today` string derivation.** `plan_auto_open.rs` / `daily_plan.rs` use `now.format("%Y-%m-%d")` rather than `meridian_core::date::today_string()`. This is actually correct here (both need the *same captured* `now` for the timestamp, avoiding a second `Local::now()` that could straddle midnight), and inline `format("%Y-%m-%d")` is the established pattern across `week.rs`/`worklog_pipeline.rs`/`scheduler.rs`. The `date` convention is about day-*bounds* math, not the trivial today-string — no change needed.

### Nits
- `MeridianTimelineShell.tsx:771` inlines the `onClose` handler with the `plan_dismissed` invoke on one line; readable but a named handler would match the file's style. Optional.

Akarsh-Hegde:
@adityaharishch review addressed in bf120cd6 - thanks, the blocking one was a real bug.

**1. Stale pending deep-link (blocking) - fixed.** Confirmed your read: with the dashboard already open, the emit delivered but the parked target was never consumed, so the next manual open would jump to the Plan modal. `deep_link::navigate_dashboard` now picks exactly one delivery half by window existence - emit to an open window, park for a fresh window's mount-time pull - and `set_pending` is private so no caller can recreate the unconditional double-write. Trade-off taken (documented on the fn): a window that exists but hasn't mounted yet (mid-creation by another path, ~sub-second) takes the emit branch and loses the event - one missed navigation, never a stale one. Both callers (auto-open, notification tap) route through it.

**Non-blocking:**
- *Midnight focus steal* - acknowledged; leaving the "skip set_focus on the boundary path" follow-up out of this PR so it stays reviewable. Happy to file an issue.
- *Hardcoded `~/.meridian`* - leaving as-is per the onboarded-marker convention, as you noted.
- *`today` derivation* - agreed, no change.

**Nit:** inline `onClose` extracted to a named `closePlan` handler.

PR #438, opened 2026-07-13
feat(tray): force-install updates below a manifest-declared minimum version

Akarsh-Hegde:
## What

Adds a **mandatory-update floor** to the DMG auto-update channel. Updates stay consent-based (banner + click) with one new exception: when the update manifest declares a minimum supported version and the running app is below it, the update downloads, installs, and relaunches automatically. This is the kill-switch for releases old versions must not keep running against (broken update paths, data-corrupting bugs).

## How

**Producer — `scripts/package-updater.sh`**
- Reads the optional `tray/minimum-version` file (plain `X.Y.Z`) at release time and ships it as a `Minimum-Version: X.Y.Z` line inside `latest.json`'s `notes`. The notes body is the transport because `tauri-plugin-updater` drops unknown manifest fields — a custom top-level key would never reach the app.
- File absent or empty (the default — the file is not committed) = every update stays consent-based. A malformed value fails the release loudly rather than silently dropping the floor.
- The staging channel inherits the marker automatically via `mirror-staging-release.sh`'s verbatim `latest.json` copy.

**Consumer — `tray/src-tauri/src/update.rs`**
- `split_minimum_version()` parses the marker (case-insensitive, checked UTF-8 slicing, unparsable value → ignored + warned, never mandatory).
- `UpdateStatus` gains `minimumVersion` + `mandatory`; the marker line is stripped from the `notes` the banners display.
- New `enforce_minimum_version()` task, spawned at launch (`lib.rs`): checks 30s after startup and then every 6h (the tray runs for weeks between relaunches, so launch-only would never catch a mandatory release published mid-run). When mandatory, it notifies ("This version is no longer supported - installing vX") and runs the existing `download_and_apply`.
- A failed forced install just waits for the next cycle — the consent banner remains as fallback. The existing `INSTALLING` single-flight guard covers races between the forced path and a banner click.

## Rollout

To make a release mandatory, commit `tray/minimum-version` containing the floor (typically the new release's own version) before cutting the release; remove or empty it afterwards if subsequent releases should go back to consent-based.

## Testing

- 7 new unit tests for marker parsing + the strictly-below comparison (`cargo test update::tests` — all pass; full tray suite 64/64).
- Release script exercised against a faked bundle layout: no file → notes unchanged; `1.68.0` → `"Meridian v1.70.0\nMinimum-Version: 1.68.0"`; garbage → exit 1.
- `cargo fmt` + `cargo clippy -- -D warnings` clean; full pre-push suite passed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -474,6 +474,17 @@ git tag "runtime-v${VER}"         <commit> && git push origin "runtime-v${VER}" 
 - **Two channels, two audiences.** A fix that must reach everyone has to land on **both** `pre-main` and `main` (or both tags) — one channel never feeds the other.
 - **It is not the app binary.** Updating the `.app` alone ships nothing under `services/`; only a runtime republish does.
 
+### Make a DMG release mandatory (force-install on old versions)
+
+DMG auto-updates are consent-based (in-app banner + click) with one exception: when the update manifest declares a **minimum supported version** and the running app is below it, the app installs the update and relaunches automatically. Use this as a kill-switch for releases old versions must not keep running against (broken update path, data-corrupting bug) — not for routine releases.
+
+1. Commit `tray/minimum-version` containing the floor as plain `X.Y.Z` (usually the new release's own version, forcing everyone older) **before cutting the release**. File absent or empty = consent-based, the default; a malformed value fails the release loudly (`scripts/package-updater.sh`).
+2. The release ships it as a `Minimum-Version: X.Y.Z` line inside `latest.json`'s `notes` — the notes body is the transport because `tauri-plugin-updater` drops unknown manifest fields. The staging channel inherits it via `mirror-staging-release.sh`'s verbatim `latest.json` copy.
+3. Installed apps below the floor force-update via `update::enforce_minimum_version` (`tray/src-tauri/src/update.rs`) — checked 30 s after launch and every 6 h thereafter, so long-running trays catch it without a relaunch. A failed forced install falls back to the consent banner and retries next cycle.
+4. Empty or remove `tray/minimum-version` afterwards if later releases should go back to consent-based — the floor ships with **every** release while the file has content.
+
+Only affects the DMG channel; npm/CLI installs still update via `meridian update`, and the MLX runtime has its own always-automatic upgrade (see the runtime task above).
+
 ---
 
 ## Coding-agent pipeline (`src/coding_agent_session_ingest/`)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4474,6 +4474,7 @@ dependencies = [
  "reqwest 0.12.28",
  "screenpipe-a11y",
  "screenpipe-screen",
+ "semver",
  "serde",
  "serde_json",
  "sha2 0.10.9",
--- a/scripts/package-updater.sh
+++ b/scripts/package-updater.sh
@@ -43,6 +43,24 @@ else
   echo "⚠ ${VERSIONED_DMG} not found — no stable-named DMG (tarball update still works)"
 fi
 
+# Optional mandatory-update floor. When tray/minimum-version contains a semver,
+# it ships as a `Minimum-Version: <v>` line inside the manifest notes — installed
+# apps running BELOW that version install this release automatically instead of
+# waiting for the banner click (tray/src-tauri/src/update.rs
+# enforce_minimum_version; the notes line is the transport because
+# tauri-plugin-updater drops unknown manifest fields). File absent or empty =
+# every update stays consent-based. A malformed value fails the release loudly:
+# a typo silently dropping the floor would defeat the point of setting it.
+MIN_FILE="tray/minimum-version"
+MINIMUM=""
+if [[ -f "${MIN_FILE}" ]]; then
+  MINIMUM="$(tr -d '[:space:]' < "${MIN_FILE}")"
+  if [[ -n "${MINIMUM}" && ! "${MINIMUM}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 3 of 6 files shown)


adityaharishch (commented):
## Review — force-install below minimum version

Reviewed all 6 changed files against repo conventions and for correctness. This is a clean, unusually well-documented change. No blocking issues found. A few things to confirm and some minor notes.

### Conventions — all pass
- **File headers:** `update.rs` and `lib.rs` both keep the required `//ambient dev tool…` first line; `package-updater.sh`, `Cargo.*`, and `CLAUDE.md` are exempt. ✅
- **User-facing text / dashes:** the one new notification body — `"This version is no longer supported - installing v{v}"` — uses a spaced plain hyphen, not an em/en-dash. Title `"Updating Meridian"` is fine. ✅ (The `-` in `tracing::*` messages is exempt — those are logs.)
- **Structured logging:** every `tracing` call puts data in fields (`version = %v`, `minimum = …`, `value`, `error = %e`) rather than interpolating into the message. ✅
- **No `unwrap()` outside tests, `update.rs` is 227 lines (< 500), no secrets.** ✅

### Correctness — looks right
- **Version compare is strictly-below** (`app.package_info().version < *m`) and the dedicated test `mandatory_is_strictly_below_minimum` pins that equal/above is *not* mandatory. Matches intent. ✅
- **Fail-open on the consumer side (good):** a manifest-fetch error → `check_status` returns `"error"` → the enforce loop's `state != "available"` guard skips forcing; an unparsable `Minimum-Version:` → ignored + warned, never mandatory. A broken/garbage manifest can't brick installs. This is the right call and the code + tests back it.
- **Fail-closed at release time (also good):** `package-updater.sh` `exit 1`s on a malformed floor so a typo can't silently drop the kill-switch. Nice asymmetry.
- **`split_minimum_version`** uses checked UTF-8 slicing (`get(..key.len())`) so a multibyte line can't panic, matches case-insensitively, and pushes the *original* (untrimmed) line to `kept` so notes formatting survives. Solid.
- **Dev runs are safe:** `check_status` short-circuits to `"unsupported"` via `is_packaged()` before any mandatory computation, and the enforce loop idles on it.
- **Race safety:** the forced path funnels through `download_and_apply`, whose existing `INSTALLING` single-flight guard (swap-true / drop-reset) covers a collision with a banner click.
- **No shell-injection in the producer:** `MINIMUM` is space-stripped, semver-regex-validated, then passed as a positional `argv` to the python heredoc (read via `sys.argv`, not interpolated into the script body). Safe.

### Please confirm (product/UX, not a code bug)
1. **Forced install has no user deferral.** On a mandatory release the flow is: toast → `download_and_apply` → `app.restart()`, with no "install now / later" choice or countdown. A user mid-work gets the app restarted underneath them with only a passive toast. That's *intended* for a kill-switch (broken update path / data-corrupting bug — you genuinely want old versions to stop), so I'm not flagging it as a defect — just confirming a toast-only, non-deferrable restart is the accepted UX rather than a short grace window.

### Minor / non-blocking
2. **`semver` type unification is load-bearing.** `app.package_info().version < *m` only compiles because the direct `semver = "1"` dep resolves to the *same* crate instance Tauri uses (Cargo.lock shows a single unified entry). It's safe (a divergence would fail at compile time, not silently), but the Cargo.toml comment "already in the tree via tauri" is doing real work — worth keeping if Tauri is ever bumped.
3. **Floor above the latest offered version.** If the `Minimum-Version` floor is higher than the newest published release, an old app will force-install the latest available (still below the floor), restart, and — since `check()` then returns no newer version — settle as `uptodate` and stop. Benign/self-terminating, but the interim toast would say "installing v{latest}" for a version still under the minimum. Operator error, just noting.
4. **Double `check()`:** `enforce_minimum_version` → `check_status().check()` then `download_and_apply().check()` — one redundant network round-trip per forced cycle. Negligible.

Overall: ship-ready pending the confirm on (1). Good test coverage on the parser + comparison, good docs on both producer and consumer.

Akarsh-Hegde:
Thanks for the thorough review — responses to each point, one fix pushed (1db2c27c):

**1. Forced install has no user deferral — confirmed as intended.** Two reasons this is acceptable for the kill-switch case:
- There is no unsaved user state to lose: the tray/dashboard render straight from `meridian.db`, and the daemon + capture keep running independently of the app restart. A restart costs the user a popover/dashboard reopen, nothing else.
- A deferral option would defeat the mechanism's purpose — the floor exists precisely for releases where "later" is the wrong answer (broken update path, data-corrupting bug). Routine releases stay consent-based; the floor is opt-in per release via `tray/minimum-version`.

If we ever want gentler ergonomics (e.g. a 60 s countdown toast before restart), it can be layered onto `enforce_minimum_version` without touching the transport. Happy to file a follow-up if you'd like it.

**2. `semver` type unification — acknowledged, comment stays.** The Cargo.toml note ("already in the tree via tauri") is exactly the breadcrumb for a future Tauri bump; as you said, a divergence fails at compile time, so there's no silent-breakage risk.

**3. Floor above the latest offered version — fixed.** `package-updater.sh` now fails the release when the floor exceeds the version that carries it (cores compared, so `1.70.0-staging.1` may carry a `1.70.0` floor). Verified in the sandbox harness: `1.68.0`/`1.70.0` floors against a `1.70.0` release pass, `1.71.0` fails with a clear message. Since every manifest is now published by a release that satisfies its own floor, the "force-install a build still below the minimum" state can no longer be produced by this script.

**4. Double `check()` — acknowledged, keeping it.** `download_and_apply` re-checks internally by design (it's the shared install path the banner click also calls, and must stay self-contained). The extra manifest GET happens at most once per 6 h cycle and only when a forced install actually fires — not worth splitting the API to save.

PR #441, opened 2026-07-14
fix(worklog): stop retrying posts that hit Jira's per-issue worklog cap

adityaharishch:
## Summary
- Jira's `WORKLOGS_PER_ISSUE_LIMIT_EXCEEDED` (per-issue 10,000-worklog hard cap) can never succeed on retry, so those posts were spamming the same failure every sweep (60s) forever
- Detect this permanent provider error and fail the row terminally instead of leaving it `approved` for endless retry
- Surface a clear "Why this failed" explanation + fix path (re-match via Edit) in the timeline detail view

## Test plan
- [x] `cargo fmt` / `cargo clippy` / `cargo test` (pre-commit + pre-push hooks passed)
- [x] UI build + UI tests (pre-push hooks passed)
- [ ] Manual verification on staging once merged: trigger a Jira worklog post against an issue at the cap and confirm the row fails terminally with the new explanation shown in the timeline

### DIFF
--- a/src/pm_worklog/post.rs
+++ b/src/pm_worklog/post.rs
@@ -20,7 +20,10 @@
 // see meridian_core::worklogs::rematch_worklog), so idempotency is scoped
 // per-row (`id`), not per (task, window) — two distinct rows may legitimately
 // share a window and both need their own post. A post failure leaves the row
-// `approved` with `last_post_error` recorded, and the next sweep retries.
+// `approved` with `last_post_error` recorded, and the next sweep retries —
+// UNLESS `is_permanent_provider_error` recognises it as unrecoverable (e.g.
+// Jira's per-issue worklog cap), in which case it's failed terminally instead
+// so it doesn't retry-spam the same doomed post forever.
 
 use anyhow::Result;
 use sqlx::SqlitePool;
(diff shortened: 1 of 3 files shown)


Akarsh-Hegde:
## Review

Solid, well-scoped fix — the terminal-failure detection is correctly checked on the full (untruncated) error before `fail_worklog` truncates it to 500 chars for storage, and it slots cleanly into the existing terminal-failure pattern (`empty comment` / `below-minimum duration`). The UI side is nicely designed too: `failureReason()` falls back to the raw message for any failure it doesn't specifically recognize, so nothing is ever hidden, and it now surfaces for *all* `failed` rows (empty-comment, below-minimum, unknown-provider), not just the new Jira-cap case — good generalization from a Jira-specific bug report.

A few things worth a look before merge:

1. **Stale doc comment** — `db::fail_worklog`'s doc (`src/pm_worklog/db.rs:300-302`) says `Record a TERMINAL post failure (empty comment, below Jira's minimum)`. This PR adds a third caller (permanent provider error) but doesn't update that list, so the doc now under-describes the function.

2. **No unit tests for the new pure functions** — `is_permanent_provider_error` (`post.rs`) and `failureReason` (`types.ts`) are both trivial, easily-testable pure functions, and every sibling file in `pm_worklog/` (jira.rs, linear.rs, azure_devops.rs, github.rs, trello.rs, collect.rs, comment.rs, ground.rs, ledger.rs) has a `#[cfg(test)]` module — `post.rs` currently has none. Since this establishes the pattern future permanent-error cases will follow, a couple of cheap tests (matches on the real Jira error text, doesn't match on an unrelated error, `failureReason` fallback) would be worth adding.

3. **Minor: doc-comment claim** — the new doc comment on `is_permanent_provider_error` states Jira returns HTTP 413 for `WORKLOGS_PER_ISSUE_LIMIT_EXCEEDED`. Detection here is string-only (doesn't gate on status code) so this doesn't affect correctness either way — but worth double-checking against a captured response, since a wrong number here could mislead someone cross-referencing logs later.

4. **Nit (non-blocking)**: the new "Why this failed" box uses `--color-state-pending` (amber) rather than `--color-state-rejected`, which is the color `failed` rows get everywhere else via `visualState()`/`stateColor()`. If that's a deliberate "this needs your attention" callout distinct from the card's own accent, ignore this — just flagging in case it wasn't intentional.

Nothing here blocks merging; (1) and (2) are the ones I'd actually want addressed.

PR #442, opened 2026-07-14
fix(worklog): fall back to Task when a project has no Bug issue type

adityaharishch:
## Summary
- Your `KAN` Jira project's scheme only has Epic/Subtask/Feature/Task — no "Bug" — so any proposal classified as a bug 400'd forever with `Specify a valid issue type` and retried every sweep (visible as the repeating `proposal ticket creation failed` warnings)
- Same failure mode exists for Azure DevOps projects on process templates (e.g. Basic) that don't define a Bug work-item type
- Before creating, check the target project's actual issue/work-item types; if "Bug" isn't available, create a "Task" instead and prefix the title with `Bug: ` so the original classification isn't lost
- Applies to both providers that model an issue type (Jira, Azure DevOps) — Linear/GitHub/Trello have no type concept and are unaffected

## Test plan
- [x] New unit tests for the `bug_fallback` decision (keep/demote/no-op cases) — `cargo test --lib pm_worklog::create`
- [x] `cargo fmt` / `cargo clippy -D warnings` / `cargo test` (pre-commit + pre-push hooks passed)
- [ ] Manual verification on staging: confirm proposals 192/194 (or new ones) now create as "Task" titled "Bug: ..." instead of failing

### DIFF
--- a/src/intelligence/providers/azure_devops/db.rs
+++ b/src/intelligence/providers/azure_devops/db.rs
@@ -9,6 +9,10 @@
 use anyhow::{Context, Result};
 use sqlx::SqlitePool;
 
+/// Delete `pm_tasks` rows no longer returned by the active-task fetch (closed,
+/// reassigned, etc.) — EXCEPT a task_key that has worklog history
+/// (`pm_worklogs`), which is kept forever so a completed work item's title
+/// never disappears from the timeline once it's closed.
 pub(super) async fn prune(pool: &SqlitePool, kept_keys: &[String]) -> Result<()> {
     if kept_keys.is_empty() {
         sqlx::query(
@@ -18,10 +22,13 @@ pub(super) async fn prune(pool: &SqlitePool, kept_keys: &[String]) -> Result<()>
         .execute(pool)
         .await
         .context("pruning all azure_devops pm_task_embeddings")?;
-        sqlx::query("DELETE FROM pm_tasks WHERE provider = 'azure_devops'")
-            .execute(pool)
-            .await
-            .context("pruning all azure_devops pm_tasks")?;
+        sqlx::query(
+            "DELETE FROM pm_tasks WHERE provider = 'azure_devops' \
+             AND task_key NOT IN (SELECT DISTINCT task_key FROM pm_worklogs)",
+        )
+        .execute(pool)
+        .await
+        .context("pruning all azure_devops pm_tasks")?;
         return Ok(());
     }
 
@@ -42,7 +49,8 @@ pub(super) async fn prune(pool: &SqlitePool, kept_keys: &[String]) -> Result<()>
 
     let sql_tasks = format!(
         "DELETE FROM pm_tasks \
-         WHERE provider = 'azure_devops' AND task_key NOT IN ({placeholders})"
+         WHERE provider = 'azure_devops' AND task_key NOT IN ({placeholders}) \
+         AND task_key NOT IN (SELECT DISTINCT task_key FROM pm_worklogs)"
     );
     let mut q2 = sqlx::query(&sql_tasks);
     for k in kept_keys {
--- a/src/intelligence/providers/github/mod.rs
+++ b/src/intelligence/providers/github/mod.rs
@@ -103,6 +103,10 @@ async fn upsert(pool: &SqlitePool, tasks: &[(GhTask, serde_json::Value)]) -> Res
 // Prune (scoped to provider = 'github')
 // ---------------------------------------------------------------------------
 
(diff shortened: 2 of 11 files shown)


adityaharishch:
Added a second, related fix to this PR: newly-created tickets were never assigned to anyone, so both the Jira (`assignee = currentUser()`) and Azure DevOps (`AssignedTo = @me`) task-sync queries never picked them up — meaning the ticket never entered `pm_tasks` and its title never appeared on the timeline, and it showed up unassigned in the tracker.

Fix: self-assign at creation time (best-effort — falls back to unassigned + a warning log if the identity lookup fails, never blocks creation).

adityaharishch:
Found the real root cause of the missing-title issue (confirmed via KAN-305, which is Done in Jira). Every provider's task sync only ever fetches **active** tasks (Jira: \`assignee = currentUser() AND statusCategory != Done\`; Azure DevOps: \`AssignedTo = @me\`; similarly scoped for GitHub/Linear/Trello), and on every sync it prunes any \`pm_tasks\` row that fell out of that fresh set — including a ticket that just got marked Done. That delete permanently erases the row the timeline's \`LEFT JOIN pm_tasks\` depends on for the title, so a completed ticket's title vanishes from the timeline forever, regardless of the assignee fix above.

Added a third commit: exclude any task_key with worklog history (\`pm_worklogs\`) from every provider's prune DELETE (including the "no live tasks at all" full-clear fallback branches). A task keeps its title permanently once it has a worklog, no matter what happens to it afterward in the tracker (Done, reassigned, deleted from the board, etc.) — only genuinely worklog-less stale cache rows get pruned. Added a regression test (\`prune_keeps_stale_task_with_worklog_history\`) plus 431 other existing tests, all green.

Akarsh-Hegde:
## Review

The core fix (probe the project's actual issue/work-item types, fall back to `Task` + `Bug: ` title prefix) is clean — `bug_fallback` is a well-isolated pure function with good test coverage (keep/demote/no-op cases), and the `has_bug`/`has_bug_type` lookups short-circuit so `Task`/`Story` creation never pays the extra API round trip. Self-assignment failures are correctly best-effort (warn + continue rather than failing ticket creation).

That said, this PR does noticeably more than its title/description say, and a couple of the extra pieces raise questions worth resolving before merge:

1. **Scope mismatch** — the description is scoped to "fall back to Task when a project has no Bug issue type," but the diff also: (a) adds self-assignment on Jira/Azure ticket creation, (b) changes `prune()` on **all five** providers (jira/linear/github/trello/azure_devops) to retain any `pm_tasks` row with worklog history forever instead of deleting it when it drops out of the fetch, and (c) flips `run_pm_sync` in `main.rs` from on-demand-only to running on every ~60s daemon tick. None of these are mentioned in the PR body. They all look like legitimate, related fixes (all in service of "a ticket's title shouldn't vanish from the timeline"), but bundling them undisclosed makes this harder to review atomically and easier to accidentally revert together if one part needs to be backed out. Worth calling all three out explicitly in the description.

2. **Self-assignment's stated justification may not hold for real Bug tickets** — the new comment says self-assigning makes the ticket "discoverable by the `assignee = currentUser()` sync query," but `jira/fetch.rs`'s JQL is `assignee = currentUser() AND statusCategory != Done AND type IN (Task, Feature)` — it also filters by `type`, and `Bug` isn't in that list. So for a project that *does* have a native Bug type (`has_bug = true`, no fallback), a self-assigned real Bug ticket still won't show up in `pm_tasks`/the timeline — the same symptom this PR sets out to fix, just via a different root cause this change doesn't touch. Worth confirming whether that's already tracked separately, since as written the self-assign fix only actually helps the fallback-to-Task path (which happens to satisfy the JQL type filter by luck, not by design).

3. **Test coverage gap** — the `pm_worklogs`-history exclusion added to `prune()` is duplicated identically across `azure_devops/db.rs`, `github/mod.rs`, `jira/mod.rs`, `linear/mod.rs`, and `trello.rs`, but only `jira/tests.rs` got a new test (`prune_keeps_stale_task_with_worklog_history`). The other four providers carry the same new SQL clause with zero coverage.

4. **Minor asymmetry** — in the "no live tasks at all" branches (azure_devops/github/linear/trello), the `pm_task_embeddings` DELETE wasn't given the same worklog-exclusion, so a `pm_tasks` row kept forever via worklog history still has its embeddings wiped on a full-clear. That may well be intentional (a closed ticket doesn't need to be embed-matched again), but it leaves `pm_tasks` rows with no corresponding `pm_task_embeddings` row — worth a quick confirmation that nothing downstream assumes those are 1:1.

(1) and (2) are the ones I'd want a response on before merging; (3)/(4) are nice-to-haves.

PR #444, opened 2026-07-14
feat(settings): add the centralised AI-provider choice

adityaharishch:
First of six PRs adding a centralised, user-selectable AI provider: the user picks which AI runs their pipeline (their own Claude / Codex / Cursor / Copilot CLI subscription, or the on-device MLX model) once during setup, switchable afterwards from Settings.

This PR is the foundation only - **nothing reads the new fields yet**. The resolver, the backends, and the wizard step follow in later PRs.

## What lands

**`LlmProvider { Claude, Codex, Cursor, Copilot, Local }`** in `meridian-core/src/llm_provider.rs`, modelled on the existing `canonical_task::Provider`. Wire forms deliberately match the coding-agent summariser's `Source::as_str()`, so the two map without a translation table. Default is `Local` - on-device is the product's pitch, and it is the only backend guaranteed to be present.

**Three settings fields**: `llm_provider`, `llm_provider_model`, `llm_local_chat_model_ready`. `#[serde(default)]` was already on the struct, so every existing `settings.json` upgrades with no migration.

**Validation in `update_settings`** - the only writer. An unrecognised provider would otherwise save cleanly and then quietly resolve back to on-device, so the user would pick Claude and keep getting the local model.

**`ui/lib/llm-providers.ts`** - the single source of truth for the UI, deliberately mirroring `integrations.ts`'s `TRACKERS`. One list, shared by the wizard and Settings later.

## The one design decision worth reviewing

`llm_provider` is stored as a **`String`, not the enum**. `load_runtime_settings()` (settings.rs:185) falls back to `RuntimeSettings::default()` on *any* deserialise error, so a single unparseable field silently resets **every** setting the daemon reads. Had this been typed as the enum, a value written by a newer build would cost an older daemon not just the provider but the user's work hours, poll interval and log level.

It is parsed with `LlmProvider::from_wire`, which returns `None` for anything unknown - the caller falls back to the default. An unknown provider costs you the provider, and nothing else. `an_unknown_llm_provider_does_not_reset_every_other_setting` pins this, and fails loudly if anyone retypes the field.

## Two dead fields deleted

- **`llm_prefer_local`** - declared in Rust, defaulted, mirrored in TS, and toggleable *with a Save button* in Advanced settings... and read by **nobody**. `grep` across `src/ services/ tray/ meridian-core/` finds only the declaration and the writer. Users can flip this switch today and it does nothing. `llm_provider == "local"` is what it was trying to say, expressed once and actually honoured.
- **`llm_model_preference`** - TS-only, with a comment claiming it "mirrors `RuntimeSettings.llm_model_preference`". That Rust field never existed.

## Tests

- `LlmProvider` round-trip: `as_str()` <-> `from_wire()` <-> serde, all five variants; unknown string -> `None`, not a panic.
- Settings upgrade: a file with no `llm_provider` loads as `local`, existing keys survive.
- The landmine test above.
- Also fixes a latent flaw in the existing settings tests: they set `MERIDIAN_SETTINGS_PATH`, a **process-global** env var, and cargo runs tests in parallel. With only one such test it never raced; adding two exposed it. They now share a lock.

`cargo test --workspace` (427 passed), `cargo clippy -- -D warnings`, `bun test` in `ui/` (181 passed), and `npm run build` are all green.


Akarsh-Hegde (commented):
Reviewed statically against the working tree (didn't re-run CI). This is a clean foundation PR and the design decisions hold up under checking — notes below, all minor.

## Verified

- **String-not-enum is the right call, and correctly defended.** `load_runtime_settings` (settings.rs:186) really does `serde_json::from_str(...).unwrap_or_else(|_| RuntimeSettings::default())`, so a single unparseable field would reset every daemon-read setting. `#[serde(default)]` on the struct (settings.rs:38) is what makes the "no migration" upgrade work, and `settings_without_llm_provider_default_to_local` pins it. Good.
- **Both dead-field deletions check out.** `grep` across `src/ services/ tray/src meridian-core/src ui/{lib,components,app}` finds `llm_prefer_local` only in the declaration/default/UI-toggle/writer (no reader), and `llm_model_preference` only in `ui/lib/settings.ts` (no Rust field ever existed, nothing consumed it in TS either). Deleting both is safe.
- **Validation is in the one true writer.** `update_settings` → `write_settings_value` is the sole production write path, and the `from_wire` check runs before the write. The `an_unknown_llm_provider_does_not_reset_every_other_setting` test is exactly the right thing to pin.
- **Good catch on `ENV_LOCK`.** `MERIDIAN_SETTINGS_PATH` is process-global and cargo runs tests in parallel — the shared lock is the correct fix, not per-test temp paths.

## Minor

1. **`cli_name()` (Rust) and `bin` (TS `llm-providers.ts`) are a second cross-boundary source of truth for the binary names** (`cursor` → `cursor-agent`, etc.), unlike the wire ids / `as_str` pair which is pinned by `wire_form_round_trips` + `serde_matches_as_str`. Nothing keeps `cli_name` and `bin` in sync today. Not actionable in this PR (nothing reads either yet), but worth wiring to a single source — or at least a mirror-test — when the resolver lands.

2. **`is_local()` and `cli_name()` are unused by any consumer in this PR** (`all()` is used by the validator; the other two aren't). Fine as scaffolding for a foundation PR, just flagging so the follow-up resolver PR actually consumes them rather than leaving speculative API.

3. **`from_wire` trims whitespace but serde/`as_str` don't.** Harmless asymmetry, but it means `update_settings` accepts `" claude "` and stores it verbatim (dirty-but-resolvable). Consider trimming at write, or dropping the trim in `from_wire` — either way one place decides. Pure nit.

Nothing blocking. Nice module docs and commit message.

adityaharishch:
**Review: correctness-focused pass** (large PR — 12,247 additions across multiple bundled workstreams; reviewed by area)

## LLM provider core (`llm_provider.rs`, `settings.rs`, `src/llm/*`)

- **Claims 1-3 in the PR description all verified true**: `llm_provider` really is stored as an unvalidated `String` in `RuntimeSettings` (`settings.rs:52`), `LlmProvider::from_wire` really does degrade gracefully to `None`→default on an unrecognized value rather than nuking the whole settings load, and `update_settings` (`tray/src-tauri/src/commands/settings.rs:96-108`) really does hard-reject an unrecognized provider server-side before writing. `llm_prefer_local`/`llm_model_preference` are fully removed on both sides (one stale mention survives only in a comment at `AdvancedSection.tsx:119` — harmless, worth a follow-up cleanup).
- **PR description is misleading on scope**: despite "nothing reads the new fields yet / resolver and backends follow [later]," `llm::complete` is already called from three live production paths in *this* PR — `src/pm_worklog/generate.rs`, `src/worklog_pipeline/hour.rs::build_report`, and `src/worklog_pipeline/workstream.rs`. The resolver's fallback chain (retry→fallback-to-local, rate-limit backoff) is sound and well-tested, and no panics were found in `detect.rs`/`claude.rs`/`resolver.rs` (proper timeouts, `kill_on_drop`, no unwraps on external command output) — but reviewers should evaluate this PR knowing the LLM calls are active now, not deferred.

## Worklog pipeline / day-task generation (`generate.rs`, `post_comment.rs`, migrations 058-060)

- **Bug — retry after a crash can double-post a comment** (`src/pm_worklog/generate.rs` `approve_inner`, ~lines 290-307): `post_comment(...)` succeeds, then `mark_posted(...)` persists `state='posted'`. If the process crashes or `mark_posted` itself errors, the row is stuck at `state='approved'` (the error path only sets `last_error`, it doesn't revert state). The idempotency check at the top of `approve()` (line 197) only short-circuits on `state == "posted"`, so a retry re-enters `approve_inner`, correctly skips re-creating the ticket (guarded by `created_task_key`), but calls `post_comment` again **unconditionally** — and `post_comment.rs` deliberately carries no dedup marker in the comment body (unlike `format_worklog_comment`'s `⏱`/`meridian-worklog` marker elsewhere), so there's no way to detect the duplicate after the fact. This breaks the "idempotent + retry-safe" claim in the module docs. Suggest persisting `posted_comment_id` atomically with the post attempt, or embedding a stable per-`(day_local, task_id)` marker in the comment body.
- Minor: `mark_created` (`day_task_worklogs.rs` ~line 240) has no `WHERE state = 'approved'` guard, unlike `mark_approved`/`mark_posted` — not exploitable today (single caller), but a latent trap for a future caller.
- Minor: `fetch_open_candidates` (`generate.rs` ~419-445) does `LIMIT 30` with no `ORDER BY` — SQLite doesn't guarantee row order here, so on boards with >30 open tickets the candidate set fed to the matcher can vary nondeterministically between calls.
- Good: `upsert_draft`'s `ON CONFLICT ... DO UPDATE ... WHERE state = 'drafted'` correctly follows the repo's no-DELETE-then-INSERT rule; `parse_answer` never panics on malformed LLM JSON.

## Ticket status picker, per provider (`ticket_update/*.rs`)

- **Bug — Jira: transition/status misalignment risk** (`jira.rs:684-699`, `pick_transition_for_choice`): `transition_status_options()` filters requiring both `to.id` **and** `to.name`; `pick_transition_for_choice()` separately filters the raw transitions on only `to.id.is_some()`, then `.zip()`s the two lists. If any transition has `to.id` but missing `to.name`, the two filtered sequences misalign and `.zip()` pairs the wrong transition id with the wrong target status — `set_status` could silently POST a transition to a *different* status than the one requested. Given Jira is the priority-1 provider, recommend building `(transition_id, StatusOption)` pairs in one pass instead of filtering two lists independently and zipping.
- **Bug — Azure DevOps: real API failures masked as "redirected"** (`azure_devops.rs:364-393`, `try_patch_state`): any non-2xx response (401/403/404/429/500) is treated identically to a legitimate workflow-reject and returned as `Ok(Some(text))` → `redirected(...)`, logged via `tracing::info!` on what should be an error path. An expired PAT or rate-limit surfaces to the user as "your board can't move it to that status" instead of a real error, and hides credential/config problems from logs/alerting.
- GitHub, Trello, Linear all look correct (GitHub correctly restricted to open/closed pseudo-states; Trello uses list-move; Linear uses `issueUpdate(stateId)` with real errors bubbling as `Err`).
- No hardcoded credentials/tokens found. Undo has no staleness/optimistic-concurrency check against the current status, so two rapid status changes could clobber each other — not blocking, but worth a note if it comes up in practice.

## DRM detector / capture (`drm_detector.rs`, `screenpipe.rs`)

- **Correctness — front-window/frame mismatch, cuts both ways on the feature's core purpose**: `resolve_url_via_applescript` always queries the **front window** of the named app, but `capture_once_ocr` iterates **all visible windows**. A background Netflix window that isn't frontmost can get captured anyway (the privacy/legal risk this feature exists to prevent), while a legitimate non-front window can get wrongly skipped if the front window happens to be a streaming URL. Worth gating on window identity, or at minimum documenting this as a known gap.
- **Performance — blocking subprocess spawns on the async capture hot path**: `resolve_url_via_applescript`→`run_osascript`→`std::process::Command::output()` (`drm_detector.rs:184-199`) is a synchronous blocking call inside the async capture loop, invoked for every frame lacking a `browser_url` on Safari/Chromium-family browsers, not just when streaming is suspected. `any_streaming_content_visible()` similarly does up to 9 `pgrep` calls plus AppleScript, once per tick, unconditionally when `pause_on_streaming` is on. Recommend `spawn_blocking` and/or caching across ticks rather than shelling out every tick in a continuously-running background daemon.
- No panic risk found (`Command` output handled via `.ok()?`/`.unwrap_or(false)` throughout). `observability/mod.rs`/`uninstall.rs` changes look intentional, no leftover WIP.

---

Given the scope here, I'd suggest splitting this into the six PRs the description originally promised rather than merging as one — at minimum the two hard bugs above (Jira zip-misalignment, worklog duplicate-comment-on-retry) should block merge until fixed.

PR #450, opened 2026-07-15
feat(dashboard): add a What's New changelog + roadmap modal

Akarsh-Hegde:
## Summary
- Adds a "What's New" modal to the dashboard: release history (highlights + fixes) and a roadmap, sourced from a small hand-curated `tray/src-tauri/resources/whats-new.json` — deliberately separate from the auto-generated `CHANGELOG.md`, which is commit-level and too internal to show end users.
- Auto-opens once per app version after an update lands, via the same poll-tick pattern as the existing "Plan your day" auto-open. Defers itself whenever the dashboard window is already open, so it never collides with the sibling Plan auto-open on the same tick and never yanks the user off whatever they're doing.
- Always reachable via a new "What's New" item in the toolbar nav pill.
- Adds a "What's New per release" recipe to `CLAUDE.md`'s Common Tasks so the content doesn't drift out of sync with real releases.

## Note for reviewers
The seeded roadmap content ("Idle & sleep detection", "Lighter local AI footprint") is a starting placeholder pulled from documented known-limitations — please confirm or replace it before this reaches users, since it reads as a product-facing promise.

## Test plan
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings` / `cargo test` (tray crate) — all green, including new `whats_new` unit tests (JSON-shape guard + seen/unseen marker logic)
- [x] `npm run build` (ui) — TypeScript + static export clean
- [x] Pre-push hooks (fmt, clippy, ui build/tests, cargo test, security audit) — all passed
- [x] `npm run tauri dev` manual pass — auto-open firing, modal render, tab switching, nav-pill open, close-marks-seen (not exercised by CI or this session — WKWebView isn't automatable in this environment)

https://claude.ai/code/session_01MrJAJ3nnr5C1iyaKGV3BiY


adityaharishch:

Overall implementation is solid — no crashes or races found in the primary paths.

1. **"Seen" tracking** (`whats_new.rs:71-90`, `whats_new_auto_open.rs:71`) — correct. Persisted to `~/.meridian/last_seen_version` (survives restart), exact-string-compared against `app.package_info().version`. Since app versions are always exact per release, this can't loop on a downgrade/rebuild, and an absent/mismatched file both correctly read as "unseen."

2. **Auto-open deferral vs. sibling "Plan your day"** (`poll/mod.rs:141,146`) — correct, not racy. Both run sequentially (`.await`) on the same tick; `whats_new_auto_open` runs strictly after `plan_auto_open` and checks `get_webview_window("dashboard").is_some()`, so if Plan opens the dashboard first, What's New correctly sees it and defers.

3. **JSON parsing** (`whats_new.rs:57-61`) — build-time malformation is caught by the `whats_new_json_parses` unit test, but at runtime `parse_whats_new()` uses `.expect(...)`. If a malformed `whats-new.json` ever shipped (hooks bypassed, hand-edit, etc.), the first `get_whats_new()` call (on modal open) would panic instead of degrading. Worth a `serde_json::from_str(..).unwrap_or_else(|_| WhatsNewData { releases: vec![], roadmap: vec![] })` fallback for defense-in-depth even though the test gate makes this low-risk in practice.

4. **Command/window wiring** — fine, `get_whats_new`/`mark_whats_new_seen_cmd` registered in `lib.rs`'s `invoke_handler!`, reuses the existing `"dashboard"` window capability — no gap.

5. **Seen-marking timing** — minor inconsistency, not a bug: the *auto-open* path (`whats_new_auto_open.rs:71`) marks the version seen **before** the window renders (documented as an intentional crash-safety tradeoff), which is subtly at odds with the doc comment at `whats_new.rs:92-94` claiming seen "always reflects the version actually viewed." A crash between the marker write and the window rendering means that version's notes are silently never shown again (toolbar pill is the only remaining path to them) — worth tightening the comment to acknowledge this, even if the tradeoff itself is fine.

6. **Types** — Rust/TS types match, except `RoadmapItem.status` is an unvalidated `String` in Rust vs. a TS literal union. A typo'd status in `whats-new.json` (e.g. `"Planned"` vs `"planned"`) parses fine on the Rust side but silently renders `undefined` label/color via the `ROADMAP_STATUS_LABEL`/`_COLOR` lookups (`WhatsNewModal.tsx:139,147`) instead of erroring loudly.

No blocking issues — nice work on the deferral logic against the sibling auto-open.

Akarsh-Hegde:
Addressed the review findings from @adityaharishch's correctness pass:

- **#3 (panic risk)** — `parse_whats_new()` no longer `.expect()`s on a malformed embedded JSON. It now logs `tracing::error!` and serves an empty `{releases: [], roadmap: []}` payload, so a bypassed hook or hand-edit degrades the modal gracefully instead of panicking on first open.
- **#6 (unvalidated status)** — `RoadmapItem.status` is now a `#[serde(rename_all = "kebab-case")]` enum (`InProgress | Planned | Considering`) instead of a bare `String`. A typo'd value in `whats-new.json` now fails the JSON parse loudly (and falls into the same empty-payload fallback above) instead of silently rendering `undefined` label/color on the frontend lookups. No frontend change needed — the wire format is unchanged.
- **#5 (doc accuracy)** — tightened `mark_whats_new_seen_cmd`'s doc comment to acknowledge that the auto-open path already writes the "seen" marker before the window renders, rather than implying this call is always what marks a version seen.

Items #1, #2, and #4 were reviewed and confirmed correct as-is, no change needed.

Also pushed the visual polish pass on `WhatsNewModal.tsx` (accent bars, a "Latest" badge, colored bullet dots, staggered pop-in) that was finished but not yet committed from earlier manual testing — data/logic unchanged.

Verified: `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (69/69 passing, including the new/existing `whats_new` tests), and `npm run build` in `ui/` all clean. Full pre-push suite (fmt, clippy, ui build + tests, cargo test, security audit) passed on push.

PR #453, opened 2026-07-16
feat(llm): real connectivity test for AI provider CLIs

Akarsh-Hegde:
## Summary
- The Intelligence panel could only ever say a CLI was *installed* — never whether it actually works. `src/llm/detect.rs` adds `test_provider()`/`test_all_installed()`, which run one real, trivial call through each backend's own `complete()` (via the existing `backend_for()` factory, so no CLI-invocation logic is duplicated) and distinguish `RateLimited` from `Failed` — a card can be correctly configured and just temporarily rate-limited, which needs a different fix than "sign in again".
- Results persist to `~/.meridian/provider_test_cache.json` (atomic write, same idiom as `settings.json`) so a verified/failed state and its timestamp survive a restart.
- Deliberately kept separate from the free, always-on install probe — a real test spends one request against the user's own subscription, so it never runs silently; only via a card's own "Test connection" button or Rescan (which now tests every *installed* provider concurrently, in addition to the existing install probe).
- Two new Tauri commands (`test_llm_provider`, `test_all_llm_providers`) expose this to `<LlmProviderPicker>`, shared by the setup wizard's Intelligence step and the dashboard's Settings panel — both surfaces get it for free.
- Cards now show `VERIFIED` / `RATE LIMITED` / `CONNECTION FAILED` with a relative timestamp and the real error message.
- `IntelligenceSection` gets a stronger warning banner that fires only when the **currently selected** provider has a known-bad last test — states what's actually happening, not just what could theoretically happen.

## Test plan
- [x] `cargo test --lib llm::` — 21 tests pass, including new cache round-trip / corrupt-cache / serde-shape tests for `detect.rs` (no live CLI calls in tests — those would be slow/costly/flaky in CI)
- [x] `cargo fmt --check` + `cargo clippy -- -D warnings` clean on both the daemon and tray crates
- [x] `npm run typecheck` — no new errors (pre-existing `bun:test` type-resolution errors in `__tests__/` are unrelated, confirmed present without this branch too)
- [x] `npm run build` (Next.js production build) succeeds
- [x] Full `pre-push` suite passed on `git push` (fmt, clippy, ui build, ui tests, `cargo test`, security audit)
- [ ] **Not done**: live click-through in a running `tauri dev` session — this feature is Tauri-IPC-dependent (`invoke()` has no bridge outside the Tauri webview), so a plain browser render can't meaningfully exercise it. Recommend a manual pass: open Settings → Intelligence, click "Test connection" on an installed CLI, and confirm the badge updates; click Rescan and confirm all installed providers test concurrently.

---

## Update: also fixes the cursor-agent silent-empty-report bug

Diagnosed earlier in this work: cursor-agent 2026.06.04 takes the `ACTIVITY_REPORT` prompt's `<HH:MM-HH:MM>` notation (prompt-writing shorthand for "fill this in") literally and emits the actual bracket characters. `parse_report_line` (`src/worklog_pipeline/hour.rs`) required each line to start with a digit, so every line silently dropped and the whole hour's report came out empty — no error anywhere in the path.

`parse_report_line` now strips an optional leading `<`/trailing `>` around the timestamp, the same way it already tolerates a stray `1. `/`- ` list marker. New test uses the actual cursor-agent output captured live while diagnosing this. **Purely additive tolerance** — it only accepts lines that previously dropped, so it cannot change how Claude/Codex/Copilot/on-device output already parses (none of them were observed emitting the literal brackets).

- [x] `cargo test --lib worklog_pipeline::` — 49 passed, including the new angle-bracket test
- [x] fmt/clippy clean; full pre-push suite passed on push

---

## Update 2: fixes the fold's idempotency guard silently dropping whole hours (`e714f931`)

Root cause of two live data losses on 2026-07-16 (the 3-4pm and 5-6pm hours both vanished from the timeline's task cards despite complete, accurate reports in `pm_worklog_hours`): `workstream::run` skipped an hour as "already folded" when the hour label appeared in any task's `hours_json` — but `hours_json` is derived from segments **touching** an hour, so a segment merely spilling past the boundary (`14:44-15:02`, `16:20-17:01`) marked the **next** hour as folded and its real work was silently dropped. No error anywhere; the hour still went `done`.

The guard (`hour_already_folded`, `src/worklog_pipeline/workstream.rs`) now judges by a prior segment **starting** inside the hour — a fold's own placements start inside the hour being folded, so that's the honest signal. An unparseable hour label fails **open** (the fold merge is retry-safe via segment coalescing + whole-story rewrite; a skipped fold is silent data loss). `DayTaskRow::hours`'s doc no longer claims to be the idempotency key.

- [x] 4 regression tests pinning both live incidents (the 16:20-17:01 and 14:44-15:02 spills) plus the fail-open edges — `cargo test --lib worklog_pipeline::` 53 passed
- [x] Verified end-to-end live: re-ran `worklog-hour 2026-07-16T17` with the fix — the previously-skipped fold executed (`workstreams built`, provider=claude) and the lost 5-6pm placements landed (`17:02-17:58`, `17:02-17:44`)
- [x] fmt/clippy clean; full pre-push suite passed on push


src/llm/detect.rs
```diff
@@ -85,15 +99,197 @@ pub async fn detect(provider: LlmProvider) -> ProviderStatus {
...
+    };
+    // A corrupt/foreign-format cache degrades to "never tested" rather than failing the
+    // whole panel — it will simply repopulate on the next test.
+    serde_json::from_str(&raw).unwrap_or_default()
+}
+
+/// Record one test result to the on-disk cache, crash-safely (temp file + atomic rename,
+/// same idiom as `settings.json`). Logged, not propagated — a failed cache write must not
+/// fail the test the user just watched succeed or fail in front of them.
+pub fn persist_test_result(result: &ProviderTestResult) {
```
Akarsh-Hegde, line 296: **Blocking — unguarded concurrent read-modify-write corrupts the test cache.**

`test_all_installed` (lines 244-257) fans every installed provider's test out via `futures::future::join_all`, and each future independently calls `persist_test_result` here, which does `load_test_cache()` → `insert` → `atomic_write_json`, with no lock. On any machine with 2+ CLI providers installed, clicking Rescan reliably runs this concurrently.

It's worse than a simple lost-update: `atomic_write_json` (`meridian-core/src/fs_utils.rs:35`) writes to a **fixed** temp path (`path.with_extension("json.tmp")`), so two concurrent writers to the same cache file share the same temp file — one can truncate the other mid-write. `load_test_cache`'s corrupt-JSON fallback then silently resets the *entire* cache to empty, discarding every provider's previously recorded VERIFIED/FAILED state, not just the racing pair's. This directly undercuts the doc comment above `test_all_installed` ("persists each result as it lands, so a slow/hanging CLI can't hold the others' verified state hostage") — it's backwards under exactly the scenario Rescan produces.

No test exercises concurrent `persist_test_result` calls. Fix: wrap cache read+insert+write in a `Mutex`, and give `atomic_write_json` a unique per-call temp suffix regardless.

Separately (lower severity): both `load_test_cache` and this fn do blocking `std::fs` I/O directly on the async call path without `spawn_blocking` — `fs_utils.rs`'s own doc comment says other callers wrap sync calls this way since they're invoked from async commands; this caller doesn't follow that stated convention.

src/worklog_pipeline/workstream.rs
```diff
@@ -40,8 +41,57 @@ use super::workstream_sanitize::apply_placements;
...
+        .next()
+        .and_then(|h| h.parse::<i64>().ok())
+    else {
+        return false;
+    };
+    let (lo, hi) = (hh * 60, (hh + 1) * 60);
+    prior
+        .iter()
+        .flat_map(|t| &t.segments)
+        .any(|s| s.start_min >= lo && s.start_min < hi)
```
Akarsh-Hegde, line 97: **Moderate — false-negative gap from segment coalescing.**

This check ("does a prior segment *start* inside this hour?") is right in isolation, but `segment::normalize` (`segment.rs:104-115`) merges a touching/overlapping segment into the earlier one and **keeps the earlier `start_min`**. So: hour N's fold creates a segment starting in hour N; hour N+1's fold extends the same task across the boundary and coalesces backward into it. The merged segment's `start_min` is still in hour N, so a later re-check of hour N+1 (crash-loop retry, LLM-Lab replay, any downstream failure that re-marks the hour pending) finds nothing starting in N+1's window and reports "not folded" — triggering an unwanted re-fold of already-completed work.

Segment coalescing makes the *time* accounting a no-op, but `apply_placements` (`workstream_sanitize.rs`) **replaces** a task's title/summary wholesale on every non-empty placement, so a spurious re-fold can genuinely rewrite the narrative with fresh (possibly different) LLM output — the doc comment's "re-applying an hour is also a no-op" claim only holds for time, not story.

None of the 4 new regression tests exercise a segment that's already been coalesced across the boundary before this guard runs. Not the two cited data-loss incidents and not new data loss, but a real gap in the fix's own core mechanism (it reintroduces a false-negative from the opposite direction of the bug it's fixing), plus wasted LLM calls and possible summary churn on retry. Worth a regression test for the coalesced-segment case before merge.

src/worklog_pipeline/hour.rs
```diff
@@ -183,12 +184,17 @@ fn parse_report(
...
-/// notes drop.
+/// "- ", "* "), a literal `<...>` wrapped around the timestamp (the prompt's
+/// `<HH:MM-HH:MM>` is prompt-writing notation for "fill this in" — Claude omits the
+/// brackets, but cursor-agent 2026.06.04 was observed emitting them literally, which
+/// otherwise made the line start with `<` instead of a digit and drop silently), the unit
+/// written as `min` / `mins` / `minutes`, and arbitrary surrounding whitespace. Returns
+/// `None` for any line without a leading minute count so headers and notes drop.
 fn parse_report_line(raw: &str) -> Option<(Option<String>, i64, String)> {
     let s = strip_list_marker(raw.trim());
+    let s = s.strip_prefix('<').unwrap_or(s);
```
Akarsh-Hegde: **Minor-moderate — the bracket strip (this line + line 197) isn't gated on an actual timestamp match, so it can misparse a non-timestamp bracketed aside.**

Traced by hand: `parse_report_line("<15 min gap, no activity>")` — the leading `<` strips unconditionally; `take_hhmm` then fails to match a real `HH:MM` and returns the string unchanged (harmless so far); but execution then falls into the existing digit/unit parser below, which happily reads "15 min" as a real entry with `prose = "gap, no activity>"` (the trailing `>` leaks in since its strip on line 197 isn't gated on whether a stamp actually matched). Before this PR, that line correctly returned `None` (first char `<` isn't a digit).

This contradicts the PR description's claim that the change is "purely additive tolerance… cannot change how existing output already parses" — it does change behavior for any bracketed non-timestamp line shaped like a duration, and a bracketed aside about an untracked gap is plausible LLM phrasing. Low likelihood but real. Fix: move the `<`/`>` stripping inside `take_hhmm` itself, only consuming them when they actually wrap a matched stamp, rather than unconditionally at the top of `parse_report_line`.

src/llm/probe.rs
```diff
@@ -0,0 +1,378 @@
...
+            }
+        }
+    }
+    Ok((SchemaRung::None, spent))
+}
+
+/// Is this failure about the ENDPOINT rather than the rung? A bad key answers nothing at
+/// any rung, so walking the ladder would spend three more requests to learn the same thing
+/// and then record a confident, wrong `None`.
+fn is_fatal(msg: &str) -> bool {
```
Akarsh-Hegde, line 224: **Minor — brittle string-matching for fatal-vs-retryable classification.**

`is_fatal` classifies purely by substring match on the error message text (`"API key"`, `"unreachable"`, `"not configured"`) — a string contract with no compile-time link to wherever those messages are actually produced. A future wording change elsewhere would silently break this classification with no signal. Contrast with `test_provider`'s `RateLimited`/`Failed` split in `detect.rs`, which is a real structured enum match — worth moving `LlmError` toward carrying that same structure so this doesn't have to guess from prose.

ui/components/LlmProviderPicker.tsx
```diff
@@ -48,10 +91,44 @@ export function useLlmProviderDetection() {
...
+        const next = { ...prev }
+        for (const r of results) {
+          if (next[r.id]) next[r.id] = { ...next[r.id], last_test: r }
+        }
+        return next
+      })
+    } catch {
+      // Same reasoning as above - stale-but-cached beats blocking the panel.
+    } finally {
+      setTestingIds(new Set())
```
Akarsh-Hegde, line 112: **Nitpick — cosmetic race with per-card tests.**

This `finally` unconditionally clears the whole `testingIds` set, so if a user clicks a card's own "Test connection" while Rescan's batch is still in flight, this clobbers that card's spinner state when Rescan finishes. No data corruption (each `status[id]` write is independently scoped), and it's already largely mitigated by the per-card disabled state upstream — not worth blocking on, but consider tracking rescan-owned vs. single-test-owned ids separately if it's ever reported as visible flicker.

src/migrations/064_llm_experiments.sql
```diff
@@ -0,0 +1,50 @@
...
+    input_json  TEXT NOT NULL DEFAULT '{}',
+    -- 'running' | 'done' | 'failed' (failed = input assembly failed or every variant errored)
+    status      TEXT NOT NULL DEFAULT 'running',
+    n_variants  INTEGER NOT NULL DEFAULT 0,
+    created_at  TEXT NOT NULL,
+    finished_at TEXT
+);
+
+CREATE TABLE IF NOT EXISTS llm_experiment_results (
+    experiment_id INTEGER NOT NULL,
```
Akarsh-Hegde, line 26: **Nitpick — no `FOREIGN KEY` to `llm_experiments(id)`.**

Harmless today since nothing ever deletes an experiment row, but would let result rows orphan if a cleanup path is added later without a matching delete. Not blocking for a dev-only harness.

dev-start.sh
```diff
@@ -94,6 +116,15 @@ if pgrep -f '\.meridian/.*bin/meridian$' >/dev/null 2>&1; then
 else
     echo "  ✓ canonical launchd daemon stopped + disabled (re-enable later with: meridian start)"
 fi
+# Say so if the port is still taken. Silence here is what let a months-old runtime
+# serve every request for a whole session: the MLX window dies in its own Terminal
+# with [Errno 48] and nothing else ever mentions it.
+if [ -n "$(lsof -ti tcp:7823 2>/dev/null || true)" ]; then
+    echo "  ⚠ something is STILL holding port 7823 — the MLX window will die with [Errno 48]" >&2
```
Akarsh-Hegde: **Nitpick — em dash in terminal output.**

CLAUDE.md's plain-hyphen rule is scoped to shipped "user-facing app text"; a dev script's stderr is a gray area rather than a clear violation, but flagging for consistency if the rule is meant to be read as blanket.

Akarsh-Hegde (commented):
Stringent pass over the diff (56 files, +7313/-356 vs. base `feat/llm-provider-enum`). Full writeup below the inline comments; summarizing here.

**Scope note:** this PR bundles four logically separate changes — the connectivity-test feature, the custom-LLM-provider/LLM-Lab subsystem, the cursor-agent report-parser fix, and the fold-idempotency fix for the 2026-07-16 data-loss incidents. Worth confirming `feat/llm-provider-enum` has its own path to `pre-main`, and note that bundling means the idempotency fix (see inline comment on `workstream.rs`) can't be reverted independently of the feature if it needs a fast follow-up.

**Blocking:** an unguarded concurrent read-modify-write on `provider_test_cache.json` (`src/llm/detect.rs`) — see inline comment. Rescan reliably races this on any machine with 2+ CLIs installed, and the shared `atomic_write_json` helper's fixed temp-file name means the race can corrupt the *entire* cache, not just the racing pair's entries.

**Moderate:** the new fold-idempotency guard (`workstream.rs`) has a false-negative gap from segment coalescing that can trigger spurious re-folds — see inline comment. Also a narrow misparse risk in the cursor-agent bracket-tolerance fix (`hour.rs`) — see inline comment.

**What's solid:** `RateLimited`/`Failed` discrimination is real (keyed off actual status codes/stderr patterns, not guessed); credential hygiene is careful throughout (no API keys logged, CRLF/header-injection rejected with tests); three previously-flagged bugs in this PR's own history (camelCase args, re-test-cut-short demotion, restore-by-position) are genuinely fixed with regression tests; LLM Lab is dev-gated server-side, not just in the UI; no command-injection risk anywhere subprocess args are built; the `coding_deadline` `Instant`→UTC-wall-clock fix is a correct root-cause fix; the two production-incident regression tests are pinned to the real timestamps that caused the losses. File headers, `anyhow::Context`, structured tracing, hyphen-only user copy, and the 7-arg limit all check out across the parts sampled.

src/llm/detect.rs
Akarsh-Hegde: Fixed in 2104c030. `persist_test_result` now holds a process-wide `Mutex` across the whole load→insert→write, so concurrent persists (a Rescan testing every installed provider at once, or a card Test overlapping a batch) serialize instead of losing updates. Separately, `atomic_write_json` now gives every call a unique `pid.seq` temp-file suffix, so two writers to the same path never share a scratch file — that closes the truncate-the-whole-cache path for **all** callers, not just this one. New regression test `concurrent_persists_do_not_lose_results` spawns real OS threads and asserts none are dropped. On the `spawn_blocking` note: kept synchronous deliberately (a single tiny JSON file under the lock) and said so in the doc comment.

src/worklog_pipeline/workstream.rs
Akarsh-Hegde: Fixed in babbd9d8. You're right that geometry can't distinguish a spill from a folded-then-coalesced segment, so I added a geometry-independent signal rather than a smarter start check. `day_hour_spans`' per-hour row — now written **after** a successful fold in `run_hour` (moved below `workstream::run`, so a failed fold writes no marker and retries) — is the fold's idempotency marker, read by `hour_folded_marker_exists`. `run`'s guard OR's it with the existing segment-start check: the marker survives coalescing, and the start check still covers the narrow crash-between-fold-and-marker window. No new migration — 058's `day_hour_spans` is reused (its `span_min` stays write-only; only presence is now load-bearing), fails open on a missing table. Regression tests added for the coalesced-segment blind spot and the DB-backed marker round-trip.

src/worklog_pipeline/hour.rs
Akarsh-Hegde: Fixed in babbd9d8. Moved the `<`/`>` tolerance inside `take_hhmm`, consuming the brackets **only** when they actually wrap a real `HH:MM`, and removed the two unconditional strips from `parse_report_line`. `parse_report_line("<15 min gap, no activity>")` now returns `None` (the exact case you traced), pinned by a new test alongside the existing cursor-agent bracket case.

src/llm/probe.rs
Akarsh-Hegde: Addressed in 4e7b402e, without the full enum refactor. The silent-break risk is now covered two-sidedly: `openai_compat`'s own tests pin that it emits the `API key`/`unreachable`/`not configured` substrings, and `is_fatal`'s test pins that it classifies those as fatal — so a producer reword fails a test rather than silently mis-classifying. Documented that contract on `is_fatal`. Promoting `LlmError::Failed` to carry a structured reason touches every backend construction site (claude/codex/copilot/cursor/local), so I've left it as a noted follow-up rather than expand this already-large PR.

ui/components/LlmProviderPicker.tsx
Akarsh-Hegde: Acknowledged — leaving as-is per your "not worth blocking" note. No data impact (each `status[id]` write is independently scoped) and the per-card disabled state already mitigates it. Will split rescan-owned vs single-test-owned ids only if visible flicker is actually reported.

src/migrations/064_llm_experiments.sql
Akarsh-Hegde: Acknowledged, won't-fix by design. Adding the FK would mean either modifying an existing migration (forbidden — it changes the sqlx checksum and crash-loops deployed DBs) or a full table rebuild, for a dev-only harness table nothing ever deletes from. Leaving it; if a cleanup path is ever added it should come with the FK in a new migration.

dev-start.sh
Akarsh-Hegde: Fixed in 4e7b402e — em dash replaced with a spaced hyphen in the port-conflict warning.

Akarsh-Hegde:
## Review fixes applied — ready to merge

All seven inline findings are addressed and their threads resolved. Three new commits on top of the branch:

| # | Finding | Severity | Disposition | Commit |
|---|---------|----------|-------------|--------|
| 1 | `detect.rs` — unguarded concurrent cache read-modify-write | **Blocking** | `persist_test_result` now holds a process-wide `Mutex` over the whole RMW; `atomic_write_json` gets a unique `pid.seq` temp suffix so concurrent writers never share a scratch file. Regression test spawns real threads. | `2104c030` |
| 2 | `workstream.rs` — fold false-negative from segment coalescing | Moderate | Added a geometry-independent marker: `day_hour_spans`' per-hour row (written *after* a successful fold) read by `hour_folded_marker_exists`, OR'd with the existing segment-start check. No new migration (058 reused). Tests for the coalesced-segment blind spot + DB marker round-trip. | `babbd9d8` |
| 3 | `hour.rs` — bracket strip not gated on a real timestamp | Moderate | `<`/`>` tolerance moved inside `take_hhmm`, gated on a real `HH:MM`; `"<15 min gap, no activity>"` now returns `None`. Test pinned. | `babbd9d8` |
| 4 | `probe.rs` — brittle `is_fatal` string matching | Minor | Documented the cross-file substring contract; two-sided test coverage (producer emits the substrings, `is_fatal` classifies them) makes a reword fail a test instead of silently misclassifying. Structured-`LlmError` refactor noted as a follow-up. | `4e7b402e` |
| 5 | `LlmProviderPicker.tsx` — cosmetic spinner race | Nitpick | Acknowledged, left as-is per your note (no data impact; per-card disabled state mitigates). | — |
| 6 | migration 064 — no FK to `llm_experiments` | Nitpick | Won't-fix by design: an FK needs either modifying an existing migration (breaks sqlx checksums on deployed DBs) or a table rebuild, for a dev-only table nothing deletes from. | — |
| 7 | `dev-start.sh` — em dash | Nitpick | Fixed. | `4e7b402e` |

**Verification:** `cargo fmt` + `cargo clippy --all-targets -D warnings` clean; affected suites pass (`worklog_pipeline` 60, `llm` 88, `meridian-core fs_utils`); full pre-push suite (fmt, clippy, UI build, UI tests, `cargo test`, security audit) passed on push.

**One thing for a human to decide before merge** (from your scope note, not something I resolved): this PR is based on `feat/llm-provider-enum`, and bundles four logically separate changes. Worth confirming `feat/llm-provider-enum` has its own path to `pre-main`, and noting that the fold-idempotency fix can't be reverted independently of the feature if it needs a fast follow-up. That's a merge-strategy call for you.

PR #454, opened 2026-07-17
feat(llm): centralised AI-provider selection - custom endpoints, connectivity testing, and a dev-only LLM Lab

adityaharishch:
## What does this PR do?

This is the `feat/llm-provider-enum` integration branch. Its through-line is giving Meridian a real, user-owned **AI-provider layer** - one place to pick which model runs the prose pipeline, add your own cloud endpoint, prove it actually works before it's trusted in production, and (in dev) compare providers head-to-head. Alongside that it folds in two adjacent pieces of pipeline work that matured on the same branch: **worklog posting to multiple tickets** and the **plan-task authoring** flow.

It targets `pre-main` and carries the merged, review-fixed work from #452, #453, #455 and #457.

### 1. Centralised AI-provider selection
- A single provider choice (`meridian-core/src/llm_provider.rs`, `settings.rs`) surfaced in **Settings -> Intelligence** and the **first-run setup wizard**, replacing the implicit "whatever CLI we found" behaviour. Consumed through `LlmProviderPicker.tsx` (restyled for readability, the "recommended" badge dropped) and `resolver.rs`.

### 2. Custom OpenAI-compatible cloud providers
- Add and select your own OpenAI-compatible endpoint straight from the provider picker (`src/llm/openai_compat.rs`, `custom_llm.rs`, `CustomProviders.tsx`).
- Before an endpoint is allowed to run production work, Meridian **probes its real schema support** (`src/llm/probe.rs`, `schema.rs`) and gates on it, so a model that can't honour structured output never silently corrupts the pipeline.
- Codex now speaks OpenAI's strict-schema dialect and reports its real error (`src/llm/codex.rs`).

### 3. Provider connectivity testing
- A real connectivity test for every installed provider CLI and custom endpoint (`src/llm/detect.rs`), wired into both Settings and the setup wizard, with results cached to disk.
- Honours the provider's **own rate-limit reset time** rather than a fixed backoff (`src/llm/reset_time.rs`), and serialises cache writes so a concurrent "Rescan" can't lose results.

### 4. LLM Lab - dev-only comparison harness
- A dev-channel-only modal (`ui/components/timeline/llmlab/*`) that replays one prose stage (hour report, workstream fold, worklog generate, day fold) across several provider/model variants and renders the outcomes as real dashboard timelines for side-by-side comparison.
- Backed by an experiment ledger (`src/llm_experiment/*`, migration **064**), a `meridian-core` reader, dev-only tray commands, and a `meridian llm-experiment` CLI.

### 5. Worklog posting to multiple tickets (migrations 061-063)
- The worklog matcher now compares a day's work against that day's **planned** tasks only, with a **manual target** override so the user can file against any board ticket (061).
- One day-task's update can post to **several tickets**, promoted into a child table with per-ticket idempotency (062) plus a pre-call post-attempt marker that closes the crash-between-post-and-record double-post gap (063). Surfaced via `WorklogTargets.tsx` / `WorklogTicketPicker.tsx` and a large rework of `pm_worklog/generate.rs`.

### 6. Plan-task authoring
- Create / draft / edit planned tasks (`src/plan_tasks/*`, `task_create.rs`, `services/prompts/plan-task-draft.md`), with a task composer and drag-and-drop plan board (`ui/components/plan/*`, `TaskComposer.tsx`).

### 7. Pipeline correctness & dev-loop fixes (folded from #453/#457)
- Geometry-independent worklog fold idempotency and bracket-gated report parsing (`worklog_pipeline/`).
- `coding_deadline` anchored to the hour's end rather than process start.
- `dev-start.sh` now watches only what rebuilds the daemon and stops a stray installed-runtime MLX server so a dev run uses its own `services/`.

## How was it tested?

- [x] `cargo test` passes (new coverage in `meridian-core/tests/readers.rs`, worklog/llm unit tests)
- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo fmt --check` passes
- [x] UI builds (`cd ui && npm run build`) - full pre-push suite green on the latest push
- [ ] Manually tested the relevant flows

## Notes for the reviewer

- **Large branch:** ~120 files. It's an integration branch that accumulated several feature merges; the sections above are the review seams.
- **Runtime version bump required before merge:** this PR changes `services/`, so `services/pyproject.toml` must be bumped above the live `runtime-staging` (currently 1.70.0) or the rebuilt MLX runtime won't ship to staging (the auto-upgrade is an equality check). The CI "Runtime version bump check" enforces this.
- New migrations 061-065 are append-only. Most are additive; **062 is destructive-but-safe** - it promotes the worklog match columns into the `day_task_worklog_targets` child table via a carry-over `INSERT...SELECT` and then `DROP COLUMN`s the six migrated fields (the data moves, nothing is lost). **065** (added in review) adds `create_attempt_at`, the write-ahead create-claim marker described in the worklog section.


Akarsh-Hegde (commented):
## Review summary

Reviewed section-by-section against the PR description's own seams (provider selection, custom endpoints/schema probing, connectivity testing, LLM Lab, worklog multi-ticket posting, plan-task authoring, pipeline/misc). Overall the connectivity-test race fix (2104c030) holds, migration hygiene is good (061/063/064 additive; 062 is a deliberate destructive-but-safe carry-over, see note below), and conventions (anyhow context, tracing, no stray `unwrap()`) are followed almost everywhere. The issues below are concrete, not style nits — grouped by severity.

### 🔴 High severity

1. **LLM Lab's CLI path isn't dev-gated and can fire real billed LLM calls in a production build.** `src/main.rs` dispatches `meridian llm-experiment run/create/exec` with no build-channel check. `cli.rs`'s doc comment claims these are "ungated... no network, no cost," but `runner::exec` → `variant_backend` → `resolver::backend_for` calls the same live Claude/Codex/custom-endpoint backends the production pipeline uses. Only `draft-task` checks `cfg!(debug_assertions)`; `run`/`create`/`exec` don't. The tray commands correctly gate via `dev_only()`, but that only protects the UI — anyone with terminal access to a packaged install can run real, metered completions today. Please add the same dev-channel check to the CLI dispatch, or gate `resolver::backend_for` itself for experiment callers.

2. **Worklog multi-ticket posting has two real idempotency gaps** (`meridian-core/src/readers/day_task_worklogs/`):
   - `mod.rs:335-336`/`392`: the parent-row `state='drafted'` guard and the child-table `targets::replace` (delete-then-insert) are two unguarded statements, not one transaction. A concurrent `approve()` landing in that gap wipes a just-posted `posted_comment_id` and lets the next approve re-post a duplicate comment. This is the exact DELETE-then-INSERT-on-a-child-table pattern CLAUDE.md forbids for idempotent tables — it just wasn't extended to this new one.
   - `generate.rs:218`: `mark_approved` is called unconditionally with no CAS, so two concurrent `approve()` calls for the same row both read `created_task_key == None` and both call `create_ticket` — two real tickets get created and commented on. Migration 063's `post_attempt_at` marker only covers the comment-post step, not ticket creation. `main.rs:628` calls `approve()` with no serialization; the only protection today is a client-side same-tab guard in `useWorklog.ts`.
   - Minor/factual: migration 062 is not additive as the PR description implies — it `DROP COLUMN`s 6 fields after a carry-over `INSERT...SELECT`. The migration itself looks correct, just flagging the description.

3. **Plan-task authoring has a TOCTOU race that both breaches the task cap and can file duplicate real tracker tickets.** `meridian-core/src/readers/plan.rs:883-923`, the `"add"` arm of `apply_plan_action`: the row-count cap check, `MAX(position)` read, and `INSERT` are three separate un-transacted queries, unlike the `"confirm"`/`"set"`/`"reorder"` arms which do use `pool.begin()`. Two concurrent adds (AI draft-task flow racing a manual board `+Add`) can both pass the cap check and land duplicate positions. Because `src/plan_tasks/create.rs`'s `ensure_plan_has_room` precedes this same racy path, two concurrent creates can each pass their own pre-check and each file a real Jira/Linear ticket, blowing past the 10-task cap with two real external side effects. Recommend wrapping the `"add"` arm in a transaction like its siblings.

4. **Custom-provider `base_url` leaks into telemetry despite the code's own stated intent not to.** `tray/src-tauri/src/commands/custom_llm.rs:1428-1429`: `#[tracing::instrument(skip(api_key), fields(has_key = ...))]` only skips `api_key` — `tracing::instrument` auto-captures every other parameter, including `base_url`, as a span field. `openai_compat.rs`'s own doc comment says a base URL can carry a key in a query string and "never log the key or the URL" — but this span ships it to the telemetry spool unconditionally and into diagnostics export bundles. Recommend `skip(api_key, base_url)` plus a redacted/host-only field if the URL is useful for debugging.

5. **Rate-limit backoff isn't cleared when the user switches providers, and can now last hours instead of a fixed 30 min.** `src/llm/resolver.rs`: `RATE_LIMITED_UNTIL` is one process-wide mutex (not per-provider); `start_backoff` now uses the provider's own parsed reset time (can be multi-hour or weekly) instead of the old fixed cap. `clear_backoff()` is documented as firing on "an explicit user-driven provider change" but is only ever called from this file's own tests — never from the settings-write path. A user who hits a long rate limit and explicitly switches CLI/endpoint stays silently pinned to the on-device model for the full window, which is a regression from pre-PR behavior and contradicts the module's stated "user's choice is sacred" invariant.

### 🟡 Medium severity

6. Same file as #4 (`custom_llm.rs`): `add_custom_llm_provider`/`probe_custom_llm_provider`/`remove_custom_llm_provider` each do an unsynchronized read-modify-write of `settings.json`'s `custom_llm_providers` array. Two concurrent calls (e.g., clicking Test on two cards) can lost-update each other. This is the same bug class already fixed for the provider-test cache in 2104c030, just not applied to this registry.
7. `src/llm/openai_compat.rs:248-251` — no explicit `redirect::Policy` on the reqwest client used against user-supplied endpoints; worth pinning `.redirect(Policy::none())` given the Bearer key is sent to an arbitrary host.
8. `src/llm/probe.rs`'s `is_fatal` (~line 834) substring-matches up to 300 chars of raw vendor error text; a 400 schema-refusal whose body happens to contain "API key" gets misclassified as endpoint-fatal, surfacing a misleading message.
9. `api_key` is protected only by convention, not the type system: `CustomLlmProvider`, `CustomEndpoint`, and `LlmConfig` all plain-`derive(Debug)` over a raw `String` key. Any future `{:?}` or `format!("{:?}", ...)` in a `.context()` call leaks it into the spool. Consider a newtype with a redacting `Debug` impl.
10. `src/llm/detect.rs`'s `persist_test_result` caches backend error text to disk verbatim with no redaction, unlike `openai_compat.rs`'s deliberate care — worth a shared sanitization step at the connectivity-test boundary too.
11. `ui/components/LlmProviderPicker.tsx`: the nested "Test connection" button only `stopPropagation()`s `click`, not `keydown`, so pressing Enter/Space while it's focused also bubbles into the card's `onKeyDown` and switches the selected provider.
12. `src/llm/reset_time.rs:126` — `"reset"` matching finds the substring inside `"resetting"` first and stops there, missing a later real `"reset at 3pm"` in the same string; falls back to flat backoff. Low real-world hit rate but worth a `word-boundary`-aware search.
13. `src/llm_experiment/store.rs`'s `load_pending` resumes rows still `running` after a killed process; if the daemon dies after a real provider call succeeded but before `finish_variant` records it, resuming re-issues the same paid call (compounds finding #1's severity since the CLI path isn't gated).

### 🟢 Low / nits

- `ui/components/integrationConnectStore.ts:189` — `'Timed out — try again'` uses an em-dash in user-facing text; CLAUDE.md requires a plain hyphen. Pre-existing string relocated during the store extraction but still a `+` line here.
- Stale comments referencing "pre-061 DB" that should say "pre-064" (`tray/src-tauri/src/commands/llm_lab.rs:131`, `ui/components/timeline/llmlab/useLlmLab.ts:34`), copy-pasted from #453's migration.
- `resolver.rs:57` doc comment still says "a minute later" against a 30-min constant.
- `src/worklog_pipeline/hour.rs`/`workstream.rs`: the fold and the `upsert_hour_span` marker write are two separate statements; a crash between them can trigger an unnecessary re-fold (extra LLM call) on retry — low severity since coalescing makes the re-fold a safe no-op, but worth a one-line note.

### Test coverage gaps worth calling out
- No mock-server test for `openai_compat::complete()`'s HTTP wiring (headers/parsing/non-2xx) — only an `#[ignore]`d live test needing real credentials.
- `ui/__tests__/plan-store.test.ts` / `task-composer.test.ts` regex-match source text rather than exercising runtime behavior (no drag-reorder, no cap-boundary, no mocked-bridge error paths).
- No test for `detect_all_with_cache()`'s id-based merge of cached results onto live providers.
- Migration 064's schema and meridian-core reader are solidly tested (including the pre-064 degrade path) — no notes there.

### What's solid
- Bracket-gated report parsing and `hour_bounds`/`coding_deadline` wall-clock anchoring fixes are correct and well-tested.
- `fs_utils.rs`'s per-call-unique temp filename correctly fixes a real concurrent-writer race.
- The per-target CAS (`begin_post`/`revert_post`/`mark_posted`, migration 063) does correctly close the crash-between-post-and-record window for the actual provider call, with thorough test coverage.
- Connectivity-test cache locking/atomicity (2104c030) holds under this PR's changes.
- No XSS/sanitisation issue found in plan-task rendering (React-escaped, no `dangerouslySetInnerHTML`).
- `Cargo.lock`/`Cargo.toml` change is just `tempfile = "3"` as a dev-dependency — benign.

---
Given the scale of this integration branch (123 files), I'd prioritize fixing #1-#5 before merging to `pre-main` — #1-#3 have real production-cost/data-integrity/external-side-effect consequences, not just theoretical races.

Akarsh-Hegde:
Thanks for the thorough review - pushed `986a307c` addressing it. Summary of what changed and what's deliberately deferred.

## High severity - all fixed

**#1 - LLM Lab CLI dev-gating.** `run`/`create`/`exec` now refuse to run in a release binary (they drive real, metered completions through `resolver::backend_for`), matching `draft-task` and the tray's `dev_only()`. `list`/`get` stay open (reads only). Corrected `cli.rs`'s false "no network, no cost" doc.

**#2 - Worklog multi-ticket idempotency.**
- `upsert_draft`/`retarget_draft` now run the parent guard-write and `targets::replace` in **one transaction** (`targets::replace` takes `&mut SqliteConnection`), so a concurrent `approve()` can't slip a comment-post between the guard matching and the child-table DELETE.
- Added a **create-claim CAS** (`begin_create`/`revert_create` + migration **065**'s `create_attempt_at`), the create-step analog of your `post_attempt_at`. Two concurrent approves - or a crash between `create_ticket` returning and `mark_created` recording the key - can no longer each file a real ticket. New tests cover exclusivity, revert-frees-retry, stale-reclaim, and success-resolves.
  - One conscious divergence from `post_attempt_at`, documented in migration 065 and `begin_create`: a dangling *create* marker **auto-reclaims** past a bounded stale window rather than being a permanent dead end. A post marker can be a dead end because its target surfaces `outcome_unknown` in the UI; an `approved` row has no regenerate path and creates aren't surfaced, so a pure dead end would wedge the draft forever. The accepted cost is a minutes-wide window where a crash-after-success + re-approve could double-create - strictly better than a permanently stuck draft. Happy to instead surface it in the read struct/UI like `outcome_unknown` if you'd prefer to match precedent exactly.

**#3 - Plan-task "add" TOCTOU.** The cap check, `MAX(position)`, and INSERT are now a **single atomic conditional INSERT** (`... HAVING COUNT(*) < cap ... ON CONFLICT DO NOTHING`), so concurrent adds can't both pass the cap. Existing cap test still green.
  - **Deferred (not closed here):** the second-order race you noted - two concurrent *creates* in `plan_tasks/create.rs` each passing `ensure_plan_has_room` and each filing a real Jira/Linear ticket. That's a cross-process network call that can't sit in a SQLite transaction; a correct fix needs a slot-reservation design and belongs in its own PR with its own tests. The DB cap is now hard, so this can at most leave **one** orphaned ticket on a genuine race, never an over-full plan.

**#4 - base_url telemetry leak.** `add_custom_llm_provider`'s span now `skip`s `base_url` too (a URL can carry a key in a query string). `endpoint_id`+`vendor` remain for debugging.

**#5 - Rate-limit backoff.** Now keyed **per backend** (custom endpoints by id), stored in a map instead of one process-wide cell. Switching to a provider that isn't limited escapes the backoff immediately - which also dissolves the cross-process problem: the daemon holds the backoff and the tray writes settings, so the documented clear-on-write could never have fired. Dropped the misleading `clear_backoff` public API (now test-only). New test asserts a Claude backoff doesn't divert a switched-to Codex.

## Medium / nits - fixed

- **#6** custom_llm `add`/`probe`/`remove` serialize their `settings.json` read-modify-write behind a `tokio::Mutex` (lost-update, same class as 2104c030; the daemon only reads that array, so a tray-process lock is sufficient).
- **#7** custom-endpoint reqwest client sets `redirect::Policy::none()` so a 3xx can't forward the Bearer key.
- **#8** `probe::is_fatal` now matches the producer's exact phrases (`"rejected the API key"`, `"custom provider unreachable"`, `"selected but not configured"`), not bare `"API key"`; added a test for a 400 schema-refusal that mentions a key.
- **#11** the Test button stops `Enter`/`Space` from bubbling to the card's `onKeyDown`.
- **#12** `reset_time` tries every `"reset"` occurrence so `"resetting"` can't shadow a real `"reset at 3pm"`; added a test.
- **#13** severity is neutralized by #1 (the resume path is now dev-only); documented the deliberate "resume re-issues one paid call after a crash" tradeoff on `runner::exec`.
- **Nits:** fixed the em-dash in the user-facing timeout string; `pre-061` -> `pre-064` comments; `resolver.rs` "a minute later" -> "half an hour"; PR description's migration-062 "additive" claim corrected (it's destructive-but-safe).

## Deferred as follow-ups (happy to file issues / a stacked PR)

- **#9 / #10 - key/error redaction hardening.** A redacting `Debug` newtype for `api_key` (#9) touches several structs + their serde, and sanitizing cached backend error text (#10) needs a shared redaction step at the connectivity-test boundary. Both are real hardening but broader than this review pass and read as "consider" - keeping them out to avoid ballooning a feature PR.
- **Low nit** - `hour.rs`/`workstream.rs` fold + marker two-statement window: left as-is since, as you noted, coalescing makes the re-fold a safe no-op.
- **Coverage gaps** you flagged (mock-server test for `openai_compat::complete()`, behavior-level plan-store/task-composer tests, `detect_all_with_cache` id-merge): worth doing, tracked as follow-up.

All local checks green (cargo fmt + clippy + test, UI build + `bun test`, integration tests). Ready for another look - leaving the merge to you.

PR #458, opened 2026-07-17
feat(llm-lab): full-screen surface with a variant switcher and task sidebar

Akarsh-Hegde:
> **Base / scope note:** this PR targets `feat/llm-provider-enum`, which trails `feat/llm-provider-connectivity-test` (the branch this was cut from) by the whole in-flight LLM-provider + Lab stack. So the raw diff is large (~127 files). The change THIS PR adds is the **top two commits** (`875641ea`, `8f8a7853`) - 11 files under `ui/components/timeline/llmlab/`, `src/llm_experiment/`, `src/pm_worklog/generate.rs`, and the tray `llm_lab` command. Re-target the base to `feat/llm-provider-connectivity-test` for a clean 2-commit review.

## What

Turns the dev-only **LLM Lab** from a 1180px modal into its **own full-screen surface**. A run now shows **one variant's timeline at a time** with a **variant-switcher** tab strip, clicking a task card in a fold variant opens a **task sidebar** beside the timeline, and the sidebar can **draft that task's worklog on demand** with the shown variant.

## Phase A - full-screen surface, switcher, sidebar

- **`LlmLabModal` -> `LlmLabScreen`**: full-viewport panel (`absolute inset-0`, own header, Escape-to-close) replacing the `ModalShell` wrapper.
- **`RunView`** (new): run header + variant-switcher tab strip + single full-width body. Switching variant **clears the task selection** - task ids (`T1`,`T2`) are per-variant (each model builds its own day), so a stale id must not carry across a switch.
- **`VariantBody`** (extracted from the former `ResultsGrid.VariantColumn`): one variant, full-width, preserving every `ok` / `failed` / `rate_limited` / `pending` branch so a non-ok variant shows its status, not a blank pane.
- **`LabTaskSidebar`** (new): Lab-local read-only detail (When + the model's per-task log) from the `DayTaskDetail` the timeline already emits. **Deliberately not `DayTaskDetailPanel`/`useWorklog`** - those drive production worklog tables for a task id, and a fold's ids are a model's simulated day, not real rows.
- `ResultsGrid.tsx` / `LlmLabModal.tsx` removed; shell import + JSX updated.

## Phase B - on-demand "Draft with this model"

The sidebar drafts THIS fold task's worklog with whichever variant the timeline is showing - a **metered, ephemeral** call, gated behind a **free/local caution**, that writes nothing.

- Because a fold task id is a model's simulated day (not a production `day_tasks` row), the draft is built from the task's **inline content**, never a production lookup:
  - `generate.rs`: `generate_request_from_task` builds the worklog-generate request from an inline title/summary/minutes, reusing the real board candidate set + the identical prompt/schema.
  - `runner.rs`: `variant_backend`'s core is now `resolve_backend(provider, model)`, shared with the draft path (`custom:<id>` included).
  - `cli.rs`: `meridian llm-experiment draft-task --day --variant --task-json` - one completion, prints `{"draft": …}`, persists nothing.
  - tray `draft_lab_worklog`: dev-gated command that shells out to it.
  - `LabTaskSidebar`: the button + free/local caution + a formatted/raw render of the drafted update (posts-to / summary / sections).

## Reconciliation carried in code

A fold `DayTask` has no drafted-worklog prose - only `summary` bullets. Real prose comes only from `worklog_generate`, which reads production `day_tasks`. So the sidebar shows the summary log for free, and the polished prose is the **on-demand, metered** draft above - never auto-run.

## Gating

Unchanged - dev-only (`channel === 'dev'` in the shell + `commands/llm_lab.rs` release refusal). This surface does not exist for users.

## Verified

- Daemon: `cargo clippy` clean; `llm_experiment` tests (20) pass.
- Tray: `cargo clippy -- -D warnings` clean.
- UI: `npm run build` (TypeScript pass) + `bun test` (227 pass).
- **Draft path exercised end-to-end** against the freshly-built binary (temp DB, free `local` variant): `meridian llm-experiment draft-task --variant local …` returns a real drafted worklog in the `{update:{summary,sections}}` shape the sidebar renders - so `generate_request_from_task` -> `resolve_backend` -> `complete` all run.
- Pre-push suite (fmt + clippy + cargo test + ui build/tests + security audit) green.

## Not yet verified

- The **in-app GUI click-through** in a live tray (full-screen layout, variant switcher, sidebar, and the draft button invoke) has not been run - it needs a dev tray build. Recommend a smoke test on a real `day_fold` run before merge.

### DIFF
--- a/src/llm_experiment/cli.rs
+++ b/src/llm_experiment/cli.rs
@@ -1,5 +1,5 @@
 //ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
-//! `meridian llm-experiment <run|create|exec|list|get>` — the LLM-Lab CLI surface.
+//! `meridian llm-experiment <run|create|exec|list|get|draft-task>` — the LLM-Lab CLI surface.
 //!
 //! * `run    --process hour-report --hour 2026-07-15T14 --variants claude,codex:gpt-5.1,local`
 //!   — create + execute + print the detail JSON (the human one-shot). A `custom:<endpoint-id>`
@@ -12,12 +12,20 @@
 //!   UI polls progress instead of holding a long invoke). Resumable.
 //! * `list   [--limit N]` / `get --id N` — read back via
 //!   `meridian_core::llm_experiments`, printed as one JSON line.
+//! * `draft-task --day D --variant p[:model] --task-json '{…}'` — the ONE exception
+//!   to the paragraph below: an ephemeral, on-demand worklog draft for a single
+//!   inline task that fires a REAL, metered LLM completion (see [`draft_task`]).
 //!
-//! Deliberately ungated in release binaries: it is useful for field debugging and
-//! only ever writes the experiment tables. The UI surface is the gated part.
+//! `run`/`create`/`exec`/`list`/`get` are deliberately ungated in release binaries:
+//! they are useful for field debugging and **only ever write the experiment
+//! tables** — no network, no cost — so an unattended run is harmless. `draft-task`
+//! is different (it calls a live provider and can incur cost), so it alone is
+//! **dev-gated** here too — a release binary refuses it, matching the tray's
+//! `draft_lab_worklog` gate. The UI surface as a whole is the gated part.
 //!
 //! # Who calls this
-//! `main.rs`'s `llm-experiment` dispatch block; the tray's `run_llm_experiment`.
+//! `main.rs`'s `llm-experiment` dispatch block; the tray's `run_llm_experiment`
+//! (and, for `draft-task`, `draft_lab_worklog`).
 
 use anyhow::{bail, Context, Result};
 use sqlx::SqlitePool;
@@ -82,7 +90,17 @@ pub async fn run(pool: &SqlitePool) -> Result<()> {
 /// SIMULATED day, not a production `day_tasks` row - so there is nothing to read
 /// by id. Fires a real, metered completion; the UI gates this behind a
 /// free/local caution. The tray `draft_lab_worklog` command shells out to it.
+///
+/// Unlike its sibling subcommands (which only write local experiment tables), this
+/// one costs money, so it is refused in a release build - it runs only from a dev
+/// binary, the same `debug_assertions` signal the tray's `dev_only` gate uses.
 async fn draft_task(pool: &SqlitePool, args: &[String]) -> Result<()> {
+    if !cfg!(debug_assertions) {
+        bail!(
+            "draft-task is a dev-only subcommand - it fires a real, metered LLM completion, \
+             so a release binary refuses it"
+        );
+    }
     let day = flag(args, "--day").with_context(|| format!("draft-task needs --day - {USAGE}"))?;
     let variant =
         flag(args, "--variant").with_context(|| format!("draft-task needs --variant - {USAGE}"))?;
--- a/src/llm_experiment/cli.rs
+++ b/src/llm_experiment/cli.rs
@@ -24,12 +24,14 @@ use sqlx::SqlitePool;
 
 use super::{
     parse_variants, request, runner, store, ExperimentInput, ExperimentProcess, ExperimentSpec,
+    Variant,
 };
 
 const USAGE: &str = "usage: meridian llm-experiment \
 run|create --process hour-report|workstream-fold|worklog-generate|day-fold \
 (--hour YYYY-MM-DDTHH | --day YYYY-MM-DD [--task-id T]) --variants p1,p2:model,custom:id,… \
-| exec --id N | list [--limit N] | get --id N";
+| exec --id N | list [--limit N] | get --id N \
+| draft-task --day YYYY-MM-DD --variant p[:model] --task-json '{\"title\":…,\"summary\":[…],\"minutes\":N}'";
 
 /// Parse argv and dispatch. Prints exactly one JSON line on success (except
 /// `create`'s `{"experiment_id":N}` — also one line); errors bubble to `main.rs`.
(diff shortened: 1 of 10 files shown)


Akarsh-Hegde:
## Review scope

Per the PR description, I reviewed the actual 2-commit diff (`875641ea`, `8f8a7853`) against `feat/llm-provider-connectivity-test` rather than the raw ~127-file diff against the stale `feat/llm-provider-enum` base — confirmed this is exactly the 11-file / 563(+)/101(-) change described.

## Overview

Solid piece of work: turns the LLM Lab modal into a full-screen surface with a variant switcher (`RunView`) and a read-only task sidebar (`LabTaskSidebar`) that can fire an on-demand, ephemeral worklog draft against the currently-shown variant. The Rust side cleanly factors `resolve_backend` out of `variant_backend` and adds a parallel `generate_request_from_task` next to `generate_request` for the inline (non-DB) case. I traced the request/response shapes end-to-end (`WorkstreamReport` fields, `DayTaskDetail` → `draft_lab_worklog` body → `--task-json` → `InlineTask` → `generate_request_from_task`, and the `custom:<id>` / `provider:model` variant-token round-trip between `RunView.variantToken()` and `Variant::parse`) — all consistent, no shape mismatches.

## Findings

**1. Module doc drift in `tray/src-tauri/src/commands/llm_lab.rs` (minor, but worth a line)**
The file header still says *"Three commands behind the LLM Lab modal"* and lists only `run_llm_experiment` / `get_llm_experiments` / `get_llm_experiment` — `draft_lab_worklog` isn't mentioned, and "modal" is now stale terminology (it's a full-screen surface post-`LlmLabScreen`). Given this repo's convention of keeping module `//!` docs authoritative (see CLAUDE.md's "Document it (required)" step), this should get a fourth bullet.

**2. A load-bearing doc claim is now false, in two places**
- `tray/src-tauri/src/commands/llm_lab.rs`'s header: *"The `llm-experiment` CLI itself stays ungated for field debugging — the UI is the gated surface."*
- `src/llm_experiment/cli.rs`'s header doesn't call this out explicitly, but its dev-gating rationale (ungated in release because it "only ever writes the experiment tables") is implicit throughout.

`draft-task` breaks that invariant: it's a CLI subcommand that fires a real, metered LLM completion and is **not** gated by `dev_only()` — only the tray's `draft_lab_worklog` wrapper is. So on a shipped release binary, `meridian llm-experiment draft-task ...` run directly from a terminal will hit a live provider and incur cost, unlike every other subcommand in this file (which only touch local experiment tables). That may well be an acceptable/intentional tradeoff (it's still explicit, deliberate CLI use — not something a user stumbles into), but the doc comments should say so, since they currently assert the opposite ("only ever writes the experiment tables").

**3. Sticky variant selection on an in-progress run (design nit, not a bug)**
`RunView`'s `firstOk` is `useMemo`'d but only consumed via `useState(firstOk)`, so it's genuinely computed once at mount (matches the comment). If you open a still-running run, `selectedIdx` starts at 0 (no `ok` yet) and never auto-advances to the first variant that finishes — you have to manually click the switcher tab even after your currently-viewed variant fails and a later one succeeds. Probably fine given the explicit "computed once" comment, but worth a decision check: was auto-follow-first-completion considered and deliberately dropped, or just not thought about?

**4. No automated coverage for the new Rust prompt-assembly / resolution paths**
`resolve_backend`, `generate_request_from_task`, and `draft_task` have no unit tests. I manually verified `generate_request_from_task`'s `WorkstreamReport` construction is field-for-field identical to `load_workstream_report`'s (so the inline draft path builds the byte-identical prompt shape modulo the DB read), but a unit test asserting `build_user_prompt` output is equal for an inline task vs. an equivalent DB-loaded one would catch future drift between the two paths cheaply (no network/DB needed for that assertion). The PR description's "Verified" section covers this manually/end-to-end, which is reasonable for a dev-only surface, but the two prompt-builders diverging silently later is exactly the kind of thing a golden-string test would catch.

## What's solid

- **Type/shape correctness**: `DayTaskDetail.{day,title,minutes,summary}` → `draft_lab_worklog` body → `InlineTask` all line up exactly; `DayTaskColumn`'s `onSelect: (detail: DayTaskDetail | null) => void` signature matches `VariantBody`'s `onSelectTask` passthrough (an improvement over the old `ResultsGrid`, which downgraded to just an id).
- **Variant token round-trip**: `variantToken()` in `RunView.tsx` and `Variant::parse` in the daemon agree on `custom:<id>` / `provider:model` / bare-`provider` forms.
- **The `let-else` refactor in `resolve_backend`** (`let Some(provider) = LlmProvider::from_wire(provider) else { ...{provider:?}... }`) is correct — the `else` branch can't see the pattern binding, so the error message correctly prints the original `&str` argument, not an unbound name. Easy thing to get wrong; it's right here.
- **No shell-injection surface**: `draft_lab_worklog`'s JSON-encoded `--task-json` arg goes through `tokio::process::Command::args(...)`, never a shell, so arbitrary task titles/summaries containing quotes/newlines are safe.
- **Gating**: `draft_lab_worklog` correctly calls `dev_only()` first, matches the existing Lab commands' pattern, registered in `lib.rs`'s `invoke_handler!`.
- Stale-reference check (`ResultsGrid`/`LlmLabModal`/`VariantColumn`) came back clean — the rename left no dangling imports.
- CSS classes (`bg-panel`, `rise`, `z-40`) match existing `ModalShell`/`ReportModal` conventions exactly.

## Suggested action

None of the findings are blocking — #1 and #2 are a couple sentences of doc fix-up, #3 is worth a one-line confirmation from you on intent, #4 is a nice-to-have. I'd merge after a quick doc pass on #1/#2, especially #2 since it's a security/cost-surface claim that's now inaccurate.

Akarsh-Hegde:
## Review addressed — `3f4f65cb`

Thanks for the thorough pass. All four findings resolved:

**#2 (metered CLI claim — the load-bearing one): fixed, and closed the hole rather than just documenting it.** Rather than leave `draft-task` ungated-with-a-caveat, I gated it: it now **refuses in a release build** (`cfg!(debug_assertions)`), matching the tray's `dev_only` gate. This is safe for the dev flow - a debug tray resolves `meridian_bin()` to `target/debug/meridian` (debug_assertions on), and in release the tray's `draft_lab_worklog` refuses before shelling out anyway, so the button never regresses. The cost-surface (a shipped binary running `meridian llm-experiment draft-task` and hitting a live provider) is gone. Both doc headers (`cli.rs`, tray `llm_lab.rs`) now state the invariant accurately: the read/run subcommands stay ungated because they only write local tables; `draft-task` is the metered exception and is dev-gated.

**#1 (module doc drift): fixed.** The tray `llm_lab.rs` header now says **four** commands (adds `draft_lab_worklog`), drops the stale "modal" wording for "full-screen surface", and the "# Who calls this" section names the draft invoke.

**#3 (sticky variant selection): implemented auto-follow.** Not intentional before - now `RunView` auto-follows the first variant to finish **until the user manually picks a tab**, then freezes on that choice. It won't jump back to a lower-index variant that finishes later (guards on the shown variant already being `ok`). So opening a still-running run lands on whichever model completes first instead of a pending/failed one.

**#4 (no coverage): added two tests in `generate.rs`.**
- `build_user_prompt_pins_the_shape_both_draft_paths_share` - a golden on the shared assembler both `generate_request` and `generate_request_from_task` feed.
- `inline_draft_request_matches_the_worklog_generate_contract` - asserts `generate_request_from_task` builds the byte-identical system/schema/max_tokens contract as the DB-backed path, and that the inline title/summary/minutes reach the workstream block.

**Verified:** daemon `pm_worklog::generate` tests (19) pass; tray `cargo clippy` clean; ui `npm run build` (TypeScript) + `bun test` (227) pass; pre-push suite green.

Ready to merge from my side. One reminder carried over from the description: the **in-app GUI click-through** in a live dev tray still hasn't been run (the draft path is verified end-to-end via the CLI, but not the React wiring) - worth a quick smoke test on a real `day_fold` run before merge.

PR #463, opened 2026-07-18
docs(whats-new): add the 1.72.0 entry for the post-1.71.0 work

Akarsh-Hegde:
## What

Adds a new **1.72.0** entry to the front of `tray/src-tauri/resources/whats-new.json`, covering the user-visible work that landed on `pre-main` after the 1.71.0 release.

`v1.71.0` was published to users on 2026-07-15 (not a prerelease, marked Latest), so **its entry is left untouched** - these notes go in a new entry rather than amending shipped release notes.

## Highlights (7)

App/website ignore rules in Settings, custom OpenAI-compatible cloud providers from the provider picker, the provider connectivity test, the AI-composed daily summary, the live-hour strip back on the day timeline, plan-scoped worklog posting, and the roadmap's "suggest a feature" link.

## Fixes (7)

The capture memory leak, duplicate/missing worklogs when an hour was processed twice, Codex worklog drafting (and it now reports the real error), the stale retry download percentage, modal/task-title/Cleanup status issues, task-card summary clipping, and provider-picker readability.

## Deliberately excluded

Everything dev-only, since none of it is user-visible: the entire LLM Lab line, `dev-start` changes, the `MERIDIAN_BIN` override, and the CI build fixes.

## Conventions

- Every bullet rewritten in plain user terms, not lifted from commit messages (per the What's New task in CLAUDE.md).
- All copy uses plain hyphens - verified no em-dash, en-dash, or double-hyphen anywhere in the file.
- `cargo test -p meridian-tray whats_new` passes: `whats_new_json_parses` ok (4 passed, 0 failed).

## Before merging - two things to confirm

1. **Version.** I assumed `1.72.0` (a minor bump above the shipped `1.71.0`, since this batch contains `feat:` commits). Note that staging is currently cutting `1.71.0-staging.N`, which is *below* the already-published `1.71.0` - the staging channel appears to compute from `1.70.0`. If that versioning is off, the next prod release may not land as `1.72.0` and this `version` field should be adjusted to match.
2. **Date.** `2026-07-18` is the drafting date; set it to the actual release day when the release is cut.

### DIFF
--- a/tray/src-tauri/resources/whats-new.json
+++ b/tray/src-tauri/resources/whats-new.json
@@ -1,6 +1,28 @@
 {
   "releases": [
     {
+      "version": "1.72.0",
+      "date": "2026-07-18",
+      "highlights": [
+        "Choose apps and websites for Meridian to ignore, so nothing from them is ever captured. Set them up in Settings.",
+        "Bring your own cloud AI provider: add any OpenAI-compatible endpoint from the provider picker and Meridian will use it for worklogs and summaries.",
+        "Test your AI provider connection from Settings or during setup, so you can confirm it works before relying on it.",
+        "A new AI-composed daily summary that reads like a summary of your day, with the charts chosen to fit what actually happened.",
+        "The day timeline shows the current hour as a live strip again, so you can watch the hour fill in as you work.",
+        "Worklogs now follow the day's plan and post to every ticket you actually moved forward.",
+        "Added a \"suggest a feature\" link to the roadmap in What's New."
+      ],
+      "fixes": [
+        "Fixed a memory leak that made Meridian's memory use climb while screen capture was running.",
+        "Fixed duplicate and missing worklogs when the same hour was processed more than once.",
+        "Fixed worklog drafting failing on the Codex provider, and it now reports the real reason when it does fail.",
+        "Fixed a failed update showing a stale download percentage when you retried it.",
+        "Fixed modals scrolling instead of fitting, clipped task titles, and changing a task's status from Cleanup.",
+        "Fixed a task card's summary clipping mid-line instead of leaving room for the \"+N more\" label.",
+        "Made the AI provider picker easier to read."
+      ]
+    },
+    {
       "version": "1.71.0",
       "date": "2026-07-15",
       "highlights": [


adityaharishch:
## Review

**Overview:** Adds a hand-curated `1.72.0` entry to the front of `whats-new.json`, covering user-visible work landed on `pre-main` after the shipped `1.71.0`. Purely additive (22 lines), no code paths touched.

**Correctness / conventions**
- JSON structure matches the existing schema (`version`/`date`/`highlights`/`fixes`), and the new entry is correctly prepended (newest-first) without touching the `1.71.0` entry — matches the "don't amend shipped release notes" rule in `CLAUDE.md`.
- Copy is rewritten in plain user terms rather than lifted from commit messages, per the What's New task convention.
- Spot-checked all added strings for dashes — plain hyphens only, no em/en-dash. Compliant with the Hard Rules.
- Covered by the existing `whats_new_json_parses` test (per PR description), which is the right guard for this file.

**Issues / things to confirm before merge**
- The author flags two open questions themselves, both real: the `1.72.0` version number is an assumption (staging currently appears to be computing from `1.70.0`, which is a separate versioning discrepancy worth someone checking), and `date: 2026-07-18` is a drafting placeholder that needs to be set to the actual release day when cut. Neither blocks the PR, but don't merge-and-forget — a maintainer should confirm the version number is right before the next release actually goes out as something other than `1.72.0`.
- Nothing else stands out — low risk, content-only change.

**Verdict:** Looks good. Recommend confirming the version-number assumption with whoever owns the release cut, otherwise ready to merge.

Akarsh-Hegde:
## Review follow-up: version assumption resolved, no changes needed

Thanks for the review. Both flagged items were the two open questions I raised in the description; the version one is now **confirmed with evidence**, so no code changes were required.

### `1.72.0` is correct

I verified against the actual tag graph:

```
v1.71.0 reachable from pre-main? NO   (main's release tag was never merged back)
v1.71.0 reachable from main?     YES  (main sees it as latest)
latest tag reachable — pre-main: v1.71.0-staging.5
latest tag reachable — main:     v1.71.0
```

Production releases run on `main`, where `v1.71.0` is the latest reachable tag. The `pre-main → main` promotion set contains `feat:` commits, so semantic-release computes a minor bump: **1.72.0**. Since What's New is keyed to the shipped release users actually run, `1.72.0` is the right value. No change to the PR.

### The staging discrepancy is real, and this diagnoses it (separate issue)

The reason staging cuts `1.71.0-staging.N` — *below* the already-shipped `1.71.0` — is that **`v1.71.0` is not an ancestor of `pre-main`**. The release commit/tag produced on `main` is never merged back, so semantic-release on `pre-main` only sees the staging prereleases and keeps computing from `1.70.0`.

Two consequences worth tracking separately (neither blocks this PR):
- A staging build reports a version *lower* than the shipped production release, which is misleading for the staging updater channel.
- The What's New auto-open matches on app version, so a staging user on `1.71.0-staging.N` will never match a `1.72.0` entry.

The fix is to get `main`'s release commit/tag merged back into `pre-main` after each release so the staging channel computes from the true latest. Happy to open that as its own issue.

### Date

`2026-07-18` remains a drafting placeholder, as noted - it should be set to the actual release day when the release is cut. Flagging rather than guessing, since the cut date isn't known yet.

PR #465, opened 2026-07-18
feat(llm): pace metered endpoints and honour 429 reset headers

Akarsh-Hegde:
## Why

Rate limiting was purely **reactive**: a 429 parked the provider for a flat 30 minutes and the pipeline dumped to local MLX. That default is justified for CLI subscriptions (5-hour windows) but is 30x too long for a per-minute quota.

The first casualty was `probe_endpoint`, which fires up to 16 metered requests back-to-back (one per schema x rung). A free-tier key (15 RPM) 429s partway through, `effective_rung()` stays `None` for a partial probe, and the production gate then refuses the endpoint - so free-tier keys were effectively unusable end to end.

## What changed

**Proactive pacing** - new `src/llm/rate_limit.rs`:
- Even-spacing slot reservation (`60/rpm` apart), not a burst token bucket - banked tokens leaving at once is exactly what trips a small RPM.
- Custom HTTP endpoints only. The four CLI providers spend a flat-rate subscription and the local model spends the GPU, neither has a per-minute quota.
- Production calls pass a 30s max wait and proceed unpaced past it (no reservation is committed when refusing); the probe waits unbounded on purpose - a probe that takes a minute and completes beats one that returns fast and leaves the endpoint unmeasured.
- New `rpm` field on `CustomLlmProvider` (`#[serde(default)]`, `0` = unmetered = pacing off), surfaced in the add-provider form.

**Graceful 429s**:
- `LlmError::RateLimited` is now a struct variant carrying `retry_after: Option<Duration>`.
- `openai_compat` lifts headers before consuming the body and parses `Retry-After` (both delta-seconds and HTTP-date forms, past-clamped), then `x-ratelimit-reset-requests`, then `x-ratelimit-reset-tokens`. Duration suffixes `6m0s` / `1.5s` / `500ms` / `1h30m` are handled, with `ms` tested before `m`.
- Backoff priority in the resolver: header duration -> provider prose (`reset_time::parse_backoff`) -> flat per-transport default (60s for custom, 30min for CLI).
- `remove_custom_llm_provider` forgets the pacing reservation.

## Known limits (documented in-module)

No TPM pacing (needs a pre-call token estimate; TPM is handled reactively from `x-ratelimit-reset-tokens`), no cross-process coordination, no adaptive learning.

## Testing

`cargo fmt` + `cargo clippy -- -D warnings` clean; `cargo test` 668 passed / 0 failed. Tray `cargo check` clean, `npx tsc --noEmit` and `npm run build` clean.

Note: `cargo clippy --all-targets` reports two failures in `src/day_summary/validate.rs:318` and `src/llm/prompts.rs:428` - both pre-existing on `pre-main` and untouched by this branch.


adityaharishch:
## Review

**Overview:** Adds proactive rate-limit pacing for custom (OpenAI-compatible) LLM endpoints, plus richer reactive 429 handling. New `src/llm/rate_limit.rs` reserves evenly-spaced call slots per endpoint (`60/rpm` apart, not a burst bucket) so a metered endpoint's own probe burst (up to 16 back-to-back requests, one per schema × rung) doesn't 429 partway through and leave the endpoint permanently unmeasured/ungated. `LlmError::RateLimited` becomes a struct carrying an optional `retry_after: Duration` parsed from `Retry-After` / `x-ratelimit-reset-*` response headers, which now outranks the existing prose-parsing backoff.

**Correctness — traced the core logic, no bugs found**
- `rate_limit::acquire`'s slot math is correct: `slot = last_reserved.filter(future).unwrap_or(now)`, `wait = slot - now`, then commits `slot + interval`. This produces genuine even spacing (verified against the `consecutive_calls_are_spaced` test logic) rather than bursty release.
- The `max_wait` refusal path does **not** commit a reservation before returning `false` — correctly avoids delaying a later caller for a request that was never sent (confirmed via the `refusing_to_wait_consumes_no_slot` test).
- The lock is dropped before the `.await` sleep (mutex is not held across an await point) — avoids the deadlock/serialization hazard the module's own doc comment calls out.
- Header parsing priority (`Retry-After` → `x-ratelimit-reset-requests` → `x-ratelimit-reset-tokens`) and the `ms`-before-`m` ordering in `parse_duration_suffix` are both correct and specifically tested (the `500ms`-read-as-500-minutes failure mode the code guards against is a real footgun, good catch).
- `resp.headers().clone()` happens before `resp.text().await` in `openai_compat.rs` — correctly avoids consuming the response body before headers are read.
- The pacing key (`rate_limit::custom_key`) and the reactive backoff key (`resolver::backoff_key`) are derived from the same function, so an endpoint can't be paced under one identity and parked under another — good defensive design given both maps are separately keyed `BTreeMap`s.
- `RateLimited` breaks the retry loop (no second attempt) while `Failed` still gets one retry — unchanged from before, and correctly preserved through the enum refactor.
- The `rpm` field is set on the row **before** the initial probe (in `add_custom_llm_provider`), not after — correctly ensures the very first (and largest) burst is paced too, per the comment explaining why.

**Minor observations (non-blocking)**
- There's no UI path to change `rpm` on an already-added custom provider — it's settable only at creation time (confirmed: `CustomProviders.tsx` has no edit/update flow, only add/remove/probe). A user who under- or over-estimates their plan's RPM at setup time has to remove and re-add the endpoint to correct it. Might be worth a follow-up if this comes up in practice; not a blocker since `0` (unmetered) is a safe default and the reactive backoff still catches real 429s either way.
- No end-to-end test exercises the full three-way backoff priority chain (`header → provider prose → flat default`) together in `resolver.rs` — each piece is well unit-tested in isolation (`openai_compat` for header parsing, `rate_limit` for pacing), but the composition itself relies on manual verification via `cargo test` passing. Given how carefully the rest of this is documented and tested, this is a nice-to-have, not a real gap.
- The in-memory-only pacing state (no cross-process coordination, so two daemons sharing one key would jointly overshoot `rpm`) is explicitly called out as an accepted limitation matching the existing backoff map's behavior — reasonable to defer.

**Test coverage:** Strong. New unit tests cover the pacing primitive (zero-rpm no-op, first-call-immediate, spacing, independent keys, refusal-doesn't-consume, forget), and the header/duration parsing (delta-seconds, HTTP-date with past-date clamping, all four duration-suffix forms, priority between `Retry-After` and vendor headers). `cargo fmt`/`clippy -D warnings`/`cargo test` (668 passing) all reported clean per the PR description.

**Verdict:** Well-designed, thoroughly documented (the module-level rationale in `rate_limit.rs` is genuinely useful for future readers), and I couldn't find a correctness issue in the pacing or header-parsing logic. Good to merge; the missing "edit rpm on an existing provider" UI path is worth a follow-up ticket but shouldn't block this.

PR #468, opened 2026-07-18
Rust-only + DMG-only: remove Python/MLX and npm, rewrite install/dev tooling, candle 0.10.2

adityaharishch:
## Summary

Makes Meridian a **Rust-only, DMG-only** product: removes the entire Python/MLX stack and the redundant npm distribution channel, rewrites the install/dev tooling around it, and folds in the latest `pre-main`. No functional capability was lost — every working path was migrated to Rust before the corresponding Python/MLX code was deleted.

Net: **252 files, +3,198 / −23,035**.

### Python / MLX removal (the bulk)
- Deleted the entire `services/` tree (MLX server, agents, eval harness, runtime packaging) and the Python coding-agent indexer/summariser — all already ported to the Rust daemon.
- Removed the on-device **Local (MLX) LLM provider** and its setup-wizard step; generation now runs only through the user's chosen CLI provider (Claude/Codex/Cursor/Copilot) or a custom endpoint.
- The session distiller's embedder runs in-process via **candle** (`src/embedder/`, BAAI/bge-small-en-v1.5) — the only on-device model left.
- Purged dead MLX/Python references from the Rust tree, health checks, docs (CONTRIBUTING, RUNBOOK, TESTING, READMEs, CLAUDE.md), and release machinery.

### Distribution: DMG-only (npm removed)
- Removed the npm channel (`@meridiona/meridian`, `@semantic-release/npm`, `install-from-bundle.sh`, `package-release.sh`, npm publish/OIDC). The signed DMG self-installs its daemon + launchd on first launch — it never needed a second channel.
- Version-update check repointed from the npm registry to the **GitHub releases API**.
- Dropped the dead npm-bundle daemon binary path (`~/.meridian/app/bin/meridian`); the DMG stages to `~/.meridian/bin/meridian`. (Credential-migration from the old bundle `.env` is intentionally kept for upgraders.)

### Install / dev tooling, Rust-only
- `install.sh` / `install-dev.sh`: removed all MLX/Python/services steps; skill install via `meridian coding-agent-install-skill`; smoke via `meridian doctor`.
- `bootstrap.sh`: now downloads + installs the DMG (no brew/node/npm/Python).
- `meridian-cli.sh`: Rust-only; **`meridian dev` now launches the full dev environment** (cargo-watch daemon + `tauri dev` tray, = `dev-start.sh`), replacing the old launchd+`next dev` model. Removed the now-dead `dev ui`/`dev daemon` sub-targets.

### Dependencies
- **candle 0.8 → 0.10.2** (latest stable that builds on our pinned Rust 1.93.1; 0.11.0 needs nightly for `stdarch_neon_f16`). No embedder code changes required.

### Test hygiene
- Fixed a pre-existing parallelism race in `meridian-oauth` tests (env-var mutation) with a crate-level serialisation guard.

### Merged from pre-main
- Notifications-plugin CI clean fix, capture autorelease-pool drain around Vision OCR / AX walks, and What's New docs. `.releaserc` conflicts resolved to keep the npm-free release **and** adopt pre-main's notifications-plugin `cargo clean` step.

## Verification
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings` (+ tray): clean
- `cargo test --workspace`: all 23 suites green
- `ui` production build: clean static export; `bun test`: 227 pass
- Full pre-push hook (fmt + clippy + tests + UI build + UI tests + security audit) passed on push
- Manually verified on a live install: daemon boots, distiller runs in Rust (candle), worklog pipeline generates, `meridian dev` runs the full hot-reload environment

## Notes
- Targets `pre-main` per repo policy. Leaving the actual merge to a maintainer.
- macOS / Apple Silicon only (unchanged).


src/embedder/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+//! When the model is not yet provisioned (first-run weights still downloading) or fails
+//! to load, [`is_ready`] returns `false` and the distiller **skips** the vector stages,
+//! degrading to lexical dedup + longest-first selection. The hour pipeline never blocks
+//! on this — a missing embedder means a lower-reduction body, not an error.
+//!
+//! # Shape
+//! - [`is_ready`] — cheap file-presence check (drives the distiller's gate + UI readiness).
+//! - [`embed_batch`] — L2-normalized vectors; lazily loads the model on first use and
+//!   caches it, running compute on a blocking thread so it never stalls the async runtime.
+//! - [`ensure_weights`] — first-run download, driven by the setup wizard.
```
adityaharishch, line 21: **Bug: the embedder is never actually provisioned in production.**

This doc comment says `ensure_weights` is "driven by the setup wizard," but nothing calls it. I grepped the whole tree — `tray/src-tauri/src/commands/`, `ui/app/setup/`, and `src/main.rs` — and there is no Tauri command wrapping `embedder::ensure_weights()`, and no UI call site.

`is_ready()` (line 45) is a pure file-presence check against `~/.meridian/models/<repo>/`, and nothing ever populates that directory for a real user. So `is_ready()` returns `false` forever, and the distiller (`src/worklog_pipeline/distiller/mod.rs:330`) permanently takes the lexical-only degrade path — the new in-process candle embedder this PR introduces as "the only on-device model left" never actually runs for anyone.

The degrade is silent by design (never errors), so this won't surface as a crash or a failing test — it just quietly never does what the module doc claims. Worth wiring a real caller (setup wizard step, or an unconditional call on daemon/tray startup) before merge, or at least filing a fast follow-up.

src/worklog_pipeline/distiller/dedup.rs
```diff
@@ -0,0 +1,153 @@
...
+        let key: String = norm(&sp.line).chars().take(80).collect();
+        if seen.insert(key) {
+            lex.push(sp);
+        }
+    }
+    lex
+}
+
+/// Dot product of two L2-normalized vectors == cosine similarity.
+fn dot(a: &[f32], b: &[f32]) -> f32 {
```
adityaharishch: **Reuse: duplicated helper.** This `dot()` cosine-similarity helper is defined identically in `select.rs:20`. Both are private `fn dot(a: &[f32], b: &[f32]) -> f32` with the same body — worth hoisting to a single shared function in the `distiller` module (`mod.rs` or a small `math.rs`) so the two copies can't drift.

src/intelligence/mod.rs
```diff
@@ -35,49 +34,15 @@ pub async fn pm_tasks_present(pool: &SqlitePool) -> bool {
...
-pub async fn pipeline_ready(pool: &SqlitePool, cfg: &Config) -> bool {
-    mlx_ready(cfg).await && pm_tasks_present(pool).await
+/// The gate for the whole task-linking + worklog pipeline: a PM tracker that has
+/// synced tasks must be present before any worklog is drafted. Failing pauses the
+/// pipeline; it resumes automatically once a tracker syncs, because the driving
+/// loops re-check each cycle.
+///
+/// (This was once ANDed with a local-classifier readiness probe; with generation now
+/// running through the user's CLI provider there is no local server to gate on.)
+pub async fn pipeline_ready(pool: &SqlitePool, _cfg: &Config) -> bool {
```
adityaharishch: **Dead code left behind by the MLX removal.** `pipeline_ready` has zero callers anywhere in the tree now that the MLX-gate half was stripped out of it (verified via grep across `src/`, `tray/`, and `meridian-core/`). It still carries an unused `_cfg: &Config` param and a doc comment describing it as "the gate for the whole pipeline," which is no longer true for anything. Worth deleting outright, or rewiring it into whatever now serves that gating role.

meridian-core/src/util/date.rs
```diff
@@ -46,6 +46,37 @@ pub fn today_string() -> String {
     Local::now().format("%Y-%m-%d").to_string()
 }
 
+/// A UTC ISO datetime string → local `HH:MM`. Accepts a trailing `Z`, a `+00:00`
+/// offset, or a bare naive datetime (treated as UTC), with or without fractional
+/// seconds. Falls back to the raw `[11..16]` slice on any parse error.
+///
+/// A faithful port of the Python services' `time_utils.utc_to_local_hhmm`, used by
+/// the session distiller to stamp each on-screen span with its local clock time.
+pub fn utc_to_local_hhmm(utc_iso: &str) -> String {
```
adityaharishch, line 55: **Reuse: this duplicates two pre-existing implementations.** `utc_to_local_hhmm()` is a third, independently-written UTC-ISO → local `HH:MM` formatter. The same conversion already exists as `local_hhmm` in `src/worklog_pipeline/hour_input.rs:54` (used by the hour-report renderer) and `src/pm_worklog/status.rs:22`. All three have different fallback behavior on unparseable input, so they can silently drift apart over time. Worth consolidating on one, ideally this new `meridian-core` one since the other two now have a shared crate to pull from.

tray/src-tauri/src/commands/version.rs
```diff
@@ -3,39 +3,39 @@
...
 use serde::Serialize;
 use std::sync::Mutex;
 use std::time::Duration;
 use tracing::Instrument;
 
-/// npm package whose published `latest` dist-tag we compare against.
-const NPM_PKG_URL: &str = "https://registry.npmjs.org/@meridiona/meridian";
-/// Re-check the registry at most hourly.
+/// GitHub releases API — its `tag_name` is the latest published version.
+const GITHUB_LATEST_URL: &str = "https://api.github.com/repos/Meridiona/meridian/releases/latest";
```
adityaharishch, line 34: **Altitude: this is a second, independent update-check mechanism.** This hardcodes its own `api.github.com/repos/.../releases/latest` call, but the tray already has a GitHub-releases-based update mechanism via the Tauri auto-updater's `latest.json` manifest (`update.rs`, configured in `tauri.conf.json`). Now there are two separate "ask GitHub for the latest version" paths that would each need touching if the release process changes again. Worth checking whether this could read from the same `latest.json` the updater already fetches instead of hitting the GitHub API directly.

src/embedder/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+use once_cell::sync::Lazy;
+use std::sync::Mutex;
+
+use candle_bert::Embedder;
+
+/// The lazily-loaded model, shared across the process. `None` until first successful
+/// load; the `Mutex` also serializes the single device's forward passes (replacing the
+/// old cross-subsystem GPU gate). Never held across an `.await` — all compute runs inside
+/// [`embed_batch`]'s blocking task.
+static EMBEDDER: Lazy<Mutex<Option<Embedder>>> = Lazy::new(|| Mutex::new(None));
```
adityaharishch, line 41: **Plausible: unrecoverable mutex poisoning.** `EMBEDDER`'s `.lock()` call (line 67-69) returns an error and bails on `PoisonError` rather than recovering, so any panic during model load or a candle forward pass (e.g. a shape mismatch) poisons this `Mutex` permanently — every subsequent call degrades to lexical-only for the rest of the daemon's process lifetime, with no restart-free recovery. Worth the same `.unwrap_or_else(|e| e.into_inner())` pattern already used for the other global mutex in this codebase (`src/llm/rate_limit.rs`'s `NEXT_SLOT`). Currently unreachable in practice since nothing provisions the weights (see the comment on line 21), but worth fixing alongside that.

src/embedder/candle_bert.rs
```diff
@@ -0,0 +1,94 @@
...
+        );
+        Ok(Self {
+            model,
+            tokenizer,
+            device,
+        })
+    }
+
+    /// Embed one string → an L2-normalized vector. CLS-pooled (first token).
+    pub fn embed_one(&self, text: &str) -> Result<Vec<f32>> {
```
adityaharishch: **Efficiency: unbatched per-item forward pass.** `embed_one` runs a full BERT forward pass for a single text; its only caller (`embed_blocking` in `mod.rs:75`) loops it via `texts.iter().map(|t| embedder.embed_one(t)).collect()` — one forward pass per span instead of batching the hour's deduped spans into a single padded `[batch, seq]` tensor. This multiplies per-hour CPU cost by span count. Also currently unreachable in practice (see the line-21 comment on `mod.rs`), but worth batching once the provisioning gap is fixed, since this now runs in-process on every hourly tick instead of in a separate Python service.

src/embedder/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+pub async fn ensure_weights() -> Result<()> {
+    provision::ensure().await
+}
+
+/// L2-normalized sentence embeddings — one row vector per input string, so downstream
+/// cosine similarity is a plain dot product. Lazily loads (and caches) the model on first
+/// use; compute runs on a blocking thread. Errors when the model can't load; callers MUST
+/// gate on [`is_ready`] and take the lexical-degrade path on failure.
+pub async fn embed_batch(texts: &[String]) -> Result<Vec<Vec<f32>>> {
+    let owned: Vec<String> = texts.to_vec();
```
adityaharishch: **Efficiency: avoidable clone.** `embed_batch` clones its whole input via `texts.to_vec()`. Its only caller, `src/worklog_pipeline/distiller/mod.rs:332`, already owns a freshly-built `Vec<String>` moments earlier — changing this signature to take `texts: Vec<String>` (owned) instead of `&[String]` would drop the extra full copy of the hour's span text on every tick. Minor, but worth it given this now runs in-process rather than over HTTP to a separate service.

src/embedder/mod.rs
adityaharishch: Fixed in de0b5cc9: `main.rs` now spawns a fire-and-forget `embedder::ensure_weights()` call on daemon boot (idempotent, gated by `is_ready()`), so the embedder actually gets provisioned instead of never being called.

src/worklog_pipeline/distiller/dedup.rs
adityaharishch: Fixed in de0b5cc9/0008dabe: hoisted `dot()` into `distiller::mod` as `pub(crate) fn dot`, and both `dedup.rs` and `select.rs` now call `super::dot` instead of carrying their own copy.

src/intelligence/mod.rs
adityaharishch: Fixed in de0b5cc9: `pipeline_ready` deleted (confirmed zero callers before removal).

meridian-core/src/util/date.rs
adityaharishch: Fixed in 0008dabe: `src/pm_worklog/status.rs` now imports and uses `meridian_core::date::utc_to_local_hhmm` (aliased as `local_hhmm`) instead of carrying its own copy. Left `src/worklog_pipeline/hour_input.rs`'s `local_hhmm` as-is since it's a distinct, still-actively-used implementation with its own doc/callers — consolidating all three onto one shared fn would be a larger, riskier change than this PR's scope; happy to file a follow-up if that's wanted.

tray/src-tauri/src/commands/version.rs
adityaharishch: Left as-is: `check_status` (the DMG updater's `latest.json` path) reports `unsupported` for any non-packaged (dev/source) build, so a dashboard running in dev would lose its version-check banner entirely if this were merged into that path. The two checks serve genuinely different purposes (current-vs-latest display vs DMG auto-update eligibility) and the existing doc comments on both already explain the distinction. Not fixing this one to avoid a riskier behavioral change outside this PR's scope.

src/embedder/mod.rs
adityaharishch: Fixed in de0b5cc9: `embed_blocking` now uses `EMBEDDER.lock().unwrap_or_else(|e| e.into_inner())`, matching the same poison-recovery pattern already used by `src/llm/rate_limit.rs`'s `NEXT_SLOT`.

src/embedder/candle_bert.rs
adityaharishch: Fixed in d3b15e1a: `Embedder::embed_batch` now tokenizes the whole batch with BatchLongest padding and runs ONE forward pass over the padded batch with an explicit attention mask (so pad tokens don't pollute CLS pooling), instead of looping a per-item unbatched pass. Verified end-to-end against the real downloaded weights (`cargo test embedder::tests -- --ignored`): near-dup cosine 0.921 vs unrelated 0.500 — same result shape as before the change.

src/embedder/mod.rs
adityaharishch: Fixed in d3b15e1a: `embed_batch` now takes `Vec<String>` by value instead of `&[String]`, so the caller's already-owned `Vec` moves into the `spawn_blocking` closure directly instead of being cloned.

PR #476, opened 2026-07-18
fix(ui): give the uninstall wizard the same SF Pro voice as the dashboard

Akarsh-Hegde:
## What

The setup wizard moved off Instrument Serif onto the timeline's SF Pro display treatment in #468, but the **uninstall** wizard was left behind - its hero heading still rendered in Instrument Serif. The two wizards share the same card chrome and the same atoms (`uninstall/page.tsx` imports `Btn`/`Check`/`Kicker`/`Row`/`Spinner` from `../setup/atoms`), so they sat side by side in two different voices.

| File | Change |
|---|---|
| `ui/app/uninstall/page.tsx` | `SERIF` → `DISPLAY`; the `h1` is now `23px / 750 / -.03em` in `var(--font-sans)` |
| `ui/app/layout.tsx` | dropped the now-unused `Instrument_Serif` next/font loader and its `--font-instrument-serif` variable |
| `ui/app/globals.css` | removed the dead `--font-serif` custom property; comment rewritten to say why there deliberately isn't one |

The heading values match the setup card header **exactly** rather than being eyeballed - the header block is structurally identical (`Kicker` + `h1` + `p`, same padding, same atoms), so the same treatment is the correct one. This is not a bare `font-family` swap: at the old serif's 26px the sans would read under-weight and loose, so weight and tracking move with it.

With uninstall converted, nothing consumes the serif any more, hence the loader and variable removal. Side benefit: **the built static export no longer ships the Instrument Serif woff2 files at all** - one less Google webfont in the bundle.

## Note for the reviewer

Rebased onto the current `pre-main` tip to resolve a conflict with #470, which landed a `JetBrains_Mono` loader on the same lines of `layout.tsx` that this removes the serif loader from. Resolution keeps #470's font fully intact and removes only the serif - verified by build output below. The comment in `layout.tsx` was reworded so the "JetBrains Mono was retired" line no longer contradicts #470's deliberate scoped exception for the hour rail.

## Verification

- Repo-wide grep for `font-serif|instrument|SERIF|Georgia` across `ui/app ui/components ui/lib` returns only prose comments
- `npm run build` succeeds; static export contains JetBrains Mono (#470 intact) and **no** Instrument Serif
- Served `/uninstall` HTML carries `font-family:var(--font-sans);font-weight:750;letter-spacing:-.03em;font-size:23px`
- Full pre-push suite green: fmt, clippy, cargo test, UI build, UI tests, security audit

Rendered visually under `next dev` by the author. Note the wizard's `invoke` calls error there (no Tauri backend), which is expected and unrelated - typography renders regardless.

### DIFF
--- a/ui/STYLESHEET.md
+++ b/ui/STYLESHEET.md
@@ -14,8 +14,10 @@ monospace face for ticket keys/times/durations; numeric alignment comes from
 `font-variant-numeric: tabular-nums` on the same family (the `.font-mono` /
 `.mt-mono-sm` classes in `globals.css` and `tray/src/style.css`), matching how
 Apple's own type system handles numeric columns — no dedicated mono face there
-either. (`Instrument_Serif` remains, unrelated — it's a legacy face used only by
-the setup wizard, not part of this type scale.)
+either. There is no serif face at all: the setup and uninstall wizards used to
+set their hero headings in `Instrument_Serif`, but both now share the same SF Pro
+display treatment (`DISPLAY` in `app/setup/atoms.tsx`), and the `next/font`
+loader plus the `--font-serif` token were removed with it.
 
 | Role | Size | Weight | Line / Tracking |
 |---|---|---|---|
--- a/ui/app/globals.css
+++ b/ui/app/globals.css
@@ -25,11 +25,11 @@
      face; numeric alignment comes from font-variant-numeric: tabular-nums —
      see the .font-mono/.mt-mono-sm rules below). -apple-system/system-ui
      resolve to the real SF Pro Text/Display already installed on macOS, so no
-     font file is bundled. --font-serif is the setup wizard's Instrument Serif
-     hero font, with the Georgia stack as a load-failure fallback. */
+     font file is bundled. There is deliberately no --font-serif: the setup and
+     uninstall wizards used to render their hero headings in Instrument Serif,
+     and now share the timeline's SF Pro display treatment instead. */
   --font-sans:  -apple-system, BlinkMacSystemFont, 'SF Pro Text', system-ui, sans-serif;
   --font-mono:  var(--font-sans);
-  --font-serif: var(--font-instrument-serif), Georgia, 'Times New Roman', serif;
 
   /* Legacy palette → Tailwind utilities (kept for Sessions/Week/setup) */
   --color-paper:       var(--paper);
--- a/ui/app/layout.tsx
+++ b/ui/app/layout.tsx
@@ -1,19 +1,11 @@
 //ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
 
 import type { Metadata } from 'next'
-import { Instrument_Serif, JetBrains_Mono } from 'next/font/google'
+import { JetBrains_Mono } from 'next/font/google'
 import './globals.css'
 import { ThemeProvider } from '@/lib/theme-context'
 import ExternalLinks from '@/components/ExternalLinks'
 import LayoutBanners from '@/components/LayoutBanners'
-
-const instrumentSerif = Instrument_Serif({
-  weight: '400',
-  style: ['normal', 'italic'],
-  subsets: ['latin'],
-  variable: '--font-instrument-serif',
-  display: 'swap',
-})
 
 // One-off exception: --font-jetbrains-mono, used ONLY by the day-task
 // timeline's hour-rail labels (DayTaskColumn.tsx), to match the marketing
@@ -30,7 +22,10 @@ const jetbrainsMono = JetBrains_Mono({
 // Meridian Timeline design's single voice is SF Pro, via --font-sans in
 // globals.css's system-font stack (-apple-system/system-ui resolve to the real
 // SF Pro on macOS) — no next/font loader needed. Plus Jakarta Sans + JetBrains
-// Mono were retired in favor of one face for both UI text and numerics.
+// Mono were retired in favor of one face for both UI text and numerics (the
+// scoped hour-rail exception above aside), and the setup/uninstall wizards'
+// Instrument Serif hero font was retired alongside them — both wizards now use
+// the same SF Pro display treatment as the timeline.
 
 export const metadata: Metadata = {
   title: 'Meridian',
@@ -39,10 +34,7 @@ export const metadata: Metadata = {
 
 export default function RootLayout({ children }: { children: React.ReactNode }) {
   return (
-    <html
-      lang="en"
-      className={`${instrumentSerif.variable} ${jetbrainsMono.variable}`}
-    >
+    <html lang="en" className={jetbrainsMono.variable}>
       <body className="min-h-screen font-sans">
         <ThemeProvider>
           {/* ExternalLinks is an invisible interceptor — mounted for every
--- a/ui/app/setup/atoms.tsx
+++ b/ui/app/setup/atoms.tsx
@@ -8,6 +8,28 @@
 
 import type { CSSProperties, ReactNode } from 'react'
 
+/**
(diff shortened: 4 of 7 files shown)


Akarsh-Hegde:
## Overview

This PR ports the uninstall wizard's hero heading off Instrument Serif onto the same SF Pro "display" treatment the setup wizard adopted in #468, and cleans up the now-dead serif plumbing (`layout.tsx`'s font loader + `--font-instrument-serif` variable, `globals.css`'s `--font-serif` token). Small, well-scoped diff (13 additions / 19 deletions across 3 files).

## Strengths

- **Values verified to match, not eyeballed.** Diffing against `ui/app/setup/page.tsx`'s header block confirms the `h1` styling (`23px / 750 / 1.1 / -.03em`, `var(--font-sans)`) is byte-for-byte identical to the setup wizard's own header — not just "close enough." Since both wizards share the exact same `Kicker` + `h1` + `p` structure and atoms, this is the right way to guarantee visual parity.
- **Follow-through on the removal.** Rather than leaving `--font-instrument-serif`/`--font-serif` as dead code once nothing consumes them, the PR removes the `next/font` loader, the CSS variable, and rewrites the now-stale comment explaining *why* there's deliberately no serif token. This matches the project's no-drift-no-dead-code preference.
- **Rebase hygiene.** The PR description calls out a conflict with #470's `JetBrains_Mono` loader on the same lines and explains how the resolution keeps both changes intact — useful context for review, and confirmed correct: `jetbrainsMono` survives untouched in the diff.
- File header (`//ambient dev tool...`) is present and untouched on both modified `.tsx` files. No `any` types introduced. No em-dash/en-dash/double-hyphen introduced in any user-facing string added by this diff (I checked byte-for-byte for U+2013/U+2014 across the touched lines).

## Issues / Suggestions

1. **`DISPLAY` constant is now duplicated verbatim, not shared.** `ui/app/setup/page.tsx` already defines `const DISPLAY: CSSProperties = { fontFamily: 'var(--font-sans)', fontWeight: 700, letterSpacing: '-.02em' }`, and this PR adds an identical second copy in `ui/app/uninstall/page.tsx`. `uninstall/page.tsx` already imports `Btn`/`Check`/`Kicker`/`Row`/`Spinner` from `../setup/atoms` — exporting `DISPLAY` from `atoms.tsx` and importing it in both places would remove the duplication and prevent exactly the kind of drift this PR exists to fix (one wizard's header style diverging from the other's because it lives in two places). Low priority but fits the PR's own stated motivation unusually well.
2. **Dead weight in the `DISPLAY` constant itself (pre-existing, not introduced here, but now propagated).** Every consumer (`setup/page.tsx` line 244 and the new `uninstall/page.tsx` line) overrides both `fontWeight` (700 → 750) and `letterSpacing` (`-.02em` → `-.03em`) at the call site, so `DISPLAY` in practice only ever contributes `fontFamily`. Not a bug, but worth a comment or trimming next time this constant is touched — this PR is a good opportunity if extracting it to `atoms.tsx` per #1.
3. **`ui/STYLESHEET.md` is now stale.** Line 17 reads: `` (`Instrument_Serif` remains, unrelated — it's a legacy face used only by the setup wizard, not part of this type scale.) `` — this is the project's own style-sheet doc explicitly documenting the font this PR deletes as "remains" and "used only by the setup wizard." Since the PR's whole point is retiring that face, this note should be removed or updated in the same change to avoid leaving a contradictory doc behind (matches the "docs + cleanup in the same change" expectation for this kind of change).
4. **Pre-existing, out of scope but adjacent:** `ui/app/uninstall/page.tsx` still has two rendered (non-comment) em-dashes in user-facing `<p>` text — "these grants from System Settings — remove them yourself…" and "drag Meridian.app to the Trash — nothing will be left running." Neither line is touched by this diff, so it's not a regression from this PR, but since this PR is specifically about giving this file a consistent "voice," it'd be a natural companion fix (Hard Rule: plain hyphen only in user-facing text).

## Risks

- Low risk. Typography-only change, no logic/behavior touched, and the new font stack (`var(--font-sans)`) resolves to the same system font already used everywhere else in the app — no new font file is bundled (in fact one is *removed*, per the PR's stated bundle-size side benefit).
- `fontWeight: 750` on a non-variable system font family may render identically to 700 in some engines (weight steps get clamped to the nearest available static weight), but that's inherited unchanged from the already-shipped setup wizard treatment, so any risk here predates this PR.

## Test coverage

No JS/TS tests added, consistent with the project having no JS/TS test suite yet (per `CLAUDE.md`) and this being visual/typography polish. The PR body documents manual verification (grep for stray serif/instrument references, `npm run build` output checked for the absence of Instrument Serif and presence of JetBrains Mono, and the rendered HTML's computed style checked directly) — reasonable verification for this kind of change.

Akarsh-Hegde:
## Review addressed - all four items fixed in b6bb08be

Thanks - every point checked out on inspection, including one I'd missed the scale of: `DISPLAY` was duplicated **three** times, not two (`setup/page.tsx`, `setup/steps.tsx`, `uninstall/page.tsx`).

### 1. `DISPLAY` duplicated → extracted to `atoms.tsx`

Now exported from `app/setup/atoms.tsx` and imported by all three files. Both wizards already shared that module, and as the review notes, this duplication *was* the mechanism behind the bug this PR fixes - setup moved off the serif while uninstall's separate copy stayed behind.

### 2. Dead weight in the constant → folded in, rather than deferred

Verified the claim across every call site, and it's stronger than "worth a comment": **all four** override `700 → 750` and `-.02em → -.03em`, so the declared defaults described no heading the app has ever rendered.

So instead of trimming to `fontFamily` only, the shared constant now carries the values actually in use (`750 / -.03em`) and call sites set only what genuinely varies - `fontSize`, `lineHeight`, `color`:

```tsx
// atoms.tsx
export const DISPLAY: CSSProperties = {
  fontFamily: 'var(--font-sans)', fontWeight: 750, letterSpacing: '-.03em',
}
// call site
<h1 style={{ ...DISPLAY, fontSize: 23, lineHeight: 1.1, color: 'var(--t-title)' }}>
```

**Verified as a no-op visually**, which is the important part for a refactor like this - the served HTML is byte-identical before and after:

```
font-family:var(--font-sans);font-weight:750;letter-spacing:-.03em;font-size:23px;line-height:1.1;color:var(--t-title)
```

Same for the 33px and 31px heroes in `steps.tsx`. Structural change only.

This also left the `CSSProperties` type import unused in all three files, so those were dropped too.

### 3. Stale `ui/STYLESHEET.md` → updated

Line 17 documented `Instrument_Serif` as a face that "remains ... used only by the setup wizard" - exactly what this PR deletes. Rewritten to state there is no serif face, and to point at `DISPLAY` in `atoms.tsx` as where the wizard hero treatment now lives.

### 4. Em-dashes in user-facing copy → fixed

Both replaced with plain hyphens per the Hard Rule. Pre-existing rather than a regression, but taken as the companion fix suggested, since this file is the one the PR is giving a consistent voice.

### Verification

- `npm run build` passes; static export still ships JetBrains Mono (#470 intact) and no Instrument Serif
- `tsc --noEmit` clean outside the pre-existing `__tests__` `bun:test` typing errors
- Rendered `/setup` and `/uninstall` heading styles byte-identical to pre-refactor
- No `const DISPLAY` definitions remain outside `atoms.tsx`; no rendered em-dashes remain in `uninstall/page.tsx`
- Full pre-push suite green: fmt, clippy, cargo test, UI build, UI tests, security audit

### One note on scope

Items 1 and 2 pulled `setup/page.tsx` and `setup/steps.tsx` into the diff - they weren't touched before, but they held two of the three duplicate copies, so the extraction isn't possible without them. Diff is now 7 files (+47/-40). Flagging in case you'd rather keep this PR to the original 3 files and split the extraction out.

PR #477, opened 2026-07-18
fix(distiller): bound embedder memory/time + related fixes

adityaharishch:
## Summary

Five commits on this branch, ahead of `pre-main`:

### `7d2b051e` — bound the in-process embedding step with a timeout
Original fix for "nothing generates past a certain hour": a heavy hour's on-device embedder had no timeout, so it could block the sequential per-hour pipeline indefinitely. Added a 45s timeout (later raised, see below) with graceful degrade to lexical-only dedup.

### `c6d68ca2` — bound memory/time on the in-process embedder, no accuracy loss
Root-caused and fixed the underlying memory/performance issues discovered while investigating the above:
- **Chunked batching** (embed spans in fixed-size batches instead of one all-at-once forward pass) — tuned empirically against the heaviest real hour on record (856 spans): unbounded → 3.22GB/timed out; chunk=16 (Python's old value) + BLAS → 1.43GB/27s; chunk=8 → 918MB/22s; **chunk=4 (final) → 586MB/20s**.
- **mimalloc** as the global allocator (cross-platform: macOS/Windows/Linux) — returns freed pages to the OS instead of retaining them, which is what let RSS climb across batches even though candle already frees each batch's tensors.
- **Candle's `accelerate` feature** enabled for macOS builds (Intel + Apple Silicon both ship Accelerate.framework free) — candle's default CPU backend has zero BLAS acceleration; this alone cut the heaviest hour's embed time from "didn't finish in 45s" to ~20-29s.
- Timeout raised 45s → 210s, plus one retry-with-backoff on a real (non-timeout) error.
- **Verified no accuracy regression**: `embedder_cosine_sanity`'s near-dup/unrelated cosine scores (0.921/0.500) are identical before and after every change, including with Accelerate enabled.

### `fbc384e8` — don't drop carryover tasks past the suggestion cap
These two `plan_tasks` commits were made directly by the repo owner in this same working tree, not authored in this session — included here only because they landed on this branch; no context/review claim is being made on them from this PR description.

### `b015dde0` — fix(ui): don't name a not-yet-reached clock time in the hour-takeover card
The in-progress-hour card said the recap would be ready "around 10 PM" while still labeling the *current* hour (e.g. 9-10 PM, mid-flight at 9:51 PM) — accurate in meaning but easy to misread as claiming it's already 10 PM. Swapped to relative phrasing ("once the hour wraps up" / "at the top of the hour") that's correct regardless of the current minute.

## Test plan
- [x] `cargo build` / `cargo build --release`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --lib` (full suite green for everything touched by this session's commits)
- [x] `embedder_cosine_sanity` — confirmed identical similarity scores across every embedder config change (chunking, mimalloc, Accelerate)
- [x] Live-tested against the heaviest real hour on record (856 spans) at each tuning step, with real before/after RSS and timing numbers
- [x] Full pre-push hook suite (fmt, clippy, ui build/tests, security audit, cargo test) passed on both pushes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 4
 
 [[package]]
+name = "accelerate-src"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "415ed64958754dbe991900f3940677e6a7eefb4d7367afd70d642677b0c7d19d"
+
+[[package]]
 name = "accessibility"
 version = "0.3.0"
 source = "git+https://github.com/eiz/accessibility.git?rev=173e898d4a52ee0afe0224ecb458d324057856e7#173e898d4a52ee0afe0224ecb458d324057856e7"
@@ -999,10 +1005,12 @@ version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6bd9895436c1ba5dc1037a19935d084b838db066ff4e15ef7dded020b7c12a4a"
 dependencies = [
+ "accelerate-src",
  "byteorder",
  "float8",
  "gemm",
  "half",
+ "libc",
  "libm",
  "memmap2",
  "num-traits",
@@ -1023,6 +1031,7 @@ version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a9317a09d6530b758990ed7f625ac69ff43653bc9ee28b0464644ad1169ada87"
 dependencies = [
+ "accelerate-src",
  "candle-core",
  "half",
  "libc",
@@ -1039,6 +1048,7 @@ version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f59d08c89e9f4af9c464e2f3a8e16199e7cc601e6f34538c2cfbb42b623b1783"
 dependencies = [
+ "accelerate-src",
  "byteorder",
  "candle-core",
  "candle-nn",
@@ -4723,6 +4733,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.49"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a45a52f43e1c16f667ccfe4dd8c85b7f7c204fd5e3bf46c5b0db9a5c3c0b8e9"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5007,12 +5026,14 @@ dependencies = [
  "dotenvy",
  "flate2",
  "futures",
+ "gethostname",
  "hex",
  "http",
  "jsonschema",
  "libsqlite3-sys",
  "meridian-core",
  "meridian-oauth",
+ "mimalloc",
  "once_cell",
  "opentelemetry",
  "opentelemetry-appender-tracing",
@@ -5114,6 +5135,15 @@ name = "micromap"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2a86d3146ed3995b5913c414f6664344b9617457320782e64f0bb44afd49d74"
+
+[[package]]
+name = "mimalloc"
+version = "0.1.52"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d4139bb28d14ad1facf21d5eb8825051b326e172d216b39f6d31df53cc97862"
+dependencies = [
+ "libmimalloc-sys",
+]
 
 [[package]]
 name = "mime"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,13 @@ name = "cat_smoke"
 path = "src/bin/cat_smoke.rs"
 
 [dependencies]
+# Global allocator — returns freed pages to the OS proactively (unlike the platform
+# default allocators, which retain freed arenas for reuse). The embedder's chunked
+# batches free a full tensor set every ~16-32 spans; without this, RSS plateaus at
+# whatever the highest transient chunk cost was instead of settling back down.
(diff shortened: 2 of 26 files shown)


Akarsh-Hegde:
## Overview

This is a small, well-scoped fix: `distiller::embed_lex` wraps its single `embedder::embed_batch` call (the in-process BGE forward pass used for SemDeDup + facility-location) in `tokio::time::timeout` (default 45s, `DISTILLER_EMBED_TIMEOUT_SECS`-overridable). On timeout it degrades the hour to lexical-only — the same degrade path already taken when the embedder isn't ready or errors — instead of letting a CPU-bound forward pass on an outsized batch (large-hour case: 725KB / 62 sessions) block the strictly-sequential hour queue indefinitely.

The root-cause narrative in the PR body (live `sample` confirming the daemon was genuinely inside `candle_core::cpu_backend`, not deadlocked) is good diagnostic work and the fix is proportionate to the bug.

## Strengths

- Reuses the existing, already-proven degrade contract (empty vectors → lexical-only) rather than inventing a new code path — low blast radius.
- Correct understanding of `tokio::time::timeout` + `spawn_blocking` semantics: the doc comment on the timeout arm accurately describes that dropping the outer future does **not** cancel the blocking OS thread — it runs to completion and the mutex is released normally, so the next hour's call just contends on the lock rather than racing a half-cancelled computation. That's the right mental model and worth having in writing.
- Env-overridable via `DISTILLER_EMBED_TIMEOUT_SECS`, consistent with the other `Lazy`/`env_num` tunables in this module (`DISTILLER_SEM_DEDUP_THR`, `DISTILLER_DF_FRAC`, etc.).
- Structured logging on the timeout path (`n_spans`, `timeout_s`) follows the project's "no bare format strings for data values" convention, and the doc comment above `EMBED_TIMEOUT` explains the *why*, not just the *what*.
- Verified live against the actual stuck hour, not just unit-level — good validation given how hard this failure mode is to reproduce synthetically.

## Issues / Suggestions

1. **Is 45s actually calibrated correctly?** The PR's own evidence cuts against it: the *prior, non-degraded* hour (256KB/34 sessions — smaller than the stuck one) "had already taken ~13 min to process." If a meaningful fraction of that 13 minutes was the embed forward pass itself (plausible, since that's the CPU-bound stage that scales with batch size), a 45s bound will silently degrade a lot of normal-but-large hours to lexical-only, permanently (hours aren't reprocessed after the fact) — not just the pathological one this PR was written for. The `distil.embed` span already records `elapsed_ms` per hour, so it should be possible to check the actual elapsed-time distribution across recent hours before locking in 45s as the default. Worth a follow-up: pull a few days of `distil.embed` timings and confirm 45s sits above the normal-large tail, not just below the pathological case that triggered this bug.

2. **`embedder::unload()` is skipped on the timeout path.** In the non-timeout branch, `unload()` still runs right after `embed_batch` resolves, matching its documented contract ("the model is used once per hour... nothing gained by keeping ~130MB resident... until the next hour"). On a timeout, that call is never reached for the current hour, and the abandoned background task — once it eventually finishes — leaves the model warm in the static `EMBEDDER` slot with nothing to unload it until some *future* hour completes within budget. This is likely self-healing in the common case (the next few hours are probably normal-sized and will hit the `Ok` branch, reuse the already-warm model, and unload as usual), but it's a real, silent violation of the eager-unload invariant on any run of consecutive oversized hours. Two reasonable fixes, either is fine:
   - Detach a `tokio::spawn` that awaits the original `embed_batch` future to completion (independent of the timeout) and always calls `unload()` afterward, so the invariant holds unconditionally; or
   - If keeping the model warm after a timeout is actually a deliberate/acceptable trade (next hour skips the reload cost), say so in the comment — right now the code silently diverges from the invariant `unload()`'s own doc comment asserts.

3. **Span/log doesn't distinguish *why* a hour degraded.** `degraded = vecs.is_empty()` collapses "embedder not ready," "embed_batch errored," and "embed timed out" into the same boolean on the `distil.embed` span/log. Given timeout is a new, distinct failure mode with its own operational implications (worth watching for on a dashboard), consider recording a `degrade_reason` field (`"not_ready" | "error" | "timeout"`) rather than relying on grepping the accompanying warn-level message.

## Risks

- **Mutex-contention cascade, not full resolution of the stall:** if a batch is large enough to time out, the *next* hour's `embed_batch` call still has to wait on the same `EMBEDDER` mutex until the abandoned task finishes — it just now waits inside its own bounded 45s window instead of forever. In a run of several consecutive oversized hours, you could see a string of hours each individually timing out (still degrading each to lexical-only) until the backlog finally drains. This is a real improvement over the pre-fix unbounded stall (the async pipeline/ETL is never blocked), but it's worth knowing this is "bounded degradation," not "the batch finishes fast now."
- **Not the "pending for next cycle" pattern** — worth being precise here since the pipeline's usual contract (per CLAUDE.md) is that a failing/rate-limited *external* provider leaves work pending for a later retry. This fix does not do that: a timed-out hour is *not* reprocessed later, it completes immediately with a **permanently lexical-only** body. That's consistent with the existing "embedder not ready" degrade (also permanent, not retried), so it's not a new inconsistency — just worth flagging so nobody assumes a degraded hour gets a second, better-quality pass later.
- Low-probability tail case: if the embedder call could ever truly hang (not just be slow — e.g. a candle/rayon deadlock rather than a long-but-finite forward pass), every subsequent hour would spawn another blocking task that also contends on the stuck mutex and times out, accumulating blocked OS threads over the daemon's lifetime. Bounded by hourly cadence so not urgent, but worth a mental note if this class of bug recurs.

## Test coverage

CI (fmt/clippy/tests) is green and `cargo test --lib distiller` covers the existing 12 tests, but none of them exercise the new timeout branch itself — understandable, since `embedder::embed_batch` isn't behind a trait/seam that's easy to mock, so simulating "slow embed" realistically is hard without the actual candle weights. A lighter-weight option that would still buy confidence: extract the `tokio::time::timeout(...)` wrapping logic (or the degrade-selection logic) around a generic/injectable future and test it with `tokio::time::pause()` + a manually-controlled sleep future standing in for `embed_batch`, asserting the degrade path fires and the span records `degraded = true` when the inner future doesn't resolve in time. Not blocking, but would pin down the one piece of new behavior this PR actually adds.

---
Overall: sound, proportionate fix for a real sequential-stall bug, using an already-trusted degrade path. Main things I'd want resolved before merge are (1) confidence that 45s doesn't over-degrade normal large hours (check `distil.embed` elapsed_ms history) and (2) closing the `unload()` gap on the timeout path — everything else here is a minor/non-blocking suggestion.

PR #479, opened 2026-07-18
release: promote pre-main to main (1.71.0 → 1.72.0)

Akarsh-Hegde:
Promotes the accumulated `pre-main` work to production. **426 files changed, +72,968 / -24,105** across ~80 non-merge commits since `v1.71.0`.

---

## ⚠️ This release force-updates existing installs (intentional)

`tray/minimum-version` is armed at `1.71.0`. `package-updater.sh` bakes `Minimum-Version: 1.71.0` into the production `latest.json`, so:

| Installed version | Behaviour |
|---|---|
| `1.70.x` and older | **Force-installs 1.72.0 and relaunches, no consent prompt** |
| `1.71.0` (current release) | Normal consent banner - `1.71.0 < 1.71.0` is false |

This is a deliberate decision, not an oversight. #480 (which would have disarmed the floor) is closed.

### Operational consequences

- **`release.yml` runs on every push to `main`** - no marker, no environment approval gate. Merging publishes 1.72.0 immediately, and the `/releases/latest/download/latest.json` endpoint resolves to it at once.
- **Timing:** `enforce_minimum_version` checks 30 s after launch, then every 6 h. An already-running app can take up to 6 h to self-evict; only a relaunch makes it prompt.
- **Irreversible per machine.** There is no downgrade path. A machine that has force-installed is on 1.72.0 permanently. The fastest mitigation to stop *further* machines is stripping the `Minimum-Version:` line from the published `latest.json`; recovery for machines already updated is a RUNBOOK §A1/§A2 forward-revert.
- **The floor persists on every subsequent release** while the file has content - 1.72.1, 1.73.0 included. Empty `tray/minimum-version` once 1.72.0 has propagated (reopen #480).
- **npm-installed users are not covered.** The updater is DMG-only and #468 removes the npm distribution, so that cohort neither force-updates nor has a channel going forward.

### What the forced cohort receives

Everything below - most consequentially #468, which removes the Python/MLX runtime and rewrites the install shape, plus nine migrations applied on first launch. They cannot defer it.

### Architecture

- **#468 - Rust-only + DMG-only.** Removes the Python/MLX runtime and the npm distribution entirely; rewrites install and dev tooling; bumps candle to 0.10.2. `services/` and `npm/` are deleted from the tree. This retires the separately-versioned MLX runtime channel that previously let a stale runtime pair with a fresh app.
- **#469 - Universal binary.** Ships as a single Apple Silicon + Intel bundle.

### AI provider layer

- **#444 - Centralised AI-provider choice** in Settings, replacing per-feature provider config.
- **Custom OpenAI-compatible cloud endpoints** - identity, registry, and backend; schema-support probing gates a custom endpoint out of production paths if it can't do strict structured output.
- **#465 - Metered-endpoint pacing** that honours each provider's own 429 reset headers.
- **Connectivity test** wired into both Settings and the setup wizard.
- **On-device MLX fallback removed** - the tool now relies on the chosen provider, with a dead-letter guard rather than a silent local fallback.

### Worklog correctness

Several independent stuck-state fixes: hours stranded by a failure across local midnight (#471 recovers hours stuck in `generating` after a daemon restart), fold idempotency judged by segment starts rather than hours touched, matching scoped to the day's plan, posting to every ticket a session advanced, and the coding wait anchored to the hour's end rather than process start.

### Dashboard / UI

Day-task cards with a status picker and on-demand worklog generation; What's New changelog + roadmap modal; update surface moved to a sidebar card; AI-composed daily summary with model-chosen visualisations; setup wizard polish and the day-task timeline hour rail (#470).

### Capture

Ignore chosen apps and websites from Settings; autorelease pools drained around per-tick Vision OCR and AX walks (#472 also unloads the embedder after its one per-hour use).

### Dev-only

**LLM Lab** - an experiment harness with a ledger, `meridian llm-experiment` CLI, a `meridian-core` reader, tray commands, and a full-screen comparison surface rendering fold outcomes as real dashboard timelines. Gated to dev builds.

---

## Migrations

Nine new migrations, `058` → `066`:

| | |
|---|---|
| `058`-`060` | day tasks, segments, worklogs |
| `061`-`063`, `065` | worklog targets, manual target, post/create attempt tracking |
| `064` | `llm_experiments` ledger (LLM Lab) |
| `066` | day summaries |

All are additive. Because they land in one promotion, a production app updating from 1.71.0 applies all nine on first launch - worth watching the first-launch logs on a real 1.71.0 install.

---

## Risk notes

- **#468 is the largest single change here.** Removing the Python/MLX runtime alters the install shape for every existing user; this is the change most deserving of end-to-end staging validation before it reaches production.
- **`main` carries 2 commits not on `pre-main`** (`chore(release): 1.71.0 [skip ci]` and its merge) - the normal semantic-release version-bump commits. Worth confirming the merge doesn't revert the 1.71.0 version bump.

## Pre-merge checklist

- [ ] `tray/minimum-version` emptied on `pre-main`
- [ ] `1.72.0-staging.1` validated end-to-end on staging machines
- [ ] Forced-install test (#473) confirmed working on the staging channel
- [ ] First-launch migration run verified on an install upgrading from 1.71.0


Akarsh-Hegde:
**Blocker tracked:** #480 empties `tray/minimum-version`. Merge that into `pre-main` first - this PR tracks the branch tip, so it will pick the change up automatically and the armed file will drop out of this diff.

Do not merge this PR while `git show origin/pre-main:tray/minimum-version` returns `1.71.0`.

Akarsh-Hegde:
**Update:** the floor is intentional - see the revised description. #480 (disarm) is closed.

The remaining blocker is mechanical: this PR is `CONFLICTING` because `main`'s 1.71.0 release commit modified three files that #468 deleted. #481 resolves that against `pre-main` by taking the delete.

**Merge order:** #481 → `pre-main`, then this PR → `main`.

PR #480, opened 2026-07-18
chore(update): disarm the minimum-version floor after the staging test

Akarsh-Hegde:
Empties `tray/minimum-version`, reverting the floor armed by #473 for the forced-install test. This is the mandatory follow-up documented in RUNBOOK.md §A2.

## Why this must land

The floor ships with **every** release while the file has content, and `pre-main` promotes to `main`. Left armed, the next production release (#479) would bake `Minimum-Version: 1.71.0` into the production manifest and force-install on every user below `1.71.0`, with no consent prompt:

| Installed version | Behaviour if the floor reaches production |
|---|---|
| `1.71.0` (current release) | Normal consent banner |
| `1.70.x` and older | **Forced install + relaunch, no click** |

That cohort is precisely the population that hasn't yet taken #468 (the Rust-only rewrite removing the Python/MLX runtime), so they'd be migrated onto a new install shape with no ability to defer.

**#479 (`pre-main` → `main`) is blocked until this merges.**

## What the test established

The floor works end to end. `1.71.0-staging.6` published with `Minimum-Version: 1.71.0` in its notes; `package-updater.sh` read the file and passed its cores-only guard (floor `1.71.0` vs a `1.71.0-staging.6` release); `mirror-staging-release.sh` propagated the notes verbatim to the rolling channel; and a machine on `1.71.0-staging.4` parsed it and flipped its update card to **"Required update available"**, confirming `mandatory = true`.

### Finding worth recording

For an **already-running** app, a mandatory update is not prompt. `enforce_minimum_version` checks 30 s after launch and then every 6 h, and the frontend uses `mandatory` only for label copy (`UpdateCard.tsx:68`) - there is no auto-install path in the UI. So a machine that is already up when a floor publishes can wait up to 6 h before self-evicting, unless the user relaunches or clicks the card.

If a floor is ever used for its real purpose - evicting a build that must stop running now - that 6 h tail is the actual eviction time, not the 30 s figure the launch check suggests. Worth knowing before relying on it in an incident.

## Implementation note

Emptied rather than deleted, per RUNBOOK §A2's wording. `package-updater.sh` guards on `[[ -f ]]` and strips whitespace, so an empty file yields `MINIMUM=""` - identical in effect to absent, while keeping the mechanism discoverable.

Deliberately **not** carrying `[staging-release]`: this needs no build of its own. The staging channel keeps the floor until its next cut, which is harmless (it only ever acts when a newer build exists). Merge with the marker instead if you want it cleared from staging immediately.


Akarsh-Hegde:
Closing - **the floor is intentional now.** The decision is to ship `Minimum-Version: 1.71.0` to production so every install below `1.71.0` force-updates to 1.72.0 without a consent prompt.

`tray/minimum-version` therefore stays armed at `1.71.0` on `pre-main`, and #479 is no longer blocked on this.

Reopen this PR after 1.72.0 has propagated - the floor ships with **every** subsequent release while the file has content, so 1.72.1 / 1.73.0 would keep forcing the same cohort unless it is emptied.

PR #484, opened 2026-07-18
fix(update): check hourly for mandatory updates, and survive suspend

Akarsh-Hegde:
Makes the mandatory-update path actually prompt. Contained to `tray/src-tauri/src/update.rs`.

## The problem

`enforce_minimum_version` slept **6 h** between checks, so an already-running app could keep a retired version alive for most of a day after a floor was published.

Nothing else compensates: `UpdateCard` checks **once on mount** (`UpdateCard.tsx:29`), with no interval. Opening the dashboard was the only other way to notice a mandatory release. This loop is the entire recurring mechanism.

Observed live during the 1.72.0 rollout: a staging machine picked up the new manifest and correctly flipped its card to "Required update available", but did not self-install - it was mid-way through its 6 h cycle.

### 1. Cadence 6 h → 1 h

24 checks/day/install instead of 4, against a six-fold cut in worst-case eviction time. Each check is one small HTTPS GET to the release CDN.

30 s and 10 min were considered and rejected - 30 s is ~2,880 requests/day/install polling a file that changes at most weekly, which is real CDN egress across an install base and needless process wakeups on battery, for no human-perceptible gain over an hour.

### 2. Wall-clock deadline instead of one long sleep

**The more consequential fix.** tokio's timer runs on a monotonic clock that does **not** advance while macOS is asleep, so the old single 6 h sleep silently stretched by however long the lid was shut. A laptop closed overnight could go well past 6 h without a check - precisely the machine a floor is meant to evict.

The loop now wakes every `ENFORCE_TICK` (60 s) and compares a `SystemTime` deadline, so a suspended machine re-checks within a tick of waking. **No network happens on a tick that isn't due** - the extra wakeups are a timer comparison, nothing more.

### 3. A failed forced install retries fast

Previously a failure fell through to the next full cycle, so a transient network blip cost 6 h on a machine actively being evicted. It now retries from `ENFORCE_RETRY_BASE` (2 min), doubling via `next_retry` to the steady interval, so a persistently failing install settles into the normal rhythm instead of hammering.

### 4. Toast fires once per episode, not per attempt

At a 2 min retry cadence, notifying on every attempt would be a notification loop on a machine that is simply offline. The `notified` flag resets when a check comes back non-mandatory.

### 5. Softer forced-install toast

| | |
|---|---|
| Before | `This version is no longer supported - installing v1.72.0` |
| After | `Installing v1.72.0 - Meridian will restart in a moment.` |

"No longer supported" scolds the user for a version they did not choose to stay on, and it is the last thing they see before the app restarts under them. It was also the tonal outlier - the other update toasts read "Meridian is up to date" / "You're on the latest version."

The restart mention is deliberately kept: the install is not consented to, so this toast is the only warning before the app quits and reopens, possibly mid-task. Only the blame was dropped, not the expectation-setting.

Title (`Updating Meridian`) is unchanged. Plain spaced hyphen per the user-facing-text hard rule.

Not changed: `UpdateCard.tsx:69` still renders "Required update available" on the sidebar card. Accurate rather than scolding, but the same stern register - flag it if it should soften too.

## Ripple

- **Nothing else calls these constants** - they are private to the module, and `enforce_minimum_version`'s signature and its single call site (`lib.rs:502`) are unchanged.
- **Rebased onto #482.** `pre-main` moved mid-change: #482 rewrote these toasts to route through the notification outbox (`notify_update`, gated by `system.update`) instead of the `sys::notify` bypass. This sits on top of that - the forced-install toast still goes through the outbox.
- **`download_and_apply` is already single-flight** (`INSTALLING`), so the faster cadence cannot produce overlapping downloads.

## Tests

Two new in-module tests covering the backoff (`cargo test -p meridian-tray`, 9 pass in `update::`):

- `retry_doubles_then_caps_at_the_steady_interval`
- `retry_backoff_reaches_the_cap_from_the_base` - guards against a base that divides the interval unevenly and sits just under the cap forever

The loop itself is not unit-tested - it needs an `AppHandle` and a live updater endpoint. The extracted `next_retry` is the part with logic worth pinning.

## Scope note

This ships in the binary users **receive**, so it governs the **next** forced update, not the one in flight. Every install below `1.71.0` today is still running the 6 h loop baked into its current build; nothing here accelerates the 1.72.0 rollout.

### DIFF
--- a/tray/src-tauri/src/update.rs
+++ b/tray/src-tauri/src/update.rs
@@ -42,6 +42,7 @@ use meridian_core::SqlitePool;
 use semver::Version;
 use serde::Serialize;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::{Duration, SystemTime};
 use tauri::{AppHandle, Emitter, Manager};
 use tauri_plugin_updater::UpdaterExt;
 
@@ -277,25 +278,75 @@ pub async fn download_and_apply(app: &AppHandle) -> Result<(), String> {
     Ok(())
 }
 
+/// Delay before the first mandatory-update check after launch.
+const ENFORCE_FIRST_DELAY: Duration = Duration::from_secs(30);
+/// Steady-state cadence for the mandatory-update check. One network round trip
+/// per hour per install — cheap enough to be unnoticeable, and it bounds how
+/// long a machine can keep running a version the manifest has retired.
+const ENFORCE_INTERVAL: Duration = Duration::from_secs(60 * 60);
+/// First retry delay after a *failed* forced install, doubling up to
+/// [`ENFORCE_INTERVAL`]. A transient network blip must not cost a full cycle on
+/// a machine we are actively trying to evict.
+const ENFORCE_RETRY_BASE: Duration = Duration::from_secs(2 * 60);
+/// Timer granularity. The loop wakes this often and compares a **wall-clock**
+/// deadline rather than sleeping one long stretch, so a machine that suspends
+/// mid-interval re-checks promptly on wake instead of resuming a timer that
+/// stood still. No network happens on a tick that isn't due.
+const ENFORCE_TICK: Duration = Duration::from_secs(60);
+
+/// The next retry delay after a failed forced install: double, capped at the
+/// steady-state interval so a persistently failing install settles into the
+/// normal cadence instead of retrying forever at speed.
+fn next_retry(current: Duration) -> Duration {
+    (current * 2).min(ENFORCE_INTERVAL)
+}
+
 /// Mandatory-update enforcement — the one path that installs without consent.
-/// Spawned once at launch; checks shortly after startup and then every 6 h
-/// (the tray runs for weeks between relaunches, so a launch-only check would
-/// never catch a mandatory release published mid-run). When the manifest marks
-/// the running version as below its minimum, it notifies and installs; on
-/// failure it just waits for the next cycle (the consent banner stays available
-/// as the fallback), and on success [`download_and_apply`] never returns.
+/// Spawned once at launch; checks [`ENFORCE_FIRST_DELAY`] after startup and
+/// then every [`ENFORCE_INTERVAL`] (the tray runs for weeks between relaunches,
+/// so a launch-only check would never catch a mandatory release published
+/// mid-run). When the manifest marks the running version as below its minimum,
+/// it notifies and installs; on success [`download_and_apply`] never returns.
+///
+/// Two properties the naive "sleep the whole interval" loop did not have:
+///
+/// - **Survives suspend.** The deadline is wall-clock ([`SystemTime`]), and the
+///   loop wakes every [`ENFORCE_TICK`] to compare against it. tokio's timer runs
+///   on a monotonic clock that does not advance while macOS is asleep, so a
+///   single long sleep silently stretched by however long the lid was shut — the
+///   common case for a laptop we are trying to evict.
+/// - **Retries a failed install quickly.** A failure previously waited the full
+///   cycle; it now retries from [`ENFORCE_RETRY_BASE`], backing off via
+///   [`next_retry`] to the steady interval. The consent banner remains the
+///   fallback throughout.
+///
+/// The "no longer supported" toast fires once per mandatory episode, not once
+/// per retry — notifying on every attempt would be a notification loop on a
+/// machine that is simply offline.
 pub fn enforce_minimum_version(app: &AppHandle) {
     let app = app.clone();
     tauri::async_runtime::spawn(async move {
-        let mut delay = std::time::Duration::from_secs(30);
+        let mut due = SystemTime::now() + ENFORCE_FIRST_DELAY;
+        let mut retry = ENFORCE_RETRY_BASE;
+        let mut notified = false;
         loop {
-            tokio::time::sleep(delay).await;
-            delay = std::time::Duration::from_secs(6 * 60 * 60);
+            // Wake at most a tick at a time; `due` is the real gate below.
+            let wait = due
+                .duration_since(SystemTime::now())
+                .unwrap_or(Duration::ZERO);
+            tokio::time::sleep(wait.min(ENFORCE_TICK)).await;
+            if SystemTime::now() < due {
+                continue;
+            }
+
             let status = check_status(&app).await;
             // `unsupported` (dev run) never flips to packaged mid-run, so the
             // loop is pure idle there; keep it anyway — it costs one no-op
             // check per cycle and avoids a second is_packaged() code path.
             if status.state != "available" || !status.mandatory {
+                due = SystemTime::now() + ENFORCE_INTERVAL;
+                retry = ENFORCE_RETRY_BASE;
+                notified = false;
                 continue;
             }
             let v = status.version.clone().unwrap_or_default();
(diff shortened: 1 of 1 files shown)


tray/src-tauri/src/update.rs
```diff
@@ -277,25 +278,75 @@ pub async fn download_and_apply(app: &AppHandle) -> Result<(), String> {
...
+            }
+
             let status = check_status(&app).await;
             // `unsupported` (dev run) never flips to packaged mid-run, so the
             // loop is pure idle there; keep it anyway — it costs one no-op
             // check per cycle and avoids a second is_packaged() code path.
             if status.state != "available" || !status.mandatory {
+                due = SystemTime::now() + ENFORCE_INTERVAL;
+                retry = ENFORCE_RETRY_BASE;
+                notified = false;
```
Akarsh-Hegde, line 417: Was `state == "error"` meant to wait the full interval here? It reads as though this arm means "nothing to do", but `check_status` folds three different situations into it: up to date, available-but-not-mandatory, and **the check itself failed** (`UpdateStatus::errored`, hit on any network error from `updater.check()`).

That last one is the case this PR is otherwise built around. Two consequences:

1. The 2 min fast retry only covers a failed *download*. A machine that is offline fails at the *check*, one step earlier, and waits the full hour - so a transient blip still costs a full cycle on a machine being actively evicted, which is what change #3 set out to stop.
2. A check error *between* download retries resets both `retry` and `notified`. So the backoff cadence is discarded and the toast re-arms, on precisely the flaky-network machine where the once-per-episode guard was supposed to hold.

A three-way split would restore the intent:

- `available` + `mandatory` - install (unchanged)
- `uptodate` / `unsupported` / available-not-mandatory - `due = now + ENFORCE_INTERVAL`, reset `retry` and `notified` (a genuine "nothing to do")
- `error` - `due = now + retry; retry = next_retry(retry)`, leaving `notified` alone

That also gives the existing `next_retry` backoff a second caller, which is a nice signal that the abstraction is the right one.

tray/src-tauri/src/update.rs
```diff
@@ -277,25 +278,75 @@ pub async fn download_and_apply(app: &AppHandle) -> Result<(), String> {
...
+        let mut notified = false;
         loop {
-            tokio::time::sleep(delay).await;
-            delay = std::time::Duration::from_secs(6 * 60 * 60);
+            // Wake at most a tick at a time; `due` is the real gate below.
+            let wait = due
+                .duration_since(SystemTime::now())
+                .unwrap_or(Duration::ZERO);
+            tokio::time::sleep(wait.min(ENFORCE_TICK)).await;
+            if SystemTime::now() < due {
```
Akarsh-Hegde: Optional hardening. `SystemTime` buys suspend-resilience but opens the mirror hole: it is settable, so a backward jump (manual clock change, dead RTC on wake, VM snapshot restore, a large NTP correction) leaves `due` sitting in the future by the size of the jump, and the loop stalls for exactly that long. Same "the timer stood still" failure class this PR exists to close, reached from the other direction.

Cheap fix is to gate on either clock, whichever fires first:

```rust
let mut due = SystemTime::now() + ENFORCE_FIRST_DELAY;
let mut due_mono = Instant::now() + ENFORCE_FIRST_DELAY;
// ...
if SystemTime::now() < due && Instant::now() < due_mono {
    continue;
}
```

The `SystemTime` arm still catches suspend (wall clock advanced, monotonic did not); the `Instant` arm catches the clock jump (monotonic advanced regardless). Firing early is harmless here - it costs one extra HTTPS GET.

Lower probability than a closed laptop lid, so entirely reasonable to note it and move on. Flagging it so it is a decision rather than an accident.

tray/src-tauri/src/update.rs
```diff
@@ -277,25 +278,75 @@ pub async fn download_and_apply(app: &AppHandle) -> Result<(), String> {
...
+///   loop wakes every [`ENFORCE_TICK`] to compare against it. tokio's timer runs
+///   on a monotonic clock that does not advance while macOS is asleep, so a
+///   single long sleep silently stretched by however long the lid was shut — the
+///   common case for a laptop we are trying to evict.
+/// - **Retries a failed install quickly.** A failure previously waited the full
+///   cycle; it now retries from [`ENFORCE_RETRY_BASE`], backing off via
+///   [`next_retry`] to the steady interval. The consent banner remains the
+///   fallback throughout.
+///
+/// The "no longer supported" toast fires once per mandatory episode, not once
```
Akarsh-Hegde: Stale after change #5 - this names the copy the PR removes. The toast now reads "Installing v{v} - Meridian will restart in a moment."

Suggest just `The forced-install toast fires once per mandatory episode, not once per retry`, which stays true through the next copy edit.

Akarsh-Hegde (commented):
Reviewed the whole change. The core premise checks out: on macOS `std::time::Instant` (and therefore tokio's timer) is backed by a clock that does not advance across sleep, so the old single 6 h sleep really did stretch by however long the lid was shut. Swapping to a wall-clock deadline with a short tick is the right shape, and the single-flight `INSTALLING` guard does cover the faster cadence as the description claims.

One thing I think is worth fixing before merge, one tradeoff worth a decision, one stale comment, and the question you asked at the end.

**Should fix - the fast retry never engages for the failure that actually happens.** `check_status` returns `state = "error"` on a network failure, which lands in the `state != "available"` arm alongside `uptodate`. So a failed *check* waits the full hour, while only a failed *download* gets the 2 min path. On an offline machine - the exact machine a floor exists to evict - the check is the first network op and the one that fails, so change #3 in your description mostly doesn't fire in the case it was written for. Details inline.

**Worth deciding - the wall clock trades one stall for another.** `SystemTime` fixes suspend but is jumpable; a backward jump parks `due` in the future for the size of the jump. Lower probability than a closed lid, so this is a hardening call rather than a blocker. Inline.

**On `UpdateCard.tsx:69`** - I'd leave "Required update available" as is. The card is a passive surface the user chose to look at, so the direct register reads as informative there. The toast was different: it arrives uninvited, immediately before the app restarts under them, which is what made "no longer supported" land as blame. Softening the toast alone is the right call.

The cadence tick I considered and have no objection to - 60 s timer wakeups with macOS coalescing are genuinely negligible, and your 30 s-vs-1 h reasoning covers the network side.

tray/src-tauri/src/update.rs
Akarsh-Hegde: Fixed - you're right, and this was the substantive one. The fast retry was covering the *second* network op while the first one silently ate the full hour, so change #3 mostly didn't fire in the case it was written for.

Split three ways as suggested:

- `error` → `due = Deadline::after(retry); retry = next_retry(retry)`, leaving `notified` untouched
- `uptodate` / `unsupported` / available-not-mandatory → full interval, reset `retry` and `notified`
- `available` + `mandatory` → install (unchanged)

The second consequence you spotted was the sharper one - a check error between download retries discarded the backoff *and* re-armed the toast, so a flapping machine would have re-toasted on every recovery. Leaving `notified` alone in the error arm closes that.

And agreed on the signal: `next_retry` having two callers reads much better than one.

tray/src-tauri/src/update.rs
Akarsh-Hegde: Taken - implemented rather than noted. Once it was written down it was hard to justify leaving a known stall in a loop whose entire purpose is not stalling, however unlikely the trigger.

Wrapped both clocks in a small `Deadline` type instead of two loose variables, so the pairing can't drift as the loop changes:

```rust
struct Deadline { wall: SystemTime, mono: Instant }
fn remaining(&self) -> Duration   // min of both; saturates to zero
fn reached(&self) -> bool         // remaining().is_zero()
```

`remaining()` also feeds the sleep, so the tick can't overshoot either clock. Three tests pin the two directions independently - wall-reached-with-monotonic-parked (suspend) and monotonic-reached-with-wall-parked (backward jump) - since neither is reachable by waiting in a test.

tray/src-tauri/src/update.rs
Akarsh-Hegde: Fixed. Rewrote the block rather than just dropping the quoted string - it also claimed "two properties" and now covers three, since the check-error retry is a distinct behaviour worth stating.

Took your phrasing for the toast line so it survives the next copy edit.

PR #498, opened 2026-07-20
feat: redesigned AI provider picker + hardened Cursor integration

adityaharishch:
## What

Three related pieces of work on the AI-provider layer: a redesigned picker shared by setup and Settings, a substantial hardening of how Meridian invokes `cursor-agent`, and a last-resort fallback so a permanently-broken agent CLI no longer costs us the summary.

## The picker

Setup and Settings now render the same component, so they cannot drift. Three recommended providers (Claude Code, Codex, Cursor) plus bring-your-own-API-key; each opens a detail view that installs the CLI, tests the connection, and sets it as default. Copilot is dropped from the chooser but stays a valid stored value, so legacy users get a switch banner rather than being stranded.

Two behavioural fixes worth calling out:

- The hook no longer bulk-tests every installed CLI on mount. That was what produced a spurious "CONNECTION FAILED" on providers the user had never configured.
- Selecting a provider **commits immediately** instead of staging a Save. This reverses an earlier staged-Save fix, so the anti-bug invariant is preserved structurally instead: `value` reads from `settings.llm_provider` on disk, and the commit only mutates state on success. The two regression tests that pinned the old shape are rewritten against the new invariants.

In-app install runs the vendor installer through the user's **login shell** (the tray is a Finder-launched `.app` with the stripped launchd PATH, which has no npm/node). Cursor sign-in uses `cursor-agent login` with the browser enabled, so the user's **subscription** is used - no API key, nothing metered.

## The Cursor hardening

`cursor-agent` is a coding agent, but Meridian only ever asks it for inference over **untrusted** text (coding-agent transcripts, screen OCR). Its defaults are wrong in both directions: too much capability, and a large irrelevant context. `src/llm/cursor_cli.rs` is now the single source of truth for invoking it.

This also closes a **real gap**: the summariser built its own argv, so it summarised untrusted transcripts with cursor-agent's default full **write + shell** tool access. It now inherits the same hardening as the provider backend.

| Lever | Effect |
|---|---|
| `--allowed-tools ""` | no tools at all; 21,040 -> 9,970 tokens |
| sandbox `HOME` | `<agent_skills>` 190 entries -> 0 |
| `--workspace <empty>` | stops Cursor injecting the user's `~/CLAUDE.md` |
| `--mode ask` | read-only, server-enforced |
| `--output-format json` | real error / usage-limit detection |
| `CURSOR_API_KEY` stripped | subscription guaranteed, never metered |

**Net: ~21,000 -> ~3,400 input tokens per call, summary quality unchanged.**

The non-obvious part of the sandbox: the four `.cursor` skill entries must be recreated **empty and read-only**, because cursor-agent self-provisions its built-in skills and plugin cache on startup and would otherwise just repopulate them. The sandbox **never touches the user's files** - it only creates entries under its own directory, and the real skill directories are simply not linked in.

Every degradation is driven by a **detected cause**, never a blind retry: unknown flag -> unhardened call, auth failure -> real HOME, unavailable model -> `auto`. A usage limit deliberately does **not** retry, because Cursor limits are account-level and a second call would burn quota to hit the same wall.

## Self-ingest fix

cursor-agent has no `--no-session-persistence`, so every call writes a chat into the store the cursor-cli source ingests from. Only the summariser was fingerprinted, so every *other* AI process routed through Cursor was ingested as developer activity and re-summarised next tick. All Meridian-issued prompts now carry a marker the ingest guard drops.

## Summariser fallback

A coding-agent transcript is still summarised by its **own** CLI - that agent holds the context and its subscription is already paid for. What changes is the outcome when that CLI is *permanently* unable to answer. Previously the row was simply left pending and dead-lettered after three drains; now it escalates once to the provider the user chose in Settings.

```
1. the session's own CLI          x2 attempts      (codex -> codex, cursor -> cursor-agent, ...)
2. rate limit?      -> STOP. back the source off, retry on a later tick
3. otherwise failed -> the global provider, ONCE
4. still nothing    -> row stays pending; dead-lettered after 3 drains
```

**Step 2 is the load-bearing one.** A `RateLimited` breaks out of the loop *before* the fallback code is reachable, so there is no path where a quota triggers a substitute - a quota refills, and is waited out. Only a `Failed` (crashed, not installed, signed out, subscription ended, unusable output) escalates, and only after every retry is spent.

Two decisions worth a reviewer's eye:

- When the global provider **is** the CLI that just failed (the normal case), `fallback::try_summarise` returns `Ok(None)` and spawns nothing. A third run of the same binary would fail identically.
- A fallback provider hitting **its own** rate limit is reported as `Failed`, not `RateLimited`. That flag drives the per-*agent-source* backoff, and the fallback is a different account - parking Codex's queue because Claude ran out of quota would punish the wrong subscription. The resolver's provider-level backoff already prevents redialling.

Fallback summaries persist as `summary_source = "fallback:<provider>"`, so a substitute's work is never mistaken for the agent's own. Nothing downstream reads that column today, so the vocabulary change is additive.

## Version pin

`cursor.com/install` is a rolling script with no version flag. Pinned to the build this was verified against, by rewriting version strings **by pattern** - a literal swap would silently no-op the day Cursor publishes a new build and hand the user "latest" again.

## Verification

- fmt, clippy `-D warnings`, **707 Rust tests**, tray check, UI build, **252 UI tests** - all green
- Live smoke tests against a real Cursor subscription throughout; token counts and the empty `<agent_skills>` block verified by decoding the actual persisted request

### DIFF
--- a/meridian-core/src/lib.rs
+++ b/meridian-core/src/lib.rs
@@ -68,7 +68,7 @@ pub use readers::{
 
 pub use canonical_task::{CanonicalTask, PersonRef, Priority, Provider, StatusCategory, TaskKind};
 
-pub use llm_provider::LlmProvider;
+pub use llm_provider::{LlmProvider, CURSOR_CLI_VERSION, CURSOR_INSTALL_CMD};
 /// The custom-endpoint registry types. `CustomLlmProvider` carries the API key and is the
 /// STORAGE form — see its docs before serialising one anywhere.
 pub use settings::{CustomLlmProvider, SchemaRung};
--- a/meridian-core/src/llm_provider.rs
+++ b/meridian-core/src/llm_provider.rs
@@ -140,6 +140,82 @@ impl LlmProvider {
             LlmProvider::Custom => None,
         }
     }
+
+    /// The shell command that installs this provider's CLI, or `None` when there is nothing
+    /// to install ([`LlmProvider::Custom`] is a cloud endpoint, not a binary).
+    ///
+    /// Cursor's is PINNED - see [`CURSOR_CLI_VERSION`].
+    ///
+    /// This is the SAME command the UI shows as its `installHint` (`ui/lib/llm-providers.ts`)
+    /// — kept here as the authoritative copy because the tray now runs it on the user's behalf
+    /// (`crate::llm::detect::install_provider` in the daemon crate). The strings are fixed
+    /// literals, never interpolated with user input, so running them through a login shell is
+    /// safe. Each is the vendor's own official, non-interactive installer.
+    pub fn install_command(self) -> Option<&'static str> {
+        match self {
+            LlmProvider::Claude => Some("npm i -g @anthropic-ai/claude-code"),
+            LlmProvider::Codex => Some("npm i -g @openai/codex"),
+            LlmProvider::Cursor => Some(CURSOR_INSTALL_CMD),
+            LlmProvider::Copilot => Some("npm i -g @github/copilot"),
+            LlmProvider::Custom => None,
+        }
+    }
+}
+
+/// The cursor-agent build Meridian installs and supports.
+///
+/// PINNED DELIBERATELY. `https://cursor.com/install` is a ROLLING script - it hardcodes
+/// whatever build is current when you fetch it, with no version flag or env var. Two
+/// reasons we don't want that:
+///
+/// 1. `crate::llm::cursor` depends on `--allowed-tools`, which is **undocumented**
+///    ("internal only", hidden from `--help`). An unpinned upgrade could drop it
+///    silently. (The backend degrades gracefully if that happens, but a pin means it
+///    doesn't happen unattended in the first place.)
+/// 2. Reproducibility: every user gets the build this was tested against, not whatever
+///    shipped that morning.
+///
+/// This value is the build all of the flag behaviour was verified on. To move it: bump
+/// this constant, re-run the cursor smoke tests, and confirm `--allowed-tools`,
+/// `--workspace` and `--mode ask` still behave.
+pub const CURSOR_CLI_VERSION: &str = "2026.07.16-899851b";
+
+/// Cursor's official installer, rewritten to install [`CURSOR_CLI_VERSION`].
+///
+/// The `sed` rewrites every version string in their script (all 5 occurrences: the temp
+/// dir, the `downloads.cursor.com/lab/<v>/<os>/<arch>/agent-cli-package.tar.gz` URL, the
+/// final dir, and both symlinks) to the pinned build. Matching the version by PATTERN
+/// rather than by literal is what makes this survive Cursor bumping their script - a
+/// literal `s/<old>/<new>/` would silently become a no-op the day they publish a new
+/// build, handing the user "latest" again, which is the exact failure being avoided.
+///
+/// Reusing their script (rather than hand-rolling a download) keeps their OS/arch
+/// detection, symlink layout and PATH guidance. Fixed literal, no user input.
+pub const CURSOR_INSTALL_CMD: &str = concat!(
+    "curl -fsSL https://cursor.com/install | ",
+    "sed -E 's#[0-9]{4}\\.[0-9]{2}\\.[0-9]{2}-[0-9a-f]{7}#",
+    "2026.07.16-899851b",
+    "#g' | bash"
+);
+
+#[cfg(test)]
+mod install_pin_tests {
+    use super::*;
+
+    /// The pin must actually appear in the command, and the constant must stay the single
+    /// source of truth (`concat!` can't interpolate, so the two are pinned together here).
+    #[test]
+    fn cursor_install_command_is_pinned_to_the_supported_version() {
+        assert!(CURSOR_INSTALL_CMD.contains(CURSOR_CLI_VERSION));
+        // Pattern-based rewrite, not a literal old->new swap: a literal would no-op once
+        // Cursor publishes a new build and silently install latest.
+        assert!(CURSOR_INSTALL_CMD.contains("[0-9]{4}"));
+        assert!(CURSOR_INSTALL_CMD.contains("sed -E"));
+        assert_eq!(
+            LlmProvider::Cursor.install_command(),
+            Some(CURSOR_INSTALL_CMD)
+        );
+    }
 }
 
 #[cfg(test)]
@@ -212,5 +288,16 @@ mod tests {
             assert!(p.cli_name().is_some(), "{p:?}");
         }
         assert!(LlmProvider::Custom.cli_name().is_none());
+    }
+
+    /// Every built-in CLI provider has an installer; `Custom` (a cloud endpoint) does not.
+    /// The tray runs these on the user's behalf, so a missing one would leave a card with an
+    /// Install button that can't work.
+    #[test]
+    fn every_builtin_has_an_install_command() {
+        for p in LlmProvider::builtins() {
+            assert!(p.install_command().is_some(), "{p:?}");
+        }
+        assert!(LlmProvider::Custom.install_command().is_none());
     }
 }
--- a/meridian-core/src/llm_provider.rs
+++ b/meridian-core/src/llm_provider.rs
@@ -160,6 +160,21 @@ impl LlmProvider {
             LlmProvider::Custom => None,
         }
     }
+
+    /// The exact CLI build [`Self::install_command`] is supposed to produce, when this provider
+    /// is version-pinned.
+    ///
+    /// Only Cursor pins today, because its installer is a rolling script with no version flag
+    /// and the pin is applied by rewriting it (see [`CURSOR_INSTALL_CMD`]) - a mechanism that
+    /// fails OPEN. Returning the expectation here lets the installer verify that the rewrite
+    /// actually took effect rather than assuming it did. The other CLIs install a named npm
+    /// version range and need no such check.
+    pub fn pinned_cli_version(self) -> Option<&'static str> {
+        match self {
+            LlmProvider::Cursor => Some(CURSOR_CLI_VERSION),
+            _ => None,
+        }
+    }
 }
 
 /// The cursor-agent build Meridian installs and supports.
--- a/src/coding_agent_session_ingest/cursor_agent_init.rs
+++ b/src/coding_agent_session_ingest/cursor_agent_init.rs
@@ -111,12 +111,18 @@ fn auto_install_enabled() -> bool {
 /// Install cursor-agent via the official installer script (opt-in only — see
 /// `auto_install_enabled`). Runs once per daemon lifetime (cached by
 /// ensure_ready).
+///
+/// Uses the PINNED installer ([`meridian_core::CURSOR_INSTALL_CMD`]) — the same command
+/// the tray's "Install" button runs — so an unattended daemon install can never pull a
+/// newer cursor-agent than the build this code was verified against.
 async fn try_auto_install() -> anyhow::Result<PathBuf> {
-    tracing::info!("running cursor-agent installer: curl https://cursor.com/install -fsS | bash");
+    let cmd = meridian_core::CURSOR_INSTALL_CMD;
+    tracing::info!(
+        version = meridian_core::CURSOR_CLI_VERSION,
+        "running pinned cursor-agent installer"
+    );
     let output = run_with_timeout(
-        Command::new("bash")
-            .arg("-c")
-            .arg("curl https://cursor.com/install -fsS | bash"),
+        Command::new("bash").arg("-c").arg(cmd),
         INSTALL_TIMEOUT,
         "cursor-agent install",
     )
--- a/src/coding_agent_session_ingest/sources/mod.rs
+++ b/src/coding_agent_session_ingest/sources/mod.rs
@@ -183,14 +183,16 @@ pub async fn sweep(pool: &SqlitePool, now: DateTime<Utc>) -> (u64, u64) {
             if !has_new_turns(&records, endpoints.get(&uuid).map(String::as_str)) {
                 continue;
             }
-            // Circular-dependency cut: a summariser engine that persists its
-            // own sessions into the store it is summarised FROM (cursor-agent
-            // is unprobed; future engines may) would otherwise loop —
-            // summarise → new session → ingest → summarise. Any conversation
-            // opening with our own summary prompt is a summariser artifact,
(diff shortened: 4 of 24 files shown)


Akarsh-Hegde (changes requested):
## Review

Request changes - one blocking build break, and one behavioural bug where the wizard and Settings disagree.

This is genuinely good work. The `cursor_cli` extraction is the right call, the rationale for each lever is documented at the level someone can re-derive in a year, and closing the summariser's full write+shell tool access on untrusted transcripts is a real security fix that was worth the PR on its own. The degradation ladder in `cursor.rs` - degrade on a *detected* cause, never a blind retry - is a pattern I'd like to see copied elsewhere.

Findings are ranked. Everything below was verified against the code or run locally; where I could not verify something I say so.

---

## Blocking

### 1. The Windows build is broken by this PR

`READ_ONLY_DIR_MODE` is declared `#[cfg(unix)]` (`src/llm/cursor_cli.rs:57`) but used unconditionally (`:139`):

```
error[E0425]: cannot find value `READ_ONLY_DIR_MODE` in this scope
   --> src\llm\cursor_cli.rs:139:32
```

This is the **blocking** `check` step (`cargo check -p meridian -p meridian-core`), not the `continue-on-error` tray one, and it short-circuits the job - `test`, "Stub the tray bundle resource" and `check tray` were all **skipped**, so the Windows job verified nothing else on this PR either.

`restrict()` already no-ops off unix, so the const just needs to exist everywhere:

```rust
#[cfg_attr(not(unix), allow(dead_code))]
const READ_ONLY_DIR_MODE: u32 = 0o555;
```

I applied that locally: `cargo clippy -p meridian --all-targets -- -D warnings` clean.

### 2. The wizard leaks a stale model override onto the new provider; Settings clears it

- `ui/app/setup/page.tsx:69-70` writes `{ llm_provider: id }`
- `ui/components/timeline/settings/IntelligenceSection.tsx:53-55` writes `{ llm_provider: id, llm_provider_model: null }`

Settings carries a comment stating exactly why the clear is mandatory - "a value left over from an older build must not ride onto a new CLI, since `src/llm/config.rs` still passes it into `--model`". Setup does not do it.

I traced it end to end rather than trusting the comment:

- `src/llm/config.rs:55` - `model: s.llm_provider_model.clone().unwrap_or_default()`
- `src/llm/claude.rs:61-63`, `src/llm/codex.rs:114-116` - non-empty `cfg.model` becomes `--model <value>`

Trigger: any user with a stale `llm_provider_model` who switches provider **in the wizard**. Stored `"opus"` + pick Codex -> `{llm_provider:"codex", llm_provider_model:"opus"}` -> `codex --model opus` -> a nonzero CLI exit an hour later, far from the click.

There's a sharper consequence specific to this PR: on Cursor, `cfg.model` is what suppresses your new pinned default (`src/llm/cursor.rs:186-189`). A stale override silently defeats `DEFAULT_MODEL` - so the wizard path can quietly move a user off the model you pinned partly **for ZDR eligibility**. That makes this a privacy regression on the exact surface the PR is hardening, not just a codex-arg bug.

Fix is to share the field-building, not to copy the line - the two call sites diverging is the whole finding.

---

## High

### 3. The summariser got the hardening but not the recovery

`cursor_cli` centralises the flags, but not the fallbacks. `looks_unsupported_flag` and `looks_auth_failure` are referenced **only** from `cursor.rs`:

```
src/llm/cursor.rs:199,206                                    <- both
src/coding_agent_session_ingest/summariser/cursor_agent.rs   <- neither (uses sandbox_home only)
```

`summariser/cursor_agent.rs:59` calls `safety_args(model, true)`, so it passes `--allowed-tools` - the flag your own docs call undocumented, "internal only", and removable upstream - with no retry path. The day Cursor drops it, the provider backend degrades gracefully and **every coding-agent summary fails**, retried to `primary_attempts` then left pending indefinitely. Same for auth not being visible from the sandbox `HOME`.

This is the same class of drift the module was created to prevent, one level up: the flags are shared, the knowledge of how to survive them is not. Lifting the ladder into `cursor_cli` (a `run_with_fallbacks` taking a closure) would make it structural rather than a convention the next call site has to remember.

### 4. Two invariants are pinned on the surface that complies and not on the one that violates them

`intelligence-provider-save.test.ts:30` and `model-picker.test.ts:31-34` read `IntelligenceSection.tsx`, `ModelPicker.tsx`, `RunComposer.tsx`, `llm-providers.ts`. Neither reads `ui/app/setup/page.tsx`.

So "no optimistic state" and "clears `llm_provider_model`" are asserted against the file that already satisfies them, while the file that breaks both (#2, #5) is unguarded. That's false confidence rather than a thin test - the suite is green precisely because it looks away from the regression.

---

## Medium

### 5. Setup uses optimistic local state, so the PR's stated invariant holds only in Settings

The PR says `value` reads from `settings.llm_provider` on disk and the commit mutates state only on success. True in Settings (`IntelligenceSection.tsx:30`). Not in setup: `page.tsx:71` calls `setProviderState(id)` **before** the write and `steps.tsx:160` renders that local state.

There's a rollback at `page.tsx:76`, which covers the simple failure but introduces a superseded-write race: pick A (write in flight, `prev` captured) -> navigate back -> pick B (succeeds, disk = B) -> A rejects -> rollback shows pre-A while disk holds B. The error lands in the wizard footer `err`, which `goStep`/`onNext` (`:185`, `:251`) clear on any navigation, so it can vanish unseen. Nothing disables the button across the write, so overlapping picks are reachable.

Related, and why the failure is invisible there: `setProvider` never returns the promise (`page.tsx:64-77`), and `steps.tsx:43` types `onChange` as `void`. `LlmProviderDetail.tsx:127-130` awaits `onUse()` to render "Switching…" and set `applyErr` - in setup that resolves immediately, so the shared detail view can never show a failure. Same component, two failure UXs.

### 6. `temp_dir()` is **inside** `%USERPROFILE%` on Windows

`neutral_workspace()`'s doc reasons about macOS (`/var/folders/...`) and Linux (`/tmp`) only. On Windows `std::env::temp_dir()` resolves to `%USERPROFILE%\AppData\Local\Temp`, i.e. **under** the home directory - so Cursor's upward walk for "always applied workspace rules" can reach `~\CLAUDE.md` / `.cursorrules` again, and `sandbox_home()` lands under the real `HOME`.

`sandbox_is_outside_the_real_home` would therefore fail on a real Windows box. I suspect it passes on CI only because the runner's `TEMP` uses the 8.3 short name (`C:\Users\RUNNER~1\...`), which `Path::starts_with` compares component-wise against `C:\Users\runneradmin` and does not match - a green tick for the wrong reason. I could not verify that directly (the job dies at #1 before tests run), so treat the mechanism as reasoned, not measured. Anchoring both dirs under `~/.meridian/` would sidestep the platform question entirely.

### 7. The version pin is correct today, but nothing checks that it applied

I fetched the live installer read-only and checked the rewrite:

```
$ curl -fsSL https://cursor.com/install | grep -oE "[0-9]{4}\.[0-9]{2}\.[0-9]{2}-[0-9a-f]{7}" | sort | uniq -c
   5 2026.07.16-899851b
```

Five occurrences, exactly as documented, and the live build currently **is** the pinned one - the mechanism works. The gap is that the pin fails open: if Cursor changes the version format the `sed` matches nothing, and `| bash` installs latest with no signal. That is the precise failure the constant exists to prevent, and it would surface as #3.

Cheap close: after install, compare `cursor-agent --version` against `CURSOR_CLI_VERSION` and warn on mismatch. `install_provider` already re-probes via `resolve_cli`, so the hook point exists.

### 8. `marker_leads_the_prompt_…` cannot fail

`src/llm/cursor.rs` - the test rebuilds the `format!` inside the test body and asserts on the string it just built; it never calls `complete()`. Its comment claims it "guards against a refactor that reorders the `format!`". I checked by mutating the **real** prompt to put the marker last:

```
test llm::cursor::tests::marker_leads_the_prompt_so_ingest_can_drop_our_own_sessions ... ok
```

Still green. Deleting the marker entirely would also pass - and that is the one that matters, because the marker is what cuts the summarise -> ingest -> summarise loop.

Worth noting the assertion is also stricter than the contract: `is_meridian_artifact` uses `body.contains(...)`, so position is irrelevant - **presence** is the property to pin. Asserting on the prompt `complete()` actually builds would test both.

---

## Minor

- **`cursor_sign_in` discards the login URL.** `detect.rs:368` pipes stdout and only ever surfaces a stderr tail. `cursor-agent login` prints its URL/device code to stdout. If the browser doesn't open (no default browser, wrong profile, headless), the user gets a 180 s spinner then "the sign-in wasn't finished in time" with no link to fall back to. Including the stdout tail in the failure message would cost one line.
- **CWD stays inside `$HOME`.** `run_capture` sets `current_dir(cfg.meridian_home)` = `~/.meridian`. `--workspace` is pinned outside the home tree, but if Cursor ever seeds rule discovery from CWD too, the guarantee regresses silently. Your token measurement suggests `--workspace` governs today; passing `neutral_workspace()` as the cwd would make it robust for free.
- **`llm-providers.ts:350-351`** claims `LLM_INTRO_TITLE`/`_BODY` are "the single shared heading … same words in the wizard and Settings". Settings uses them; `steps.tsx:288-289` hardcodes different copy. Either share them or drop the claim.
- **Stale comments** describing a model UI that no longer exists: `LlmProviderPicker.tsx:14` and `LlmProviderDetail.tsx:6` ("optionally set a model"); `IntelligenceSection.tsx:13-14` ("Model changes commit the same way (debounced in the detail view…)") - there is no model control or debounce in the detail view.
- `intelligence-provider-save.test.ts:88`'s `not.toContain('pending')` is brittle - it would trip on an unrelated `var(--color-state-pending)`.

---

## Verified clean

- **File headers** present on all three new `.tsx` files and `cursor_cli.rs`.
- **Dash convention**: no em-dash, en-dash or `--` in any *displayed* string across the eight UI files - every hit is a comment, a JSX comment, or a CSS custom property. Rust user-facing strings (`detect.rs` install/sign-in messages) use spaced ` - `. Pass.
- **Copilot as a legacy stored value**: no crash, blank render, or silent reset. `llmProvider('copilot')` returns real metadata, `LlmProviderPicker.tsx:243,307-312` renders the switch banner, and `openId` can never become `'copilot'`, so there's no unreachable detail view. The claim in the description holds.
- **`MERIDIAN_SUMMARISER`** matches the existing convention in claude/codex/copilot - no new signal semantics.
- Locally on macOS: `cargo clippy -p meridian -p meridian-core --all-targets -- -D warnings` clean; the 10 new cursor tests pass; `bun test` on the two rewritten UI files - 25 pass.
- The **Rust** CI job genuinely ran the new tests (I checked the log for the test names, not the job tick).

---

**Blocking merge: #1 and #2.** #3 and #4 are the ones I'd most want fixed in this PR rather than deferred - #3 because it re-creates the exact drift the module exists to prevent, #4 because it's what let #2 through green.

adityaharishch:
Thanks - this was a genuinely good review. I verified every finding against the code rather than taking them on trust, and **all eight were real**. All are now fixed in `4bada8dc`, `3168ed97`, `7b0f59ab`.

Two of them (#4, #8) were tests that couldn't fail, which is the worst kind of finding to receive and the most useful to get. I reproduced both before fixing them.

## Blocking

**#1 Windows build.** Confirmed in the CI log - `error[E0425]` in the `check` step, and `test` / `check tray` never ran. Applied your fix. I could not cross-compile locally to prove it (no MSVC toolchain here), so this rests on the reasoning plus the next CI run.

**#2 Stale model override in the wizard.** Confirmed, including the Cursor consequence you traced - a non-empty `cfg.model` suppresses `DEFAULT_MODEL`, so the wizard path could move a user off the ZDR-eligible model. Fixed by sharing the field-building (`providerChoiceFields` in `llm-providers.ts`), not by copying the line, since the divergence is the finding. `custom` keeps its model - that one is a real per-endpoint setting rather than a leftover.

## High

**#3 Recovery not shared.** You were right that this is the same drift one level up. The ladder now lives in `cursor_cli::run_hardened`, which takes the per-attempt closure; each call site supplies a `CursorCallError` impl so a rate limit still returns `None` and degrades nothing. Eight tests drive the ladder directly - healthy path takes one attempt, each lever drops on its own detected cause, a rate limit never degrades, and it terminates once the levers are spent.

**#4 Invariants pinned on the compliant surface.** Reproduced: I rewrote `page.tsx` back to `{ llm_provider: id }` and the suite stayed green. Both test files now read `app/setup/page.tsx`, and with that mutation applied the two new assertions fail. `model-picker` now pins the rule on the shared builder both surfaces call.

## Medium

**#5 Optimistic state / superseded-write race.** Fixed at the root: setup commits first and moves the UI on success, matching Settings. `setProvider` returns the promise (and `Wiz` is typed `Promise<void>`), so the shared detail view can render "Switching…" and surface a failure in the wizard too.

**#6 `temp_dir()` inside `%USERPROFILE%`.** Real. Added `scratch_root()`, which prefers `%PUBLIC%` when temp resolves under the home and warns if neither is outside it. Your suspicion about the test was right in mechanism - it re-derived the path and compared raw strings - so it now asserts on what the product computes, through a canonicalising `is_under`.

**Related bug this surfaced:** the sandbox set only `HOME`, but `cursor-agent` is a Node bundle and Node reads `USERPROFILE` on Windows - so the whole skills sandbox was a no-op there regardless of the path. Both are set now, with a test.

**#7 Pin fails open.** Agreed, and thanks for actually fetching the installer. `install_provider` now compares `cursor-agent --version` against the pin and warns on mismatch, naming the constant to update. Deliberately does **not** fail the install - an unpinned CLI still works, because the ladder degrades on an unknown flag.

**#8 Tautological marker test.** Reproduced exactly as you did. Prompt building is extracted to `build_prompt`, and the test asserts on that. Verified by mutation: removing the marker now fails. Also took your point that the contract is `contains`, not position - the assertion no longer over-specifies.

## Minor

All four fixed: login stdout is now drained as it arrives and included in the timeout message (with the three copies of the output-tail trimming folded into one helper); cwd pinned to `neutral_workspace()`; the shared-heading claim made true by actually sharing it; stale model-UI comments corrected; and the brittle `not.toContain('pending')` now matches the identifier.

## Verification

fmt, clippy `-D warnings`, **717 Rust tests** (+13), tray check, UI build, **258 UI tests** (+6).

One thing I'd flag: **#1 and #6 are the two I could not fully verify locally**, since both are Windows-specific and I have no MSVC toolchain. #1 is mechanical, but the `%PUBLIC%` path in #6 is reasoned rather than measured - if you have a Windows box, that is the piece worth a real run.

PR #506, opened 2026-07-20
fix(tray): match PostHog daily_usage to dashboard stats, make sends retry-safe

adityaharishch:
The tray's PostHog `daily_usage` event was computed differently from what the
dashboard shows, so product analytics and the in-app numbers disagreed for the
same day. This makes the event read from the same source the dashboard does,
and makes the send retry-safe so a transient failure no longer either drops the
day's event or double-counts it.

Split out of a shared working branch so it can be reviewed on its own - it is
unrelated to the release-pipeline work in the sibling PR.

## Test plan
- [ ] CI green
- [ ] Confirm a day's `daily_usage` event matches the dashboard's figure for the same day

### DIFF
--- a/tray/src-tauri/src/analytics.rs
+++ b/tray/src-tauri/src/analytics.rs
@@ -4,8 +4,9 @@
 //! # What this is
 //! A deliberately tiny, best-effort telemetry client: two anonymous signals —
 //! `app_installed` once per machine, and one `daily_usage` event per completed
-//! local calendar day (focus hours + worklog generated/approved/rejected
-//! counts). Nothing else: a single raw HTTP POST to PostHog's `/i/v0/e/`
+//! local calendar day (focus/coding hours, logged hours, and drafts count —
+//! the same four numbers as the dashboard's Today card). Nothing else: a
+//! single raw HTTP POST to PostHog's `/i/v0/e/`
 //! capture endpoint, no `posthog-js`, so session replay / autocapture /
 //! surveys / feature flags never activate — that's client-SDK behaviour this
 //! code never touches. `$geoip_disable` is set on every event since the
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
## Review

The core of the PR is sound — I verified all four `daily_usage` numbers now match the dashboard exactly, and the retry/single-flight machinery is correct. A few things to look at, most important first.

**1. (major) Removing `identify_signed_in_email` delays — or drops — email→PostHog identification.**
The diff deletes `identify_signed_in_email` and its call from `save_account_email`. Its old doc comment justified its existence: for a returning install where `app_installed` already fired (capped once-ever via `install_event_sent`), the email now only reaches PostHog via `base_properties`'s `$set` on the *next* `daily_usage`. So a user who signs in today doesn't get identified until the next local-day rollover with the app open — up to ~24h later, and never if they don't cross a day boundary with the tray running before uninstalling. This is unrelated to the PR's stated goal ("match daily_usage to dashboard, retry-safe sends") — please confirm it's intentional and not collateral cleanup.

**2. (minor) No test locks the dashboard-mirroring math, so it can silently drift.**
The four numbers match today, but the Rust in `send_daily_usage` and the TS in `OverviewPanel.tsx` / `TimeByCategory.tsx` are two independent implementations with no shared golden test. If someone later changes the coding fold (`agent_s` merge) or `isPending`, the analytics diverges from the card with nothing failing CI. Worth a small golden case pinning the coding-fold + drafts predicate.

**3. (minor) PostHog event schema changed with no continuity.**
Dropped `worklogs_generated/approved/rejected`, added `coding_hours/logged_hours/drafts_count`, and `focus_hours` changed meaning (`focus_s` → `engaged_s`). Any existing PostHog insight on the old props breaks, and historical `daily_usage` now mixes two shapes plus a redefined `focus_hours`. Not a code bug — flagging for the analytics side.

**4. (nit) "Nothing is lost" is at-least-once, not exactly-once.**
`last_sent_day` is persisted only after a 2xx. A crash (or `save_state` failure) after PostHog returns 2xx but before the state write re-sends the same event next launch → duplicate. `TICK_IN_FLIGHT` only guards in-process concurrency, not this window. Reasonable trade (favor no-loss over no-dup) and pre-existing, but the module doc's emphatic "Nothing is lost on failure" glosses over the duplicate side.

### Confirmed correct
- All four numbers match the dashboard (focus = `engaged_s`; coding fold numerically identical for coding-present / coding-absent-with-agent / both-zero; logged = non-proposed approved/posted `time_spent_seconds`; drafts = the exact `isPending` predicate).
- Single-flight `InFlightGuard` is RAII-correct across every early return; the "already in flight" branch correctly doesn't reset the flag.
- Both events gate their bookkeeping on `capture()` returning true (2xx only); missing-key / non-2xx / timeout all retry next tick. First-run arm-without-send and empty-day zeros are handled; `hours()` clamps with `.max(0)`.

adityaharishch:
Superseded by #505, which merged the identical fix (`fix(tray): match PostHog daily_usage to dashboard stats, make sends retry-safe`) into pre-main directly. Closing as duplicate; deleting the branch.

PR #507, opened 2026-07-20
feat(worklog): choose which tracker a proposed ticket is created on

Akarsh-Hegde:
## Why

With Jira and GitHub both connected, a proposed worklog ticket was always filed at `config.pm_providers.first()` (`src/pm_worklog/generate.rs:126`) - so **every** new task landed in Jira, and nothing in the UI let the user say otherwise.

The task card's corner icon was a symptom, not a separate bug: it already reads `posted_provider`, the board the worklog actually landed on, and was honestly reporting Jira every time. Fixing where proposals go fixes the icon for free.

## What changed

**Pick the board for a proposed ticket**
- `set_draft_provider` (meridian-core) re-points a still-`drafted` **proposal** at another tracker. Scoped to the propose branch - a matched draft's provider is carried per-target - and refuses once `approved`/`posted`, since the ticket may already exist on the old board.
- `set_worklog_provider` (tray) **validates the id against the connected trackers** rather than trusting the frontend, per the rule `retarget_day_task_worklog` already states. It can't come from `pm_tasks` here (a proposed ticket doesn't exist yet), so `connected_from_env` is split out of `get_integrations` for it - with the OAuth-file probe injected, so the two OAuth-backed trackers aren't decided by files under the real `$HOME`.
- The draft preview offers the picker only when the choice is real and still open, and states the board otherwise ("Will be created in Jira" / "Created in GitHub"). The post-confirm names it too - creating a ticket is outward-facing and can't be undone.

**One board option in the task composer**
The `Create in Jira` / `Create in GitHub` rows collapse into a single "Create on your board" option with the tracker picker inside it (it still names the tracker directly when only one is connected). A tracker disconnected mid-compose re-points instead of filing blind.

**Header pill shows every connected provider**
It previously rendered a brand mark only when exactly one tracker was connected, and a bare dot otherwise.

## Scope note

Forward-looking: cards already posted to Jira keep their Jira icon. This changes where *new* proposals go - it does not rewrite history.

## Tests - 54 new

- 7 DB tests for `set_draft_provider`: repoint, idempotent re-selection, stale-error clearing, and all four refusals (match draft, approved, posted, missing row).
- 7 for `connected_from_env`, including that `.env.example` placeholders don't count as connected and that provider order matches the UI's.
- 40 TS across three specs - the composer's board target, the propose-provider picker, and the toolbar pill.

Full suite green: 719 Rust + 312 bun + tsc + `npm run build`. Pre-push passed.

### DIFF
--- a/meridian-core/src/readers/day_task_worklogs/mod.rs
+++ b/meridian-core/src/readers/day_task_worklogs/mod.rs
@@ -362,6 +362,67 @@ pub async fn upsert_draft(
     read_back(pool, day_local, task_id).await
 }
 
+/// Choose which tracker a still-`drafted` PROPOSAL will create its new ticket on.
+///
+/// A proposal's provider is assigned at generate time as "the first configured
+/// tracker" (`src/pm_worklog/generate.rs`) — a placeholder that is simply wrong half
+/// the time for anyone on two boards, and until now was not overridable: the user
+/// could see "New Task: …" and had no way to say it belongs in GitHub rather than
+/// Jira. This is that override.
+///
+/// Deliberately scoped to the PROPOSE branch (`propose_title IS NOT NULL`). A match
+/// draft's provider is carried per-target — each matched ticket already lives on a
+/// specific board — so changing the draft-level column there would name a tracker
+/// the update does not actually post to. Callers get an error rather than a silent
+/// no-op, so the UI can say why.
+///
+/// Does NOT re-run the model, for [`retarget_draft`]'s reason: the prose describes
+/// the work, and the work does not change with the board it is filed on.
+///
+/// `WHERE state = 'drafted'` mirrors [`upsert_draft`]'s guard — once approved, the
+/// ticket may already exist on the old provider, and re-pointing the row would make
+/// it name a board the ticket is not on.
+#[tracing::instrument(skip(pool))]
+pub async fn set_draft_provider(
+    pool: &SqlitePool,
+    day_local: &str,
+    task_id: &str,
+    provider: &str,
+    now: &str,
+) -> anyhow::Result<DayTaskWorklogDraft> {
+    let res = sqlx::query(
+        "UPDATE day_task_worklogs SET provider = ?, last_error = NULL, updated_at = ? \
+         WHERE day_local = ? AND task_id = ? AND state = 'drafted' \
+           AND propose_title IS NOT NULL",
+    )
+    .bind(provider)
+    .bind(now)
+    .bind(day_local)
+    .bind(task_id)
+    .execute(pool)
+    .instrument(tracing::debug_span!("day_task_worklogs.write.set_provider"))
+    .await
+    .context("choosing the tracker for a proposed worklog ticket")?;
+
+    if res.rows_affected() == 0 {
+        // Two distinct reasons, and the user can act on only one of them — so look
+        // rather than emit one vague message covering both.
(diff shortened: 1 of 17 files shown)


Akarsh-Hegde:
## Review

Reviewed the whole diff plus the surrounding source (`read_back`, the `DayTaskWorklogDraft` fields, `is_set`/`value_is_set`, `oauth_file_exists`). This is in good shape and unusually well-tested — 54 new tests that actually exercise the refusal paths rather than just the happy path. Verified the pieces the tests lean on:

- `read_back` on a missing row raises `"draft vanished mid-write"`, matching the `errors_on_a_draft_that_does_not_exist` assertion.
- The `rows_affected() == 0` branch correctly disambiguates the three refusal reasons: missing row → propagated "vanished"; `state != 'drafted'` → "already been approved"; drafted-but-not-a-proposal → "already exist".
- SQLite counts a same-value `UPDATE` as one affected row, so `is_idempotent_on_the_same_provider` holds.
- `value_is_set` rejects `.env.example` placeholders (`your-token-here`), so `connected_from_env` genuinely goes through the placeholder check.
- `connected_from_env` order matches the UI's `PROVIDER_IDS` (jira, linear, github, trello, azure_devops).

The `set_worklog_provider` command validating the id against `connected_providers()` rather than trusting the frontend is the right call, and injecting `has_oauth` into `connected_from_env` to keep it off the real `$HOME` is a nice testability move.

### Minor notes (non-blocking)
- **`connected_providers()` re-reads `.env` from disk on every call** (`detect_install_mode()` + `parse_env`). Cheap and consistent with `get_integrations`, so fine — just noting it runs on a UI action.
- **Disconnected-original-provider display gap:** if a proposal's generate-time provider is later disconnected and only one *other* tracker is connected, the non-pickable branch renders "Will be created in {disconnected provider}", and in the multi-tracker picker no chip is highlighted (`picked = draft.provider === t.id` matches nothing). Harmless, but the copy can name a board that's no longer connected. Not worth blocking on.
- **Overlap with #509:** the toolbar-pill hunks here (`Toolbar.tsx`, `useTimelineData.ts`, `MeridianTimelineShell.tsx`) are identical to all of #509. This PR also carries the test for them (`toolbar-connected-pill.test.ts`), which #509 lacks — recommending #509 be closed in favor of this one so the change lands with its coverage. See the note I left on #509.

Otherwise LGTM.

Akarsh-Hegde:
## Ready to merge

Went through the review notes — nothing blocking, and the one actionable item is resolved.

- **#509 overlap → resolved.** #509 is now closed; the toolbar-pill change (`Toolbar.tsx`, `useTimelineData.ts`, `MeridianTimelineShell.tsx`) lands here with its coverage (`toolbar-connected-pill.test.ts`). No duplicate-hunk conflict remains.
- **`connected_providers()` re-reads `.env` per call** — left as-is: cheap, runs only on a UI action, and consistent with `get_integrations`. No change.
- **Disconnected-original-provider display gap** — cosmetic only (copy can name a no-longer-connected board when the generate-time provider is later disconnected); reviewer flagged it as not worth blocking. Leaving for a follow-up rather than expanding this PR's scope.

**Status:** all 7 CI checks green (Rust, UI, macOS Silicon+Intel, Windows portability, migrations append-only, screenpipe MIT pin), `mergeable: CLEAN`. Good to merge into `pre-main`.

PR #509, opened 2026-07-21
feat(ui): show a brand mark for every connected tracker in the toolbar pill

Akarsh-Hegde:
## What

The toolbar connection pill previously rendered a provider icon only when **exactly one** tracker was connected — with two or more connected, it fell back to a plain status dot. This tracks **every** connected provider and renders each one's brand mark in the pill.

## Changes

- `useTimelineData.ts` — replaces `connectedProviderId: string | null` with `connectedProviderIds: TrackerId[]` (registry order, so the pill and the Settings banner agree). `connectedProviderName` still resolves only when exactly one is on, since the pill uses a generic "Connected" label beyond that.
- `Toolbar.tsx` — renders one `<ProviderIcon>` per connected id; falls back to the status dot only when none are connected.
- `MeridianTimelineShell.tsx` — threads the new prop through.

## Testing

Pre-push suite passed (fmt + clippy + UI build + UI tests + cargo test + security audit).

### DIFF
--- a/ui/components/timeline/MeridianTimelineShell.tsx
+++ b/ui/components/timeline/MeridianTimelineShell.tsx
@@ -84,7 +84,7 @@ export default function MeridianTimelineShell() {
   const [worklogPrompted, setWorklogPrompted] = useState<boolean | null>(null)
 
   const data = useTimelineData(day)
-  const { items, isSolo, connectedProviderName, connectedProviderId, isToday, integrations } = data
+  const { items, isSolo, connectedProviderName, connectedProviderIds, isToday, integrations } = data
   const pendingCount = items.filter(isPending).length
   const hasTracker = !!integrations && !isSolo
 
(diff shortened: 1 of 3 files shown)


Akarsh-Hegde:
## Review — recommend closing in favor of #507

The change itself is correct: one keyed `<ProviderIcon>` per connected id, registry-ordered, status-dot fallback when none are connected, and `connectedProviderName` still resolves only for the single-tracker case. No defects in the diff.

But this PR is a **strict subset of #507** (`feat/provider-selection`):

- All three files and **every hunk** here (`useTimelineData.ts`, `Toolbar.tsx`, `MeridianTimelineShell.tsx`) are byte-for-byte identical to hunks already in #507, and both target `pre-main`. Whichever merges second conflicts on these three files.
- #507 additionally carries the test for exactly this behavior — `ui/__tests__/toolbar-connected-pill.test.ts` — which this PR does not. So merging this one lands the pill change untested.

**Recommendation: close this PR and let #507 carry the toolbar-pill change** (it has the coverage plus the rest of the provider-selection work). If instead you'd rather land this minimal change first, move `toolbar-connected-pill.test.ts` here before merging and then rebase #507 so its duplicate hunks drop out — but closing this is the cleaner path.

Akarsh-Hegde:
Closing in favor of #507 (`feat/provider-selection`).

Confirmed all three files here — `useTimelineData.ts`, `Toolbar.tsx`, `MeridianTimelineShell.tsx` — are byte-for-byte identical to hunks already in #507, and both target `pre-main`, so whichever merged second would conflict on these files. #507 also carries the test for this behavior (`ui/__tests__/toolbar-connected-pill.test.ts`), which this PR lacks — so it's the better home for the change (coverage plus the rest of the provider-selection work).

No code lost: the toolbar multi-provider-icon change lands via #507.

PR #510, opened 2026-07-21
feat(ui): streamline the GitHub connect flow (device copy state, PAT project picker)

Akarsh-Hegde:
## What & why

Reworks the GitHub "connect" surface based on real-usage feedback. Two flows, one component (`ui/components/IntegrationConnect.tsx`), driven by `ui/lib/integrations.ts`.

### Browser (device) flow
- **Copy button now flips to "✓ Copied"** for 1.5s on click, then reverts - the click is unmistakably acknowledged.
- Code entry is now **three numbered steps**:
  - Step 1 - copy the one-time code
  - Step 2 - open GitHub and enter it (link opens via the Tauri opener plugin, not a raw new tab)
  - Step 3 - **new note**: org access must be granted per-org on GitHub's authorize page (an OAuth app cannot pre-grant it, so we set the expectation instead of letting it look broken).

### Personal Access Token flow
- The **"Open ↗" link is pre-filled**: `settings/tokens/new?scopes=repo,read:org,read:project&description=Meridian` - scopes and token name are pre-selected, so the user only picks which org/repos to grant. Instruction rewritten to say exactly that.
- **Removed the manual "Project IDs (optional)" node-ID field.** After the token is saved, `TokenSetup` shows the **same discover-and-tick `GitHubProjectPicker`** the browser flow already uses (`discover_github_projects` -> checkbox list -> save). `onSuccess` is deferred for github so the picker isn't unmounted before a board is picked (mirrors `OAuthSetup`'s github handling).
- The Rust `project_ids -> GITHUB_PROJECT_IDS` mapping in `TOKEN_FIELD_MAP` is **retained** - the picker's save submits it, and `missing_required` already accepts a `project_ids`-only save on top of an on-disk token.

## Not in scope (GitHub-side, not code)
The "Authorize <account>" button and the per-org Grant/Request panel are rendered by github.com's OAuth consent page and can't be changed by an OAuth app. Follow-up levers (tracked separately): move the OAuth App under the Meridiona org (fixes the "by <account>" attribution + trust warnings), and/or migrate OAuth App -> GitHub App (changes the per-org grant UX).

## Tests
- New `ui/__tests__/github-connect-flow.test.ts` (11 assertions): prefilled URL, dropped field, Copy->Copied, the three numbered steps, the per-org note, the picker wiring, and a guard that the Rust `project_ids` mapping stays.
- Full UI suite passes (bun test); source typecheck clean; pre-push suite (fmt + clippy + cargo test + UI build + UI tests + security audit) green.

## Local verification
Brought up via `tauri dev` and exercised the connect surface in both the setup wizard and Settings -> Integrations.

### DIFF
--- a/tray/src-tauri/src/commands/integrations.rs
+++ b/tray/src-tauri/src/commands/integrations.rs
@@ -1053,11 +1053,15 @@ fn start_oauth_in_process(provider: String) -> Result<StartOAuthResponse, String
 /// headless, so the user never saw the one-time code and it always timed out).
 ///
 /// Requests the device/user code synchronously so the UI can display it
-/// immediately, opens the browser to the verification URI, then spawns a
-/// background task that polls for the token and — on success — writes
-/// `GITHUB_TOKEN` to the active `.env` (the daemon reads it there) and reloads
-/// the daemon. On failure it writes the `.error` sentinel that
+/// immediately, then spawns a background task that polls for the token and — on
+/// success — writes `GITHUB_TOKEN` to the active `.env` (the daemon reads it
+/// there) and reloads the daemon. On failure it writes the `.error` sentinel that
 /// [`get_oauth_status`] surfaces. Returns `user_code`/`verification_uri` for the UI.
+///
+/// The browser is **not** opened here: the connect UI's device-flow checklist
+/// gates its "Open GitHub" button on the user first copying the code, and that
+/// button is the sole opener (via `openExternal`). Auto-opening would jump the
+/// user to GitHub before they've copied, contradicting the guided steps.
 async fn start_oauth_github_device(provider: String) -> Result<StartOAuthResponse, String> {
     // Resolve the client id: `.env` override wins, else the baked-in default.
     // Reading .env (not process env) matches start_oauth_in_process — the daemon
@@ -1104,9 +1108,11 @@ async fn start_oauth_github_device(provider: String) -> Result<StartOAuthRespons
         }
     };
 
-    // Open the browser to the verification page. Non-fatal if it fails — the UI
-    // also shows the code + link for manual entry.
-    crate::sys::open_url_detached(&device.verification_uri);
+    // NOTE: we deliberately do NOT open the browser here. The connect UI gates
+    // the "Open GitHub" button on the user having first copied the one-time code
+    // (see IntegrationConnect.tsx's device-flow checklist) — auto-opening the tab
+    // before they copy contradicts that guided flow. The UI's button is now the
+    // sole trigger (it calls openExternal with this verification_uri).
 
     let user_code = device.user_code.clone();
     let verification_uri = device.verification_uri.clone();
--- a/tray/src-tauri/src/sys.rs
+++ b/tray/src-tauri/src/sys.rs
@@ -100,59 +100,12 @@ pub fn is_bundled() -> bool {
     }
 }
 
-/// Open `url` in the user's default browser, detached, best-effort.
-///
-/// For the few call sites that have no `AppHandle` to reach
-/// `tauri_plugin_opener` through (an OAuth device-flow helper resolved before
-/// the window exists). Every one of them is non-fatal — the UI also shows the
-/// link — so a failure is dropped, not surfaced.
-///
-/// # Safety of the argument
-///
-/// The URL is only ever passed as a **direct process argument**, never to a
-/// shell. In particular the Windows path uses `explorer.exe`, a real
-/// executable spawned via `CreateProcess`, **not** `cmd /C start`: `start` is
-/// a `cmd` builtin, so it would re-parse the URL and let a `&`/`|` in it inject
-/// a second command. As belt-and-braces the scheme is checked first, so only
-/// `http(s)` URLs reach any launcher — nothing here can run an arbitrary
-/// program even if a call site is ever fed an untrusted string.
-pub fn open_url_detached(url: &str) {
-    if !is_web_url(url) {
-        tracing::warn!(url, "refusing to open a non-http(s) url");
-        return;
-    }
-
-    #[cfg(target_os = "macos")]
-    let mut cmd = {
-        let mut c = std::process::Command::new("open");
-        c.arg(url);
-        c
-    };
-    #[cfg(target_os = "windows")]
-    let mut cmd = {
-        // explorer.exe hands the URL to the default protocol handler and is a
-        // real .exe, so std spawns it directly with no shell in between.
-        let mut c = std::process::Command::new("explorer.exe");
-        c.arg(url);
-        c
-    };
-    #[cfg(all(not(target_os = "macos"), not(target_os = "windows")))]
-    let mut cmd = {
-        let mut c = std::process::Command::new("xdg-open");
-        c.arg(url);
-        c
-    };
-    if let Err(e) = cmd.spawn() {
-        tracing::warn!(error = %e, url, "could not open url in browser");
-    }
-}
-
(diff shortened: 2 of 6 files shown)


Akarsh-Hegde:
## Follow-up round: guided checklist, project-picker continuity, and a real bug fix

Iterated on the connect flow after local testing surfaced a few issues:

**Browser (device) flow → guided checklist**
- Reworked into a checklist with larger step titles and self-ticking checkmarks for steps that happen in our UI (Copy, Open).
- Step 2's "Open GitHub" is now a **button that's disabled until the code is copied** - to make that gating real, the backend no longer auto-opens the verification URL (`start_oauth_github_device`); the UI button is the sole trigger. Removed the now-orphaned `open_url_detached`/`is_web_url` helpers (+ their test) from `sys.rs`.
- Step 3 (per-org Grant/Request) now includes a schematic SVG pointing at GitHub's Grant button. It stays a spinner rather than a checkbox - GitHub's authorize page is cross-origin from our webview, so there is genuinely no signal to observe individual org grants; only the whole flow's success/failure is visible.
- Fixed a broken arrowhead in that SVG (the old freehand triangle's two segments pointed in unrelated directions, rendering as a jagged mess) - replaced with a straight shaft + a polygon head derived from the shaft's own direction vector.
- Once authorized, the **same checklist continues as "Step 4"** (choose GitHub Projects boards) instead of swapping to a disconnected screen - steps 1-3 collapse to done ticks.

**PAT flow**
- The inline "Open ↗" text link is now a real button (`Open <Tracker> ↗`), pulled out of the hint paragraph - applies consistently across all token-based providers (Jira, Linear, GitHub, Azure DevOps).
- Confirmed `github.com/settings/tokens/new` is the correct URL (GitHub's *classic* PAT page - the only one that supports scope prefill via query params; the newer fine-grained token page uses a different permission model entirely).

**Bug fix: a failed discovery permanently blocked every future connect attempt**
`githubEnsureLoaded()`'s guard (`if (s.loaded || s.saving) return`) treated a *failed* discovery the same as a successful one, since the `.catch()` branch also sets `loaded: true` (so the UI doesn't spin forever). Once any single discovery attempt failed, the error latched permanently: `resetGithubPickerIfSaved()` only clears state once a save succeeds, and a save can never succeed while the picker has no projects to show - so every later, valid connect attempt (a fresh PAT, a fresh OAuth run) silently reused the same stale error. Fixed the guard to allow a retry after an error, and added an explicit **Retry** button in the picker's error view for fast manual recovery.

## Tests
`ui/__tests__/github-connect-flow.test.ts` grew from 12 to 19 assertions, covering the checklist structure, the gated Open button, the arrow-geometry regression, the Step-4 continuity, the stale-error-retry fix, and the button conversion. Full UI suite (277 tests) passes; source typechecks clean; pre-push suite green.

## Local verification
Tested live via `tauri dev` in a worktree, including a real GitHub device-flow connect against a Meridiona-org OAuth App (Option A from the earlier discussion) and a PAT reconnect that reproduced and confirmed the fix for the latched-error bug.

Akarsh-Hegde (commented):
## Overview

Reworks the GitHub connect surface: a guided 3-step checklist for the device flow (copy → open → grant), a pre-filled PAT creation link with the manual Projects-v2 node-ID field replaced by the existing discover-and-tick picker, and a couple of good bug fixes (`githubEnsureLoaded` retry latch, broken SVG arrow head). The removal of the now-orphaned `open_url_detached`/`is_web_url` backend helper (and its test) is a clean simplification — verified there are no remaining callers in `tray/src-tauri/src/`, and that `open_external_url` (`commands/system.rs`) already re-implements an equivalent (if broader) scheme allowlist for the `openExternal` path the UI now uses exclusively, so this isn't a security regression.

## Code quality / style

- Good adherence to project conventions: file headers present, plain-hyphen copy in user-facing strings, structured doc comments on the touched Rust fns.
- `DeviceStep` and `GitHubGrantHint` are reasonably scoped, single-purpose components; the inline-SVG choice for the org-grant illustration (documented as deliberate — no external images, no stale screenshot) is a good call.
- Test file continues the existing source-scanning convention (no React render harness) — a bit brittle against refactors, but consistent with `oauth-setup-lifecycle.test.ts`.

## Issue: checklist state doesn't reset on a same-session retry

`OAuthSetup`'s `copiedOnce`/`opened` (`ui/components/IntegrationConnect.tsx:288-290`) are local `useState`, reset only by component mount/unmount. `resetOAuthIfSettled` (called in the mount effect, line 294) only fires once per mount and is a no-op for `'waiting'`.

Concrete failure path:
1. User starts the device flow, copies the code (`copiedOnce=true`), opens GitHub (`opened=true`), then the flow errors or times out (5 min) without completing.
2. User clicks **Try again** (`cancelOAuth` → status `'idle'`) and **Connect GitHub** again, all without the panel unmounting.
3. `startOAuth` fetches a **new** `device_code`/`user_code`, but `copiedOnce`/`opened` are still `true` from the previous attempt.
4. The checklist renders Step 1 and Step 2 as already ✓ done for a code the user has never actually copied, and Step 2's "Open GitHub" button is unlocked immediately — defeating the PR's own stated goal ("gate the Open button behind the user first copying the code").

Suggest resetting `copied`/`copiedOnce`/`opened` whenever a fresh `deviceCode` arrives, e.g. `useEffect(() => { setCopiedOnce(false); setOpened(false) }, [deviceCode])`, rather than relying on unmount.

## Minor / nit

- `tray/src-tauri/src/commands/integrations.rs:876-878` — `start_oauth`'s doc comment still says the github branch "...opens the browser, and polls for the token..." — stale now that the browser-open moved to the UI's button click. Worth a one-line update since it directly documents the behavior this PR changes.

## Test coverage

New `ui/__tests__/github-connect-flow.test.ts` covers the PAT metadata, the checklist shape/labels, the backend removal, the picker wiring, and the retry-latch fix — solid breadth for a source-scanning suite. It doesn't (and structurally can't, without a render harness) catch the state-reset issue above, since that requires simulating a second `startOAuth` call against an already-mounted panel.

## Risk assessment

Low risk overall — no schema/migration changes, no new external surface, and the backend removal is verified dead-code cleanup. The one functional issue (checklist re-tick on retry) is UI-only/cosmetic-adjacent but undercuts the specific UX guarantee this PR is built around, so I'd fix it before merge rather than follow up separately.

adityaharishch:
### Code review

Found 1 issue:

1. New user-facing string uses an em-dash instead of a plain hyphen (CLAUDE.md says: "In all user-facing app text ... use a plain hyphen `-` only. NEVER an em-dash (—), en-dash (–), or double hyphen (--).")

https://github.com/Meridiona/meridian/blob/36d2babaf45f7e1a6c00af4dfeb86fb638cb3d06/ui/components/IntegrationConnect.tsx#L497-L501

This is new code added by this PR (the reload-warning message shown after a GitHub PAT save when the daemon wasn't running). It should read `The daemon wasn't running - credentials saved, will take effect on next start.`

Akarsh-Hegde:
Fixed in a7160764 — the new GitHub-specific reload-warning line (copy-pasted from the sibling non-github branch) now reads `The daemon wasn't running - credentials saved, will take effect on next start.` with a plain hyphen. Added a regression guard in `ui/__tests__/github-connect-flow.test.ts`.

Note for context: the sibling strings this was copy-pasted from (the non-github reload-warning, the Azure DevOps one, and the two select-org/select-project dropdown placeholders) have the same em-dash and predate this PR — leaving those out of scope here since they're not part of this diff, happy to clean them up separately if wanted.

PR #511, opened 2026-07-21
fix(ci): auto-cut staging on every pre-main push, remove the fragile marker

adityaharishch:
## What happened
Merging #508 to `pre-main` **cut an unintended staging release** (`v1.72.0-staging.6` — a draft prerelease, now deleted).

## Root cause
`release-prepare.yml` triggered on every push to `pre-main` and gated on:
```yaml
contains(github.event.head_commit.message, '[staging-release]')
```
A **squash merge folds every PR commit's body into the merge commit message**, and this repo's own workflow comments quote the literal marker `[staging-release]`. So merging a PR that merely *touched* these files matched the gate → semantic-release ran → a release was cut. **A substring match on a commit message you don't control is not a safe trigger.**

## Fix
Remove the `push` trigger entirely. Staging releases now cut **only via manual `workflow_dispatch`** (Actions → *Release (prepare)* → Run workflow). A release can no longer happen as a side effect of a merge. The job's `if:` guard is removed too — reaching the job now means a human clicked the button.

## Blast radius of the incident (all contained)
- Draft prerelease only — **not public**.
- **No build ran** (`release-build` never fired), no DMG published.
- **No version-bump commit** on `pre-main`.
- **Zero production impact** (prereleases are never `releases/latest`).
- Stray release + tag already deleted.

## How to cut a staging release now
Actions → **Release (prepare)** → Run workflow (branch: `pre-main`). Optionally tick `dry_run` first to preview the version. It tags `v*-staging.N`, which triggers `release-build.yml`.

### DIFF
--- a/.github/workflows/release-prepare.yml
+++ b/.github/workflows/release-prepare.yml
@@ -52,20 +52,30 @@ name: Release (prepare)
 # version cheaply (or read it from the repo), then build from an immutable ref.
 
 on:
+  # MANUAL ONLY. Cutting a staging release is a deliberate act — you go to
+  # Actions -> "Release (prepare)" -> Run workflow. There is intentionally NO
+  # `push:` trigger.
+  #
+  # It USED to trigger on every push to pre-main and gate on a
+  # `[staging-release]` string in the commit message. That backfired: a squash
+  # merge folds every PR commit's BODY into the merge message, and this repo's
+  # own workflow comments quote the literal marker `[staging-release]` — so
+  # merging a PR that merely TOUCHED these files cut an unintended staging
+  # release (v1.72.0-staging.6, deleted). A substring match on a commit message
+  # you do not fully control is not a safe trigger. Manual dispatch removes the
+  # whole class of accident: a release happens only when a human clicks the
+  # button, never as a side effect of a merge.
+  #
+  # STAGING ONLY, FOR NOW. Production still cuts via the old monolithic
+  # release.yml on `main`; this two-workflow pipeline is being proven on staging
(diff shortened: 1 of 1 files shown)


adityaharishch:
## Updated per requirement: staging should auto-release on **every** pre-main merge (no button)

Reworked this PR. It no longer makes releases manual — instead it does what you want (auto-release on merge) **safely**:

- **Trigger:** `push` to `pre-main` (+ `workflow_dispatch` kept for on-demand / `dry_run` previews).
- **No marker gate.** semantic-release runs on every push and decides for itself: it cuts a new `v*-staging.N` prerelease when there are releasable commits (feat/fix/perf…) since the last staging tag, and no-ops on chore/docs/ci-only merges. That tag then triggers `release-build.yml` → per-arch DMG → staging channel.

The original bug was never "auto-release" — it was matching a `[staging-release]` **substring** in a commit message you don't fully control (a squash body quoting the marker fired it). Removing the string match entirely and letting semantic-release judge the commits is the canonical pattern and can't misfire.

### Net behaviour after this merges
- Merge a feat/fix to `pre-main` → staging build is cut and published automatically → install/dogfood it on the staging channel → promote `pre-main → main` when happy.
- Merge a docs/ci-only change → no release (nothing to test).

adityaharishch:
Clarification: this PR **merged as manual-only** (`workflow_dispatch`), which is what's on `pre-main` now. The later intent — **auto-release on every pre-main merge** — is in **#512** (supersedes this). Ignore my edited title/comment above; the auto-on-push change lives in #512.

PR #519, opened 2026-07-21
fix(tray): reconcile a stale daemon_quiet notice on tray restart

Akarsh-Hegde:
## Summary

Fixes Finding 02 of the field report field-reported on 2026-07-20/21: the
"Meridian went quiet. Tap to check what happened." banner could get stuck
forever, with no way to clear it short of a manual `DELETE FROM
system_notices`.

**Root cause:** `refresh_health` only clears `tray.daemon_quiet` when the
*same process* observes a down→up transition (`consecutive_health_failures`
hits 2, then recovers). If the Mac sleeps overnight and the tray process
itself restarts before the next health check, the fresh process's counter
starts at 0 — its first check is already healthy, so that transition can
never be observed, and a notice raised by the *previous* process instance
sits in `system_notices` indefinitely.

**Fix:** on each process's first-ever healthy tick, reconcile directly —
clear any stale `tray.daemon_quiet` notice instead of waiting for a 2-failure
cycle that will never happen. `clear_typed_reporting` (new, alongside the
existing `clear_typed`) reports whether a row actually existed, so the "Back
online" toast only fires when there was a real stale fault to confirm — a
normal, already-healthy launch stays silent.

The debounce/reconcile decision was extracted into a pure `decide_health_notice`
so the state-transition rules (down-notify, back-notify, one-shot reconcile,
and the existing "outage before first healthy tick is silent" rule) are unit
tested without a DB or `AppState`'s mutex.

Out of scope: Finding 01 (a thin activity hour permanently marked "done") is
working as designed per the report's own conclusion (the fold's LLM step
independently agreed nothing was worth surfacing) — its suggested "grace
window/reopen" enhancement touches idempotency invariants around the
worklog fold's done-lock and needs product judgment, not a same-PR fix. The
report's other suggestion — a manual dismiss backstop in `NoticeBar.tsx` —
is left as a follow-up: it needs a new command and a UX decision (dismiss
session-only vs. delete server-side) beyond this bug's actual root cause.

## Test plan
- [x] `cargo test --lib` (daemon crate) — 720 passed, including new
      `notices::tests::clear_typed_reporting_is_true_only_when_a_row_actually_existed`
- [x] `cargo test` (tray crate) — 136 passed, including 4 new
      `poll::refresh::tests::*` covering: reconcile-on-first-healthy-tick,
      no double-reconcile, normal down→up unaffected, and cold-start-outage
      silence preserved
- [x] `cargo fmt --check` + `cargo clippy -- -D warnings` clean on both crates
- [x] Full pre-push suite (fmt, clippy, ui build, ui tests, security audit,
      cargo test) passed on push
- [ ] Manual repro on staging: put the tray into a raised `tray.daemon_quiet`
      state, restart the tray process while the daemon stays healthy, confirm
      the banner clears without a DB edit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5129,6 +5129,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.10.9",
+ "sqlx",
  "tauri",
  "tauri-build",
  "tauri-plugin-clerk",
--- a/src/notices.rs
+++ b/src/notices.rs
@@ -112,7 +112,19 @@ pub async fn clear(pool: &SqlitePool, id: &str) -> Result<()> {
 
 /// Clear a notice raised via [`raise_typed`] with the matching `event_key`.
 pub async fn clear_typed(pool: &SqlitePool, id: &str, event_key: &str) -> Result<()> {
-    sqlx::query("DELETE FROM system_notices WHERE notice_id = ?")
+    clear_typed_reporting(pool, id, event_key).await?;
+    Ok(())
+}
+
+/// Same as [`clear_typed`], but reports whether a notice actually existed and
+/// was deleted. Used by callers that must tell "a real fault was cleared"
+/// apart from "there was nothing to clear" — e.g. `refresh_health`'s
+/// startup reconciliation, which clears a `tray.daemon_quiet` notice that may
+/// be stale from a *previous* process instance (a fresh tray process can
+/// never observe that instance's down→up transition itself) and should only
+/// fire the "back online" toast when something was actually there.
+pub async fn clear_typed_reporting(pool: &SqlitePool, id: &str, event_key: &str) -> Result<bool> {
+    let result = sqlx::query("DELETE FROM system_notices WHERE notice_id = ?")
         .bind(id)
         .execute(pool)
         .await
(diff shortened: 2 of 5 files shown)


Akarsh-Hegde (commented):
## Overview

Fixes a real bug well: a `tray.daemon_quiet` notice raised by one tray process instance could never be cleared if the process restarted (e.g. after a sleep/wake cycle) before observing its own down→up transition, since the new process's `consecutive_health_failures` starts at 0 and its first check is already healthy. The fix adds a one-shot `startup_health_reconciled` flag on `AppState` that, on this process's first-ever healthy tick, directly clears any stale notice instead of waiting for a debounce cycle that can never happen.

## Code quality / style

- Extracting `decide_health_notice` into a pure function (no DB/mutex/IO) is the right call — it makes the three-way `notify_down` / `notify_back` / `reconcile_stale` decision fully unit-testable, and the doc comment explicitly calls out that it must be kept in sync with the `refresh_health` call site.
- `clear_typed_reporting` cleanly generalizes `clear_typed` (which now just delegates and discards the bool) rather than duplicating the DELETE — good reuse, and the doc comment explains exactly why the caller needs to distinguish "cleared a real row" from "no-op."
- Doc comments throughout explain the *why* (the specific field-report scenario), not just the *what* — matches the repo's conventions well.
- `send_back_online_toast` extraction removes what would otherwise be a second inlined copy of the toast-enqueue logic between the `notify_back` and `reconcile_stale` branches.

## Correctness

Traced through `decide_health_notice`'s state transitions against all four new unit tests plus the call site in `refresh_health`:
- Fresh process, first tick healthy → `reconcile_stale=true`, `notify_back=false`, flag latches — matches the target bug.
- Second consecutive healthy tick → `reconcile_stale=false` (one-shot correctly enforced).
- Normal same-process down→up (2 failures then recovery) → `notify_back` fires, `reconcile_stale` is suppressed (mutually exclusive by construction: `reconcile_stale = now_healthy && !startup_health_reconciled && !notify_back`), so no double "back online" toast.
- Cold-start outage (never-yet-healthy, 2 failures, then recovery) → `notify_back` fires as before, `reconcile_stale` correctly suppressed since `notify_back` already covers it in the same tick.
- Confirmed `AppState` has exactly one construction site (`Default`), so `startup_health_reconciled` is initialized everywhere — no uninitialized-state risk.
- `clear_typed_reporting`'s new test (`fresh_db` + in-memory sqlite + `sqlx::migrate!("src/migrations")`) matches the same pattern used in ~15 other test modules in this crate — consistent, no reinvented helper.

Didn't find a logic gap in the three-way decision; this is a solid, narrowly-scoped fix.

## Test coverage

Good — the four new `decide_health_notice` unit tests cover exactly the transition matrix that matters (first-healthy-tick reconcile, one-shot suppression, unaffected normal down→up, unaffected cold-start-silence), and the `notices.rs` test explicitly checks the "only true once" semantics of `clear_typed_reporting`. The PR body's manual staging repro step is still unchecked — worth confirming before merge since it's the one path (tray process restart mid-outage-then-recovery) that's hardest to unit test.

## Scope

The PR explicitly and reasonably defers Finding 01 (thin-hour lock) and the manual-dismiss backstop to follow-ups, with clear reasoning for each (product judgment needed / new command + UX decision needed). Good judgment call to keep this PR to the actual root cause.

## Risk assessment

Low risk. No schema/migration changes, the new field defaults safely, the decision function is pure and exhaustively tested, and the DB write path (`clear_typed_reporting`) reuses the existing DELETE query unchanged. Nothing here looks blocking.

Akarsh-Hegde:
Addressed the one open item from review — the manual staging-repro checklist item ("restart the tray process while the daemon stays healthy, confirm the banner clears") is the hardest path to verify without a packaged staging build, so I added the closest automatable equivalent instead of leaving it unchecked:

**`poll::refresh::tests::stale_notice_from_a_prior_process_instance_is_actually_cleared`** — seeds a real in-memory, schema-migrated DB with a `tray.daemon_quiet` notice (as a prior process instance would have raised it), then runs the exact `decide_health_notice` → `clear_typed_reporting` → `send_back_online_toast` sequence `refresh_health`'s `reconcile_stale` branch runs, and asserts against the real `system_notices`/`notifications` rows afterward — no mocking either DB call. This is genuinely stronger than the existing pure-logic unit tests since it exercises the actual write path end-to-end, though it's still short of driving a packaged app through a real sleep/restart cycle.

No other findings from review — it was a clean pass ("didn't find a logic gap... low risk... nothing blocking"). Also cleaned up a latent `clippy --tests` warning (test module wasn't at the end of the file — only surfaces under `--tests`, which the plain `cargo clippy` in CI doesn't run, so it was invisible before).

Full suite green: 138 tray tests (1 new) + 720 daemon tests, `cargo fmt`/`clippy` clean on both, full pre-push suite (fmt, ui build, ui tests, clippy, security audit, cargo test) passed on push.

From my side this is ready to merge — the one thing I couldn't do is the literal packaged-staging click-through, so flagging that as the last manual check if you want full confidence before merging.

PR #532, opened 2026-07-21
Rebuild daily summary with modern UI and planned/actual tracking

adityaharishch:
## Summary

Completely rebuilt Meridian's daily summary screen to provide a modern, aesthetic interface that handles both planned and unplanned days. Replaced Vega-based charting with custom SVG components and established a two-layer system where database facts lock task outcomes while AI fills gaps and identifies workstreams.

## Key Changes

**Core Logic**
- Two-layer task matching: database facts (posted worklogs, linked tickets) are deterministic and lock outcomes; AI model fills gaps and points at relevant workstreams
- New `Adherence` type with achievement percentage: `(earned * 100 + planned) / (planned * 2)` with proper rounding
- 30-minute minimum threshold applied consistently across frontend, backend, and themes
- Staleness-triggered auto-refresh (regenerates summaries older than 20 minutes on open)

**Schema & Database**
- Migration 068: dropped `panels_json`, added `headline`, `plan_json`, `adherence_json`, `themes_json`, `evidence_at`
- New types: `DaySummaryInsight` (text + learned flag), `PlanVerdict`, `Adherence`, `DayTheme`
- Prompt rewritten with tone contract (no "drifted"/"failed"/"wasted"), `**emphasis**` formatting, locked pre-matches rule

**UI Components** (all new)
- `AchievementRing`: signature visualization of plan adherence with done/partial/not-touched segments and percentage
- `PlanLedger`: detailed ledger of planned tasks with outcome glyphs, evidence, time, and certainty indicators
- `DayShape`: no-plan hero showing inferred themes as proportional bands
- `Insights`: unlabelled insight rows with learned items in a separate "Picked up along the way" card
- `Workstreams`: task list filtered to ≥30 minutes with proportional bars and duration
- `Composing`: full-screen generation state with animated progress and step captions

**Layout**
- Modal card over dimmed blurred backdrop (house overlay pattern)
- Structured sections with labels and proper spacing using new `Section` component
- Boxed presentation for lists (plan ledger, workstreams); unboxed for prose (insights, hero)
- Escape-to-close and detail dialog accessible from within the card

## Test & Build Status
- `npm run build`: clean
- `bun test`: 332 pass / 0 fail
- No Vega references remain in codebase
- Verified end-to-end on both planned (6→2 done, 2 partial, 2 untouched) and unplanned days

## Commits

- `c679adb0` - feat(summary): rebuild the daily summary around planned vs actual
- `c3ec7791` - feat(summary): give composing its own screen instead of one static line
- `76fe559b` - refactor(summary): put the daily summary in a card with real structure

🤖 Generated with Claude Code


Akarsh-Hegde (changes requested):
Strict review of the full diff (backend adherence/summary logic, worklog escalation, the cross-provider prune fix, and both frontend slices). Confirmed every finding below by reading the actual file on `feat/daily-summary-rebuild`, not just the patch text.

## Blockers / Majors

**1. `src/pm_worklog/escalate.rs:229-249` (`escalate_create`) and `:291-295` (`escalate_match`) — no idempotency guard around the tracker write, so a crash/timeout between the external call and the DB commit causes a duplicate ticket or duplicate comment on retry.**
`create_ticket`/`post_comment` run against the real tracker *before* `graduate_local_task` (the DB transaction that flips the task off `provider = 'local'`) commits. If the process dies, the DB write errors, or the tray's 150s `run_meridian_json` timeout (`tray/src-tauri/src/commands/worklog_generate.rs:33`) fires while the daemon is still finishing, the personal task is still `provider = 'local'` when control returns to the caller. A retry (manual re-click, or client-side retry-after-timeout) re-enters `escalate_create`/`escalate_match` and files a second ticket / posts a second comment for the same work. The module's own doc comment claims "the op is one transaction: it fully lands or not at all" — true only for the DB repoint, not the tracker call that precedes it. This PR already solves the identical crash-window elsewhere (`generate.rs`'s `auto_log_local_targets` uses a claim step via `targets::begin_post`/`revert_post` specifically so "the crash-safety and idempotency guarantees are identical" to approve) — `escalate.rs` needs the same treatment: claim/lock the task before the external call, or make the tracker call itself safely retryable (e.g. check-before-create).

**2. `ui/components/summary/DaySummaryOverlay.tsx:158-160` — Escape closes the whole day-summary overlay even when the nested ticket-detail dialog (`selected`, opened at line 413) is open.**
```js
function onKey(e: KeyboardEvent) { if (e.key === 'Escape') onClose() }
```
There's no check against `selected` before calling `onClose()`. `DayTaskDetailPanel` (rendered when `selected` is set) has no Escape handling of its own, so today: open the day summary → click into a ticket's detail → press Escape expecting to back out of the detail view → the entire summary closes instead. This is a real modal-stacking regression newly introduced by this diff (there was no Escape handling at all before this PR). Fix: `if (selected) { setSelected(null); return }` before `onClose()`.

**3. `DaySummaryOverlay.tsx:241` — `planned` derivation can show the wrong screen (no-plan layout) for a day that *does* have a plan, if the plan was committed after the last compose.**
```js
const planned = (sc?.planned ?? false) && (summary?.plan.length ?? 0) > 0
```
The comment directly above this line says the intent is: "The plan branch comes from the LIVE scalars rather than the stored summary, so a day planned after its summary was composed still gets the right screen." But the actual expression ANDs the live `sc.planned` with the *stored* `summary.plan.length` — so if a plan is committed after the last generate, `sc.planned` is `true` but `summary.plan` is still `[]`, and `planned` evaluates `false`. The UI falls back to the no-plan stat block, which is the opposite of the stated fix. `is_stale` on the Rust side doesn't track plan edits either, so nothing self-corrects until the user manually hits Regenerate. This looks like the AND should be dropped in favor of `sc?.planned ?? false` alone (or the empty-plan case should be handled inside the "planned" branch, not by routing to the wrong branch).

**4. `DaySummaryOverlay.tsx:166-198` — stale-response race: `generate`/`generateNow` can overwrite the currently-displayed day with a late response for a day the user has since navigated away from.**
Both callbacks unconditionally `setSummary(s)`/`setData(...)` when their `invoke` resolves, with no check that `day` is still the day being viewed. If a background compose (staleness auto-refresh, manual Regenerate, or "Generate now") is in flight and the user pages to a different day via `onShiftDay` before it resolves, the late response silently overwrites the new day's state — header shows the new day, body reverts to the old day's content. The initial-load effect one screen down already guards exactly this class of bug with a `cancelled` flag; `generate`/`generateNow` have no analogous guard. This is the same overlapping-stale-response pattern already fixed once in this repo for `ProviderTasks.fetchTasks` (`e5110ba6`) — worth applying the same fix here.

## Design/scalability concern worth a second look

**5. `daily_plan` retention in the 5-way prune fix is unbounded, not scoped to "today's plan."**
All five providers (`src/intelligence/providers/{jira,github,linear,trello}/mod.rs`, `azure_devops/db.rs`) add the identical clause:
```sql
AND task_key NOT IN (SELECT DISTINCT task_key FROM daily_plan)
```
with no `plan_date` bound, and the doc comment above each (`.../jira/mod.rs:312`, `github/mod.rs:108`, `linear/mod.rs:371`, `trello.rs:220`, `azure_devops/db.rs:14`) states this is intentional — "kept forever," mirroring `pm_worklogs` retention. But the bug this fixes is scoped to *today*: "closing a Today's Focus item ... doesn't delete the row its Undo needs." As written, any task ever added to any day's plan, ever, becomes permanently unpruneable in `pm_tasks` for the life of the install, even years after that day and its Undo window are gone — `daily_plan` itself has no TTL/GC (checked `src/migrations/041_daily_plan.sql` and the rest of the plan module; only explicit per-row "remove from today's plan" deletes exist). Worklog-history retention is legitimately permanent (it's real history); plan-membership retention as implemented goes further than the stated justification requires. Recommend scoping the subquery to a recency window (today, or today plus a short trailing window, using local-day bounds) rather than all-time — please confirm if unbounded retention here is actually the intended tradeoff, since the docstring reads like a deliberate choice, but the growth is unbounded and worth being explicit about.

## Minor / must-fix-before-merge (migration content can never be edited again once merged)

**6. `src/migrations/068_day_summary_rebuild.sql:40` — comment is factually wrong about the shipped shape.**
> `` `insights_json` held bare strings; it now holds `{text, learned}` objects. ``
The actual shipped struct (confirmed in `meridian-core/src/readers/day_summaries.rs` and `src/day_summary/generate.rs`'s parse struct) is `DaySummaryInsight { title, text }` — there is no `learned` field anywhere in this PR's insight type. Since migration files can't be edited after merge, this stale comment (left over from an earlier design iteration) will be permanently wrong in history. Please fix before merging.

**7. `src/migrations/068_day_summary_rebuild.sql:23` — `themes_json` comment describes behavior that doesn't exist in this PR.**
"for a day with NO plan: what it turned out to be about, grouped by the model, pointing at real day_task ids" — but `src/day_summary/generate.rs` hardcodes `themes: Vec::new()` unconditionally, and the LLM schema doesn't even offer the model a `themes` field. `themes_json` is dead weight for now (kept for wire back-compat per the generator's own comment) — the migration comment should say so rather than describing live grouping behavior.

## Minor

- `meridian-core/src/readers/board.rs:36-48` — the new plan-snapshot fallback in `provider_for_key` swallows any query error via `.await.ok().flatten()` with no `tracing::warn!` on the failure path, unlike this codebase's usual failure-boundary logging convention. If the snapshot query starts silently failing (bad JSON, pre-044 DB), nobody would notice.
- `src/pm_worklog/escalate.rs` — `graduate_local_task` (the function doing the actual repoint, the riskiest logic in this slice) has no tracing of its own — no span, no per-branch log, none of its 5 queries wrapped the way `day_task_worklogs/targets.rs` wraps its writes. Given finding #1 above, this is exactly the function you'd want to debug in production.
- `meridian-core/src/readers/task_detail.rs:9-12` doc comment states `local_worklog_text`/`local_worklog_posted_at` are "only ever set for `provider = 'local'` tasks" — contradicted by `escalate.rs`'s `graduate_local_task`, which deliberately writes them onto the real ticket's row on both the create-new and match-existing paths. Doc/contract mismatch introduced within this same PR.
- `Cargo.toml`/`Cargo.lock` — the `opentelemetry-otlp` reqwest-client → reqwest-rustls swap and the `rustls-native-certs` addition are unrelated to this PR's stated scope (daily-summary rebuild / worklog escalation / prune fix). Also, the lock diff only *adds* `rustls-native-certs` in two places and removes nothing, which doesn't corroborate the Cargo.toml comment's claim of dropping native-tls/hyper-tls/openssl entirely — worth double-checking that claim, and consider splitting this into its own PR.
- `src/main.rs` (day-summary CLI handler doc comment) and `src/llm/prompts.rs:240-245` (`DAILY_SUMMARY` doc comment) both still describe the pre-rebuild behavior (narrative + Vega-Lite panel fallback / verdict-per-ticket from the model) that this PR removed. Stale docs, not migration files, so fixable anytime, but worth cleaning up now while it's fresh.
- `ui/components/timeline/DatePicker.tsx` (new, ~215 lines) has no live entry point in this PR: it's only reachable through `CleanupCard` → `CleanupOverlay` → `MeridianTimelineShell`'s cleanup modal branch, and that branch plus its only trigger (`OverviewPanel`'s CTA) are both commented out in this same PR (last commit, "temporarily hide the Board Cleanup view"). Not a bug, but this PR ships new interactive UI that can't currently be exercised end-to-end in the running app — worth calling out explicitly so it isn't assumed to be manually verified.
- `MeridianTimelineShell.tsx` / `OverviewPanel.tsx` — commenting out (not deleting) the Board Cleanup entry points leaves `import { CleanupModal } from './CleanupModal'` and the destructured `cleanupIssueCount` (still actively computed in `useTimelineData.ts` on every tasks/integrations change) orphaned. Doesn't break the build today (no eslint installed, no `noUnusedLocals`), but it's dead code/dead computation left behind by the hide.
- `TaskDetailDialog.tsx` — `EscalateResponse` is declared as a local `interface` rather than in `ui/lib/api-types.ts`, diverging from this codebase's established convention (every other command response type lives there; CLAUDE.md's porting playbook explicitly says response types go in `api-types.ts`, "never a route file").
- `DaySummaryOverlay.tsx:138` — `hovered` state is declared and reset but never read anywhere; dead state.
- `DaySummaryOverlay.tsx` — the pre-compose "Generate now" button has no `disabled`/re-entry guard (unlike the header's "Regenerate", which is `disabled={generating}`), so a fast double-click can fire two concurrent `generate_day_summary_now` calls. Low impact since drafting is idempotent and never auto-posts, but inconsistent with the guard one click over.

## Verified clean (called out since this review was asked to be strict — these were real risk areas that held up)

- `adherence.rs`'s achievement-percentage formula, rounding, and division-by-zero guard are correct; the "key the plan list by ticket" fix (`WorkList.tsx` + `resolve_deterministic`) genuinely fixes the described dropped-ticket bug and is covered by tests on both sides.
- The 5-provider prune fix (aside from finding #5's scope) was applied identically and correctly everywhere — no provider missed the shared `mark_retained_offboard` helper or diverged in its query.
- Auto-compose (`day_summary/auto.rs`) is wired so a failing/rate-limited LLM call can't take down the daemon's `run_loop` — every error path is caught and logged, never unwrapped.
- Per-ticket worklog update bodies are one LLM call with a structured multi-target response, not N calls (no cost/rate-limit multiplier), and a missing per-ticket update falls back to the shared body rather than posting nothing.
- No XSS surface, no unjustified `any`, no stray em/en-dashes in user-facing strings, file headers intact across all reviewed files.

🤖 Reviewed by Claude Code (6 parallel agents covering backend core logic, summary generation/auto-compose, the worklog escalation pipeline, the cross-provider retention fix, and both frontend slices — findings above were independently re-verified against the actual files on `feat/daily-summary-rebuild` before being included).

adityaharishch:
Thanks for the strict review - addressed all of it in `f99fe5b1` (plus a merge of `pre-main` in `598abfcc` that resolved the conflicts). Summary by finding:

**Blockers**
1. **escalate.rs idempotency** - the non-idempotent tracker call now runs behind a claim on the personal task row (`escalation_attempt_at` + `escalation_ticket_key`, migration `073`), mirroring the day-task worklog's begin_create/mark_created: a concurrent escalation is refused, a stale claim is re-takeable, and a retry reuses the created/posted key instead of filing a second ticket / posting a second comment. `graduate_local_task` is now `#[instrument]`'d with per-query spans and a summary log, and there are 5 new unit tests (claim won-once/stale-retake/clear, graduate create + match).
2. **Escape closes nested dialog** - Escape now backs out of `selected` first, only closing the overlay when no ticket detail is open.
3. **`planned` derivation** - dropped the `&& summary.plan.length` AND; it's `sc?.planned ?? false` now, so a day planned after compose gets the plan screen (the ledger may lag one recompose, which is the right screen vs. the no-plan one).
4. **stale-response race** - `generate`/`generateNow` now capture the day and bail before every `setState` if the user navigated away (via a `dayRef`), plus a shared `composing` re-entry guard.

**Design concern**
5. **Unbounded `daily_plan` retention** - kept intentional, now documented explicitly in `providers/mod.rs`: scoping to a recency window would break historical plan/worklog title resolution for older tickets, and the growth is human-paced (planned tasks/day), not ticket-churn-driven. Confirmed as the intended tradeoff rather than silently unbounded.

**Migration must-fixes** (done while unreleased - these migrations were also renumbered above pre-main's 068/069, which otherwise collided after the merge and would break the sqlx migrator)
6. `insights_json` comment fixed to `{title, text}`.
7. `themes_json` comment now says it's unused/wire-back-compat only; `plan_json` corrected to binary done/not_touched.

**Minors**
- `board.rs` snapshot fallback now `warn!`s at the failure boundary instead of `.ok().flatten()`.
- `task_detail.rs` doc updated - `local_worklog_text` is also carried onto a real ticket on escalation.
- `EscalateResponse` moved to `api-types.ts`.
- Dead `hovered` state removed; orphaned `CleanupModal` import + `cleanupIssueCount` destructure commented out alongside the disabled feature.
- Cargo.toml native-tls comment corrected (it no longer claims native-tls is purged workspace-wide; it isn't - just kept off our OTLP/reqwest path). Left the dep change in place rather than splitting, since reverting it would reintroduce the native-tls-via-otlp path.
- Stale docs refreshed: `main.rs` day-summary CLI + `DAILY_SUMMARY` prompt doc.
- "Generate now" double-fire is covered by the shared `composing` re-entry guard from #4.

Not changed (informational): `DatePicker.tsx` has no live entry point because Board Cleanup is intentionally hidden this release - flagging as expected, not manually verifiable end-to-end until that re-enables.

All green locally: fmt + clippy (workspace + tray), full `cargo test`, `ui` build, and `bun test` (348 pass). Pre-push suite passed on the push.

PR #540, opened 2026-07-21
fix(windows): resolve HOME via meridian_core::paths, quote OAuth browser-launch URLs, fix duplicate ClerkProvider

Akarsh-Hegde:
## Summary

Three bugs found live-testing the just-shipped `v1.75.0-staging.2` build, two Windows-specific and one platform-agnostic (reproduced live on this dev machine too, running the real staging binary).

### 1. Raw `HOME` reads across `tray/src-tauri`

`HOME` is never set on Windows. `backend_install.rs` already had this fixed earlier in the cycle (`meridian_core::paths::home_dir()`, which falls back to `dirs::home_dir()` -- the correct `%USERPROFILE%`-based resolution), but that migration missed a dozen other call sites still doing `std::env::var("HOME")` directly:

- `install.rs`: `detect_install_mode()`, `canonical_env_path()`, `meridian_db_path()`'s default fallback, and the CLI bin/cwd resolvers (`meridian_bin()`, `cli_cwd()`)
- `commands/account.rs`: `account_path()` -- the signed-in-email persistence
- `commands/integrations.rs`: `home()` -- tracker credential resolution
- `commands/dashboard.rs`, `commands/whats_new.rs`, `commands/diagnostics.rs`
- `analytics.rs`, `poll/whats_new_auto_open.rs`, `poll/plan_auto_open.rs`

On Windows every one of these silently fell through to its no-op/Bare/default path instead of erroring loudly. This is what explains **"Database not found"** on the health banner -- `meridian_db_path()`'s HOME-dependent fallback never resolved a usable path.

Call sites already `#[cfg(target_os = "macos")]`-gated or marked `dead_code` off macOS (`lib.rs`'s Dock-reopen handler, both `app_icons.rs` occurrences) were left untouched -- they're correctly scoped already.

Side note found while chasing this: `detect_install_mode()`'s release-path fallback also walks up to 8 parent directories from the process's cwd looking for a bare `.env` -- on a machine that also has a dev checkout, launching the *packaged* binary from (or near) that checkout directory silently borrows the checkout's dev credentials (dev Clerk key, etc.) instead of the compiled-in production ones. Not fixed here (it's an edge case that doesn't affect real end-user machines, which have no dev checkout), but flagged since it caused a confusing dead-end during this investigation.

### 2. OAuth browser launch truncates the authorize URL on Windows

`meridian-oauth`'s `open_browser()` spawned:

```rust
Command::new("cmd").args(["/C", "start", "", url]).spawn()
```

`Command`'s automatic Windows argument quoting only wraps an argument in `"..."` when it contains a space, tab, or `"`. An OAuth authorize URL has none of those -- only `&`-joined query params -- so it went out **unquoted**. `cmd.exe` then parsed the command line itself and treated every unescaped `&` as a command separator: only the query string up to the *first* `&` (`response_type=code`) ever reached the browser. `client_id`, `redirect_uri`, `scope`, `state`, `code_challenge` -- everything after -- were silently dropped as bogus follow-on "commands".

This is exactly Atlassian's *"We're having trouble logging you in -- the authorize request was incomplete or invalid"* error on Jira sign-in. Fixed by building the raw `cmd /C start "" "<url>"` line with `CommandExt::raw_arg`, so the URL is genuinely quoted and `&` is protected from cmd.exe's tokenizer. The URL can never contain a literal `"` (`encode()` percent-encodes everything outside the unreserved set), so the quoting is always safe.

(GitHub's reported "stuck rotating" Authorize page uses a device-code flow that doesn't go through this function at all, so it's very likely unrelated/a separate GitHub-side issue -- flagged in the original PR description as a guess, now walked back since it doesn't hold up on closer reading of `github.rs`.)

### 3. Duplicate `<ClerkProvider>` in Settings -> Account -- the actual Clerk bug

This is **not Windows-specific** and not a build-config issue at all, despite `ClerkErrorBoundary`'s generic "Sign-in isn't configured for this build" message strongly suggesting one. Root-caused by attaching WebView2's remote debugging (`WEBVIEW2_ADDITIONAL_BROWSER_ARGUMENTS=--remote-debugging-port=9222`) to the actual running `v1.75.0-staging.2` binary and driving it via the Chrome DevTools Protocol. The real thrown error:

```
@clerk/react: You've added multiple <ClerkProvider> components in your React
component tree. Wrap your components in a single <ClerkProvider>.
```

`RequireSignIn` (`app/page.tsx`) wraps the *entire dashboard* -- including Settings -- in one `ClerkGate`/`<ClerkProvider>`, established the moment the dashboard is reachable at all (sign-in is compulsory app-wide). `AccountAuthControl` (Settings -> Account) independently wrapped itself in *another* `ClerkGate`/`<ClerkProvider>`, nested inside the first. `@clerk/react` hard-refuses that combination, `initClerk()`'s promise rejects, and `ClerkErrorBoundary` papered over what was actually a component-tree bug with a message about missing configuration. The setup wizard's `SignInWidget` was unaffected because it runs in a separate, non-`RequireSignIn`-gated window/context -- its `ClerkGate` is the only provider in that tree, which is why sign-in "worked" there and only broke once you opened Settings afterward.

Fixed by having `AccountAuthControl` render directly against the dashboard's already-live session (`useClerk()`/`useUser()` via `useSignedInEmail()`) instead of mounting a redundant provider. The `knownEmail`/`KnownAccountFallback` optimistic-render plumbing existed only to paper over that redundant provider's own async re-init roundtrip, so it's removed along with it -- `AccountStatus`'s existing `!isLoaded` check already covers the (now near-instant, already-warm) loading state. `AccountSection`'s now-unused `signedInEmail` local state and its `get_account_email` prefetch go for the same reason; the `save_account_email`/`clear_account_email` side-effect calls stay (analytics still needs them for the `$set` email upgrade).

### 4. Health banner hardcoded macOS instructions

`commands/health.rs`'s "database not found" message and its `ui/components/HealthBanner.tsx` consumer both hardcoded `launchctl load ~/Library/LaunchAgents/com.meridiona.daemon.plist` regardless of platform. Made the message platform-aware on the Rust side (`start_daemon_hint()`, cfg'd per OS) and had the frontend actually use the backend's message instead of unconditionally overriding it with the macOS default for the "not found" case.

## Verification

- `cargo check --workspace` (root) and `cargo check` (tray/src-tauri, separate workspace): clean.
- `cargo fmt --check`: clean in both workspaces.
- `cargo test -p meridian-oauth`: all 20 tests pass.
- `npx tsc --noEmit` (ui/): no errors in any changed file (pre-existing, unrelated `bun:test` type-decl errors in `__tests__/` are environment noise on this machine, not caused by this change).
- The `ClerkProvider` fix (#3) was reproduced AND verified fixed live, against the actual running staging binary, via WebView2 remote debugging + CDP -- not just inferred from reading code.
- The `HOME` (#1) and OAuth URL (#2) fixes are not yet re-verified against a live Windows 10 install with a fresh build.

## Test plan

- [ ] Cut a new staging build from `pre-main` including this fix.
- [ ] Fresh-install on the Windows 10 machine that surfaced these bugs; confirm the daemon starts and the DB is found without manual intervention.
- [x] Confirm Settings -> Account shows the signed-in state (no Clerk "not configured" error) after visiting the dashboard post-wizard -- verified live on this dev machine against the real staging binary before/after the fix.
- [ ] Retry Jira OAuth sign-in end-to-end; confirm it completes without Atlassian's "incomplete or invalid" error.
- [ ] Retry GitHub OAuth; confirm whether Authorize still hangs (now believed unrelated to this PR's fixes -- needs separate investigation if it persists).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5097,6 +5097,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
+ "dirs 5.0.1",
  "rand 0.8.6",
  "reqwest 0.12.28",
  "serde",
--- a/meridian-oauth/Cargo.toml
+++ b/meridian-oauth/Cargo.toml
@@ -25,6 +25,10 @@ reqwest = { version = "0.12", features = ["json", "rustls-tls"], default-feature
 # the tray stays lean; the daemon's `full` unifies over this.
 tokio = { version = "1.15", features = ["net", "io-util", "time", "sync"] }
 tracing = "0.1.40"
+# Home-directory fallback for store::oauth_dir() when $HOME is unset (Windows).
+# Same crate/version meridian-core's util::paths already uses — adds nothing
+# new to the workspace lockfile.
+dirs = "5"
 # Cross-process advisory file lock for the rotating refresh token uses std's native
 # File::try_lock/unlock (stable since Rust 1.89; toolchain is pinned to 1.93.1), so
 # no external locking crate is needed — see store::lock_provider.
--- a/meridian-oauth/src/flow.rs
+++ b/meridian-oauth/src/flow.rs
@@ -375,6 +375,13 @@ else{document.querySelector('h2').textContent='No token in URL. Try again.';}\
 /// tokenizer the way it would from an interactive prompt. The URL can't itself
 /// contain a `"` — [`encode`] percent-encodes anything outside the unreserved
 /// set — so this quoting is always safe.
+/// The `raw_arg` line itself, pulled out as a pure function so the quoting
+/// shape is unit-testable without mocking `Command::spawn`.
+#[cfg(target_os = "windows")]
+fn windows_start_line(url: &str) -> String {
+    format!("start \"\" \"{url}\"")
+}
+
 #[cfg(target_os = "windows")]
 fn open_browser(url: &str) {
     use std::os::windows::process::CommandExt;
@@ -384,7 +391,7 @@ fn open_browser(url: &str) {
     );
     let result = std::process::Command::new("cmd")
         .arg("/C")
-        .raw_arg(format!("start \"\" \"{url}\""))
+        .raw_arg(windows_start_line(url))
         .spawn();
     if let Err(e) = result {
         tracing::warn!(error = %e, "failed to launch the system browser");
@@ -482,6 +489,32 @@ mod tests {
         assert_eq!(decode("a%20b"), "a b");
         assert_eq!(decode("abc-_123"), "abc-_123");
         assert_eq!(decode("x%2Fy"), "x/y");
+    }
+
+    /// The regression this whole PR exists for: a URL with `&`-joined query
+    /// params must come out fully wrapped in real quotes, so cmd.exe's
+    /// tokenizer can never see an unescaped `&` and split the command line.
+    #[cfg(target_os = "windows")]
+    #[test]
+    fn windows_start_line_quotes_the_whole_url() {
+        let url = "https://auth.atlassian.com/authorize?response_type=code&client_id=abc&redirect_uri=http%3A%2F%2F127.0.0.1%3A9123%2Fcallback&state=xyz";
+        let line = windows_start_line(url);
+        assert_eq!(line, format!("start \"\" \"{url}\""));
+        // The URL's own `&`s must all land strictly inside the quoted span.
+        let open = line.find('"').unwrap();
+        let close = line.rfind('"').unwrap();
+        assert!(
+            close > open + 1,
+            "URL must be wrapped in a non-empty quoted span"
+        );
+        for (i, c) in line.char_indices() {
+            if c == '&' {
+                assert!(
+                    i > open && i < close,
+                    "found an `&` outside the quoted span at byte {i}: {line}"
+                );
+            }
+        }
     }
 
     #[test]
--- a/meridian-oauth/src/flow.rs
+++ b/meridian-oauth/src/flow.rs
@@ -360,15 +360,43 @@ else{document.querySelector('h2').textContent='No token in URL. Try again.';}\
 /// argument is a window-title slot, not part of the command — the empty `""`
 /// there is required, otherwise `start` misreads a quoted URL as the title and
 /// never opens it.
+///
+/// The whole `/C ...` line is built with [`CommandExt::raw_arg`] rather than
+/// passed as normal `args` elements: `Command`'s automatic Windows quoting
+/// only wraps an argument in `"..."` when it contains a space, tab, or `"` —
+/// an authorize URL has none of those, only `&`-joined query params, so it
+/// went out unquoted. `cmd.exe` then parsed each unescaped `&` as a command
+/// separator and only the query string up to the FIRST `&` ever reached the
+/// browser (`response_type=code`, with `client_id`/`redirect_uri`/`scope`/
+/// `state` silently dropped as bogus follow-on commands) — the exact shape of
+/// Atlassian's "authorize request was incomplete or invalid" and GitHub's
+/// consent page hanging on a `redirect_uri`-less Authorize click. Wrapping the
+/// URL in real quotes inside the raw command line protects `&` from cmd.exe's
+/// tokenizer the way it would from an interactive prompt. The URL can't itself
+/// contain a `"` — [`encode`] percent-encodes anything outside the unreserved
+/// set — so this quoting is always safe.
+#[cfg(target_os = "windows")]
+fn open_browser(url: &str) {
+    use std::os::windows::process::CommandExt;
+    tracing::info!(
+        url,
+        "if it doesn't open automatically, open this URL yourself"
+    );
+    let result = std::process::Command::new("cmd")
+        .arg("/C")
+        .raw_arg(format!("start \"\" \"{url}\""))
+        .spawn();
+    if let Err(e) = result {
+        tracing::warn!(error = %e, "failed to launch the system browser");
+    }
+}
+
+#[cfg(not(target_os = "windows"))]
 fn open_browser(url: &str) {
     tracing::info!(
         url,
         "if it doesn't open automatically, open this URL yourself"
     );
-    #[cfg(target_os = "windows")]
-    let result = std::process::Command::new("cmd")
-        .args(["/C", "start", "", url])
-        .spawn();
     #[cfg(target_os = "macos")]
     let result = std::process::Command::new("open").arg(url).spawn();
     #[cfg(not(any(target_os = "windows", target_os = "macos")))]
--- a/meridian-oauth/src/store.rs
+++ b/meridian-oauth/src/store.rs
@@ -44,9 +44,23 @@ impl OAuthTokens {
     }
 }
 
+/// The current user's home directory. `$HOME` is normally unset on Windows,
+/// so `dirs::home_dir()` (the platform profile-directory API, `%USERPROFILE%`
+/// there) is the fallback — the same resolution order as
+/// `meridian_core::paths::home_dir()`, duplicated here rather than taken as a
+/// dependency: `meridian_core` pulls in sqlx/libsqlite3-sys for its DB layer,
+/// which this config-free, daemon-and-tray-shared crate deliberately has no
+/// reason to carry for one directory lookup.
+fn home_dir() -> PathBuf {
+    std::env::var_os("HOME")
+        .filter(|h| !h.is_empty())
+        .map(PathBuf::from)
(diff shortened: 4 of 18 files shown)


Akarsh-Hegde:
## Review: #540 — Windows HOME resolution + OAuth browser-launch quoting

### Overview
Solid, well-documented fix for two real, silent Windows bugs: (1) consolidating raw `std::env::var("HOME")` reads onto `meridian_core::paths::home_dir()` across `tray/src-tauri`, and (2) fixing `cmd.exe` argument-splitting on `&` in the OAuth authorize URL via `CommandExt::raw_arg`. I verified the `open_browser` fix is correct: `Command`'s automatic Windows quoting only kicks in on space/tab/`"`, so the unquoted URL really did get truncated at the first `&` by `cmd.exe`'s tokenizer, and wrapping it in a real `"..."` inside a `raw_arg` line fixes that. I also confirmed every interpolated value that reaches `open_browser`/`run_fragment_relay_flow` (state, PKCE challenge, redirect_uri, scope, Trello key) goes through a percent-encoder first, so the "URL can never contain a literal `\"`" safety claim in the doc comment holds.

I spot-checked the `home_dir()` swap across all 9 files the PR body lists (`analytics.rs`, `account.rs`, `dashboard.rs`, `diagnostics.rs`, `health.rs`, `integrations.rs`, `whats_new.rs`, `install.rs`, `poll/plan_auto_open.rs`, `poll/whats_new_auto_open.rs`) — all correctly swapped, and the `#[cfg(target_os = "macos")]`-gated / `dead_code` sites in `lib.rs` and `app_icons.rs` that the PR says were deliberately left alone are indeed correctly scoped.

### Issue: `meridian-oauth/src/store.rs::oauth_dir()` still reads raw `HOME` — likely still broken on Windows

```rust
fn oauth_dir() -> PathBuf {
    let home = std::env::var("HOME").unwrap_or_else(|_| ".".to_string());
    PathBuf::from(home).join(".meridian").join("oauth")
}
```

This is the single choke point for every OAuth token read/write (`store::path/load/save/lock_provider`) — i.e. exactly the code that runs right after the browser-launch fix in this same PR gets a user through Jira's "Authorize" click. It isn't in the PR's list of already-correctly-scoped call sites, and it isn't `cfg`-gated to a platform. On Windows, `HOME` is normally unset (as the PR description itself explains at length for the other call sites), so this falls through to `"."` — the token file lands at `./.meridian/oauth/jira.json` relative to whatever the daemon/tray's current working directory happens to be, not `%USERPROFILE%\.meridian\oauth\jira.json`. That's a different location than `commands/integrations.rs::home()` (now fixed) and everything else expects, and it isn't guaranteed stable across restarts. Net effect: a user could complete the OAuth dance (now that the URL truncation is fixed) and still end up with tokens that don't get found on next launch, or get written somewhere unexpected.

`meridian-oauth` doesn't currently depend on `meridian_core` (checked `Cargo.toml`), so this isn't a one-line `s/env::var/meridian_core::paths::home_dir()/` — it needs either that dependency added, or the same `$HOME`-then-`dirs::home_dir()` fallback duplicated locally in `store.rs`. Given this PR's whole premise is "Windows HOME resolution was missed in a bunch of places," this looks like the one spot most directly in the OAuth flow this PR is fixing.

### Minor / nits

1. **`install.rs::meridian_db_path()`** still builds the default path with a hardcoded separator:
   ```rust
   let home = meridian_core::paths::home_dir_or_cwd();
   let p = format!("{}/.meridian/meridian.db", home.display());
   ```
   The sibling fix a few lines up in the same file (`meridian_bin()`) was specifically changed from string-concat to `home.join(rel)` to avoid a hardcoded `/`. This one line reverts to `format!` with `/`. Forward slashes are generally tolerated by Windows path APIs so this probably isn't a functional bug, but it's inconsistent with the PR's own stated rationale and with `paths.rs`'s "join via `Path::join`, not `format!`" guidance. Same pattern also appears in `health.rs::check_a11y_trusted`'s `format!("{}/.meridian/logs/...", home)`, though that path is `#[cfg(not(feature = "capture"))]` and likely dead in production builds.

2. **Redundant text in the health banner.** `check_database()`'s error is now `"Database not found — start the daemon: {hint}"`, and `HealthBanner.tsx` renders both `bannerTitle = "Database not found"` *and* this full string as the detail line underneath — so "Database not found" is shown twice. Previously the detail only showed "Start the daemon: …" without repeating the title. Worth trimming the redundant prefix on one side.

3. **Em dash in user-facing string.** `"Database not found — start the daemon: ..."` uses an em-dash, which the repo's CLAUDE.md hard rule says should be a plain hyphen in user-facing text. Pre-existing (not introduced by this PR), but this PR touches that exact line and now also surfaces it as raw, unstyled frontend text (previously it was only ever shown behind the frontend's own hardcoded macOS string) — free win to fix while it's already being edited.

4. **`start_daemon_hint()` on Windows** says "run the \"Meridian Daemon\" task in Task Scheduler, or restart Meridian." `backend_install::register_service` has a fallback (`install_startup_folder_launcher`) for machines where `schtasks /Create` is blocked by policy — on those, no such Task Scheduler entry exists. Not a regression since this whole message is new, but the hint may not always apply; consider "restart Meridian" as the Windows-universal instruction.

5. **cmd.exe `%VAR%` expansion** (informational, not a real risk today): quoting protects against `&`/pipe/redirection splitting, but `cmd.exe` still expands `%NAME%` pairs inside quoted segments. Every value reaching `open_browser` today is hex-percent-encoded by a codebase-controlled `encode()`, so this isn't currently exploitable — just flagging in case a future caller ever feeds a less-constrained string through this path.

### Test coverage
`cargo test -p meridian-oauth` (20 tests) covers the encoder and URL-building logic, but there's no test exercising the actual `raw_arg` command-line string built by `open_browser` on Windows (understandably hard to unit test a `spawn()` call directly — could extract the `format!("start \"\" \"{url}\"")` line-building into a small pure function for the string-shape assertion, `cfg`-independent, if you want coverage here without mocking `Command`).

### Summary
The core Windows fixes (HOME consolidation, `raw_arg` quoting) are correct and well-reasoned. The main thing I'd want addressed before merge is `meridian-oauth/src/store.rs::oauth_dir()` — it's the one HOME-dependent call site in the exact flow this PR is fixing that was missed, and it'll likely produce a confusing "tokens not found" symptom on Windows immediately after the browser-launch fix starts working. The rest are minor polish items.

Akarsh-Hegde:
Addressed the review comments and resolved the merge conflict against `pre-main` — ready to merge.

**Review comments (all addressed):**
- `meridian-oauth/src/store.rs::oauth_dir()` was still reading raw `HOME` (unset on Windows), the one HOME-dependent call site in the OAuth flow this PR fixes that got missed. Added a local `home_dir()` (`$HOME` then `dirs::home_dir()`, matching `meridian_core::paths::home_dir()`'s resolution order) rather than taking a `meridian_core` dependency, since that would pull in sqlx/libsqlite3-sys for one directory lookup in this deliberately dependency-light crate.
- `install.rs::meridian_db_path()`'s default-path fallback now uses `PathBuf::join` instead of a hardcoded `/` in a `format!`, consistent with the sibling `meridian_bin()` fix.
- Trimmed the redundant "Database not found" (was showing as both the banner title and, verbatim, the detail line) and swapped the remaining em dash for a hyphen.
- `start_daemon_hint()`'s Windows message no longer names the "Meridian Daemon" Task Scheduler entry, which doesn't exist on machines where `register_service` fell back to a Startup-folder launcher — collapsed to the same "restart Meridian" hint every other non-macOS platform uses.
- Extracted the `open_browser` string-building into a pure function and added a regression test — though see below, this got superseded by a better fix during the merge.

**Merge conflict:** `meridian-oauth/src/flow.rs` conflicted with `pre-main`, which had independently picked up an `explorer.exe`-based fix for the exact same `open_browser` truncation bug (via #539/#541's CodeRabbit pass) instead of my `cmd /C start` + `raw_arg` quoting. `explorer.exe` is strictly better here — it takes the URL as a literal argument with no shell re-parsing at all, so it's immune to `&`, `%`, `^`, and every other cmd.exe metacharacter, not just the `&` this PR set out to fix. Took `pre-main`'s version and dropped mine (including its now-orphaned unit test, since there's no comparable pure function left to test once the shell is out of the picture entirely).

**Verified after resolving:**
- `cargo check --workspace` (root) and `cargo check` (tray/src-tauri): clean.
- `cargo fmt --check`: clean in both workspaces.
- `cargo test -p meridian-oauth`: 20/20 pass.
- All 5 CI checks green on the merge commit (`Rust (macOS Apple Silicon)`, `UI`, `Detect changes`, `screenpipe MIT pin`, `Migrations append-only`).
- `mergeable_state: clean`.

PR #545, opened 2026-07-22
feat(observability): centralised error observability for packaged installs

Akarsh-Hegde:
## What & why

Gives packaged Meridian installs a path to **auto-ship redacted, error-only telemetry to a central, team-visible OpenObserve** (via an OTel Collector gateway), so we can debug production without asking every user to hand-export a diagnostics bundle. Reuses the existing store-and-forward spool (`pending → shipper → sent/quarantine`, retention, retry) — the genuinely new parts are a **central endpoint, distributable auth, and on-device redaction**, not a policy flip.

Local capture and `meridian logs` stay **full-fidelity** throughout — redaction produces a separate stripped copy for egress only. The whole thing **fails safe**: with no endpoint/token/DSN baked in, all shipping is inert and only local capture runs.

## Architecture — two shipping modes (`src/observability/otlp_target.rs`)

| | Dev / Bare checkout | Canonical (packaged) install |
|---|---|---|
| Ships when | `otlp_enabled` + OO creds | `error_reporting_enabled` (opt-out, **on by default**) |
| Target | the engineer's own local OO | central gateway (release-baked endpoint) |
| Auth | `Basic base64(email:pass)` | `Bearer <write-only token>` (release-baked) |
| Redaction | none (full fidelity) | **error-only + allowlist** |

The central endpoint + token are **release-injected `option_env!` constants** (Dayflow-style), inert in source builds so a dev build can never ship to prod by accident. `ship_one` now takes a scheme-aware `Authorization` header (Basic or Bearer).

## The privacy boundary — `src/telemetry_spool/redact.rs`

Runs on the **ship leg only**, immediately before egress. Two fail-closed rules:

1. **Error-only** — WARN+ log records and ERROR-status spans; everything else dropped (this is also where the high-volume `INFO`/`SPAN` records carrying OCR / a11y / LLM I/O get dropped wholesale). A tight `NOISE_SUBSTRINGS` denylist additionally drops known-benign WARN repeaters that would flood the backend.
2. **Allowlist by value-type + key** — numeric/bool always kept (can't hold content); strings only if the key is on `SAFE_STRING_KEYS`; bytes/array/kvlist dropped. Free-text fields (`*.message`, stacktraces, log body, **span `status.message`**) additionally get URL/email/token scrubbing + a length clamp. Undecodable payloads are **never** shipped raw (quarantined).

> **Empirical finding that shaped this:** probing the real spool showed **0 ERROR-status spans / 18,785**; the error signal is **~10% of log records (WARN+), correlated by `trace_id`**. So dashboards/gateway target error **logs**, not traces. The trace-redaction path stays (correct, cheap, future-proof) but ships near-empty today.

## Consent — opt-out, on by default

Matches the reference apps (screenpipe, Dayflow): `error_reporting_enabled` defaults **true**. Surfaced as a **disclosure note + off-switch** in Settings → Capture & Privacy → Error reporting (it was drafted into Advanced, but that tab is hidden on `pre-main` — an opt-OUT default with an unreachable off-switch is not consent, so the merge moved it next to the other "what leaves my machine" controls), plus a first-run note on the setup wizard's Completion screen ("Meridian sends error reports to help fix bugs (change in Settings)"). Deliberately does **not** claim the reports are anonymised — they are redacted, not anonymous (that's a later phase). The `MERIDIAN_TELEMETRY_DISABLED` hard kill still stops all capture.

## Gateway ops stack (`ops/central-observability/`)

Deployable turnkey stack for the central instance, **separate** from the per-developer local OO (`scripts/install-openobserve-daemon.sh`):
- `docker-compose.yml` — OpenObserve + OTel Collector (contrib) + Caddy (auto-TLS).
- `otel-collector-config.yaml` — bearer-token auth on the OTLP receiver (validates the rotatable write-only ingest token), basic-auth export to OO (root creds live only here).
- `Caddyfile` + `.env.example` — TLS reverse proxy + config template. Ops guidance is in file comments (no standalone README, per repo convention).

## Tray coverage (Phase 2)

- **2A** — `meridian::observability::init("meridian-tray")` is now **unconditional** (was a dev-only `otel` feature), so the tray's logged errors reach the spool → central pipeline. The tray was otherwise dark in release, and it's the process the user actually clicks. No new deps (OTel machinery already linked via the always-present `meridian` dep).
- **2B** — `tray/src-tauri/src/crash.rs` adds `tauri-plugin-sentry` 0.5 + re-exported `sentry-rust-minidump` for the crash class OTLP can't see (segfault, panic-before-init, OOM), with fingerprint grouping. Gated **identically** to central shipping (consent + kill-switch + baked DSN, gate at *init* not by muting). `before_send` **reuses the same `scrub_text`** as OTLP, and the `sentry` dep omits the `tracing` feature so logged errors aren't double-reported. DSN release-injected (`MERIDIAN_SENTRY_DSN`), inert without it.

## Release injection (`.github/workflows/release-build.yml`)

Both build steps (macOS + Windows) bake in, from repo **variables**:
- `MERIDIAN_CENTRAL_OTLP_ENDPOINT` (daemon, `option_env!`)
- `MERIDIAN_SENTRY_DSN` (tray, `option_env!`)

The write token (`MERIDIAN_CENTRAL_TELEMETRY_TOKEN`) is now wired in too, as a **secret** rather than a variable — unlike the endpoint and the DSN it is the one credential a shipped install carries, so it must not surface in logs or the Actions UI. Unset still means inert: `resolve_otlp_target` needs endpoint AND token together.

## Verification

- `cargo test` — 38 telemetry tests incl. redaction (allowlist / error-only / free-text scrub / clamp / noise-drop / status.message).
- fmt + clippy clean across the daemon; the tray isn't in pre-push (only the `meridian` package is), so it was validated directly: `cargo check` + `clippy` + `fmt -p meridian-tray` under the real `capture` feature.
- Redaction allowlist reviewed against real spool content.

## Deferred activation (infra-gated — separate follow-up)

1. Deploy `ops/central-observability/` on a Linux VM (creds + `INGEST_TOKEN` + domains + DNS + `docker compose up -d`).
2. Repo variable `MERIDIAN_CENTRAL_OTLP_ENDPOINT` = `https://<gateway>/v1/traces`.
3. Repo secret `MERIDIAN_CENTRAL_TELEMETRY_TOKEN` (= `INGEST_TOKEN`). The build steps already read it.
4. Create a Sentry project → repo variable `MERIDIAN_SENTRY_DSN`.
5. Cut a release; confirm redacted error logs in OO + a test crash in Sentry.

## Known follow-ups

- Real **anonymisation** of shipped error text (v1 is redacted, not anonymous).
- Sentry **release-health sessions** (`sentry::start_session()`; 0.42 dropped the `auto_session_tracking` option).
- Optional **server-side redaction** second pass at the gateway (on-device allowlist is the sole net today, by design).


Akarsh-Hegde:
## Review

### Overview
Adds a second telemetry-shipping mode: packaged installs now auto-ship a redacted, error-only OTLP stream to a central, team-visible OpenObserve gateway (Bearer auth, release-baked endpoint/token), plus native-crash capture via Sentry/minidump in the tray. Reuses the existing spool/shipper. New redaction module (`src/telemetry_spool/redact.rs`) is the stated privacy boundary. Includes a turnkey `ops/central-observability/` Docker stack, CI env wiring, and a Settings toggle (opt-out, default on).

I verified the `tauri-plugin-sentry`/`sentry-rust-minidump` re-exec mechanics and the `sentry-contexts` source against the actual upstream crates (not just the in-repo comments) to check the claims in this PR's docstrings.

---

### 🔴 Critical: the redaction boundary does not cover the machine hostname — leaks a likely-identifying string to the central backend

Both new egress paths ship the raw OS hostname, and neither is scrubbed:

1. **OTLP path.** `src/observability/mod.rs` (pre-existing, ~line 295-301) unconditionally attaches a resource attribute:
   ```rust
   KeyValue::new("host.name", gethostname::gethostname().to_string_lossy().into_owned())
   ```
   `redact.rs`'s `SAFE_STRING_KEYS` allowlists `"host.name"` and only runs it through `scrub_paths` (which only strips `/Users/<name>` style path segments) — a bare hostname like `akarshs-macbook-pro.local` passes straight through untouched.

2. **Sentry path.** `tray/src-tauri/src/crash.rs`'s `scrub_event` only touches `event.message` and each `exception.values[].value`. It never clears `event.server_name`. I confirmed against `sentry-contexts` (the crate `sentry`'s `"contexts"` feature pulls in, which this PR's `Cargo.toml` explicitly enables) that its `ContextIntegration` **auto-populates `server_name` from `hostname::get()`** whenever the app hasn't already set one — which this PR doesn't. `send_default_pii: false` does **not** cover this field; it's a separate integration.

On macOS, the Unix hostname is very commonly auto-derived from the account holder's real name during setup (`Akarshs-MacBook-Pro.local`, `johns-imac.local`, etc.). This PR's entire premise is "redacted, error-only, on-device redaction, reviewed against real spool content" and the in-app consent copy explicitly promises stripped content ("File paths, URLs, emails, and captured content are stripped on your device first"). Shipping the literal device name for every WARN+ log/ERROR span, from every default-opted-in user, to a shared team-visible backend, contradicts that. Before this PR, `host.name` only ever reached a developer's *own* local OO (dev-gated); this PR is what turns it into default-on, team-shared egress — so the risk is created here even though the vulnerable line predates the PR.

**Fix:** drop `host.name` from `SAFE_STRING_KEYS` (or scrub/hash it) in `redact.rs`, and add `event.server_name = None;` (or a hash) in `crash.rs::scrub_event`. `service.instance.id` already exists for correlation, so `host.name` isn't even needed downstream.

---

### 🟡 Design / correctness

- **`scrub_text`'s "long opaque token" regex can mangle ordinary file paths in free-text fields.** The blob regex `[A-Za-z0-9+/=_\-]{32,}` includes `/`, so a typical Rust error like `"failed to open /Users/<user>/Documents/Meridiona/meridian/src/telemetry_spool/redact.rs"` — after the home-dir scrub — still has runs like `Documents/Meridiona/meridian/src/telemetry_spool/redact` (56 chars, no `.` to break it) that exceed 32 chars and get replaced with `<redacted>`. This only affects `error.message`/`exception.stacktrace`/`status.message`/log bodies (not `code.filepath`, which uses the lighter `scrub_paths`), so it's a debuggability regression rather than a leak — but it will make a good fraction of real error messages far less useful once shipped. Worth tightening (e.g. require the blob to *not* contain `/`, or apply the blob regex only to whitespace-delimited tokens).

- **`NOISE_SUBSTRINGS` is a hardcoded string duplicate of a log message defined elsewhere** (the worklog "provider not configured" WARN). There's no test or type-level coupling between the two — if that message's wording changes in the worklog pipeline, this denylist silently stops matching and the noisy WARN starts flooding the central backend again, with no compiler or test signal. Low cost now (one entry), but worth gating on a stable marker/error code instead of message text, or at least cross-referencing the source log call site in a comment.

### 🟡 Documentation drift

`CLAUDE.md` still says: *"Dev-only `--features otel` on the tray exports spans to OpenObserve (`service.name = meridian-tray`) — release builds omit it."* This PR removes the `otel` feature entirely and makes tray telemetry unconditional in all builds — the opposite of what's documented. Given the repo's own conventions require keeping `CLAUDE.md` in sync with observability behavior, this should be updated as part of this PR.

### 🟢 Verified as correct (no issue, despite looking suspicious at first glance)
- The minidump reporter re-exec: checked `sentry-rust-minidump`'s actual source — `MinidumperChild::spawn()` internally detects and takes over the crash-reporter child process, so `crash::init_client()` being called first in `run()` (before any Tauri setup) is sufficient; no extra "am I the reporter process" check is needed here.
- `RuntimeSettings::error_reporting_enabled` backfill for existing `settings.json` files: confirmed the struct-level `#[serde(default)]` plus the `Default` impl correctly yields `true` for upgraders, as claimed.
- CI env wiring (macOS + Windows) is duplicated but internally consistent; secret vs. variable classification (token=secret, endpoint/DSN=variable) is correctly reasoned.

### Minor
- Staging and production DMG builds share the same `MERIDIAN_CENTRAL_OTLP_ENDPOINT`/`MERIDIAN_SENTRY_DSN` — staging test traffic and real user error data will land in the same central OO/Sentry project with no separation. May be fine for a small team, but worth an explicit acknowledgment.
- `resolve_otlp_target()` now loads `settings.json` on every shipper tick unconditionally (previously skipped entirely on canonical installs before the central path existed). Necessary for the feature, negligible cost, just noting the behavior change.

### Test coverage
Redaction unit tests are solid for what they cover (severity filtering, status filtering, allowlist, path scrub, noise-drop, undecodable handling) — but there's no test exercising a *free-text field containing an embedded path*, which is exactly the scenario in the blob-regex finding above.

### Security
Bearer-token gateway auth, root creds isolated to the collector, no exposed OpenObserve port, CR/LF header-injection guards on both token and email/password — all good. The `.env.example`'s `$`-escaping warning for the bcrypt hash is a nice catch of a real Compose footgun.

---

**Bottom line:** solid architecture and mostly careful engineering, but the redaction boundary — the one thing this PR is supposed to guarantee — has a real gap (hostname) that undercuts the privacy claims made to users in the consent copy. I'd block merge on the `host.name`/`server_name` fix; the rest are non-blocking but worth addressing.

Akarsh-Hegde:
## Review addressed — ready to merge

All findings resolved in `05aa3138`. CI green, branch `CLEAN`. Two places where I deliberately diverged from the review's suggested fix are called out below with reasoning.

---

### 🔴 Blocker — hostname leak on both egress paths: **fixed**

Confirmed both halves of the finding against the code:
- `"host.name"` was on `SAFE_STRING_KEYS` (redact.rs:90) and only got `scrub_paths`, which strips `/Users/<name>` segments and leaves a bare hostname untouched.
- `scrub_event` touched only `event.message` and `exception.values[].value` — never `server_name`, which `sentry-contexts` auto-populates from `hostname::get()`.

**Diverged from the suggested fix: pseudonymised rather than dropped.** The review's rationale for dropping was that "`service.instance.id` already exists for correlation" — that turns out not to hold. `service.instance.id` is on the allowlist but is **never populated anywhere** (`grep -rn "service.instance.id" src/ tray/src-tauri/src/` returns only the allowlist entry). The only resource attribute that identifies a machine is `host.name` itself, so dropping it would make *"one user hitting this 500 times"* indistinguishable from *"500 users hitting it once"* — which is most of the value of a central error backend.

So `redact::pseudonymize_host` replaces the value with a salted, truncated SHA-256. Grouping survives, the identifying string does not. Both egress paths call the same helper, so a crash in Sentry and an error log in OpenObserve resolve to the same machine and stay correlatable.

Stated plainly, because it matters for the consent copy: this is **pseudonymous, not anonymous**. Someone who already knows a candidate hostname can confirm it by hashing. That's consistent with the PR's existing position (redacted, not anonymous); true anonymisation stays a later phase. The raw hostname is untouched in the local spool, so `meridian logs` still shows a developer their own machine.

### 🟡 `scrub_text` blob regex mangling paths: **fixed**

Reproduced exactly as described — `failed to open /Users/<user>/Documents/…/redact.rs` collapsed to `failed to open /Users/<user><redacted>.rs`.

**Diverged from the first suggested fix, and this one is worth flagging as a trap.** "Require the blob to not contain `/`" looks equivalent to the token-wise option and is not — it silently introduces a leak:

```
input : auth failed: aB3xY9kLmNp2QrS7t/UvW1zA4bC6dE8fG0hIjK5lMnO=
no-`/`: auth failed: aB3xY9kLmNp2QrS7t/UvW1zA4bC6dE8fG0hIjK5lMnO=   ← unredacted
```

A 44-char base64 secret with one mid-token slash splits into two ~22-char runs, neither reaching the 32-char threshold, so the whole secret ships in clear. Roughly half of slash-containing base64 lands that way. That trades a debuggability bug for a leak in the module whose entire job is not leaking.

Went with the second suggestion instead — apply the rule per whitespace-delimited token, skipping path-shaped ones, keeping `/` in the character class so secrets still die whole. Residual documented in the code: a base64 blob that happens to begin with `/`, or whose final segment contains a `.`, is treated as a path and survives. Accepted because this rule is defence-in-depth; the actual boundary is the error-only filter plus the key allowlist.

### 🟡 Test coverage gap: **fixed**

Three new tests, covering exactly the scenario identified as missing plus its counterpart:
- `free_text_keeps_embedded_paths` — absolute, relative, and extension-less directory paths survive, while the home-dir username is still scrubbed.
- `free_text_still_redacts_every_secret_shape` — four shapes die, including both slash cases above.
- `host_name_is_pseudonymised_not_shipped_raw` — raw hostname never egresses; pseudonym is stable across calls (grouping survives restarts) and distinct per machine.

### 🟡 `NOISE_SUBSTRINGS` prose coupling: **documented, not restructured**

Comment now names the source `tracing::warn!` and states the failure mode explicitly — rewording that log silently resumes the flood, with no compiler or test signal. Deferred the durable fix (matching a stable marker/error code) because it means threading an identifier through the log call site, and there's exactly one entry today. Flagging it rather than pretending it's solved.

### 🟡 `CLAUDE.md` drift: **fixed, and there was more of it**

The flagged `--features otel` line is corrected. Two further sites said the opposite of what this PR does:
- The Environment Variables table described `MERIDIAN_OTLP_ENDPOINT` as *"a Canonical/packaged install never ships"*.
- The Observability section header and body both asserted *"Shipping to OpenObserve is dev-only"* and that the shipper *"refuses to do so entirely for a Canonical install, regardless of settings"*.

All three updated, plus the two redaction guarantees (error-only egress, hostname pseudonymisation) documented so they're harder to silently regress.

### Minor — staging/prod sharing one backend: **fixed rather than acknowledged**

The review suggested an explicit acknowledgment. Fixed instead, because the first end-to-end test of this feature *is* a staging release, and its synthetic errors would land in the same OpenObserve streams and Sentry project as real user data with no way to filter them apart — in a system whose whole value is signal-to-noise on rare errors.

`deployment.environment` was already on the allowlist but never populated. It's now set from the release workflow's job-level `MERIDIAN_CHANNEL`, which already exists for both the macOS and Windows build jobs — so this needed **no workflow change**, only the resource attribute. Source builds report `dev`.

### Minor — `resolve_otlp_target()` loading settings every tick

Acknowledged, no change. Necessary for the feature and negligible cost, as noted.

---

### Verification

- `cargo fmt` + `cargo clippy -- -D warnings` + `cargo test` clean.
- `meridian-tray` verified separately (it isn't covered by pre-push): fmt, clippy, and 161 tests green.
- Redaction suite now 15 tests, all passing.

### Note for whoever merges

The gateway is deployed and verified end-to-end — a test OTLP record traverses TLS → Caddy → bearer auth → collector → basic auth → OpenObserve and is visible in the UI. `MERIDIAN_CENTRAL_OTLP_ENDPOINT`, `MERIDIAN_CENTRAL_TELEMETRY_TOKEN`, and `MERIDIAN_SENTRY_DSN` are all set on the repo, so the first release cut after this merge will begin shipping. Worth watching that release for: only WARN+ arriving, `host.name` appearing as a 16-hex pseudonym rather than a device name, and `deployment.environment` correctly distinguishing staging from prod.

PR #562, opened 2026-07-23
feat(setup): show a founder's welcome note with a live download number

adityaharishch:
## Summary
- The setup wizard's Welcome screen now shows a short note that names which downloader you are ("You're the 214th person to bring Meridian into their day...") - the goal is to make the very first screen feel personal, not like a tool booted up.
- The number is the sum of `download_count` across every asset on the latest GitHub release, read from the public GitHub Releases API - the same source `get_version` already reads. No Resend, no companion website endpoint, no secrets: Meridian's DMG/installer are published as GitHub release assets and the website's `/dl` redirect already points there, so GitHub is already counting every download for us.
- Fails silently: if GitHub is unreachable and there's no cached count yet, the card just doesn't render - this is a warm decoration, never something setup depends on.

## Design notes
- No email/sign-in required to show this - it renders on the very first Welcome screen, before the wizard steps begin.
- Count is cached process-wide in the tray for 5 minutes.
- (Earlier revision of this PR sourced the count from a Resend audience via a new website endpoint - dropped once Resend was ruled out as a dependency, in favor of this simpler zero-secret approach.)

## Test plan
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings`
- [x] `cargo build` / `cargo check` on the tray crate
- [x] `npm run build` (ui) - static export builds clean, `tsc --noEmit` shows no errors in changed files
- [x] Full pre-push suite passed (fmt, clippy, ui build/tests, security audit, cargo test)
- [ ] Manual check in a packaged/dev build: open the setup wizard and confirm the welcome-note card renders with a real number

### DIFF
--- a/tray/src-tauri/src/commands.rs
+++ b/tray/src-tauri/src/commands.rs
@@ -16,6 +16,8 @@
 //!   (also the local read path for `meridian logs`'s OTel-spool decode — see
 //!   `src/telemetry_spool/render.rs` — since the old JSONL-tailing `logs`
 //!   module and its dashboard consumer are both gone).
+//! - [`downloads`] — live Meridian download count for the setup wizard's
+//!   Welcome screen (`meridiona.com/api/downloads-count`).
 //! - [`integrations`] — which trackers are connected (`/api/integrations`).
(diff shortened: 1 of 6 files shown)


adityaharishch:
Dropping this feature per request.

PR #563, opened 2026-07-23
fix(windows): auto-restart a daemon that's already down at tray startup

Akarsh-Hegde:
## Problem

Meridian repeatedly showed "Meridian is offline / Not recording - restart the daemon" with no automatic recovery.

Two coupled gaps, both Windows-specific in practice:

- On machines where `schtasks /Create` is blocked by policy at install time, the daemon is launched only by a Startup-folder VBScript that fires once at login. Nothing supervises it afterward, so any exit (sleep/resume, VM suspend, a crash) leaves it down until the next login.
- The tray is meant to be that backstop: `refresh_health` restarts the daemon on a confirmed outage. But the restart was gated behind `notify_down`, which requires `daemon_was_healthy`. That flag only flips true after the tray observes a healthy tick, so a daemon that was already down when the tray started never triggered the restart - the user was stranded on the manual "Restart daemon" button.

This completes the prior work on this branch (02785320), which added the restart call and the `spawn_staged_daemon` fallback but left the trigger gated off for exactly this cold-start-down case.

## Fix

Decouple the recovery action from the went-quiet notification in `tray/src-tauri/src/poll/refresh.rs`:

- New `attempt_restart` decision fires one restart on the 2nd consecutive confirmed-down poll (~2 min), regardless of `daemon_was_healthy`.
- The "went quiet" notification keeps its cold-start silence (unchanged).
- Fires once per down-episode (the failure counter only equals 2 on a single tick), so no restart loop.

## Testing

- Added `cold_down_daemon_is_restarted_even_though_notice_stays_silent` plus assertions on the existing decision tests.
- `cargo test -p meridian-tray --no-default-features --lib poll::refresh` - 6 passed.
- Full `.githooks/pre-push` suite passed (fmt, UI build, clippy, UI tests, cargo test, security audit).

Diagnosed live on the affected Windows/Parallels machine: the daemon binary runs fine when launched; the problem was purely the missing supervision plus the gated auto-restart.


adityaharishch (commented):
Reviewed the diff on `tray/src-tauri/src/poll/refresh.rs`. Overall the decoupling of `attempt_restart` from `notify_down` is the right shape — it correctly identifies that `daemon_was_healthy` should gate the *notification* debounce, not the *recovery* action, and the `notify_down ⇒ attempt_restart` invariant holds (both key off `second_failure`, and `notify_down` additionally requires `daemon_was_healthy`), so reusing `restart_result` instead of calling `restart()` twice is a nice cleanup on top of the fix.

One correctness/completeness gap, though, in exactly the scenario this PR targets:

**Cold-start restart failure is permanently silent.** `attempt_restart` only fires once per down-episode (`consecutive_health_failures == 2`), and in the cold-start case (`daemon_was_healthy == false`) `notify_down` never fires at all — so if `daemon_control::restart()` fails (e.g. on Windows, `schtasks /Run` fails *and* the `spawn_staged_daemon()` fallback also fails because there's no staged binary at `~/.meridian/bin/meridian.exe`), the user gets:
- no retry (the counter keeps climbing past 2, so `attempt_restart` never re-fires),
- no notice/toast (that requires `notify_down`, gated on `daemon_was_healthy`),
- nothing in the UI at all, ever, for that process lifetime.

This is precisely the "`schtasks` blocked by policy, tray is the sole supervisor" machine the PR describes. Pre-PR, this case also had no restart and no notice, so it's not a regression, but the PR's stated goal is "no more indefinite stranding on the manual restart button" — and a failed auto-restart in the cold-start path still leaves the user indefinitely stranded, just now without even the option of noticing a banner. Worth considering: raise a (rate-limited/one-shot) notice on `attempt_restart` failure regardless of `daemon_was_healthy`, since a failed recovery attempt is itself news the user needs, independent of the went-quiet debounce.

Minor nit: the `match &restart_result { Some(Err(_)) => ..., _ => ... }` in the `notify_down` block leans on a comment-documented invariant (`notify_down` implies `Some`) rather than making illegality unrepresentable — a `debug_assert!(restart_result.is_some())` before the match would catch a future refactor that breaks the invariant instead of silently mis-rendering the toast text as the success case.

Tests look solid otherwise — `cold_down_daemon_is_restarted_even_though_notice_stays_silent` and the updated existing cases correctly pin down the intended decision-table behavior for `decide_health_notice`.

Akarsh-Hegde:
Thanks @adityaharishch - good catch on the cold-start silent-failure gap. Addressed in ba217cf2.

**Cold-start restart failure is no longer silent.** `refresh_health` now raises the `tray.daemon_quiet` notice when the one automatic restart *fails*, regardless of `daemon_was_healthy` - a failed recovery is news in its own right, independent of the went-quiet cold-start debounce. It reuses the same notice id, so it dedupes with the warm-path notice and is cleared by the existing `notify_back` / `reconcile_stale` paths on recovery. New test `cold_start_failure_notice_is_cleared_on_recovery` pins that clear-on-recovery invariant.

**On the retry half of the point:** this PR also adds a dedicated `run_daemon_watchdog` (poll/mod.rs) that probes the daemon every 5s and re-attempts the restart every ~45s while it stays down. So a failed cold-start restart is now both *retried* (watchdog, indefinitely) and *surfaced* (notice) - no longer stranded for the process lifetime.

**Nit:** added `debug_assert!(restart_result.is_some())` before the `notify_down` match, so a future refactor that breaks the `notify_down => attempt_restart` invariant trips instead of silently mis-rendering the toast as the success case.

All 7 `poll::refresh` tests green; full pre-push suite (fmt / ui build / clippy / ui tests / cargo test / audit) passed. Ready to merge.

PR #565, opened 2026-07-23
fix(windows): stop the running daemon before restaging its binary

Akarsh-Hegde:
## Summary
- `ensure_backend_installed` re-stages the bundled daemon into `~/.meridian/bin/meridian.exe` on every version bump, but the previous build's daemon can still be running from that exact path (via the scheduled task, or the Startup-folder VBScript fallback on machines where `schtasks` is policy-blocked). Windows keeps a running exe's pages mapped, so the copy fails with "the process cannot access the file" (os error 32), surfaced to users as "Meridian couldn't finish installing."
- Added a Windows-only `stop_running_daemon_before_stage` step in `backend_install::install`, run right before `stage_binary`: best-effort `schtasks /End` (stops it if the scheduled task launched it) plus `taskkill /F /IM meridian.exe` (also catches the Startup-folder-launcher case), then polls `tasklist` for up to ~3s so the file handle actually clears before the copy is attempted.
- macOS is unaffected (block is `#[cfg(target_os = "windows")]`).

## Test plan
- [x] `cargo check` / `cargo clippy -- -D warnings` on the tray crate
- [x] Full `pre-push` hook suite (fmt, ui build, clippy, ui tests, cargo test, security audit) — all green
- [ ] Verify on a real machine that a version update no longer raises the "Meridian couldn't finish installing" notice

### DIFF
--- a/tray/src-tauri/src/backend_install.rs
+++ b/tray/src-tauri/src/backend_install.rs
@@ -195,6 +195,14 @@ async fn install(backend: &Path, home: &Path) -> Result<(), String> {
         migrate_legacy_bundle_env(home).await;
     }
 
+    // Windows keeps a running exe's pages mapped, so overwriting it fails with
+    // "the process cannot access the file" (os error 32) — and this isn't just
+    // a first-install concern: `ensure_backend_installed` re-stages on every
+    // version bump while the *previous* build's daemon is still alive from an
+    // earlier login. Stop it first so the copy below can succeed.
+    #[cfg(target_os = "windows")]
+    stop_running_daemon_before_stage().await;
+
     let daemon_bin = home.join(".meridian").join("bin").join(DAEMON_FILE);
     stage_binary(&backend.join(DAEMON_FILE), &daemon_bin).await?;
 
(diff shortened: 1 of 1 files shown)


adityaharishch (changes requested):
Strict review of `stop_running_daemon_before_stage` / `daemon_process_running` (tray/src-tauri/src/backend_install.rs).

**1. Force-kill bypasses the daemon's graceful shutdown path (blocking concern)**
`schtasks /End` and `taskkill /F /IM meridian.exe` both terminate the process without giving it a chance to run its normal shutdown sequence (per the daemon's architecture, `SIGINT`/`SIGTERM` triggers `run_etl()` completion + pool close — see main.rs's poll loop). On Windows there's no equivalent signal delivered here: the daemon can be mid-`run_etl()` or mid-write to `meridian.db` when it's force-killed. WAL mode makes SQLite crash-tolerant, but that's a mitigant, not a justification — this is a real behavior change (previously the daemon was never killed out from under itself during an update) and should at least be called out in the PR description/comments as an accepted trade-off, or better, given a chance to exit cleanly first (e.g. a named event / a lightweight local check the daemon polls) before escalating to `/F`.

**2. The 3s poll budget likely undersells the problem it's fixing**
The doc comment says this "mirrors the launchd bootout wait in `register_agent`," but `register_agent` polls for up to **15s** (15 × 1s), while this polls for only **~3s** (10 × 300ms) — for what the PR itself frames as the same class of problem (a held file handle not releasing promptly). The PR description explicitly calls out slow-to-release cases like AV scanning; on such a machine, 3s may not be enough, and `stage_binary`'s subsequent copy will still hit os error 32, silently reproducing the exact bug this PR claims to fix, just less often. Either match the macOS budget or, more robustly, wrap `stage_binary`'s copy itself in a short retry/backoff rather than relying solely on a pre-check.

**3. `taskkill /F /IM meridian.exe` is unscoped**
It kills every process with that image name system-wide, not just the one this module staged/registered. A developer manually running a locally-built `meridian.exe` for debugging on the same machine would be silently killed by an unrelated tray update. Worth at least a comment acknowledging this, since it's a real (if narrow) footgun.

**4. No test coverage for the new logic**
Every other helper added to this module historically gets at least a partial unit test (`migrate_bundle_env_copies_but_never_clobbers`, `cleanup_legacy_a11y_helper_removes_stale_files`, `apply_subs_replaces_every_occurrence`). `daemon_process_running`'s CSV-parsing logic (`contains(DAEMON_FILE)` against captured "found" vs "no tasks are running" tasklist output) is pure string handling and testable without a real Windows process — right now it ships with zero coverage.

Given #1 and #2 touch data-integrity and defeat-the-purpose-of-the-fix risk respectively, I'd like these addressed (or explicitly justified) before merge.

PR #566, opened 2026-07-23
fix: 4 memory-growth/leak fixes (ETL session-text, capture retention, MCP DB cache, summariser ledger, UI stores)

adityaharishch:
## Summary

Fixes 4 confirmed memory-growth/leak issues found by a prior read-only audit. Each fix is its own commit, tested before moving to the next.

### Fix 1 (HIGH): unbounded whole-session OCR reload on block close
**Files:** `src/db/screenpipe.rs`, `src/etl/block_ops.rs`, `src/etl/text_filter/mod.rs`, `src/etl/text_merge.rs`, `tests/etl_session_text.rs`

`get_frame_full_texts` did a single `fetch_all` across a block's entire frame-id range when the continuation-merge path in `close_block` rebuilt `session_text` from scratch — a session that stays in one app for hours never closes early, so at close time this could load hours of OCR text into RAM at once.

Added `fetch_frame_text_page` (bounded `id`-driven pagination), `ChromeFreqAccumulator`, and `SessionTextBuilder` so the chrome-detection and text-build passes fold incrementally over pages instead of materializing the whole range as one `Vec`. **No accuracy loss**: output is byte-for-byte identical to the prior non-paginated build for the same input — verified by a new integration test spanning 1600 frames (3 pages) that asserts equality against a direct non-paginated fetch.

### Fix 2 (HIGH): no retention on capture tables + MCP re-reads the whole DB file every call
**Files:** `src/etl/capture_retention.rs` (new), `src/etl/mod.rs`, `src/main.rs`, `src/db/meridian.rs`, `CLAUDE.md`, `packages/meridian-mcp/src/db-cache.ts` (new), `packages/meridian-mcp/src/index.ts`, `packages/meridian-mcp/src/__tests__/db-cache.test.ts` (new)

**(a)** `capture_frames` / `capture_ui_events` / `capture_secondary_screens` had no retention and grew forever. Added an age + processed-based sweep (`MERIDIAN_CAPTURE_RETENTION_DAYS`, default 30) wired into the poll loop right after `run_etl` — only deletes rows both older than the window AND already consumed by `etl_cursor.last_frame_id`, so an unprocessed frame is never pruned regardless of age. Also runs a page-bounded `PRAGMA incremental_vacuum` on a coarser cadence to reclaim freed pages (no-op on already-existing DBs still in `auto_vacuum=NONE` — we deliberately never force a full `VACUUM` on a live daemon; new DBs get `auto_vacuum=INCREMENTAL` so this actually reclaims space for them).

**(b)** The MCP server did `fs.readFileSync` + `new SQL.Database(...)` (two full copies) on every single tool call, and emscripten never returns freed WASM pages to the OS. Extracted DB-open logic into `db-cache.ts`, which reuses one long-lived `Database` per path, invalidated only when a cheap `fs.statSync` shows the file's mtime/size changed. Kept sql.js (not better-sqlite3) per the project's prior native-ABI distribution pain.

### Fix 3 (LOW): summariser retry-attempts map never evicted
**File:** `src/coding_agent_session_ingest/summariser/mod.rs`

The per-row failure ledger (`attempts: HashMap<i64, u32>`) was only ever inserted into, never removed from. Extracted a `record_attempt()` helper that removes a row's entry on success and increments it on failure, so the map only ever holds currently-pending rows with an active failure count.

### Fix 4 (LOW): webview module-level Maps never evict
**Files:** `ui/lib/lru.ts` (new), `ui/components/plan/planStore.ts`, `ui/components/timeline/useWorklog.ts`, `ui/lib/app-icons.ts`, `ui/__tests__/lru.test.ts` (new), `ui/__tests__/plan-store.test.ts`

Added a shared `LruMap` (get/set/has/delete — the subset these stores use) and capped all three module-level stores at 100 entries, evicting least-recently-touched past the cap. No behavior/API change below the cap.

## Why this is safe (no accuracy loss)
- Fix 1: verified byte-for-byte identical output via test, not a truncation.
- Fix 2a: never deletes a frame/event the ETL hasn't consumed yet (id/timestamp-gated against the live cursor).
- Fix 2b: only invalidates the cache when the underlying file actually changes; small staleness window is acceptable for a read-mostly MCP server against a separate-process writer.
- Fix 3: only evicts entries for rows that have already succeeded — no behavior change to retry/dead-letter logic.
- Fix 4: 100-entry cap is far beyond realistic usage cardinality (days/tasks/apps seen in one session) — pure hygiene.

## Test plan
- [x] `cargo fmt --check` (after every commit)
- [x] `cargo clippy --all-targets -- -D warnings` (after every commit)
- [x] `cargo test` — full suite, including new integration/unit tests (after every commit)
- [x] `cd ui && npm run build` + `bun test` (356 tests)
- [x] `cd packages/meridian-mcp && npm run build` + `npm test` (node:test, 4 new tests)
- [x] Pre-push hook (fmt + clippy + ui build + ui tests + cargo test + security audit) passed on push

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -236,6 +236,7 @@ There are no JS/TS test suites yet. When adding them, place them under `ui/__tes
 | `MERIDIAN_EMBEDDER_DIR` | `~/.meridian/models/<repo-basename>/` | Override the on-disk directory the session-distiller embedding weights live in (`src/embedder/provision.rs`). |
 | `MERIDIAN_EMBEDDER_REPO` | `BAAI/bge-small-en-v1.5` | Override the HuggingFace repo the embedder weights are fetched from. |
 | `DISTILLER_SEM_DEDUP_THR` | `0.86` | Cosine threshold for the distiller's semantic dedup (`src/worklog_pipeline/distiller/`). |
+| `MERIDIAN_CAPTURE_RETENTION_DAYS` | `30` | Age floor for the capture_frames/capture_ui_events/capture_secondary_screens retention sweep (`src/etl/capture_retention.rs`) — only prunes rows both older than this AND already consumed by the ETL cursor. |
 
 Tilde expansion is handled by `Config::from_env()`. Never hardcode paths.
 
--- a/packages/meridian-mcp/package-lock.json
+++ b/packages/meridian-mcp/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "meridian-mcp",
-  "version": "1.28.1",
+  "version": "1.74.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "meridian-mcp",
-      "version": "1.28.1",
+      "version": "1.74.0",
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "1.27.1",
--- a/packages/meridian-mcp/package.json
+++ b/packages/meridian-mcp/package.json
@@ -12,6 +12,7 @@
     "start": "node dist/index.js",
     "install-mcp": "node dist/install.js",
     "dev": "ts-node src/index.ts",
+    "test": "tsc && node --test \"dist/__tests__/*.test.js\"",
     "prepublishOnly": "npm run build"
   },
   "keywords": [
--- a/packages/meridian-mcp/src/__tests__/db-cache.test.ts
+++ b/packages/meridian-mcp/src/__tests__/db-cache.test.ts
@@ -1,0 +1,97 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+
+// Tests for db-cache.ts's mtime/size-gated sql.js Database cache — the fix
+// for the MCP server re-reading + re-allocating the whole meridian.db file
+// on every single tool call. Run via `node --test` (built-in Node test
+// runner, no extra test-framework dependency) against the compiled output:
+// `npm run build && node --test dist/__tests__/`.
+
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import initSqlJs from "sql.js";
+import { openCachedDb, _resetDbCacheForTests, _dbCacheSizeForTests } from "../db-cache.js";
+
+async function makeTempDb(dir: string, name: string, seedRows: number): Promise<string> {
+  const SQL = await initSqlJs();
+  const db = new SQL.Database();
+  db.run("CREATE TABLE t (id INTEGER PRIMARY KEY, val TEXT)");
+  for (let i = 0; i < seedRows; i++) {
+    db.run("INSERT INTO t (val) VALUES (?)", [`row-${i}`]);
+  }
+  const bytes = db.export();
+  db.close();
+  const dbPath = path.join(dir, name);
+  fs.writeFileSync(dbPath, bytes);
+  return dbPath;
+}
+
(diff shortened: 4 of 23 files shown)


Akarsh-Hegde (changes requested):
## Review: 4 memory-growth/leak fixes

Solid PR — four independent, well-isolated fixes for memory growth found by the prior audit, each backed by tests that target the specific property being changed (byte-identical pagination output, LRU eviction semantics, ledger eviction on success, retention-cursor behavior).

### Bug: capture-retention cursor can delete its own reference row (`src/etl/capture_retention.rs`)

`prune_capture_tables` reads `etl_cursor.last_frame_id`, looks up that frame's `timestamp` (`cursor_ts`), then runs:

```rust
DELETE FROM capture_frames WHERE id <= ?1 AND timestamp < ?2
```
bound to `(last_frame_id, cutoff_ts)`. Because the comparison is `id <= last_frame_id` (not `<`), if the cursor's own frame is itself older than the retention cutoff, this DELETE removes the cursor's own row in the same tick that just read it.

On the *next* tick, `SELECT timestamp FROM capture_frames WHERE id = last_frame_id` returns nothing, and the function hits the "cursor frame not found — skipping this tick" early return — retention silently stops running until the ETL cursor advances to an id that still exists. Self-healing once new frames get processed, but during any stretch where the cursor is stuck (daemon paused, machine asleep, catching up on a backlog), pruning quietly stops enforcing the retention window instead of continuing to work off the age boundary alone.

This is exercised (accidentally) by `processed_old_frame_is_pruned`: it sets `last_frame_id = 1` pointing at the *only* row, which is old, and asserts the table ends up empty — that's the cursor deleting itself. The test just doesn't go on to check tick N+1 with the same cursor value, where it would hit the "not found" branch.

Suggested fix: exclude the cursor's own row from the age-based delete (`id < last_frame_id`), or fall back to `id >= last_frame_id ORDER BY id ASC LIMIT 1` when the exact cursor row is missing, so the reference frame survives its own sweep.

### Minor / worth a look

- `ui/components/timeline/useWorklog.ts`: the LRU cap (100) evicts by recency of `store.set`/`store.get` touches, not by "is this mid-flight." An entry with `phase: 'generating'`/`'approving'` is only protected while something actively touches it — if a user browses 100+ other distinct day/task panels while a generation sits idle in-flight, it could be evicted, and the `store.get(key)?.phase === 'generating'` dedup guard would silently stop preventing a duplicate concurrent generation for that key. Low likelihood given real usage, and already partly acknowledged in the comment, but the "never the eviction candidate while active" claim isn't fully guaranteed.
- `rebuild_session_text_paginated` reads the frame range from the DB twice (chrome-frequency pass, then text-build pass) to keep memory bounded — correct and tested, but doubles DB I/O per block close on long sessions. Reasonable tradeoff, just flagging it's there.
- `packages/meridian-mcp/package.json`/`package-lock.json` and `ui/package-lock.json` carry version bumps (`1.28.1→1.74.0`, `1.71.0→1.74.0`) unrelated to this PR's description — looks like drift from `pre-main` moving forward rather than an intentional change. Worth confirming before merge.
- `db-cache.ts`: checked that the MCP server never `await`s between `db = await openDb()` and using `db` within a tool-call handler (only one `await` in the whole request handler), so the shared cached instance can't be closed out from under an in-flight call by a concurrent request. Holds up.

### Strengths
- Tests target the actual property changed, not just smoke coverage.
- The retention sweep's "never delete unprocessed rows" invariant is correctly enforced in the normal case and clearly documented.
- `record_attempt` extraction in the summariser is a clean, minimal fix.
- `LruMap` is appropriately minimal — only the subset of `Map` actually used by callers.

PR #567, opened 2026-07-23
fix(windows): suppress the console-window flash on child process spawns

Akarsh-Hegde:
## Summary
On Windows, every `Command::new` spawn of a **console-subsystem** program pops a visible console window unless the process is created with `CREATE_NO_WINDOW` — piping stdio does not suppress it. The tray is a GUI app and the daemon runs console-less, so each such spawn flashed a window. Two user-visible symptoms:
- a popup on nearly **every tray operation** (the tray shelling out to `meridian.exe` for tasks-sync, ticket-update, ticket-parents, uninstall, daemon control, backend install)
- a popup on **"claude verification"** (the coding-agent summariser's `claude`/`codex`/`cursor-agent` CLI calls, which npm installs as `.cmd` shims routed through `cmd.exe`)

## What changed
- New `meridian_core::proc_ext::NoWindow` trait: a chainable `.no_window()` implemented for **both** `std::process::Command` and `tokio::process::Command`. Sets `CREATE_NO_WINDOW` on Windows, no-op everywhere else, so call sites chain it unconditionally.
- Applied `.no_window()` to every console-subsystem spawn across the daemon and tray: `schtasks`/`taskkill`/`tasklist`/`powershell`, the CLI provider calls (summariser, detect, cursor-agent init), the `meridian.exe` shell-outs, and the daemon binary itself.
- Deliberately **not** applied to `health::fix::exec` — `meridian doctor --fix` runs in a real terminal and inherits stdio; suppressing it would swallow output the user wants.
- Promotes `tokio` from a dev-dependency to a real dependency of `meridian-core` (`process` feature only) for the tokio `Command` impl. tokio is already in the tree via sqlx's runtime, so feature-unification means this adds nothing new to compile.
- Removes a `needless_return` in `app_icons.rs` that only trips clippy on the non-macOS build (the macOS CI never compiles that cfg arm) and would otherwise block the Windows pre-push clippy gate.

## Note on the base
This branch is **stacked on `fix/windows-stage-copy-daemon-lock` (#565)** because both change `backend_install.rs`. Until #565 merges, this PR's diff will also show #565's commit; once #565 merges into `pre-main`, it will show only the console-flash change. Merge #565 first.

## Test plan
- [x] `cargo check` + `cargo clippy -- -D warnings` on the daemon and tray, compiled for the `aarch64-pc-windows-msvc` target (so the `#[cfg(windows)]` arms were actually type-checked and linted)
- [x] Full `pre-push` hook suite (fmt, ui build, clippy, ui tests, cargo test, security audit) — all green
- [ ] Confirm on a real Windows run that tray operations and summariser runs no longer flash a console window

### DIFF
--- a/meridian-core/Cargo.toml
+++ b/meridian-core/Cargo.toml
@@ -22,6 +22,13 @@ dirs = "5"
 # Instrumentation facade only (~no-op without a subscriber) — the heavy OTLP
 # exporter is wired by the binary (daemon, or the tray's `otel` dev feature).
 tracing = "0.1"
+# Only the `process` feature, only for `proc_ext::NoWindow`'s impl on
+# `tokio::process::Command` (the std impl needs no dep). tokio is already in the
+# workspace tree via sqlx's tokio runtime and every binary that depends on this
+# crate (daemon, tray) already builds it with `full`, so feature-unification
+# means this adds nothing new to compile — it just brings the `process` module
+# into scope here.
+tokio = { version = "1", features = ["process"] }
 
 [dev-dependencies]
 # For the #[ignore]d golden-compare smoke test (runs get_today against the real DB).
--- a/meridian-core/src/lib.rs
+++ b/meridian-core/src/lib.rs
@@ -50,6 +50,10 @@ pub mod plan_marker;
 /// tray writes `capture_frames`, the daemon's ETL reads it.
 pub mod capture;
 
+/// `.no_window()` for `Command` — suppresses the Windows console-window flash
+/// on a console-subsystem child spawn. No-op on every other OS.
+pub mod proc_ext;
+
 // ── Curated public API: flat module paths, stable across file moves ──────────
 pub use db::{get_active_session, open_existing, ActiveSession};
 
--- a/meridian-core/src/proc_ext.rs
+++ b/meridian-core/src/proc_ext.rs
@@ -40,3 +40,42 @@ impl NoWindow for tokio::process::Command {
         self
     }
 }
+
+// Windows-only: the trait does nothing off Windows, and `cmd.exe` (the console
+// child these exercise) exists only there. Both tests assert `.no_window()`
+// still produces a working spawn — the flag suppresses the console window, it
+// must not break launching the child. That the window is actually suppressed
+// was confirmed out-of-band on a real Windows host (a nested `cmd.exe →
+// powershell` grandchild reports a console `HWND` of 0 with the flag set and a
+// non-zero one without it); there is no non-flaky way to assert the window's
+// absence from a headless test, so these lock in the spawn-still-works half.
+#[cfg(all(test, windows))]
+mod tests {
+    use super::NoWindow;
+
+    #[test]
+    fn create_no_window_has_the_documented_value() {
+        assert_eq!(super::CREATE_NO_WINDOW, 0x0800_0000);
+    }
+
+    #[test]
+    fn no_window_std_command_still_spawns() {
+        let status = std::process::Command::new("cmd")
+            .args(["/C", "exit 0"])
+            .no_window()
+            .status()
+            .expect("cmd should spawn with CREATE_NO_WINDOW set");
+        assert!(status.success());
+    }
+
+    #[tokio::test]
+    async fn no_window_tokio_command_still_spawns() {
+        let status = tokio::process::Command::new("cmd")
+            .args(["/C", "exit 0"])
+            .no_window()
+            .status()
+            .await
+            .expect("cmd should spawn with CREATE_NO_WINDOW set (tokio)");
+        assert!(status.success());
+    }
+}
--- a/meridian-core/src/proc_ext.rs
+++ b/meridian-core/src/proc_ext.rs
@@ -1,0 +1,42 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+//! Suppress the console-window flash Windows creates when a GUI process (the
+//! tray) or a console-less daemon spawns a console-subsystem child — every
+//! plain Win32 console app (`schtasks`, `taskkill`, `tasklist`, `powershell`)
+//! and every npm-shimmed `.cmd`/`.bat` CLI (`claude`, `codex`, …, which
+//! `Command::new` routes through `cmd.exe`) is one. Piping the child's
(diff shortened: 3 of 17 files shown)


adityaharishch:
## Review — strict pass

**Scope note:** this PR is stacked on #565 (`fix/windows-stage-copy-daemon-lock`), which hasn't merged yet, so `gh pr diff 567` shows both commits. I've reviewed the whole diff but attributed findings to the commit that actually owns each hunk — the `backend_install.rs` stop-daemon-before-restage logic is #565's, not #567's own scope.

### What this PR does (567's own commit, `2d733964`)
Adds `meridian_core::proc_ext::NoWindow`, a `.no_window()` extension trait implemented for both `std::process::Command` and `tokio::process::Command` that sets `CREATE_NO_WINDOW` on Windows (no-op elsewhere), and applies it at every console-subsystem spawn site that runs on a Windows host: the cursor-agent init flow, the coding-agent summariser's CLI spawn, `health::platform::cmd_output`, the four `llm::detect` spawn sites, `platform::windows::list_process_argvs`, `uninstall.rs`'s `schtasks /Delete`, and the tray's `meridian.exe` shell-outs (tasks-sync, ticket-update, ticket-parents, uninstall, daemon control). Also promotes `tokio` to a real (process-feature-only) dependency of `meridian-core`, and drops a stray `return` that only trips clippy on non-macOS.

### Correctness
- **Coverage is genuinely complete.** I grepped every `Command::new`/`tokio::process::Command::new` call site in `src/`, `tray/src-tauri/src/`, `meridian-core/src/` and cross-checked each Windows-reachable one against the diff. `cursor_agent_init.rs`'s four call sites all funnel through the single `run_with_timeout` helper, which got `.no_window()` added once — smart, avoids four separate edits. Every other Windows-reachable console spawn (schtasks/taskkill/tasklist/powershell, the CLI provider probes, the `meridian.exe` shell-outs) is covered. macOS/Unix-only spawns (`launchctl`, `plutil`, `pkill`, `tccutil`, `osascript`, `pgrep`, `ps`, `df`, `id`, `open`) are correctly left untouched.
- **`health::fix::exec` is correctly excluded** — matches the PR description's stated rationale (interactive `doctor --fix` needs its terminal), and I confirmed no `.no_window()` was added there.
- **No `creation_flags` collision risk** — grepped for other `creation_flags`/`CREATE_NEW_PROCESS_GROUP`/`DETACHED_PROCESS` usage; there is none, so `.no_window()`'s `creation_flags()` call (which *sets*, not ORs, the flags field on `std::process::Command`) can't clobber an existing flag elsewhere.
- **`tokio::process::Command::creation_flags`** is an inherent Windows-only method on tokio's `Command`, so not needing the `CommandExt` import there (unlike the `std::process::Command` impl, which correctly scopes that import to `#[cfg(windows)]`) is right, not an oversight.
- The `app_icons.rs` `return` → tail-expression clippy fix is valid: the `#[cfg(not(target_os = "macos"))]` block is genuinely the function's last statement on non-macOS builds.

### Style / conventions
- Doc comments follow the file-header + module `//!` + per-fn `///` convention consistently.
- `tokio` feature-gating (`features = ["process"]` only) is minimal and the comment justifying it (feature-unification via sqlx's `full`) is verifiable and correct given the workspace already pulls tokio `full` elsewhere.

### Risks / gaps (flagging per "strict")
1. **Unverified on real Windows.** The PR's own test plan leaves "confirm on a real Windows run" unchecked. `cargo check`/`clippy` for the `aarch64-pc-windows-msvc` target catches type/lint errors but can't catch a wrong flag value or an environment where `CREATE_NO_WINDOW` doesn't suppress the flash for a particular child (e.g. some `.cmd` shims spawn a *nested* `cmd.exe` that isn't covered by the parent's creation flags). Given this only ships to Windows users, I'd want the manual check before merge, not just static analysis.
2. **No test coverage added** for `proc_ext::NoWindow` — understandable (there's nothing meaningfully unit-testable about a Windows-only flag with no macOS CI runner), but worth calling out explicitly since none exists even as a `#[cfg(windows)]`-gated smoke test.
3. Minor: the module doc for `proc_ext.rs` calls piping-stdio-doesn't-suppress-the-flash a fact worth double-checking empirically once on Windows (point 1) rather than taking on faith, since the whole PR's premise rests on it.

No blocking issues found in `2d733964`'s own diff. Recommend merging #565 first as already noted, then giving this one an actual Windows smoke test before merging.

Akarsh-Hegde:
Thanks for the strict pass. All three raised gaps are now addressed; pushed as `5db76867`.

### Gap 1 & 3 — empirical Windows verification (incl. the nested-`cmd.exe` concern)
Verified on a real Windows host. I spawned a **nested `cmd.exe -> powershell` grandchild** — deliberately the same shape as the npm `.cmd` shims that run `claude`/`codex`/`cursor-agent` (`Command::new("claude.cmd")` -> `cmd.exe /C` -> node) — and had the grandchild report its console window handle via `GetConsoleWindow()`:

| creation flag | grandchild `GetConsoleWindow()` |
|---|---|
| none | `HWND=66566` (a real console window — this is the flash) |
| `CREATE_NO_WINDOW` | `HWND=0` (no console window) |

So the flag **is** inherited by grandchildren: the nested `cmd.exe` attaches to the flagged parent's (absent) console, and so does whatever it launches. The concern that a `.cmd` shim's nested `cmd.exe` might escape the parent's flag does not hold — that exact chain is suppressed.

### Gap 2 — test coverage
Added `#[cfg(all(test, windows))]` tests in `proc_ext.rs`: they assert the `CREATE_NO_WINDOW` value and that `.no_window()` still produces a working spawn (flag suppresses the window without breaking the child launch) for **both** the `std` and `tokio` `Command` impls. They run under the Windows pre-push `cargo test`, and pass on this host:

```
test proc_ext::tests::create_no_window_has_the_documented_value ... ok
test proc_ext::tests::no_window_std_command_still_spawns ... ok
test proc_ext::tests::no_window_tokio_command_still_spawns ... ok
```

(There's no non-flaky way to assert the window's *absence* from a headless unit test — that's what the out-of-band `GetConsoleWindow` check above covers — so the tests lock in the spawn-still-works half.)

### Stacking
Still stacked on #565 (both touch `backend_install.rs`). **Merge #565 first**, then this; once #565 lands in `pre-main` the diff here collapses to just the console-flash change.

**Ready to merge** once #565 is in. No open items on this branch.

PR #571, opened 2026-07-23
fix(tray): quiet the Windows ARM64 'update check failed' toast (missing platform asset)

Akarsh-Hegde:
## Summary
- `install_provider` (src/llm/detect.rs) unconditionally spawned $SHELL-or-/bin/zsh to run vendor installers. On Windows there is no SHELL env var and no /bin/zsh, so the shell itself failed to start with "could not start a shell to install: os error 3" before the installer (even a plain `npm i -g ...`) ever ran, breaking every Install button (Codex, Cursor, etc.) on Windows. `installer_command()` now branches per platform: `cmd /C` on Windows, the existing login-shell behavior everywhere else.
- The tray's "Check for Updates..." toasted a raw `None of the fallback platforms [...] were found in the response platforms object` error on Windows ARM64 builds. `scripts/package-updater-windows.sh` only ever publishes `windows-x86_64` / `windows-x86_64-nsis` keys (the CI `windows` release job builds on `windows-latest`, i.e. x86_64, only - there is no ARM64 Windows release job), so an ARM64 install can never match anything in `latest.json`, forever. `check_status` (tray/src-tauri/src/update.rs) now recognizes that specific `tauri-plugin-updater` error (`TargetNotFound`/`TargetsNotFound`) and reports `state: "unsupported"` (the same bucket as an unbundled dev run) instead of a scary, unfixable `"error"`, with friendlier toast copy.

## Test plan
- [x] `cargo test -p meridian --lib llm::detect` - 14/14 pass, including new `installer_command_actually_runs_on_this_platform`
- [x] `cargo test --lib update::` (tray/src-tauri) - 14/14 pass, including new `missing_platform_asset_matches_target_not_found_variants` and `other_updater_errors_are_not_treated_as_missing_platform_asset`
- [x] Full pre-push suite (fmt, clippy, ui build/tests, cargo test, security audit) green
- [ ] Manual: install a provider (Codex/Cursor) from a Windows ARM64 staging build and confirm it completes instead of erroring
- [ ] Manual: "Check for Updates..." on the same ARM64 build now shows "Auto-update not available" instead of the raw fallback-platforms error

### DIFF
--- a/tray/src-tauri/src/update.rs
+++ b/tray/src-tauri/src/update.rs
@@ -188,6 +188,40 @@ fn is_packaged() -> bool {
     crate::sys::is_bundled()
 }
 
+/// Marks an `"unsupported"` [`UpdateStatus`] whose `error` field is a reason
+/// code (`"no-platform-asset"`), not an error to alarm the user with — see
+/// [`is_missing_platform_asset`].
+const NO_PLATFORM_ASSET: &str = "no-platform-asset";
+
+/// True for the `tauri-plugin-updater` errors that mean "this platform has no
+/// published update artifact in `latest.json`" rather than a real check
+/// failure. Concretely this is Windows ARM64 today: `scripts/package-updater-windows.sh`
+/// only ever writes the `windows-x86_64` / `windows-x86_64-nsis` keys (the
+/// `.github/workflows/release-build.yml` `windows` job builds on `windows-latest`,
+/// i.e. x86_64, only — there is no ARM64 Windows release job), so an ARM64
+/// install's updater lookup can never match anything the manifest carries.
+///
+/// Surfacing that as `state: "error"` toasts a scary, unactionable "Update
+/// check failed" on *every* check, forever, for every ARM64 Windows install —
+/// there is no retry that fixes it. Bucketing it with `"unsupported"` (the
+/// same state used for an unbundled dev run) makes both the tray toast and the
+/// dashboard banner go quiet instead, same as the DMG updater already does for
+/// npm-installed CLIs.
+///
+/// Deliberately narrow: only `TargetNotFound`/`TargetsNotFound` map here.
+/// Every other error (network, signature, 404 on `latest.json` itself) still
+/// reports `"error"` — CI's `verify-release-bundle.sh` / `compose-updater-manifest.py`
+/// already fail the release loudly if an *expected* platform key (darwin-aarch64,
+/// windows-x86_64, …) goes missing, so a real regression there won't get
+/// silently swallowed by this.
+fn is_missing_platform_asset(e: &tauri_plugin_updater::Error) -> bool {
+    matches!(
+        e,
+        tauri_plugin_updater::Error::TargetNotFound(_)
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde (commented):
## Review — strict pass

### 🔴 Confirmed: doesn't actually fix Cursor's Windows install
`installer_command()`'s Windows branch runs the command via `cmd /C`. That fixes the npm-based installers (Claude, Codex, Copilot), but `LlmProvider::install_command()` for Cursor is:

```
curl -fsSL https://cursor.com/install | sed -E 's#...#...#g' | bash
```

`sed` and `bash` don't exist on stock Windows, and `cmd.exe` doesn't interpret single-quoted args the way `sed` expects. This pipeline will still fail under `cmd /C` — with a different, still-broken error than before. The PR summary says this fixes "every install button (Codex, Cursor, etc.)" — that's not accurate for Cursor.

Not speculative: the already-open **#576** independently confirms this — *"overlaps in spirit with the still-open #571 (which fixed the `cmd /C` half of the installer-shell bug but not Cursor's Windows-specific installer...)"* — and ships the real fix (PowerShell + a Windows-specific pinned Cursor installer).

The new test `installer_command_actually_runs_on_this_platform` only exercises `echo hello-from-installer`, so it can't catch this — it proves the shell spawns, not that the real provider commands work.

**Suggestion:** narrow the PR claim to "npm-based providers" or close this in favor of #576, which supersedes it.

### 🟡 Duplicate with #574
`#574`'s branch independently re-implements this exact same diff (`installer_command` + `is_missing_platform_asset`) — byte-identical hunks in both `src/llm/detect.rs` and `tray/src-tauri/src/update.rs`. Both are open/mergeable; landing both leaves duplicated history with no clear source of truth. Recommend picking one (or rebasing #574 to drop the overlap) before merging.

Akarsh-Hegde:
## Update - review resolved ?

Both findings from the strict-pass review are addressed by **slimming this PR**:

- **?? Cursor Windows install gap** + **?? duplicate of #574** - the `src/llm/detect.rs` `installer_command` change (the part that only fixed the `cmd /C` half and left Cursor's `curl | sed | bash` still broken on Windows) has been **removed**. The real Windows Cursor fix ships in **#576** (a `#[cfg(windows)]` PowerShell installer via `irm 'cursor.com/install?win32=true' | iex`, spawned through `powershell.exe`). Dropping the hunk here kills both the duplication and the incomplete-fix concern in one move.

This branch now carries **only** its unique contribution: the ARM64 update-check messaging - `is_missing_platform_asset` buckets `TargetNotFound`/`TargetsNotFound` as `"unsupported"` (the same quiet path the DMG updater already uses for npm CLIs) instead of toasting an unactionable "Update check failed" forever on every Windows ARM64 install. No overlap with #574 or #576 remains.

Force-pushed `d58545d3`. Full pre-push suite (fmt / UI build / clippy / UI tests / audit / cargo test) green; `mergeable_state: clean`. Merge-order-independent from #576/#574 (no shared files). **Ready to merge.**

PR #574, opened 2026-07-24
feat(tray): self-heal launch-at-login

Akarsh-Hegde:
## Summary

- **feat(tray)**: register the tray as a login item once ever, via `tauri-plugin-autostart` — self-heals both fresh and pre-existing installs so the tray comes back after a reboot/re-login without a manual relaunch. Mirrors the daemon's own `LaunchAgent` self-heal in `backend_install.rs`.
- **fix(llm)**: `install_provider` was unconditionally spawning `$SHELL`-or-`/bin/zsh`, which fails to start at all on Windows (`os error 3`) before any installer command runs — broke every provider's install button on Windows, including plain `npm i -g …`. Now branches to `cmd /C` on Windows.
- **fix(tray)**: `tauri-plugin-updater`'s `TargetNotFound`/`TargetsNotFound` (no published artifact for this platform — Windows ARM64 today) was surfacing as `state: "error"`, toasting an unactionable "Update check failed" on every check forever. Now bucketed into the existing `"unsupported"` state with copy that reflects the real reason.

The latter two were found and fixed incidentally while bringing up and testing the autostart feature on a Windows ARM64 VM.

## Test plan

- [ ] `cargo test` (detect.rs, update.rs both have new unit tests covering the fixes)
- [ ] Manual: fresh bundled install on macOS — confirm login item registered, marker written, no re-registration on second launch
- [ ] Manual: bundled install on Windows — confirm login item (Run key) registered
- [ ] Manual: click an LLM provider install button on Windows — confirm it now runs instead of failing with "could not start a shell"
- [ ] Manual: Windows ARM64 build — confirm update check shows the quiet "Auto-update not available" state instead of an error toast

https://claude.ai/code/session_0136cCDNbNc6SsX1hYNTVXaG

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -545,6 +545,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "auto-launch"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1f012b8cc0c850f34117ec8252a44418f2e34a2cf501de89e29b241ae5f79471"
+dependencies = [
+ "dirs 4.0.0",
+ "thiserror 1.0.69",
+ "winreg 0.10.1",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2198,6 +2209,15 @@ dependencies = [
 
 [[package]]
 name = "dirs"
+version = "4.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ca3aa72a6f96ea37bbc5aa912f6788242832f75369bdfdadcb0e38423f100059"
+dependencies = [
+ "dirs-sys 0.3.7",
+]
+
+[[package]]
+name = "dirs"
 version = "5.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "44c45a9d03d6676652bcb5e724c7e988de1acad23a711b5217ab9cbecbec2225"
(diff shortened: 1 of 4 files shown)


Akarsh-Hegde (commented):
## Review — strict pass

### 🔴 Confirmed: carries the same Cursor-Windows gap as #571
This branch's `installer_command`/`update.rs` changes are byte-identical to #571's. Same issue applies: `cmd /C` fixes the npm-based providers but not Cursor, whose installer is `curl ... | sed -E '...' | bash` — neither `sed` nor `bash` exist on stock Windows. The PR body says *"broke every provider's install button on Windows"* — Cursor is still broken after this fix, just with a different error. Confirmed by the already-open **#576**, which explicitly calls out this exact gap in #571/this-diff and ships the real fix (PowerShell-based Windows installer for Cursor).

### 🟡 Duplicate with #571
The `installer_command`/`is_missing_platform_asset` portion of this diff is a wholesale duplicate of #571 (both open, both mergeable, byte-identical hunks). Recommend closing/rebasing one before merge so this PR carries only the autostart delta.

### 🟡 Autostart: `is_bundled()` doesn't rule out running from the mounted DMG
`autostart::ensure_enabled_once` registers a login item pointing at `current_exe()`'s path, gated only by `sys::is_bundled()` (checks the exe sits inside `*.app/Contents/MacOS`, not that the `.app` lives somewhere stable like `/Applications`). If a user launches the app straight off the mounted disk image before dragging it in, the LaunchAgent gets pinned to a path that vanishes on eject. Since `autostart_configured` is written on first success and never re-checked, self-heal silently and permanently stops working for that user, with no retry. This is a known failure class for macOS auto-launchers (Sparkle guards against it explicitly). Worth a location check before writing the marker, given how carefully this codebase already handles similar macOS path gotchas (TCC/cdhash churn, the daemon-deploy note about never symlinking into `~/Documents`).

Akarsh-Hegde:
## Update - review resolved ?

All three findings addressed:

- **?? Cursor Windows gap** + **?? duplicate of #571** - the shared `detect.rs` / `update.rs` diff has been **dropped**. The installer fix now lives in **#576** (the real PowerShell Cursor installer) and the ARM64 update-check messaging in **#571**. This PR is now scoped to **just the autostart feature** - no byte-identical hunks remain across the three.
- **?? autostart pinned to a path that vanishes off the mounted DMG** - added a location guard. On macOS, `ensure_enabled_once` now refuses to register the login item while `current_exe()` sits under `/Volumes/.` (mounted disk image) or `./AppTranslocation/.` (Gatekeeper quarantine), **deferring without writing the marker** so it retries on the next launch - by which time the user has dragged the app to a stable path (e.g. `/Applications`). This prevents the "self-heal silently breaks forever" failure the review called out. The string test is unit-covered (`path_is_stable`); non-macOS is unaffected (Windows Run-key / Linux desktop entries already point at the installed path).

Force-pushed `23625ca9`. Local pre-push suite green; CI green except the **Rust (macOS Apple Silicon)** job still running - that's where the new test executes. **Ready to merge** once that job goes green.

PR #575, opened 2026-07-24
fix(ui): hide the Advanced tab and simplify notification settings

Akarsh-Hegde:
## Summary
- Comment out the Advanced nav entry in the Settings sidebar and its section render in the Settings modal, so the tab is no longer reachable from the UI (`AdvancedSection` component itself is left intact, just unreferenced, for easy re-enable later)
- Simplify Settings -> Notifications down to two controls: the master on/off switch and quiet hours
- Removed the per-category toggles (plan reminder, worklog ready, system fault, pause/resume, daemon health, updates, summarisation issues, board hygiene) and the "Open planner each day" toggle
- Every notification event is now gated only by `notifications_enabled` + quiet hours, decided at drain time in `meridian-core::notifications::event_allowed` — producers already enqueued unconditionally, so this is a pure policy simplification, not a behavior loss for any individual event type
- The daily planner auto-open keeps running (still governed by its marker/onboarded/plan-handled gates) since it's no longer user-toggleable — flagging this explicitly since it's the one behavior-visible change: users can no longer turn off the daily planner pop-up
- Removed the now-dead `notify_*` / `auto_open_plan` fields from `RuntimeSettings` (Rust + TS mirror) and updated stale doc comments across the tray/daemon that referenced the old per-type toggles
- Added a Rust unit test (`event_allowed_respects_only_the_master_switch`) and a new UI test file (`ui/__tests__/notifications-settings-simplified.test.ts`) covering the simplified surface

## Test plan
- [x] `cargo test --workspace` — 733+ tests pass, 0 failures
- [x] `cargo clippy --workspace -- -D warnings` and `cargo fmt --check` — clean
- [x] `npm run build` in `ui/` succeeds with no TypeScript/lint errors
- [x] `bun test` in `ui/` — 361 tests pass, 0 failures
- [x] pre-push checks pass (fmt, clippy, ui build/tests, security audit, cargo test)
- [ ] Manually open Settings in the tray and confirm: no "Advanced" entry in the sidebar; Notifications shows only the master switch + quiet hours

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -382,7 +382,7 @@ spooled files to OpenObserve, and it refuses to do so entirely for a
 Only a Dev/Bare checkout with `otlp_enabled` + `oo_email`/`oo_password`
 configured ships live, for an engineer debugging against their own
 OpenObserve instance. A shipped install's only path to a developer's
-OpenObserve is the tray Settings → Advanced → **Export Diagnostics** button
+OpenObserve is the tray Settings → Account → **Export Diagnostics** button
 (or `meridian telemetry export`), which bundles the spool + the launchd
 crash-safety-net logs into a `.tar.gz` the user hands to support, imported by
 hand with `meridian telemetry import <bundle> --endpoint <url> --auth
--- a/NOTIFICATIONS.md
+++ b/NOTIFICATIONS.md
@@ -55,7 +55,7 @@ use crate::notifications::{self, NewNotification};
 
 notifications::enqueue(pool, NewNotification::event(
     &format!("my.event:{scope}"),   // dedup: '<event_key>:<scope>'
-    "my.event",                     // event_key — preference lookup
+    "my.event",                     // event_key — dedup/category grouping, not a preference lookup
     "Title", "Body",
 )
 .link("/route")                     // optional dashboard deep link
@@ -64,9 +64,11 @@ notifications::enqueue(pool, NewNotification::event(
 ).await?;
 ```
 
-That's the whole producer. Delivery policy (master switch, per-type toggle,
-quiet hours) is enforced at drain time by `meridian-core`, never by producers —
-**always enqueue; the user's settings decide whether it surfaces.**
+That's the whole producer. Delivery policy (master switch, quiet hours) is
+enforced at drain time by `meridian-core`, never by producers — **always
+enqueue; the user's settings decide whether it surfaces.** There is no
+per-type toggle — every event is treated the same, gated only by those two
+knobs, kept deliberately simple for the user.
 
 ## Add an interactive notification (buttons / inline reply)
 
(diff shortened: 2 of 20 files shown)


Akarsh-Hegde (commented):
## Review — strict pass

### 🔴 Confirmed regression: hiding "Advanced" removes the only GUI path to Export Diagnostics
`AdvancedSection.tsx` is the sole home of Export Diagnostics, and per this repo's own `CLAUDE.md`: *"A shipped install's only path to a developer's OpenObserve is the tray Settings → Advanced → Export Diagnostics button (or `meridian telemetry export`)."* Commenting out the nav entry + render branch removes that path entirely, along with GUI access to log level, ETL poll interval, task-routing confidence floors, LLM budget, and the Jira-updater toggle.

The PR description frames this as pure UI cleanup ("component itself is left intact... for easy re-enable later") without acknowledging it also breaks the documented support-diagnostics flow for packaged installs. The stated fallback (`meridian telemetry export`) is a terminal command — not realistically something a typical end user being walked through a support ticket can use, and not obviously on `PATH` for a DMG-only install per this repo's own launchd/PATH notes elsewhere.

**Suggestion:** confirm with stakeholders this is intentional (e.g. if the in-flight central-error-observability work is meant to replace manual export) before merging — otherwise keep Export Diagnostics reachable from somewhere (tray menu, a different settings tab) rather than dropping it silently.

### Verified clean
- `cargo check` on `meridian-core`, `meridian`, and `tray/src-tauri` all compile cleanly with the removed settings fields — no stray references left behind.
- No caller passes `initialSection="advanced"`, so there's no dead/blank-modal path from elsewhere in the UI.
- The removed-fields default handling is safe for old `settings.json` files (no `deny_unknown_fields`, extra keys are just ignored).

### Note (disclosed, not a bug)
Losing all per-category notification toggles (plan reminder, worklog ready, system fault, etc.) and the "Open planner each day" toggle is a real reduction in user configurability, not just a "no behavior loss" simplification as worded in the description — though to the PR's credit this is explicitly called out for the planner toggle.

Akarsh-Hegde:
Addressed the review:

**Confirmed regression (fixed in c7e920f5):** hiding the Advanced tab silently took "Export Diagnostics" with it — the only GUI path a packaged install has to hand a developer a local telemetry bundle. Moved that control to Settings → Account (always reachable, next to the version line under "About") instead of leaving it stranded behind the hidden tab. `AdvancedSection.tsx` no longer renders it (avoids a future duplicate if the tab is ever re-enabled), and the doc references in `CLAUDE.md` / `diagnostics.rs` / `commands/settings.rs` that pointed at "Settings → Advanced → Export Diagnostics" now point at the new location. Added `ui/__tests__/advanced-tab-hidden.test.ts` to guard both the tab-hiding and that the diagnostics export stays wired up in Account.

All other Advanced-tab knobs (log level, ETL poll interval, task-routing floors, LLM budget, Jira-updater toggle) remain hidden per the original ask — those are internal/dev-facing, unlike Export Diagnostics which is a support-facing action end users get walked through.

**Noted, not changed:** the reduction in per-category notification configurability was already called out explicitly in the PR description as an intentional simplification, not an oversight — leaving as-is per the original request.

Verified: `cargo test --workspace`, `cargo clippy --workspace -- -D warnings`, `cargo fmt --check`, `npm run build` (ui/), and `bun test` (ui/, 365 passing) all green, plus the full pre-push suite (fmt, clippy, ui build/tests, security audit, cargo test).

Ready for another look / merge.

PR #576, opened 2026-07-24
fix(windows): repair CLI install/detect for Claude, Codex, and Cursor

Akarsh-Hegde:
## Summary
- `install_provider` unconditionally spawned `$SHELL`-or-`/bin/zsh` to run the vendor installer, which fails to even start on Windows (os error 3) before any installer (including a plain `npm i -g ...`) gets to run. `installer_command` now spawns `powershell.exe` directly on Windows (not nested inside `cmd /C`, which can't run PowerShell-only syntax and would need a second, fragile layer of quote-escaping), with an explicit `exit $LASTEXITCODE` so a failing install is reported as failed.
- Cursor's official installer is a bash script that only recognizes Linux/Darwin and has no `sed` - unrunnable on Windows regardless of shell. Cursor publishes a separate PowerShell installer for Windows (`cursor.com/install?win32=true`); `CURSOR_INSTALL_CMD` now has a Windows variant that fetches and pins it via `-replace` instead of `sed`.
- The daemon's own cursor-agent init (`cursor_agent_init.rs`) hardcoded `which`/`bash -c`, neither of which exist on Windows, so a correctly installed cursor-agent was never found by the background summariser - every Cursor Agent session was left pending forever on Windows. It now goes through the same `resolve_cli`/`installer_command` machinery every other CLI lookup in the codebase already uses.
- `candidate_dirs()` was missing cursor-agent's actual Windows install location (`%LOCALAPPDATA%\cursor-agent` - it doesn't install via npm there), and the candidate-dir fallback probe wasn't `PATHEXT`-aware like the PATH-tier probe is, so a just-installed CLI could still resolve to an unspawnable extensionless shim on the immediate post-install re-probe.
- Platform-specific user-facing hints (bail messages, health check remedies) now point at a command that actually works on the running platform instead of always suggesting `curl | bash`.

Note: overlaps in spirit with the still-open #571 (which fixed the `cmd /C` half of the installer-shell bug but not Cursor's Windows-specific installer or the daemon-side `which`/`bash` breakage) - this branch was built from `pre-main` directly rather than stacked on it, so either can land first; happy to rebase/close the redundant one once you've looked at both.

## Test plan
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` - clean
- [x] `cargo test --workspace` - clean except 2 pre-existing, unrelated failures in `tray/src-tauri/src/install.rs` (Windows path-separator assumptions in tests, confirmed via `git diff` this branch never touches that file)
- [x] 47 new/updated unit tests covering: Windows `CURSOR_INSTALL_CMD`/`CURSOR_INSTALL_HINT` (pin, PowerShell not bash), `installer_command` (spawns, propagates success/failure exit codes), `candidate_dirs`/`probe_candidates` (LOCALAPPDATA\cursor-agent, PATHEXT priority), `cursor_agent_init` (`auto_install_enabled` parsing edge cases, `find_cursor_agent` wiring)
- [x] Full pre-push suite (fmt, clippy, ui build/tests, cargo test, security audit) green
- [x] Built a real staging NSIS installer on Windows ARM64 and ran it locally
- [ ] Manual: install Claude/Codex/Cursor from the tray's Settings on a clean Windows machine
- [ ] Manual: confirm a Cursor Agent session actually gets summarised by the daemon on Windows (the `cursor_agent_init` fix)

Co-authored with [Claude Code](https://claude.ai/code/session_0136cCDNbNc6SsX1hYNTVXaG)

### DIFF
--- a/meridian-core/src/lib.rs
+++ b/meridian-core/src/lib.rs
@@ -73,7 +73,7 @@ pub use readers::{
 
 pub use canonical_task::{CanonicalTask, PersonRef, Priority, Provider, StatusCategory, TaskKind};
 
-pub use llm_provider::{LlmProvider, CURSOR_CLI_VERSION, CURSOR_INSTALL_CMD};
+pub use llm_provider::{LlmProvider, CURSOR_CLI_VERSION, CURSOR_INSTALL_CMD, CURSOR_INSTALL_HINT};
 /// The custom-endpoint registry types. `CustomLlmProvider` carries the API key and is the
 /// STORAGE form — see its docs before serialising one anywhere.
 pub use settings::{CustomLlmProvider, SchemaRung};
--- a/meridian-core/src/llm_provider.rs
+++ b/meridian-core/src/llm_provider.rs
@@ -192,11 +192,14 @@ impl LlmProvider {
 ///
 /// This value is the build all of the flag behaviour was verified on. To move it: bump
 /// this constant, re-run the cursor smoke tests, and confirm `--allowed-tools`,
-/// `--workspace` and `--mode ask` still behave.
+/// `--workspace` and `--mode ask` still behave. Shared by both platform variants of
+/// [`CURSOR_INSTALL_CMD`] below - `downloads.cursor.com/lab/<version>/<os>/<arch>/...`
+/// keys the same version string across every OS, only the artifact differs.
 pub const CURSOR_CLI_VERSION: &str = "2026.07.16-899851b";
 
 /// Cursor's official installer, rewritten to install [`CURSOR_CLI_VERSION`].
 ///
+/// # Unix
 /// The `sed` rewrites every version string in their script (all 5 occurrences: the temp
 /// dir, the `downloads.cursor.com/lab/<v>/<os>/<arch>/agent-cli-package.tar.gz` URL, the
 /// final dir, and both symlinks) to the pinned build. Matching the version by PATTERN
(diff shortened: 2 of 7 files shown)


Akarsh-Hegde (commented):
## Review — strict pass

This is a solid, well-scoped fix — genuinely resolves what #571/#574 didn't (Cursor's Windows installer, the daemon-side `which`/`bash` breakage, and PATHEXT-aware candidate resolution). I traced each claim against the source rather than taking the PR description at face value:

- Verified `powershell.exe` is spawned directly (not nested in `cmd /C`), avoiding the double-quote-escaping trap — correct call, and the doc explains why clearly.
- Verified the `; exit $LASTEXITCODE` suffix is necessary and correctly placed (an explicit `exit N` inside `cmd` terminates the process before the suffix runs, so `installer_command("exit 7")` still yields exit code 7 — confirmed by tracing the semantics, not just trusting the test).
- Verified the regex escaping in the PowerShell `-replace` (`\\.` → literal `\.` in the emitted script) matches the Unix `sed` pattern's intent.
- Verified `home_dir_or_cwd()` actually honors `$HOME` first on every platform (including Windows, via `meridian-core/src/util/paths.rs`), so the Windows-gated tests that set `$HOME` to isolate themselves are legitimate, not accidentally testing nothing.
- Confirmed via an existing comment in this same file (`resolve_cli_returns_an_absolute_existing_path`'s doc, "the Windows portability job") that there's real CI coverage for the `#[cfg(windows)]` tests added here, so the 47 new/updated tests aren't just compiling-but-never-running.
- `PATHEXT` ordering (`.COM;.EXE;.BAT;.CMD` before the bare name) correctly prefers the `.CMD` shim over an extensionless POSIX script, matching the new `probe_candidates_prefers_pathext_over_a_bare_extensionless_shim` test.

### Minor note (not blocking)
If Cursor's Windows install script (fetched via `irm` and run via `iex`) is ever pure-PowerShell-cmdlets with no external process invocation, and it fails without an explicit `exit <code>`, `$LASTEXITCODE` could remain unset/stale and the appended `; exit $LASTEXITCODE` would report success on a failed install. This is an inherent limitation of trusting a remote script's own exit signaling rather than a flaw specific to this PR — just flagging since the doc comment reads as a stronger guarantee than it fully is.

### Process
Appreciate that this PR explicitly calls out the overlap with #571 and offers to reconcile, rather than silently duplicating it (unlike #574, which does duplicate #571 wholesale without mentioning it). Given this PR is the more complete fix (also covers Cursor + the daemon's `cursor_agent_init` breakage), I'd suggest landing this one and closing/rebasing #571 and the overlapping half of #574 once this merges.

Akarsh-Hegde:
## Canonical Windows installer fix - ready to merge ?

Flagging this as the single source of truth for the Cursor-on-Windows install gap that the reviews on #571 and #574 raised. Confirmed it:

- adds a `#[cfg(windows)]` `CURSOR_INSTALL_CMD` that fetches Cursor's **PowerShell** installer (`irm 'https://cursor.com/install?win32=true' | iex`), version-pinned via a .NET regex `-replace` - necessary because Cursor's Unix script is bash (`uname -s` only ever answers `Linux`/`Darwin`, so it bails "unsupported OS" even under Git-Bash/WSL) and stock Windows has no `sed`/`bash` on PATH;
- spawns it through `powershell.exe` **directly**, not `cmd.exe`, since `irm`/`iex` are PowerShell aliases with no `cmd` equivalent;
- ships a platform-correct `CURSOR_INSTALL_HINT` so the "cursor-agent not found" hints never point a Windows user at a `curl | bash` command that can't run.

As part of resolving those reviews, **#571** has been slimmed to just its ARM64 update-check messaging and **#574** to just autostart - both had their overlapping (and Windows-incomplete) `detect.rs` installer hunks removed, leaving this PR as the one place the installer fix lives. `mergeable_state: clean`. **Recommend merging this first**, then #571 / #574 (order-independent - no shared files).

PR #577, opened 2026-07-24
fix(jira): stop flapping OAuth sync error on transient refresh failures

Akarsh-Hegde:
## Problem

The `Sync failed: Jira auth failed - re-run `meridian oauth-login jira`` notice pops up at random, then clears itself on the next sync. It is not a dead token - it is a **transient refresh failure treated as fatal**.

`meridian-oauth::flow::post_token` (the OAuth access-token refresh) used a tight **6s timeout** and treated **every** non-2xx status and network error as fatal, **with no retry**. So a momentary network blip or an Atlassian `429`/`5xx` at the ~30-min refresh tick surfaced a scary `Reconnect Jira` sync error - which then vanished on the next successful poll. Classic flap.

## Fix

- **Classify** token-endpoint failures (`TokenError` / `TokenFailure`): network / timeout / `429` / `5xx` => **Transient** (refresh token still valid); a real `4xx` (`invalid_grant`, revoked/rotated, bad client) => **Terminal** (user must re-auth). Carried in the `anyhow` chain so callers classify via `meridian_oauth::is_transient` without string-matching.
- **Retry transient failures** 3x with backoff (400ms -> 1200ms); per-attempt timeout 6s -> 8s. A dead grant still fails on the first try - no backoff wasted.
- **Only surface the re-auth notice for terminal failures**: the sync path (`providers/jira`) keeps the stale cache and stays quiet on a transient blip instead of raising the `Reconnect Jira` notice; the health check downgrades transient from *critical* to *warn*.

## Test / verification

- New unit test `is_transient_classifies_through_context_layers` (chain-walk through `.context()` layers).
- `cargo test -p meridian-oauth` green; daemon `cargo check` + `cargo clippy -D warnings` clean.
- Full pre-push suite (fmt / UI build / clippy / ui tests / audit / cargo test) passed.

Files: `meridian-oauth/src/flow.rs`, `meridian-oauth/src/lib.rs`, `src/intelligence/providers/jira/mod.rs`, `src/health/jira.rs`.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/meridian-oauth/src/flow.rs
+++ b/meridian-oauth/src/flow.rs
@@ -48,6 +48,68 @@ pub struct TokenResponse {
     pub scope: String,
 }
 
+/// Whether a token-endpoint failure is worth retrying, or means the grant is
+/// dead. This is the distinction the background refresh loop needs: a passing
+/// network blip must not be surfaced to the user as "re-authenticate".
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TokenFailure {
+    /// The grant itself was rejected and won't recover on its own — an OAuth 4xx
+    /// (`invalid_grant`, a revoked/expired refresh token, a bad client). The
+    /// stored refresh token is dead; the user must re-authenticate.
+    Terminal,
+    /// A passing condition — a network error, request timeout, HTTP 429, or a
+    /// 5xx from the provider. The stored refresh token is still valid and a
+    /// later retry will succeed.
+    Transient,
+}
+
+/// A classified token-endpoint error. It is threaded into the `anyhow` chain
+/// (via `?`/`.context`) so the sync loop can tell "re-authenticate" from "try
+/// again later" without string-matching messages — see [`is_transient`].
+#[derive(Debug)]
+pub struct TokenError {
+    pub failure: TokenFailure,
+    detail: String,
+}
+
+impl TokenError {
+    fn transient(detail: impl Into<String>) -> Self {
+        Self {
+            failure: TokenFailure::Transient,
+            detail: detail.into(),
+        }
+    }
+    fn terminal(detail: impl Into<String>) -> Self {
+        Self {
+            failure: TokenFailure::Terminal,
+            detail: detail.into(),
+        }
+    }
+    fn is_transient(&self) -> bool {
+        matches!(self.failure, TokenFailure::Transient)
+    }
+}
+
+impl std::fmt::Display for TokenError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&self.detail)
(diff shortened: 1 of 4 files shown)


Akarsh-Hegde (commented):
## Review — strict pass

Good root-cause fix for the flapping notice — the transient/terminal split on 429/5xx/network/timeout vs. real 4xx rejections is well-reasoned, and `is_transient`'s chain-walk through `.context()` layers is correctly tested (`is_transient_classifies_through_context_layers`). Two things worth weighing before merge, both judgment calls rather than clear bugs:

### 1. Retries now run *inside* the dual-locked critical section in `ensure_fresh`
`meridian-oauth/src/jira.rs::ensure_fresh` holds both the intra-process `refresh_lock()` mutex AND the cross-process advisory file lock (`store::lock_provider`) for the full duration of `flow::refresh(...)`. With this PR, a fully-transient run can now take up to ~25.6s worst case (3 attempts × 8s timeout + ~1.6s backoff) instead of the previous single ~6s timeout — roughly a 4x increase in how long that critical section can stay held. That lock exists specifically to prevent double-spending Atlassian's rotating refresh token across concurrent callers (this repo's own memory notes a prior incident here), and at least two call sites already contend for it in this diff alone (`src/health/jira.rs`'s health check and `src/intelligence/providers/jira/mod.rs`'s sync path). Worse network conditions — exactly when transient retries kick in — are also when contention from other processes/callers is most likely. Worth confirming this extended hold time is an acceptable tradeoff, or whether the retry loop should sit outside the lock (only take the lock once a token is actually in hand to persist).

### 2. "2xx with an unparseable body" is bucketed as Transient
That heuristic (proxy/captive-portal interstitial) is reasonable for *some* cases, but it treats a fundamentally different failure mode — an API contract break or corrupted response on an otherwise-successful HTTP call — the same as a network blip. Because there's no cross-tick escalation (every ~30-min tick gets a fresh set of 3 attempts with no memory of prior ticks), a *permanently* broken-but-still-2xx endpoint would never re-surface the old critical "Reconnect Jira" notice at all — it would only ever show as a passive `warn` health check, which is a much easier signal for a user to miss than the active notice this used to raise. Consider either keeping "2xx but unparseable" terminal (distinct from network/429/5xx), or adding an escalate-after-N-consecutive-transient-ticks path so a genuinely permanent failure still eventually surfaces loudly.

Neither of these blocks the core fix — the flapping bug this PR targets is real and the classification for the common cases (network error, timeout, 429, 5xx, real 4xx) is correct. Confirmed via `grep` that Jira is currently the only provider actually exercising this shared `flow::refresh`/`ensure_fresh` path (Trello uses a static token, GitHub uses device-code polling), so this is scoped correctly — no other integration is silently missing the same treatment.

PR #581, opened 2026-07-26
fix(llm): provider install/sign-in detection + dashboard health banners

Akarsh-Hegde:
## Summary

Improvements to LLM-provider management on the Intelligence tab: more honest install/sign-in detection, **in-app browser sign-in for all three CLI providers (Cursor, Codex, and Claude)**, and dashboard surfacing when the in-use provider is unavailable or rate-limited.

## Commits

1. **fix(llm): re-validate `resolve_cli` cache hits with `exists()`** — a CLI uninstalled/moved while the tray runs was reported "installed" forever (process-lifetime cache never revalidated), then failed with a raw OS/shell error on the next invocation. Now revalidates a cache hit with a cheap `Path::exists()` and evicts on a miss.
2. **fix(llm): scope the `resolve_cli` cache read to avoid a re-lock deadlock** — the revalidation initially held the `RESOLVED_BINS` mutex guard across the `if let` body, so the eviction re-lock deadlocked on the same thread. Caught by the pre-push `cargo test`; scoped the read so the guard drops first.
3. **feat(llm): in-app Codex browser sign-in** — one-click "Sign in to Codex" button (`codex login`), mirroring Cursor.
4. **feat(health): dashboard banner when the in-use provider is unavailable** — warns when the selected provider's CLI is missing or its last test failed (hourly summaries paused).
5. **feat(health): soft rate-limit notice, distinct from the unavailable alarm** — a rate-limited provider gets a softer "catching up" notice, since it self-recovers.
6. **test(llm): update connection-phase assertions for the `signInProvider` generalization.**
7. **feat(llm): in-app Claude sign-in (parity with Cursor and Codex)** — "Sign in to Claude" button driving `claude auth login --claudeai` (browser OAuth against the user's Anthropic account, on their Claude subscription). Brings Claude to full parity: install, in-app sign-in, availability banner, and bespoke failure handling.
8. **fix(review): probe cadence + banner dismissal + sign-in mapping** — see the review responses below: TTL-cache the provider-health probe so it isn't re-run every 60s tick; per-alert banner dismissal so dismissing one banner can't suppress a later, worse one; explicit `{cursor,codex,claude}` sign-in mapping that fails loudly instead of defaulting to Cursor.

## Scope note

The in-app sign-in is added for **all three** subscription CLIs — Cursor, Codex, **and Claude** (`claude_sign_in` / "Sign in to Claude"). An earlier revision of this description under-stated the Claude piece; it is included and called out here.

## Testing

- **Cursor and Codex:** verified live on Windows end-to-end (install → sign-in → test), including the availability banner firing on a live uninstall and clearing after reinstall/sign-in.
- **Claude:** `claude auth login --claudeai` is confirmed a real invocation against an installed Claude Code CLI (`auth login`/`logout`/`status` are real subcommands; `--claudeai` is the documented default). The code path is a deliberate line-for-line mirror of the live-verified Cursor/Codex flow, compiles, and passes tests. A live end-to-end browser sign-in **for Claude specifically** is still pending — the OAuth completion needs a human at the browser — and I'm happy to drive that pass before merge.
- **Automated:** `resolve_cli` staleness/deadlock tests, the new `classify_provider_health` ladder test, and the full 755-test lib suite pass. All pre-push waves green (`cargo fmt`, `cargo clippy`, ui build, ui tests, `cargo test`, security audit).

### DIFF
--- a/src/llm/detect.rs
+++ b/src/llm/detect.rs
@@ -828,6 +828,116 @@ pub async fn codex_sign_in() -> InstallOutcome {
     }
 }
 
+/// Run the interactive `claude auth login` on the user's behalf — the "Sign in to Claude"
+/// button in the provider detail view.
+///
+/// `claude auth login` (default `--claudeai`) opens the user's browser to sign into their
+/// Anthropic account on their **Claude subscription** — no API key, nothing metered. A
+/// deliberate mirror of [`cursor_sign_in`]/[`codex_sign_in`]: the browser is enabled (this is
+/// an explicit click, not the daemon's unattended path), stdout is drained live so the
+/// verification URL is surfaced if the browser can't open, and the wait is bounded by
+/// [`INTERACTIVE_LOGIN_TIMEOUT`]. Once it completes, Claude Code persists the auth and every
+/// later run just adopts it — the same "sign in once" model as the other two.
+#[tracing::instrument(skip_all, fields(ok = tracing::field::Empty, cli_path = tracing::field::Empty))]
+pub async fn claude_sign_in() -> InstallOutcome {
+    let label = "claude auth login";
+    let Some(path) = resolve_cli("claude").await else {
+        return install_failed(
+            label,
+            "claude isn't installed yet - install it first".into(),
+        );
+    };
+
+    tracing::info!("llm: launching interactive claude auth login");
+    let mut cmd = Command::new(&path);
+    cmd.arg("auth")
+        .arg("login")
+        .arg("--claudeai")
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::piped())
+        .kill_on_drop(true)
+        .no_window();
+
+    let mut child = match cmd.spawn() {
+        Ok(c) => c,
+        Err(e) => return install_failed(label, format!("couldn't start claude: {e}")),
+    };
+
+    // Drain stdout as it arrives - `claude auth login` prints its verification URL to STDOUT
+    // while it waits, which is exactly what the user needs if the browser doesn't open for them.
+    // Same reasoning as the matching drain in `cursor_sign_in`.
+    let seen = std::sync::Arc::new(std::sync::Mutex::new(String::new()));
+    if let Some(out) = child.stdout.take() {
+        let seen = seen.clone();
+        tokio::spawn(async move {
+            use tokio::io::{AsyncBufReadExt, BufReader};
+            let mut lines = BufReader::new(out).lines();
+            while let Ok(Some(line)) = lines.next_line().await {
+                tracing::info!(line = %line, "claude auth login");
+                let mut buf = seen.lock().unwrap_or_else(|e| e.into_inner());
+                buf.push_str(&line);
+                buf.push('\n');
+            }
+        });
+    }
+    let login_output = |seen: &std::sync::Mutex<String>| -> String {
+        let buf = seen.lock().unwrap_or_else(|e| e.into_inner());
+        tail(buf.trim(), 300)
+    };
+
+    let output = match tokio::time::timeout(INTERACTIVE_LOGIN_TIMEOUT, child.wait_with_output())
+        .await
+    {
+        Ok(Ok(out)) => out,
+        Ok(Err(e)) => return install_failed(label, format!("couldn't run claude auth login: {e}")),
+        Err(_) => {
+            let printed = login_output(&seen);
+            return install_failed(
+                label,
+                if printed.is_empty() {
+                    "the sign-in wasn't finished in time - click Sign in to Claude again".into()
+                } else {
+                    format!(
+                        "the sign-in wasn't finished in time. Finish it by hand with what \
+                             claude printed:\n{printed}"
+                    )
+                },
+            );
+        }
+    };
+
+    if output.status.success() {
+        let span = tracing::Span::current();
+        span.record("ok", true);
+        span.record("cli_path", tracing::field::display(path.display()));
+        tracing::info!("llm: claude auth login succeeded");
+        InstallOutcome {
+            ok: true,
+            message: "Signed in to Claude.".into(),
+            path: Some(path.display().to_string()),
+            command: label.to_string(),
+        }
+    } else {
+        // stderr first (that is where the reason goes), but fall back to what the CLI printed on
+        // stdout - on some failures the URL is the only useful thing said.
+        let reason = tail(String::from_utf8_lossy(&output.stderr).trim(), 400);
+        let reason = if reason.is_empty() {
+            login_output(&seen)
+        } else {
+            reason
+        };
+        install_failed(
+            label,
+            if reason.is_empty() {
+                "claude auth login failed".to_string()
+            } else {
+                reason
+            },
+        )
+    }
+}
+
 /// The last `n` characters of `s`, char-safe.
 ///
 /// The tail is the useful end of CLI output - npm, curl and cursor-agent all print the real
--- a/src/llm/detect.rs
+++ b/src/llm/detect.rs
@@ -934,13 +934,17 @@ static RESOLVED_BINS: std::sync::OnceLock<std::sync::Mutex<HashMap<String, PathB
 /// the slow shell spawn on macOS whenever the current process's `PATH` is already good (e.g.
 /// the daemon's launchd-configured one, per the `resolve_cli` doc above).
 pub async fn resolve_cli(bin: &str) -> Option<PathBuf> {
-    if let Some(hit) = RESOLVED_BINS
+    // Read the cache under a SCOPED lock and drop the guard on this `let` before the block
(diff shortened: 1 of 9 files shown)


Akarsh-Hegde:
## Review

### Overview
Six commits: a `resolve_cli` staleness fix (+ a real deadlock fix caught by `cargo test`), in-app Codex sign-in, a dashboard "provider unavailable" banner, and a softer rate-limit notice. I verified the `claude auth login --claudeai` invocation against an installed Claude Code CLI and traced the new health-check call path end-to-end.

---

### 🔴 Undisclosed scope: a third, untested sign-in flow shipped alongside Codex

The PR description says: *"Codex gets a one-click 'Sign in to Codex' button... mirroring Cursor"* and the Testing section claims *"Verified live on Windows (install → sign-in → test for **Codex and Cursor**...)"*. But the diff also adds a complete, separate `claude_sign_in()` path — a new `detect.rs` function, a new tray command, `lib.rs` wiring, and UI plumbing in `LlmProviderDetail.tsx`/`LlmProviderPicker.tsx` (the `signInProvider` ternary explicitly branches on `providerId === 'claude'`). This is a real, user-facing feature (a live browser OAuth flow against Anthropic) that isn't mentioned anywhere in the PR body, and per the stated testing scope it was never live-verified the way Codex and Cursor were.

I checked whether `claude auth login --claudeai` is even a real invocation (a legitimate risk given it's new and unmentioned) — it is: confirmed against an installed `claude` CLI (`auth login`/`logout`/`status` are real subcommands, `--claudeai` is the documented default auth method). So the command itself isn't fabricated, but that's exactly why this deserves a mention in the PR — please add it to the description and confirm it got the same install → sign-in → test pass on at least one platform before merge.

---

### 🟡 Health-check regression: the "cheap" provider probe isn't cheap in exactly the case it exists for

`in_use_provider_health()` (called from `check_health()` on every 60s poll tick **and** on every on-demand `get_health` command, e.g. whenever the dashboard opens) calls `resolve_cli(bin).await` for any CLI-backed provider. `resolve_cli`'s cache **only stores hits** — the existing doc comment and test (`resolve_cli_does_not_cache_misses`) make this deliberate, so a user installing the CLI mid-session is picked up without a restart. That trade-off was fine when `resolve_cli` was only called from on-demand actions (Test Connection, Rescan). This PR puts it on a recurring timer.

The consequence: for a user whose selected provider's CLI is **not installed** — precisely the state that makes `llm_provider_ok == false` and triggers the new "provider unavailable" banner — `probe_current_path` finds nothing (a packaged `.app`'s launchd `PATH` is stripped), so every single tick falls through to `probe_login_shell`, which spawns a real login-shell subprocess bounded by a 6-second timeout. That happens **every 60 seconds, indefinitely**, for as long as the banner is showing (which could be the entire session). The code comment says *"a cheap install probe + cached last-test read, no metered call... so it's fine on the 60 s health cadence"* — true that it's not metered, but "cheap" doesn't hold for the exact failure mode this feature is built to detect.

It's also not run inside the existing `tokio::join!` for the other three checks — it's a separate sequential `.await` afterward — so on the interactive `get_health` path (dashboard open), a user now waits however long that shell probe takes on top of the existing fast local checks.

Given the fix in commit 1 of this same PR is specifically about avoiding a stale/expensive probe, it'd be worth either (a) caching negative results for a short TTL (e.g. re-probe every N minutes instead of every tick) or (b) only running `in_use_provider_health` on a slower cadence than the rest of the health check.

---

### 🟡 `HealthBanner`'s single `dismissed` flag now covers 4 alert types, not 2

`HealthBanner.tsx`'s `dismissed` state is one shared boolean, set by any of the four banners' Dismiss button and never reset (pre-existing design, unchanged by this PR). Before this PR that meant dismissing the a11y warning could hide a later database error; this PR adds two more banner types onto the same flag. Concretely: if a user dismisses the (lowest-priority) a11y-helper banner, and *later* their configured LLM provider becomes unavailable — meaning **hourly summaries silently stop generating** — the new banner will never show for the rest of the session, since `dismissed` is already `true`. That's a meaningfully worse failure mode than the a11y case, since there's no other UI surfacing "your summaries stopped." Worth considering per-problem-type dismissal (e.g. dismiss keyed to `llm_provider_name + detail`, reset when the signature changes) rather than one global flag, especially now that a real functional regression (paused summaries) rides on it.

---

### 🟢 Verified correct
- The `resolve_cli` deadlock fix (commit 2) is correct: the cache read is scoped to its own `let` so the `MutexGuard` drops before the eviction re-lock — no re-entrant lock on the same thread. The accompanying test (`resolve_cli_drops_a_stale_cache_entry_whose_path_no_longer_exists`) exercises the exact regression described.
- `claude auth login --claudeai` is a real CLI invocation, not a typo/fabrication — confirmed against the installed CLI.

### Minor
- `LlmProviderPicker.tsx`'s `signIn(id)` maps unknown `id`s to `cursor_sign_in` as a fallback. Today this is unreachable (the sign-in button only renders for `cursor`/`codex`/`claude`, gated by the same `signInProvider` list in `LlmProviderDetail.tsx`), but if a fourth provider is ever added to one list and not the other, it'll silently run Cursor's login for the wrong provider instead of failing loudly. Consider an explicit `null`/throw default instead of falling back to `cursor_sign_in`.

### Test coverage
The `resolve_cli` staleness fix has solid, targeted unit test coverage. The UI test update correctly generalizes the Cursor-specific assertion to cover all three subscription providers. The health-banner priority logic and the new `in_use_provider_health` ladder (unavailable → rate-limited → fine) have no automated coverage that I could find in this diff — worth at least a unit test for the priority/precedence logic in `HealthBanner.tsx` given it now has 4 mutually-exclusive states instead of 2.

---

**Bottom line:** the core fixes (stale-cache revalidation, the deadlock) are solid and well-tested. The main things I'd want addressed before merge: disclose and verify the Claude sign-in flow the same way Codex was, and reconsider the probe cadence so the banner's own detection mechanism doesn't spawn a shell process every minute for the users it's meant to help.

Akarsh-Hegde:
## Review addressed — pushed `2710cd13`

Thanks for the thorough pass — a couple of these were genuine, and the deadlock one you flagged as "verified correct" is exactly why the pre-push `cargo test` earned its keep. Point by point:

### 🔴 Undisclosed Claude sign-in scope — disclosed; verification status made honest
You're right it wasn't in the description. Updated the PR body to call out the Claude flow explicitly (commit 7) alongside Cursor/Codex. On verification: `claude auth login --claudeai` is confirmed a real invocation against an installed CLI (as you also independently checked), the code is a line-for-line mirror of the live-verified Cursor/Codex path, and it compiles + passes tests. The one thing I can't self-complete is the browser OAuth handshake itself — that needs a human at the browser, the same way the Cursor/Codex passes were driven by hand. I've flagged it as *pending* in the Testing section rather than claim a live pass that didn't happen, and I'm happy to drive that verification before merge.

### 🟡 Health-check probe cadence — fixed (`2710cd13`)
Agreed, a real regression: `in_use_provider_health` ran `resolve_cli` on every 60s tick **and** every `get_health`, and for an uninstalled provider that can't hit `resolve_cli`'s hit-only cache, so it fell through to the login-shell probe each time — a shell spawn every minute for exactly the users the banner is meant to help. Took your option (a): the result is now cached on a 5-minute TTL keyed by provider id (switching providers invalidates immediately), so it's at most one probe per window while still raising/clearing within the hourly-summary cadence it guards. Also split the probe from a pure classifier, so the sequential `.await` on the interactive `get_health` path only pays the probe cost once per window, not every open.

### 🟡 Shared `dismissed` flag — fixed (`2710cd13`)
Good catch, and agreed the new "summaries paused" banner makes this materially worse than the a11y case. Replaced the single global boolean with a per-alert **signature** (`db:… | provider-unavailable:name:detail | provider-ratelimit:… | a11y`). Dismissal is scoped to that exact signature; when the active problem changes, the old dismissal no longer matches and the banner reappears — so dismissing a11y can no longer suppress a later provider-unavailable banner.

### Minor: `signIn(id)` cursor fallback — fixed (`2710cd13`)
Replaced the ternary that defaulted unknown ids to `cursor_sign_in` with an explicit `{cursor, codex, claude}` map; an unmapped id now returns a loud failure outcome instead of silently running Cursor's login for the wrong provider.

### Test coverage — added
Extracted the unavailable → rate-limited → fine ladder into a pure `classify_provider_health()` and added a unit test pinning every rung (`classify_provider_health_ranks_unavailable_over_rate_limited_over_fine`). The HealthBanner priority/dismiss logic is React-side with no pure extraction yet, so I left a dedicated banner-priority unit test out of this round — can follow up if you want it.

---

**Bottom line:** all four code findings (probe cadence, banner dismissal, sign-in mapping, ladder test coverage) are fixed and green — all pre-push waves pass. The single remaining item before merge is the **live Claude sign-in pass** (browser OAuth needs a human). Say the word and I'll drive that; otherwise the flow is command-verified and at parity with the live-tested Cursor/Codex path.

PR #584, opened 2026-07-27
feat(observability): ship platform attributes and surface the Support ID

Akarsh-Hegde:
Two gaps found while verifying the first staging release of central error reporting (#545).

## 1. Platform attributes were absent, not merely unqueried

`os.type`, `host.arch` and `app.install_mode` are all on `redact::SAFE_STRING_KEYS`, so reading the allowlist suggests they were already available. They were not — `observability::init`'s `Resource::new` set exactly four attributes, and `Resource::new` runs no detectors (unlike `Resource::default`), so the SDK contributed nothing either. Confirmed against a live spool payload, not just by reading the code:

```
deployment.environment
host.name
service.name
service.version
```

A Windows error and a macOS error were therefore indistinguishable in the backend — on the one path that carries the actual error signal. Sentry has OS and arch via `sentry-contexts`, but Sentry only sees tray crashes, so the platform breakdown existed precisely where there was least data.

This is the same **allowlisted-but-never-populated** trap as `service.instance.id` in the #545 review. Documented in CLAUDE.md so the next person checks `Resource::new` rather than the allowlist.

Populating these needs no change to the redaction boundary — they are enum-like build facts that structurally cannot carry user content.

Shipped as the raw Rust constants (`macos`/`windows`, `aarch64`/`x86_64`) rather than the OTel semconv spellings (`darwin`, `arm64`). Nothing downstream parses them as semconv enums, and a mapping layer is a lossy translation with no consumer asking for it.

## 2. The host pseudonym was unreachable by support

`host.name` ships as a one-way 16-hex pseudonym (the #545 blocker fix). It is the **only** identifier on an error row, and nothing in the app showed a user their own value — so a customer writing in about a crash could not be matched to their errors at all. Aggregate dashboards worked; helping a specific person did not.

Now surfaced as **Support ID** in Settings → Account (with copy-to-clipboard, since a mistyped 16-hex string silently matches nothing) and in a synthesized `bundle-info.txt` inside export bundles.

**Why this shape rather than automatic attribution.** The alternative considered was adding the existing analytics `device_id` to the OTel resource, which would join error rows to PostHog Persons and out falls the signed-in email, with no user involvement. Rejected: it re-identifies the error stream by the back door, defeats the pseudonymisation entirely, and would make the consent copy ("stripped on your device first") describe something no longer true in effect. Surfacing the ID keeps attribution per-incident and consented. If proactive outreach is wanted later, the honest version is an explicit opt-in toggle, not a quiet field.

`bundle-info.txt` deliberately carries nothing the automatic ship leg wouldn't already send — no raw hostname, no account email, no paths.

## The drift hazard, and the test that pins it

Both the displayed and the shipped value now come from one new `redact::local_host_pseudonym`. This is the load-bearing part: they are computed in different crates, and any divergence in how the hostname is read (`to_string_lossy` vs `to_str`, a trimmed `.local`, case) yields an ID that matches no rows — and nothing fails loudly. `displayed_pseudonym_matches_shipped` asserts the Settings value equals what `keep_attribute` writes for `host.name`.

## Tests

- `displayed_pseudonym_matches_shipped` — the parity invariant above.
- `machine_shape_attributes_survive_verbatim` — the three new attributes survive `keep_attribute` **unaltered**. Allowlisted strings still pass through `scrub_paths`, so "on the allowlist" is not the same as "arrives intact" — the same shape of trap as the one this PR fixes.
- `build_export_bundle_includes_uncounted_bundle_info` — the bundle carries the pseudonym, does **not** carry the raw hostname, and the synthetic header is not counted in `file_count` (which reports telemetry files collected).

## Verification

Daemon `cargo fmt` + `clippy -D warnings` + 770 tests. Tray verified separately since pre-push does not cover that workspace: fmt, clippy, 161 tests. UI build + typecheck (45 pre-existing `bun:test` type errors on `pre-main`, unchanged) + 381 bun tests.

**Not verifiable until a release rebuilds and reinstalls** — these are compile-time-resolved resource attributes in the packaged daemon, so the current staging install will not show them.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -367,6 +367,30 @@ supported way to read logs locally, replacing the old JSONL-tailing UI and the
 old bash `meridian logs` (which used to tail launchd-redirected stdout/stderr
 text).
 
+**Two couplings that silently delete error coverage.** Both have bitten once;
+neither fails loudly, and neither is visible from the call site.
+
+1. **The `EnvFilter` decides what is captured at all — before the spool, before
+   severity, before redaction.** `EnvFilter` matches directives against targets
+   by string **prefix**, so the `meridian` directive covers `meridian_core::*`,
+   `meridian_tray_lib::*` and `meridian_oauth::*` for free — but nothing else.
+   Any crate whose target does not start with `meridian` is discarded entirely
+   unless named in `build_default_filter`. That is why `CAPTURE_DIRECTIVES`
+   exists (`screenpipe_screen`/`screenpipe_a11y` — the OCR + accessibility
+   stack, ~48 error sites that were invisible). **Adding a dependency that does
+   real work means adding its target there**, and auditing its `warn!`/`error!`
+   bodies first: third-party crates were written to print to a terminal and may
+   interpolate content that must never egress.
+2. **The attribute allowlist is keyed on the names you actually type.**
+   `tracing::warn!(error = %e, …)` emits the attribute key `error`, not the
+   semconv `error.message`. Anything not on `redact::SAFE_STRING_KEYS` is
+   dropped from the ship leg, so a well-instrumented error can arrive as a bare
+   static string with its cause stripped. When adding a field to a WARN/ERROR
+   site that you would need in order to debug it remotely, check it is on that
+   list — and if it names the user's data (ticket key, file path, app name,
+   window title, hour/day) it must stay off, so put the diagnostic value in the
+   message or an allowlisted key instead.
+
 The only thing this pipeline structurally can't capture is a hard crash
 (panic before `init` runs, segfault, OOM kill) — for that, launchd's own
 stdout/stderr redirect (`~/.meridian/logs/<service>.log` /
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -400,13 +400,34 @@ spans egress at all, and `host.name` is replaced by a stable **pseudonym**
 macOS is routinely the account holder's real name. The tray's Sentry
 `before_send` applies the identical pseudonym to `event.server_name`.
 
+That pseudonym is the ONLY identifier on an error row — nothing associates it
+with an account, and the hash is one-way. So it is surfaced to the user as
+**Support ID** (Settings → Account, and `bundle-info.txt` inside an export
+bundle), which is what makes a support ticket traceable to its error rows at
+all. Both the displayed value and the shipped one come from
+`redact::local_host_pseudonym` **on purpose** — computing either side
+separately would let them drift, and the user would quote an ID matching
+nothing, silently (`displayed_pseudonym_matches_shipped` pins this).
+
+**Which resource attributes actually ship** is a separate question from the
+allowlist, and the two are easy to confuse: `redact::SAFE_STRING_KEYS` lists
+many keys that nothing populates (`service.instance.id`, `os.version`,
+`app.version`, …). `observability::init`'s `Resource::new` sets exactly:
+`service.name`, `service.version`, `host.name`, `deployment.environment`,
+`os.type`, `host.arch`, `app.install_mode` — and `Resource::new` runs NO
+detectors, so the SDK contributes nothing either. **Allowlisted ≠ present:
+before relying on an attribute in a query or dashboard, check it is set here.**
+
 **Export Diagnostics remains the manual path**, unchanged and independent of
 consent: tray Settings → Account → **Export Diagnostics** (or `meridian
 telemetry export`) bundles the spool + the launchd crash-safety-net logs into
 a `.tar.gz` the user hands to support, imported by hand with `meridian
-telemetry import <bundle> --endpoint <url> --auth <base64>`. Retention
-(default 7 days, `MERIDIAN_TELEMETRY_RETENTION_DAYS`) applies to both
-`pending/` and `sent/`, regardless of shipping status.
+telemetry import <bundle> --endpoint <url> --auth <base64>`. The bundle also
+carries a synthesized `bundle-info.txt` (Support ID, version, channel, os,
+arch) so a hand-delivered archive can be joined to that machine's already-
+ingested rows; it deliberately contains nothing the automatic ship leg wouldn't
+already send. Retention (default 7 days, `MERIDIAN_TELEMETRY_RETENTION_DAYS`)
+applies to both `pending/` and `sent/`, regardless of shipping status.
 
 - **Rust**: `tracing::info!/warn!/error!/debug!` with **structured fields** — never format data values into the message string (already enforced).
 - **Wrap discrete operations in spans** (`tracing::info_span!` / `debug_span!`) and put the meaningful inputs, outputs, and metrics as **span attributes**, not buried in log lines. For an LLM/model call, capture the EXACT input as sent and output as received (post-cap/post-template — reflect any truncation that actually happened), plus real token counts/latency. See `src/llm/resolver.rs`'s `llm.call` span tree (request → infer → response) and `src/worklog_pipeline/distiller`'s `distil.run` span for the reference shape.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -412,11 +412,14 @@ nothing, silently (`displayed_pseudonym_matches_shipped` pins this).
 **Which resource attributes actually ship** is a separate question from the
 allowlist, and the two are easy to confuse: `redact::SAFE_STRING_KEYS` lists
 many keys that nothing populates (`service.instance.id`, `os.version`,
-`app.version`, …). `observability::init`'s `Resource::new` sets exactly:
-`service.name`, `service.version`, `host.name`, `deployment.environment`,
-`os.type`, `host.arch`, `app.install_mode` — and `Resource::new` runs NO
+`app.version`, `app.install_mode`, …). `observability::init`'s `Resource::new`
+sets exactly: `service.name`, `service.version`, `host.name`,
+`deployment.environment`, `os.type`, `host.arch` — and `Resource::new` runs NO
 detectors, so the SDK contributes nothing either. **Allowlisted ≠ present:
 before relying on an attribute in a query or dashboard, check it is set here.**
+Note `observability::init` runs in BOTH the daemon and the tray, so any
+attribute added there must be correct for both — `app.install_mode` is
+deliberately left unset for exactly this reason (see the comment there).
 
 **Export Diagnostics remains the manual path**, unchanged and independent of
 consent: tray Settings → Account → **Export Diagnostics** (or `meridian
--- a/src/observability/install_mode.rs
+++ b/src/observability/install_mode.rs
@@ -44,11 +44,28 @@ pub(super) fn is_canonical_install() -> bool {
         return false;
     };
     match std::env::current_exe() {
-        Ok(exe) => exe == meridian_dir.join("bin").join("meridian"),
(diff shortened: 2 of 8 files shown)


Akarsh-Hegde:
## Review pass on my own first commit - two defects fixed, one design question raised

### `app.install_mode` would have been wrong in every packaged install (fixed, removed)

Its only available signal is `is_canonical_install()`, which compares `current_exe()` against `~/.meridian/bin/meridian`. That is a **daemon-shaped test**, but `observability::init` also runs in the tray - `lib.rs:92` calls `meridian::observability::init("meridian-tray")` unconditionally - and the tray's exe lives inside the `.app` bundle. Every tray-originated row on a fully packaged install would have been labelled `source`.

A wrong platform attribute is worse than an absent one: nothing contradicts it, and it gets trusted in a dashboard filter for months. Removed, with a comment at the site explaining why it stays unset - the key is on `SAFE_STRING_KEYS`, so its absence will otherwise read as an oversight and get "fixed" back in.

`deployment.environment` already separates source builds (`dev`) from released ones, which was the distinction actually wanted. `os.type` and `host.arch` are unaffected - both are compile-time constants, correct in either process.

### "changes if you rename the device" was false (fixed, clause cut)

The Support ID description asserted this to users. The pseudonym hashes `gethostname()`, which on macOS is the Unix hostname, **not** the Computer Name that a rename in System Settings edits. Verified on this Mac:

```
gethostname()          Unknown_a2:97:03:78:a9:6a
scutil --get HostName          not set
scutil --get LocalHostName     Mac-Studio-2
scutil --get ComputerName      Mac Studio
```

The two are unrelated here. Wrong copy in the one place a user would check it, so the clause is gone; the rest of the description holds.

### Pre-existing, raised not fixed: is the hostname a stable machine identifier?

That output is the uncomfortable part. With `HostName` unset - the macOS default - `gethostname()` fell back to a **MAC-address-derived string**, not to LocalHostName. #545's design assumes `host.name` is stable per machine; that assumption underpins both the "one user hitting this 500 times vs 500 users once" argument for pseudonymising rather than dropping, and the Support ID feature in this PR. If that value can shift (interface change, DHCP state), the pseudonym shifts with it and both break quietly.

I have not established whether it actually does shift, and it is out of scope here - this PR surfaces the existing identifier, it does not choose it. Flagging it rather than quietly building on top: if it needs fixing, the change is to derive `host.name` from a stable platform ID (macOS `IOPlatformUUID`, Windows `MachineGuid`), which is a separate decision with its own Sentry-correlation implications.

One incidental note in its favour: on this machine the raw hostname carries no personal name at all, so the privacy motivation for pseudonymising is untouched either way.

### Also worth stating

The resource set is built once per process in `Resource::new` at init, so these attributes are fixed for the process lifetime. Nothing re-reads them if the install changes underneath a running daemon.

Verified after the fixes: daemon `fmt` + `clippy -D warnings` + 770 tests, tray 161 tests, UI build + 381 bun tests. Pre-push suite passed.

Akarsh-Hegde:
## Windows audit - found central error reporting was inert there entirely

Checked every piece this PR touches against Windows. One real bug, pre-existing from #545, now fixed in `5345ac03`.

### 🔴 `is_canonical_install()` could never match a packaged Windows install

```rust
exe == meridian_dir.join("bin").join("meridian")   // hardcoded, no suffix
```

The tray stages the daemon as **`meridian.exe`** (`backend_install::DAEMON_FILE`, correctly cfg-gated for Windows). So `current_exe()` ends in `.exe`, the comparison never matched, and `resolve_otlp_target()` fell through to the **DEV** branch - which requires `otlp_enabled` plus local OpenObserve credentials a packaged install never has. It resolved no target, so **no Windows install has ever shipped a single error record.**

The failure was invisible from both ends. The shipper logs at `debug!` when it finds no target, spool files aged out under normal retention, and the backend can't report traffic it never expected. Everything else on the Windows path was already right - the release job bakes `MERIDIAN_CENTRAL_OTLP_ENDPOINT` and `MERIDIAN_CENTRAL_TELEMETRY_TOKEN` into `target/release/meridian.exe` exactly as macOS does. It was one absent suffix.

Fixed via `std::env::consts::EXE_SUFFIX` rather than a second cfg-selected constant, so the daemon can't drift from the tray again if a platform is added. `staged_daemon_file_name_carries_the_platform_exe_suffix` pins it on whichever OS CI runs.

Worth noting this is only detectable by reading both crates together - the tray's constant is correct and well-commented, and the daemon's looks fine in isolation. Two crates, no shared constant, silent on mismatch.

### 🟢 Everything else checks out

**`os.type` / `host.arch`** - compile-time constants, correct on Windows (`windows` / `x86_64`), and the Windows job sets `MERIDIAN_CHANNEL` the same as macOS, so `deployment.environment` works too.

**Redaction** - correct on Windows, but nothing pinned it, so I added `windows_paths_are_scrubbed_but_not_mangled`. `C:\Users\<name>` is scrubbed (the regex is `(?i)([/\\](?:Users|home)[/\\])`, backslash-aware), the path survives the blob rule intact, and a secret on the same line still dies whole. It holds for a slightly lucky reason worth recording: `\` is **not** in the blob character class `[A-Za-z0-9+/=_\-]{32,}`, so backslash-separated paths can't reach the 32-char threshold the way `/`-separated ones do - Windows paths are structurally safer here than Unix ones. `looks_like_path` covers drive letters independently.

**Support ID** - `gethostname()` on Windows returns the computer name, which is *more* stable than the macOS case flagged in the previous comment (where an unset `HostName` fell back to a MAC-derived string). The macOS stability question stands; Windows is fine.

**Export bundle** - `bundle-info.txt` is platform-neutral. The launchd-log fold is macOS-only by nature and already degrades to "no such files", so a Windows bundle simply carries the spool.

### Impact

Windows telemetry starts flowing on the first release carrying this. Until then, `deployment.environment` plus the now-populated `os.type` is what will tell you whether Windows installs actually appear - which, given the above, is the specific thing to check rather than assume.

Akarsh-Hegde:
## Error-coverage audit - two defects that deleted most of this feature's value

Followed up on "is every operation's error actually reaching the backend". Answer was no, in two independent ways. Both fixed in `9ec1900e`.

### 🔴 1. The OCR + accessibility stack was never captured at all

`EnvFilter` matches directives against targets by **string prefix**. `meridian=info` therefore covers `meridian_core::*`, `meridian_tray_lib::*` and `meridian_oauth::*` for free - every first-party target starts with those bytes. The in-process capture crates do not:

| target | matched a directive? |
|---|---|
| `meridian_core::readers::tasks` | yes, via `meridian` prefix |
| `meridian_tray_lib::commands::health` | yes, via `meridian` prefix |
| `screenpipe_screen::*` | **no** |
| `screenpipe_a11y::*` | **no** |

All ~48 `warn!`/`error!` sites in those two crates were discarded **at the subscriber** - before the spool, before the severity filter, before redaction. Not shipped, not in `meridian logs`, not in an export bundle. Screen-capture bail-outs, Apple Vision OCR handler failures, `AXObserver`/`CGEventTap` registration failures, monitor enumeration errors: all structurally invisible.

This is the worst of the two, because capture is the daemon's **only data source**. A silent failure there is indistinguishable downstream from "the user did nothing today" - which is exactly the support case you cannot currently diagnose.

Fixed with `CAPTURE_DIRECTIVES` (`warn`, dropping to `error` when the user selects ERROR).

**Privacy audit done before enabling**, since these are third-party crates written to print to a terminal: every `warn!`/`error!` body in both crates is an infrastructure failure - monitor ids, image dimensions, word-range indices, error `Display`s. **None** interpolates a window title, app name, URL, or OCR text. Noted at the constant that bumping the fork revision requires re-auditing, since this is precisely the change that could start shipping captured content.

The tests assert the **negative** too - that the old filter did not pass capture targets - because otherwise the regression check would be vacuous, and that unrelated third-party targets (`reqwest`) stay out, so this is a narrow widening rather than a global one.

### 🔴 2. Every shipped error arrived with its cause stripped

`SAFE_STRING_KEYS` carried `error.message`. The key this codebase actually emits is bare **`error`** - CLAUDE.md's own mandated convention, used by **203 of ~419** warn/error sites. The allowlist dropped it.

Decoded from the live staging spool:

```
engine  = claude
attempt = 2
error   = "claude timed out after 60s"
```

What reached the backend: the static message, plus `attempt` (numeric, kept unconditionally). Not the engine. Not the cause. Every AI-provider failure, summariser failure and ETL failure has been shipping as a bare string.

Allowlisted `error` as **free text** - an `anyhow` chain splices arbitrary `Display` output, so it needs the full URL/email/token scrub, not just the path scrub - plus `provider` and `engine`, enum-like values naming our own components.

**Deliberately stopped there.** The census also surfaced `task`/`task_key` (ticket ids), `file`/`path`, `app_name`, `hour`/`day`. Those are the user's data, are not needed to diagnose a failure, and stay dropped - a test asserts they still are, so the widening cannot quietly grow by precedent.

Also allowlisted `code.filename`, the only one of the four `code.*` attributes missing.

### Follow-ups, with assessment rather than a list

**Daemon panics never reach the backend** - real, deliberately not fixed here. The tray has Sentry with the `panic` feature; the daemon has nothing, so a panic reaches only launchd's stderr file. A `panic::set_hook` emitting `tracing::error!` would *look* like coverage but largely isn't: `SpoolClient::send` defers the write via `spawn_blocking`, and the OTel batch processor batches before that, so a panicking process is unlikely to flush. The honest fix is Sentry in the daemon (new dep, DSN wiring, redaction parity) - worth doing, not worth faking.

**`let _ =` sites: 84 hits, not a finding.** Sampled twelve: `remove_dir_all` on temp dirs, `stdout().flush()`, `launchctl` during uninstall, `stamp_sync_error` (already recording an error), `force_flush` on shutdown. Best-effort cleanup where the error is genuinely uninteresting. No evidence of swallowed operational failures; reporting the count without the sample would have been misleading.

**Volume needs one staging round before prod.** Enabling 48 new WARN sites on a stack that runs at frame cadence is the flood risk. Two candidates to watch: `"Apple Vision OCR skipped {} words with missing bounding boxes"` and `"tesseract: OCR failed"`. If either turns out to be per-frame, it belongs in `NOISE_SUBSTRINGS`. Staging is the right place to measure that, and it is the reason to check record counts by body on the first release rather than assuming.

### Scope note

This widens PR #584 beyond its original title. The changes are in the same files (`redact.rs`, `observability/mod.rs`) and the same feature area, and #584 has not merged, so splitting them into a stacked PR would add process without letting either land independently. Happy to split if you would rather review them separately.

Akarsh-Hegde:
## Superseded - split into three PRs

Closing this in favour of three focused PRs, as discussed. The branch is left in place (nothing force-pushed or deleted); all four commits were cherry-picked out.

| PR | Scope | Depends on |
|---|---|---|
| #586 | Platform attributes (`os.type`, `host.arch`) + Support ID | - |
| #587 | Windows packaged-install detection (`.exe` suffix) | - |
| #588 | OCR/a11y capture directives + `error` cause allowlist | stacked on #586 |

**Why this shape.** #587 is a standalone bug - Windows has never shipped a single error record - so it should not wait behind a feature review. #588's test anchors on a test added by #586, so those two stack; review only `fbc3cf3b` on #588, and once #586 merges that diff collapses to the single commit.

The two conflicts hit while splitting were both test-placement, resolved so each branch carries only its own test: the Windows redaction test lives on #587, the pseudonym-parity and error-cause tests on #586 and #588 respectively. Each branch verified independently after resolution (fmt, clippy `-D warnings`, full daemon suite; tray suite for #586, the only one touching it).

All review discussion above stays readable here for history.

PR #586, opened 2026-07-27
feat(observability): ship platform attributes and surface the Support ID

Akarsh-Hegde:
Split out of #584 (closed). This is the feature half; the two bug fixes it was bundled with are #587 and #588.

Two gaps found while verifying the first staging release of central error reporting (#545).

## 1. Platform attributes were absent, not merely unqueried

`os.type` and `host.arch` are on `redact::SAFE_STRING_KEYS`, so reading the allowlist suggests they were already available. They were not - `observability::init`'s `Resource::new` set exactly four attributes, and `Resource::new` runs no detectors (unlike `Resource::default`), so the SDK contributed nothing either. Confirmed against a live spool payload rather than by reading code:

```
deployment.environment
host.name
service.name
service.version
```

A Windows error and a macOS error were therefore indistinguishable on the one path that carries the actual error signal. Sentry has OS and arch via `sentry-contexts`, but Sentry only sees tray crashes, so the platform breakdown existed precisely where there was least data.

Same **allowlisted-but-never-populated** trap as `service.instance.id` in the #545 review. CLAUDE.md now says to check `Resource::new`, not the allowlist.

Shipped as the raw Rust constants (`macos`/`windows`, `aarch64`/`x86_64`) rather than OTel semconv spellings (`darwin`, `arm64`) - nothing downstream parses these as semconv enums, and a mapping layer is a lossy translation with no consumer asking for it.

`app.install_mode` was in the first draft and **removed on review**: its only signal is `is_canonical_install()`, an exe-path check shaped for the daemon, but `observability::init` also runs in the tray (`lib.rs:92`) whose exe lives in the `.app` bundle - so every tray row on a packaged install would have been labelled `source`. A wrong platform attribute is worse than an absent one. A comment at the site records why it stays unset, since the key is on `SAFE_STRING_KEYS` and its absence otherwise reads as an oversight.

## 2. The host pseudonym was unreachable by support

`host.name` ships as a one-way 16-hex pseudonym (the #545 blocker fix). It is the **only** identifier on an error row, and nothing showed a user their own value - so a customer writing in about a crash could not be matched to their errors at all. Aggregate dashboards worked; helping a specific person did not.

Now surfaced as **Support ID** in Settings -> Account (with copy-to-clipboard, since a mistyped 16-hex string silently matches nothing) and in a synthesized `bundle-info.txt` inside export bundles.

**Why this shape rather than automatic attribution.** The alternative was adding the existing analytics `device_id` to the OTel resource, joining error rows to PostHog Persons and yielding the signed-in email with no user involvement. Rejected: it re-identifies the error stream by the back door and would make the consent copy ("stripped on your device first") untrue in effect. Surfacing the ID keeps attribution per-incident and consented. If proactive outreach is wanted later, the honest version is an explicit opt-in toggle, not a quiet field.

`bundle-info.txt` carries nothing the automatic ship leg wouldn't already send - no raw hostname, no account email, no paths.

## The drift hazard, and the test that pins it

Both the displayed and shipped values come from one new `redact::local_host_pseudonym`. They are computed in different crates, and any divergence in how the hostname is read (`to_string_lossy` vs `to_str`, a trimmed `.local`, case) yields an ID matching no rows - silently. `displayed_pseudonym_matches_shipped` asserts the Settings value equals what `keep_attribute` writes for `host.name`.

A copy claim was also corrected on review: "changes if you rename the device" was false. The pseudonym hashes `gethostname()`, the Unix hostname, not the Computer Name a rename edits. Verified locally - `gethostname()` returned `Unknown_<mac-address>` while ComputerName was "Mac Studio".

## Known issue, raised not fixed

That check surfaced a pre-existing question. With `HostName` unset - the macOS default - `gethostname()` fell back to a **MAC-address-derived string**. #545 assumes `host.name` is stable per machine; that assumption underpins both the grouping argument for pseudonymising rather than dropping, and the Support ID feature here. If the value can shift (interface change, DHCP state), both break quietly.

Not established that it does shift, and out of scope - this PR surfaces the existing identifier, it does not choose it. The fix, if needed, is deriving `host.name` from a stable platform ID (`IOPlatformUUID` / `MachineGuid`), with its own Sentry-correlation implications. Windows is unaffected: `gethostname()` there returns the computer name.

## Tests

- `displayed_pseudonym_matches_shipped` - the parity invariant.
- `machine_shape_attributes_survive_verbatim` - the new attributes survive `keep_attribute` **unaltered**. Allowlisted strings still pass through `scrub_paths`, so "on the allowlist" is not "arrives intact".
- `build_export_bundle_includes_uncounted_bundle_info` - bundle carries the pseudonym, not the raw hostname, and the synthetic header is not counted in `file_count`.

## Verification

Daemon `fmt` + `clippy -D warnings` + tests. Tray verified separately since pre-push does not cover that workspace: fmt, clippy, 161 tests. UI build + typecheck (45 pre-existing `bun:test` type errors on `pre-main`, unchanged) + 381 bun tests.

**Not verifiable until a release rebuilds and reinstalls** - these are compile-time-resolved resource attributes in the packaged daemon.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -396,9 +396,19 @@ Redaction applies to the **ship leg only** — it produces a separate stripped
 copy to POST, so local capture and `meridian logs` stay full-fidelity. Two
 things it guarantees that are easy to regress: only WARN+ logs / ERROR-status
 spans egress at all, and `host.name` is replaced by a stable **pseudonym**
-(`redact::pseudonymize_host`) rather than shipping the raw hostname, which on
+(`redact::local_host_pseudonym`) rather than shipping the raw hostname, which on
 macOS is routinely the account holder's real name. The tray's Sentry
 `before_send` applies the identical pseudonym to `event.server_name`.
+
+**The pseudonym is seeded from a hardware id, NOT the hostname** — and the ship
+leg *replaces* `host.name` rather than hashing whatever was captured. This is
+not a stylistic choice: on macOS `HostName` is unset by default, so the kernel
+hostname is derived from the network (measured on a dev Mac, `hostname` was
+byte-identical to the router's reverse-DNS record for the current IP, itself
+derived from a per-network randomised Wi-Fi MAC). Seeding from it gave the same
+machine a new identity on every network change, silently breaking the grouping
+the value exists for. See `telemetry_spool::machine_id` before touching any of
+this; `pseudonym_is_independent_of_the_captured_hostname` pins it.
 
 That pseudonym is the ONLY identifier on an error row — nothing associates it
 with an account, and the hash is one-way. So it is surfaced to the user as
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -400,13 +400,34 @@ spans egress at all, and `host.name` is replaced by a stable **pseudonym**
 macOS is routinely the account holder's real name. The tray's Sentry
 `before_send` applies the identical pseudonym to `event.server_name`.
 
+That pseudonym is the ONLY identifier on an error row — nothing associates it
+with an account, and the hash is one-way. So it is surfaced to the user as
+**Support ID** (Settings → Account, and `bundle-info.txt` inside an export
+bundle), which is what makes a support ticket traceable to its error rows at
+all. Both the displayed value and the shipped one come from
+`redact::local_host_pseudonym` **on purpose** — computing either side
+separately would let them drift, and the user would quote an ID matching
+nothing, silently (`displayed_pseudonym_matches_shipped` pins this).
+
+**Which resource attributes actually ship** is a separate question from the
+allowlist, and the two are easy to confuse: `redact::SAFE_STRING_KEYS` lists
+many keys that nothing populates (`service.instance.id`, `os.version`,
+`app.version`, …). `observability::init`'s `Resource::new` sets exactly:
+`service.name`, `service.version`, `host.name`, `deployment.environment`,
+`os.type`, `host.arch`, `app.install_mode` — and `Resource::new` runs NO
+detectors, so the SDK contributes nothing either. **Allowlisted ≠ present:
+before relying on an attribute in a query or dashboard, check it is set here.**
+
 **Export Diagnostics remains the manual path**, unchanged and independent of
 consent: tray Settings → Account → **Export Diagnostics** (or `meridian
 telemetry export`) bundles the spool + the launchd crash-safety-net logs into
 a `.tar.gz` the user hands to support, imported by hand with `meridian
-telemetry import <bundle> --endpoint <url> --auth <base64>`. Retention
-(default 7 days, `MERIDIAN_TELEMETRY_RETENTION_DAYS`) applies to both
-`pending/` and `sent/`, regardless of shipping status.
+telemetry import <bundle> --endpoint <url> --auth <base64>`. The bundle also
+carries a synthesized `bundle-info.txt` (Support ID, version, channel, os,
+arch) so a hand-delivered archive can be joined to that machine's already-
+ingested rows; it deliberately contains nothing the automatic ship leg wouldn't
+already send. Retention (default 7 days, `MERIDIAN_TELEMETRY_RETENTION_DAYS`)
+applies to both `pending/` and `sent/`, regardless of shipping status.
 
 - **Rust**: `tracing::info!/warn!/error!/debug!` with **structured fields** — never format data values into the message string (already enforced).
 - **Wrap discrete operations in spans** (`tracing::info_span!` / `debug_span!`) and put the meaningful inputs, outputs, and metrics as **span attributes**, not buried in log lines. For an LLM/model call, capture the EXACT input as sent and output as received (post-cap/post-template — reflect any truncation that actually happened), plus real token counts/latency. See `src/llm/resolver.rs`'s `llm.call` span tree (request → infer → response) and `src/worklog_pipeline/distiller`'s `distil.run` span for the reference shape.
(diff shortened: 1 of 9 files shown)


Akarsh-Hegde:
## Hostname stability: checked, confirmed broken, fixed in `dbacabef`

The open question from the PR body was whether `gethostname()` is a stable per-machine value. It is not, on macOS, and the mechanism is worse than "might drift".

### Measured, not inferred

```
hostname                    Unknown_a2:97:03:78:a9:6a
dig -x 192.168.1.13         Unknown_a2:97:03:78:a9:6a.     <- byte-identical
en1 ether                   a2:97:03:78:a9:6a              <- locally administered
scutil --get HostName       not set
scutil --get ComputerName   Mac Studio                     <- unrelated
```

`HostName` is unset by default on macOS - a rename edits `ComputerName` and `LocalHostName`, not this - so configd derives the kernel hostname from the **network**. Here it is exactly the router's reverse-DNS record for the current IP, which the router derived from `en1`'s MAC. That MAC is locally-administered: it is a macOS **private Wi-Fi address**, which rotates per network.

So the chain is: join a different Wi-Fi → new randomised MAC → new DHCP/reverse-DNS name → new kernel hostname → **new pseudonym for the same machine**.

That invalidates both things the pseudonym exists for. Grouping splits, so "one user hit this 500 times" becomes indistinguishable from "500 users hit it once" - which was the entire argument in #545 for pseudonymising rather than dropping `host.name`. And a Support ID a user already quoted goes stale. Neither failure is visible from either end.

### Fix

Seeds from `IOPlatformUUID` (new `telemetry_spool::machine_id`), falling back to the hostname only where no better id exists. **No Windows probing added**: `gethostname()` there returns the computer name, which is already stable, so untested registry reads for a working case would be risk without benefit.

Shells out to `ioreg` at an absolute path rather than pulling in `core-foundation`/`io-kit-sys` for one startup string read, cached in a `OnceLock`. Any failure returns `None` rather than a junk id - a stable-looking wrong value would be worse than the status quo.

Two consequential details:

- **The ship leg now REPLACES `host.name` rather than hashing the captured value.** Hashing inherited the instability at a second point: a spool file written on one network and drained after joining another carried a different pseudonym. Safe because spool files are always written locally, and `telemetry import` - the one path that ships another machine's payload - bypasses redaction entirely by design, so it cannot be mislabelled by this.
- **Sentry's `server_name` is now set unconditionally** from the same function rather than mapped over the existing value. Otherwise a crash would carry a hostname-derived pseudonym while that machine's error logs carried a hardware-derived one, silently decorrelating the two egress paths that #545 deliberately made agree.

### This also strengthens the pseudonymisation

Not just stabilises it. This is a one-way hash of a **low-entropy** input, and hostnames are enumerable - an attacker who suspects a machine can hash `Akarshs-MacBook-Pro.local` and compare. A 128-bit hardware UUID is not brute-forceable that way. The previous scheme was weaker than it looked.

### Tests

- `pseudonym_is_independent_of_the_captured_hostname` - two records captured under different hostnames (the real scenario: one Mac, two networks) ship as one machine.
- `distinct_seeds_still_yield_distinct_pseudonyms` - guards the obvious wrong fix; returning a constant would pass the test above and destroy the feature.
- `probe_succeeds_on_this_machine` - asserts `IOPlatformUUID` actually resolves on macOS, so a silent fallback to the hostname cannot quietly reintroduce this. Nothing else would report it.
- `parses_the_uuid_out_of_real_ioreg_output` / `malformed_output_yields_none_rather_than_a_junk_id`.

### One-time discontinuity

Rows already in the backend keep their old hostname-derived pseudonym, so a machine's history does not join across this change. Accepted - the old value was not stable anyway, so that history was already fragmented.

Daemon 775 tests, tray 161 tests, UI build clean.

Akarsh-Hegde:
## Status: ready to merge, no review findings outstanding

CI green. No review comments were left on this one; the only changes since it opened are ones I raised against my own work, both already written up above:

- `app.install_mode` **removed** - its only signal is a daemon-shaped exe-path check, but `observability::init` also runs in the tray, so every tray row on a packaged install would have read `source`. A wrong platform attribute is worse than an absent one.
- The Support ID copy claim "changes if you rename the device" **corrected** - false, `gethostname()` is not what a rename edits.
- The open hostname-stability question **resolved** in `dbacabef` - it was real. The pseudonym now seeds from `IOPlatformUUID` rather than a hostname that on macOS is derived from the router's reverse DNS and changes with the network. Details in the comment above.

### Merge-order note

**Merge this before #588**, which is stacked on it. #587 is independent of both.

### What to check on the release that carries this

These are compile-time resource attributes in the packaged daemon, so nothing is observable until a release rebuilds and reinstalls. On that release:

1. `os.type` / `host.arch` present, and Windows rows actually appearing - which is also the only way to confirm #587's fix, since without `os.type` a Windows record is indistinguishable from a macOS one.
2. `host.name` a 16-hex value, and **stable for a machine across a network change** - the specific thing `dbacabef` fixes, and the one regression that would be invisible without looking for it.
3. Sentry `server_name` matching the same pseudonym, or the two egress paths have decorrelated.

Expect a one-time discontinuity: rows already ingested keep their old hostname-derived pseudonym, so a machine's history will not join across this change. That history was already fragmented by the bug being fixed.

adityaharishch:
### Code review

No issues found. Checked CLAUDE.md compliance (file headers, plain-hyphen user-facing text, Rust tracing/error-handling conventions), shallow-scanned the diff for bugs (tar archive construction in `append_bundle_info`, the Rust↔TS `AppInfo`/`supportId` serialization, the `copySupportId` clipboard handler), verified the pseudonym-parity and platform-attribute doc claims against the actual `observability::init`/`redact.rs` code, and checked git history/prior-PR context on `redact.rs` and `observability/mod.rs` for regressions of previously-fixed issues (e.g. the `service.instance.id` allowlisted-but-unpopulated trap, `Resource::new` vs `Resource::default`).

Also noted (not a defect): the PR as originally opened flagged the hostname-instability issue on macOS as a known, out-of-scope caveat, and a follow-up commit (`dbacabef`) landed on this PR before merge that fixes exactly that by seeding the pseudonym from `IOPlatformUUID` instead of `gethostname()` - resolving it within the same PR.

PR #587, opened 2026-07-27
fix(observability): recognise the packaged Windows install (.exe suffix)

Akarsh-Hegde:
Split out of #584 (closed). Standalone and independently mergeable - no dependency on #586 or the error-coverage PR.

## Central error reporting has never worked on Windows

`is_canonical_install()` compared `current_exe()` against a hardcoded path:

```rust
exe == meridian_dir.join("bin").join("meridian")   // no suffix
```

The tray stages the Windows daemon as **`meridian.exe`** (`backend_install::DAEMON_FILE`, correctly cfg-gated on its side). So `current_exe()` ends in `.exe`, the comparison never matched, and `resolve_otlp_target()` fell through to the **DEV** branch - which requires `otlp_enabled` plus local OpenObserve credentials a packaged install never has. It resolved no target, so **no Windows install has ever shipped a single error record.**

Everything else on the Windows path was already correct: the release job bakes `MERIDIAN_CENTRAL_OTLP_ENDPOINT` and `MERIDIAN_CENTRAL_TELEMETRY_TOKEN` into `target/release/meridian.exe` exactly as the macOS job does. It was one absent suffix.

## Why it stayed invisible

Nothing failed loudly at either end. The shipper logs "no OTLP target configured" at `debug!`, spool files aged out under normal retention, and a backend cannot report traffic it never expected. The symptom is indistinguishable from "Windows users hit fewer errors".

It is also only visible reading two crates together - the tray's constant is correct and well-commented, the daemon's looks fine in isolation, they share no constant, and a mismatch is silent.

## Fix

Derived from `std::env::consts::EXE_SUFFIX` rather than a second cfg-selected constant, so the daemon cannot drift from the tray again if a platform is added. `staged_daemon_file_name_carries_the_platform_exe_suffix` pins the name against the platform suffix on whichever OS CI runs.

## Also: Windows coverage for the redaction suite

The suite was entirely Unix-shaped. Behaviour on Windows was already correct, but nothing pinned it, so `windows_paths_are_scrubbed_but_not_mangled` now asserts: `C:\Users\<name>` is scrubbed, the path survives the blob rule intact, and a secret sharing the line still dies whole.

Worth recording *why* it holds - `\` is **not** in the blob character class `[A-Za-z0-9+/=_\-]{32,}`, so backslash-separated paths cannot reach the 32-char threshold that was mangling `/`-separated ones before the #545 review fix. Windows paths are structurally safer here than Unix ones. `looks_like_path` covers drive letters independently.

## Impact

Windows telemetry starts flowing on the first release carrying this. Until then there is no Windows data at all, so on that release the specific thing to verify is that Windows installs appear - which needs `os.type` from #586 to be checkable at all.

## Verification

Daemon `fmt` + `clippy -D warnings` + full test suite.

### DIFF
--- a/src/observability/install_mode.rs
+++ b/src/observability/install_mode.rs
@@ -44,11 +44,28 @@ pub(super) fn is_canonical_install() -> bool {
         return false;
     };
     match std::env::current_exe() {
-        Ok(exe) => exe == meridian_dir.join("bin").join("meridian"),
+        Ok(exe) => exe == meridian_dir.join("bin").join(staged_daemon_file_name()),
         // `current_exe()` failing is rare (permissions, exotic sandboxing) —
         // fall back to the machine-wide marker file rather than guessing.
         Err(_) => meridian_dir.join(".env").exists(),
     }
+}
+
+/// File name the tray stages the daemon under in `~/.meridian/bin/`.
+///
+/// **Must** track `backend_install::DAEMON_FILE` in the tray crate, which is
+/// `meridian.exe` on Windows and `meridian` elsewhere. This was hardcoded to
+/// `"meridian"`, so on Windows the comparison above could never match a real
+/// packaged install: `current_exe()` ends in `.exe`, so `is_canonical_install()`
+/// returned false, `resolve_otlp_target()` fell through to the DEV branch, and
+/// that branch requires `otlp_enabled` + local OpenObserve credentials which a
+/// packaged install never has. Net effect: central error reporting was silently
+/// inert on every Windows install, with no warning on either side.
+///
+/// Derived from `EXE_SUFFIX` rather than a `cfg`-selected constant so it cannot
+/// drift again if another platform is added.
+fn staged_daemon_file_name() -> String {
+    format!("meridian{}", std::env::consts::EXE_SUFFIX)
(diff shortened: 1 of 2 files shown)


adityaharishch:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Akarsh-Hegde:
## Ready to merge

Review came back clean, CI green. Nothing to address.

Recording what merging this actually turns on, since the change is one line of logic and the impact is easy to understate: **no Windows install has ever shipped an error record.** `is_canonical_install()` compared `current_exe()` against `~/.meridian/bin/meridian` while the tray stages `meridian.exe`, so every packaged Windows install fell through to the dev branch and resolved no OTLP target. The endpoint and token were already baked in correctly by the release job - it was only the suffix.

Independent of #586 and #588; merge in any order.

**On the first release carrying this, verify Windows rows actually appear** rather than assuming. That needs `os.type`, which comes from #586 - until that also ships there is no attribute distinguishing a Windows record from a macOS one, so "it works now" would be unfalsifiable. Given this bug went undetected since #545, treating Windows presence as an explicit check is the point.

PR #588, opened 2026-07-27
fix(observability): capture OCR/a11y failures and stop dropping error causes

Akarsh-Hegde:
Split out of #584 (closed).

> **Stacked on #586.** Its two commits are included here because this PR's test anchors on one they add. Review only `fbc3cf3b`; merge #586 first and this diff collapses to that commit alone. #587 is independent of both.

Audit of end-to-end error coverage after the first staging release, prompted by "is every operation's error actually reaching the backend". Answer was no, in two independent ways.

## 🔴 1. The OCR + accessibility stack was never captured at all

`EnvFilter` matches directives against targets by **string prefix**. `meridian=info` therefore covers `meridian_core::*`, `meridian_tray_lib::*` and `meridian_oauth::*` for free - every first-party target starts with those bytes. The in-process capture crates do not:

| target | matched a directive? |
|---|---|
| `meridian_core::readers::tasks` | yes, via `meridian` prefix |
| `meridian_tray_lib::commands::health` | yes, via `meridian` prefix |
| `screenpipe_screen::*` | **no** |
| `screenpipe_a11y::*` | **no** |

All ~48 `warn!`/`error!` sites in those crates were discarded **at the subscriber** - before the spool, before the severity filter, before redaction. Not shipped, not in `meridian logs`, not in an export bundle. Screen-capture bail-outs, Apple Vision OCR handler failures, `AXObserver`/`CGEventTap` registration failures, monitor enumeration errors: all structurally invisible.

This is the worse of the two, because capture is the daemon's **only data source**. A silent failure there is indistinguishable downstream from "the user did nothing today" - exactly the support case that is currently undiagnosable.

Fixed with `CAPTURE_DIRECTIVES` at `warn`, dropping to `error` when the user selects ERROR.

**Privacy audit done before enabling.** These are third-party crates written to print to a terminal, so this is the one fix in the set that could *create* a leak. Every `warn!`/`error!` body in both crates is an infrastructure failure - monitor ids, image dimensions, word-range indices, error `Display`s. **None** interpolates a window title, app name, URL, or OCR text. Noted at the constant that bumping the fork revision requires re-auditing, since that is precisely the change that could start shipping captured content.

The tests assert the **negative** too - that the old filter did not pass capture targets, or the regression check would be vacuous - and that unrelated third-party targets (`reqwest`) stay out, so this is a narrow widening rather than a global one.

## 🔴 2. Every shipped error arrived with its cause stripped

`SAFE_STRING_KEYS` carried `error.message`. The key this codebase actually emits is bare **`error`** - CLAUDE.md's own mandated convention, used by **203 of ~419** warn/error sites. The allowlist dropped it.

Decoded from the live staging spool:

```
engine  = claude
attempt = 2
error   = "claude timed out after 60s"
```

What reached the backend: the static message, plus `attempt` (numeric, kept unconditionally). Not the engine. Not the cause. Every AI-provider failure, summariser failure and ETL failure has been shipping as a bare string.

Allowlisted `error` as **free text** - an `anyhow` chain splices arbitrary `Display` output, so it needs the full URL/email/token scrub, not just the path scrub - plus `provider` and `engine`, enum-like values naming our own components.

**Deliberately stopped there.** The census also surfaced `task`/`task_key` (ticket ids), `file`/`path`, `app_name`, `hour`/`day`. Those are the user's data, are not needed to diagnose a failure, and stay dropped - a test asserts they still are, so the widening cannot quietly grow by precedent.

Also allowlists `code.filename`, the only one of the four `code.*` attributes missing, so records carried the full path but not the bare filename.

## Follow-ups, with assessment rather than a list

**Daemon panics never reach the backend** - real, deliberately not fixed here. The tray has Sentry with the `panic` feature; the daemon has nothing, so a panic reaches only launchd's stderr file. A `panic::set_hook` emitting `tracing::error!` would *look* like coverage but largely isn't: `SpoolClient::send` defers the write via `spawn_blocking`, and the OTel batch processor batches before that, so a panicking process is unlikely to flush. The honest fix is Sentry in the daemon (new dep, DSN wiring, redaction parity) - worth doing, not worth faking.

**`let _ =` sites: 84 hits, not a finding.** Sampled twelve: `remove_dir_all` on temp dirs, `stdout().flush()`, `launchctl` during uninstall, `stamp_sync_error` (already recording an error), `force_flush` on shutdown. Best-effort cleanup where the error is genuinely uninteresting. No evidence of swallowed operational failures; reporting the count without the sample would have been misleading.

## Volume needs one staging round before prod

Enabling 48 new WARN sites on a stack that runs at frame cadence is the flood risk. Two candidates to watch: `"Apple Vision OCR skipped {} words with missing bounding boxes"` and `"tesseract: OCR failed"`. If either is per-frame it belongs in `NOISE_SUBSTRINGS`. Staging is the right place to measure that - check record counts grouped by body on the first release rather than assuming.

## Verification

Daemon `fmt` + `clippy -D warnings` + 773 tests.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -367,6 +367,30 @@ supported way to read logs locally, replacing the old JSONL-tailing UI and the
 old bash `meridian logs` (which used to tail launchd-redirected stdout/stderr
 text).
 
+**Two couplings that silently delete error coverage.** Both have bitten once;
+neither fails loudly, and neither is visible from the call site.
+
+1. **The `EnvFilter` decides what is captured at all — before the spool, before
+   severity, before redaction.** `EnvFilter` matches directives against targets
+   by string **prefix**, so the `meridian` directive covers `meridian_core::*`,
+   `meridian_tray_lib::*` and `meridian_oauth::*` for free — but nothing else.
+   Any crate whose target does not start with `meridian` is discarded entirely
+   unless named in `build_default_filter`. That is why `CAPTURE_DIRECTIVES`
+   exists (`screenpipe_screen`/`screenpipe_a11y` — the OCR + accessibility
+   stack, ~48 error sites that were invisible). **Adding a dependency that does
+   real work means adding its target there**, and auditing its `warn!`/`error!`
+   bodies first: third-party crates were written to print to a terminal and may
+   interpolate content that must never egress.
+2. **The attribute allowlist is keyed on the names you actually type.**
+   `tracing::warn!(error = %e, …)` emits the attribute key `error`, not the
+   semconv `error.message`. Anything not on `redact::SAFE_STRING_KEYS` is
+   dropped from the ship leg, so a well-instrumented error can arrive as a bare
+   static string with its cause stripped. When adding a field to a WARN/ERROR
+   site that you would need in order to debug it remotely, check it is on that
+   list — and if it names the user's data (ticket key, file path, app name,
+   window title, hour/day) it must stay off, so put the diagnostic value in the
+   message or an allowlisted key instead.
+
 The only thing this pipeline structurally can't capture is a hard crash
 (panic before `init` runs, segfault, OOM kill) — for that, launchd's own
 stdout/stderr redirect (`~/.meridian/logs/<service>.log` /
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -375,12 +375,13 @@ neither fails loudly, and neither is visible from the call site.
    by string **prefix**, so the `meridian` directive covers `meridian_core::*`,
    `meridian_tray_lib::*` and `meridian_oauth::*` for free — but nothing else.
    Any crate whose target does not start with `meridian` is discarded entirely
-   unless named in `build_default_filter`. That is why `CAPTURE_DIRECTIVES`
-   exists (`screenpipe_screen`/`screenpipe_a11y` — the OCR + accessibility
-   stack, ~48 error sites that were invisible). **Adding a dependency that does
-   real work means adding its target there**, and auditing its `warn!`/`error!`
-   bodies first: third-party crates were written to print to a terminal and may
-   interpolate content that must never egress.
+   unless named in `observability::filter`. That is why `CAPTURE_TARGETS` exists
+   (`screenpipe_screen`/`screenpipe_a11y` — the OCR + accessibility stack, ~48
+   error sites that were invisible). **Adding a dependency that does real work
+   means adding its bare crate name to that list** — every log level then picks
+   it up automatically — and auditing its `warn!`/`error!` bodies first:
+   third-party crates were written to print to a terminal and may interpolate
+   content that must never egress.
 2. **The attribute allowlist is keyed on the names you actually type.**
    `tracing::warn!(error = %e, …)` emits the attribute key `error`, not the
    semconv `error.message`. Anything not on `redact::SAFE_STRING_KEYS` is
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -400,13 +400,34 @@ spans egress at all, and `host.name` is replaced by a stable **pseudonym**
 macOS is routinely the account holder's real name. The tray's Sentry
 `before_send` applies the identical pseudonym to `event.server_name`.
 
+That pseudonym is the ONLY identifier on an error row — nothing associates it
+with an account, and the hash is one-way. So it is surfaced to the user as
+**Support ID** (Settings → Account, and `bundle-info.txt` inside an export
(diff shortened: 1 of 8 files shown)


adityaharishch:
### Code review

Found 2 issues:

1. `src/observability/mod.rs` grows from 465 to 622 lines, crossing the repo's file-size limit (CLAUDE.md Hard Rules: "Keep files under 500 lines; split when a file grows beyond that")

https://github.com/Meridiona/meridian/blob/fbc3cf3b838ebe3cf10410c2ffe4a2d7b4554742/src/observability/mod.rs#L1-L1

2. The `"ERROR"` branch of `build_default_filter` hardcodes the capture-crate targets instead of deriving them from `CAPTURE_DIRECTIVES`, so a future capture crate added to `CAPTURE_DIRECTIVES` (per this PR's own new CLAUDE.md guidance to do so) will silently have no coverage at the ERROR log level, while DEBUG/INFO/WARN pick it up correctly. The new test `capture_directives_present_at_every_log_level` only checks for the two literal crate names, so it stays green even if this drifts.

https://github.com/Meridiona/meridian/blob/fbc3cf3b838ebe3cf10410c2ffe4a2d7b4554742/src/observability/mod.rs#L418-L436

Akarsh-Hegde:
## Both findings fixed in `d87db154`

Thanks - finding 2 is the same silent-drift class this PR exists to fix, and my test would not have caught it. Worth calling that out rather than glossing it.

### 2. `ERROR` branch hardcoded the capture crate names

Correct, and the consequence was precisely as described: a target added to the list - which **this PR's own CLAUDE.md guidance instructs someone to do** - would have had coverage at DEBUG/INFO/WARN and silently none at ERROR. Also correct that `capture_directives_present_at_every_log_level` asserted on the two literal crate names, so it would have stayed green through exactly that drift.

Fixed at the root rather than by patching the branch. `CAPTURE_DIRECTIVES` (a pre-formatted directive string) is now `CAPTURE_TARGETS` (bare crate names), and each branch supplies only a level:

```rust
const CAPTURE_TARGETS: &[&str] = &["screenpipe_screen", "screenpipe_a11y"];

let (base, capture_level) = match log_level.to_uppercase().as_str() {
    "DEBUG"            => ("meridian=debug,sqlx=warn", "warn"),
    "WARNING" | "WARN" => ("meridian=warn,sqlx=warn",  "warn"),
    "ERROR"            => ("meridian=error,sqlx=error", "error"),
    _                  => (/* INFO default */,          "warn"),
};
format!("{base},{}", capture_directives(capture_level))
```

There is now no site where a target and a level are written together, so the drift is not expressible rather than merely tested-for.

**The test now iterates the constant, and I verified it actually fails on your scenario** rather than assuming it would - reintroduced the hardcoded branch plus a third target:

```
ERROR filter has no directive for screenpipe_capture:
  meridian=error,sqlx=error,screenpipe_screen=error,screenpipe_a11y=error
```

Restored, and green again. One honest limitation recorded in a comment: the *behavioural* half of that test still names two targets, because `tracing::error!` requires a literal target. The directive loop is the part that scales, and the comment says so, so the next reader does not mistake the literals for the guard.

Also added `error_level_suppresses_capture_warnings`, since "ERROR honours the user's choice for the capture stack too" was an intentional behaviour with nothing pinning it.

### 1. `src/observability/mod.rs` over the 500-line cap

Correct. Extracted the filter into `observability/filter.rs` - `EnvFilter` construction, the reload handle, and their tests - following the same split `install_mode.rs` and `otlp_target.rs` already use. "What may be captured" is self-contained and independent of the subscriber/exporter wiring the parent owns.

```
mod.rs           622 -> 447   (below its 465 pre-PR size)
filter.rs              244
otlp_target.rs         276
install_mode.rs         59
```

Every module is now under the cap, and `mod.rs` ends up smaller than it was before this PR rather than merely back under the line.

CLAUDE.md updated to match: it referenced `CAPTURE_DIRECTIVES` and now points at `observability::filter`'s `CAPTURE_TARGETS`, with the instruction changed to "add the bare crate name and every log level picks it up automatically" - which is only true after this fix.

### Verification

`cargo fmt` + `clippy -D warnings` + 774 daemon tests.

### Ready to merge, with two caveats already in the PR body

- **Stacked on #586** - review only the commits after `ddcf6a57`; merging #586 first collapses this to its own commits.
- **Volume needs one staging round before prod.** 48 newly-enabled WARN sites on a stack that runs at frame cadence. `"Apple Vision OCR skipped {} words with missing bounding boxes"` and `"tesseract: OCR failed"` are the two to watch; if either proves per-frame it belongs in `NOISE_SUBSTRINGS`. Check record counts grouped by body on the first staging release rather than assuming.

PR #589, opened 2026-07-27
fix(tray): stage the daemon binary via atomic rename, not in-place overwrite

Akarsh-Hegde:
## Summary

Diagnosed the macOS "the app crashes and can't reopen" reports by pulling this dev machine's own crash reports from `~/Library/Logs/DiagnosticReports/`. Found a real, reproducible daemon crash-loop and fixed its root cause.

## What was happening

Every recent `meridian` (daemon) crash on this machine — 8 reports spanning 2026-07-20 through 2026-07-27, roughly one every 1-2 days — has the **identical** signature:

```
"type": "EXC_BAD_ACCESS", "signal": "SIGKILL (Code Signature Invalid)"
"termination": {"namespace": "CODESIGNING", "indicator": "Invalid Page"}
```

That's macOS's hardened-runtime code-signing monitor killing a *running* process because a page of its executable image no longer matches the signature it was validated against at launch.

## Root cause

`backend_install::ensure_backend_installed` re-stages `~/.meridian/bin/meridian` from the bundled `Resources/backend/meridian` on every app update — **before** `register_service` stops/relaunches the daemon via launchd. `stage_binary` did this with `tokio::fs::copy(src, dest)`.

On Unix, `fs::copy` opens the destination with `O_TRUNC` and writes into the **same inode** — it does not unlink-and-recreate. If the previous daemon process is still running (it usually is: `register_agent`'s `launchctl bootout` happens *after* staging, not before), that process has `dest` mapped as its executable text segment. Truncating and rewriting the file corrupts those pages out from under it. macOS's code-signing enforcement validates pages on demand as they're faulted in, so the process gets `SIGKILL`ed the next time it touches an affected page — which can be immediately or, per the observed reports, up to ~26 minutes later (matches a case where the binary was re-staged at 03:13 and the daemon was killed at 03:39).

The code already had a mirror-image guard for this — `stop_running_daemon_before_stage` — but it's `#[cfg(target_os = "windows")]`-only, because Windows fails loudly (file locked, copy errors out) where macOS fails silently and asynchronously (the copy "succeeds," the crash happens later, disconnected in time from the update). The comment on that function frames the underlying problem as Windows-specific ("Windows keeps a running exe's pages mapped"), which is the misconception this PR corrects — Unix has the equivalent problem with a different failure shape.

## Fix

`stage_binary` now copies into a temp file in `dest`'s own directory, `chmod`s it, then `rename(2)`s it into place. `rename` swaps the directory entry without touching the old inode's content, so a process that already has `dest` open/mapped keeps running against its original, untouched pages — only a *fresh* exec of the path picks up the new binary. No macOS-side "stop first" dance needed; this is the standard safe pattern for replacing an executable that might be live.

Windows is unaffected — `stop_running_daemon_before_stage` still runs first there (rename-over-a-running-exe fails on Windows same as a direct copy would), this is purely additive.

## Testing

- `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (tray crate) — all pass.
- New regression test `stage_binary_does_not_disturb_a_reader_of_the_old_file`: opens `dest` before staging (standing in for a running process's mapped executable), stages a longer replacement, and asserts the already-open handle still reads back the complete, unmodified old content while a fresh read of the path sees the new content. This fails against the old `tokio::fs::copy` implementation (truncates the held-open file) and passes against the rename-based one.
- Full `pre-push` suite (fmt + clippy + ui build/tests + security audit + cargo test) passed locally.

## Note: this doesn't cover everything in the original report

This fixes the **daemon**'s crash-loop, which is real and reproducible. It does **not** fully explain the *tray's* "unexpectedly quit while reopening windows" dialog — the daemon is a windowless LaunchAgent, so killing it can't be what shows that dialog. A separate, single `meridian-tray` crash report on this machine (2026-07-23) shows an unrelated bug: a double-free in `screenpipe_a11y::platform::macos::run_app_observer`'s `NotificationGuard` drop path (libmalloc `BUG_IN_CLIENT_OF_LIBMALLOC`), which lives in the `Meridiona/screenpipe-fork` git dependency, not this repo — needs a separate fix there. Flagging for follow-up; not blocking this fix.

### DIFF
--- a/tray/src-tauri/src/backend_install.rs
+++ b/tray/src-tauri/src/backend_install.rs
@@ -179,6 +179,13 @@ async fn install(backend: &Path, home: &Path) -> Result<(), String> {
             .await
             .map_err(|e| format!("mkdir {}: {e}", p.display()))?;
     }
+
+    // A staging attempt killed between `stage_binary`'s copy and its rename
+    // (tray crash, force-quit, OOM) leaves an orphaned
+    // `<DAEMON_FILE>.staging-<pid>` temp file behind — nothing else ever
+    // scans for these, so sweep them here before staging runs again.
+    cleanup_stale_staging_files(&home.join(".meridian/bin")).await;
+
     // Purge leftovers from a pre-cutover **bundle** install before staging the
     // in-process backend. macOS-only by construction: Windows has no install
     // history to migrate from, so there is nothing to boot out and no legacy
@@ -721,6 +728,51 @@ async fn stage_binary(src: &Path, dest: &Path) -> Result<(), String> {
     Ok(())
 }
 
+/// Remove any `<DAEMON_FILE>.staging-<pid>` leftovers in `bin_dir` from a
+/// staging attempt that [`stage_binary`] never finished (killed between its
+/// copy and its rename) — see that function's doc comment for why the temp
+/// file exists. Best-effort and non-fatal: a listing or removal failure is
+/// only logged, since a lingering temp file is harmless clutter, not a
+/// correctness problem, and must never block `install()`.
+async fn cleanup_stale_staging_files(bin_dir: &Path) {
+    let prefix = format!("{DAEMON_FILE}.staging-");
+    let mut entries = match tokio::fs::read_dir(bin_dir).await {
+        Ok(e) => e,
+        Err(e) => {
+            tracing::debug!(
+                dir = %bin_dir.display(), error = %e,
+                "backend_install: could not list bin dir for stale staging sweep"
+            );
+            return;
+        }
+    };
+    loop {
+        let entry = match entries.next_entry().await {
+            Ok(Some(e)) => e,
+            Ok(None) => return,
+            Err(e) => {
+                tracing::debug!(error = %e, "backend_install: stale staging sweep readdir error");
+                return;
+            }
+        };
+        let Some(name) = entry.file_name().to_str().map(str::to_owned) else {
+            continue;
+        };
+        if !name.starts_with(&prefix) {
+            continue;
+        }
+        let path = entry.path();
+        match tokio::fs::remove_file(&path).await {
+            Ok(()) => {
+                tracing::info!(path = %path.display(), "backend_install: removed stale staging temp file")
+            }
+            Err(e) => {
+                tracing::debug!(path = %path.display(), error = %e, "backend_install: could not remove stale staging temp file")
+            }
+        }
+    }
+}
+
 /// `chmod u+rwx,go+rx` (0755) on a freshly staged binary.
 #[cfg(unix)]
 async fn set_executable(path: &Path) -> Result<(), String> {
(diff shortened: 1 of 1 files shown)


adityaharishch (commented):
Strict review of the fix — logic checks out and matches the crash diagnosis in the PR description. Root cause (in-place `tokio::fs::copy` truncating a live process's mapped executable image, tripping macOS's codesigning page-validation SIGKILL) and fix (temp-file-in-same-dir + `rename(2)`) are correct and the standard safe pattern here. Verified:

- `tmp` is placed via `dest.with_file_name(...)`, i.e. in `dest`'s own directory, so the `rename` is guaranteed same-filesystem and atomic — no silent fallback to a cross-device copy.
- `.meridian/bin` is created (in `install()`) before `stage_binary` is ever called, so the temp file's parent directory is guaranteed to exist.
- `chmod` happens on the temp file *before* the rename, so `dest` never becomes visible with wrong permissions.
- Every failure branch (`copy`, `set_executable`, `rename`) best-effort removes the temp file before returning `Err`, so a failed stage doesn't error out *and* leak — modulo the crash case noted below.
- The Windows path is unaffected/additive as claimed: `stop_running_daemon_before_stage` still runs first there, so by the time `stage_binary` renames, nothing has `dest` open.
- New regression test genuinely exercises the bug: opens `dest` before staging, asserts the held-open handle still reads the untouched old bytes and a fresh read sees the new ones. This fails against the old `tokio::fs::copy`-only code and passes against this fix.
- The `same`-content fast path is unchanged and still correct — skips staging entirely when bytes match, so the code hash / TCC grant stays stable as documented.

One minor, non-blocking nit:

`tray/src-tauri/src/backend_install.rs` line ~703: the temp filename is `{DAEMON_FILE}.staging-{pid}` with no cleanup path anywhere in the module. If the tray is killed (crash, force-quit, OOM) between the `copy`/`set_executable` and the `rename`, that temp file is orphaned in `~/.meridian/bin/` permanently — nothing on a later run scans for and removes stale `*.staging-*` entries. Low severity (rare window, harmless leftover, doesn't affect the running daemon), but worth a follow-up: a glob-and-remove-stale-`.staging-*` sweep at the top of `install()` would close the loop. Not blocking this PR given how narrow the window is and that it doesn't reintroduce the crash being fixed.

Also a very minor style nit: the test's directory name embeds a literal `"norace"` string (`format!("meridian-stage-binary-{}-{}", pid, "norace")`) — reads a bit oddly since it's not actually guarding against a race (the dir is already made unique by pid). Could just drop that segment.

No correctness, security, or clippy/fmt issues found otherwise. CI is green. Approving the approach; the nit above is a suggestion, not a blocker.

Akarsh-Hegde:
Addressed both nits from the review in `31d40c73`:

- **Orphaned temp file**: `install()` now calls a new `cleanup_stale_staging_files` at the top, before staging runs, which sweeps any leftover `<DAEMON_FILE>.staging-<pid>` files out of `~/.meridian/bin/` — closes the window where a tray crash/force-quit/OOM between `stage_binary`'s copy and its rename would leave a permanent orphan. Best-effort/non-fatal, matching the rest of `install()`'s error posture. Added `cleanup_stale_staging_files_removes_only_staging_leftovers`, which asserts it removes the stale temp file and leaves the real staged binary + an unrelated file untouched.
- **Test naming**: dropped the dead `"norace"` literal from `stage_binary_does_not_disturb_a_reader_of_the_old_file`'s temp dir name, and gave it and the new sweep test distinct directory names (`-reader-`/`-sweep-`) so two `#[tokio::test]`s can't collide on pid alone if temp-dir naming is ever reused elsewhere.

`cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test` (6/6 in `backend_install`, including both new tests) all pass, and the full `pre-push` suite (fmt + clippy + ui build/tests + security audit + cargo test) passed on push. Ready to merge from my side.

PR #599, opened 2026-07-28
ci(release): warm the Windows release cache on main + nightly

Akarsh-Hegde:
## Summary
- The Windows release cache (`windows-release-x86_64-pc-windows-msvc`) had never once been saved to `main` — every tag-triggered Windows release build compiled OpenSSL-from-source + SQLCipher + the daemon/tray from a stone-cold cache, every time.
- macOS already gets a warm `main`-scoped cache via `cache-warm.yml`'s push-to-`main` and nightly (04:00 UTC) triggers. This extends the same workflow with a matching `warm-windows` job, so both platforms warm the same way on the same schedule.
- Mirrors `release-build.yml`'s Windows job exactly where it matters for cache-key parity: same toolchain pin, same NASM setup, same `pwsh`-for-the-compile-step fix (Git Bash forces MSYS Perl ahead of Strawberry Perl, which breaks OpenSSL's `Configure`), same `shared-key`. Uses `--no-bundle` so it doesn't need any signing secrets — it only compiles and writes the cache, same as the macOS job.

## Test plan
- [x] `cargo fmt --check` / `cargo clippy -- -D warnings` (pre-commit)
- [x] `cargo test`, UI build + tests, security audit (pre-push)
- [ ] First real run (push to `pre-main`/`main`, nightly cron, or manual `workflow_dispatch`) actually saves `windows-release-x86_64-pc-windows-msvc` under `refs/heads/main` — verify via a subsequent tag-triggered `release-build.yml` Windows job restoring it and finishing fast (this repo's `main` builds already proved the pattern works for macOS at 8m and now Windows at 12m off a manually-seeded cache).

### DIFF
--- a/.github/workflows/cache-warm.yml
+++ b/.github/workflows/cache-warm.yml
@@ -42,6 +42,19 @@ name: Cache warm (release)
 #
 # Neither job signs, notarizes, bundles or publishes. Each is a compile and a
 # cache write.
+#
+# ── KNOWN GAP: NOT EVERY BAKED SECRET INVALIDATES THIS CACHE ─────────────────
+#
+# Neither job here sets release-build.yml's secret-bearing env vars (Clerk key,
+# Sentry DSN, PostHog key, GitHub/Jira OAuth ids, central OTLP endpoint/token,
+# channel), so a tag build restoring this cache compiles those constants in for
+# real for the first time. Cargo only recompiles a crate on an env var change
+# if that crate's OWN build.rs declares `cargo:rerun-if-env-changed=VAR` — most
+# of the vars above have no such declaration anywhere in the workspace, so a
+# rotated secret can silently stay baked at its old (here: absent) value across
+# a cache restore. Pre-existing (predates this file), not introduced by either
+# job below — tracked in
+# https://github.com/Meridiona/meridian/issues/601.
 
 on:
   # Every push to main re-seeds the cache with whatever production is at, so a
@@ -104,8 +117,12 @@ jobs:
           shared-key: macos-release-aarch64-apple-darwin
           cache-directories: |
             ~/.cargo/registry/src
-          # Unconditional: this job only ever runs on main (push/schedule) or by
-          # hand, and writing the main-scoped cache is its entire purpose.
+          # Unconditional: this job's push/schedule triggers only ever fire on
+          # `main`, and writing that cache is its entire purpose. A manual
+          # `workflow_dispatch` CAN target any branch, but GitHub scopes a
+          # cache save to the triggering ref, so a dispatch from elsewhere
+          # just writes a side entry under that ref — it can't clobber the
+          # `main`-scoped cache release-build.yml restores from.
           save-if: true
 
       - uses: actions/setup-node@v4
@@ -189,9 +206,13 @@ jobs:
       # source and needs NASM on PATH. windows-latest's bundled Strawberry
       # Perl already covers the Perl half — do not install a second one.
       - name: Install NASM (for vendored OpenSSL's assembly routines)
-        # Pinned to the commit behind v1.5.2, not the mutable `v1` tag — same
-        # supply-chain reasoning as release-build.yml (this job authenticates
-        # as GH_TOKEN and can write to the shared cache).
+        # Pinned to the commit behind v1.5.2, not the mutable `v1` tag: this
+        # job authenticates as GH_TOKEN and can write to the shared cache, so
+        # a repointed mutable tag is a supply-chain path worth closing here.
+        # NOTE: release-build.yml's own Windows job still uses the mutable
+        # `v1` tag as of this writing — this job is intentionally the more
+        # conservative of the two on this one line, not a byte-for-byte match.
+        # NASM's action choice doesn't affect cache-key parity either way.
         uses: ilammy/setup-nasm@72793074d3c8cdda771dba85f6deafe00623038b # v1.5.2
 
       - name: Install Rust 1.93.1
@@ -207,8 +228,12 @@ jobs:
           # reasoning as the macOS job above.
           add-job-id-key: false
           shared-key: windows-release-x86_64-pc-windows-msvc
-          # Unconditional: this job only ever runs on main (push/schedule) or
-          # by hand, and writing the main-scoped cache is its entire purpose.
+          # Unconditional: this job's push/schedule triggers only ever fire on
+          # `main`, and writing that cache is its entire purpose. A manual
+          # `workflow_dispatch` CAN target any branch, but GitHub scopes a
+          # cache save to the triggering ref, so a dispatch from elsewhere
+          # just writes a side entry under that ref — it can't clobber the
+          # `main`-scoped cache release-build.yml restores from.
           save-if: true
 
       - uses: actions/setup-node@v4
--- a/.github/workflows/cache-warm.yml
+++ b/.github/workflows/cache-warm.yml
@@ -1,6 +1,6 @@
-name: Cache warm (macOS release)
+name: Cache warm (release)
 
-# SEEDS THE RELEASE COMPILE CACHE ON `main`. SHIPS NOTHING.
+# SEEDS THE RELEASE COMPILE CACHES ON `main` FOR BOTH PLATFORMS. SHIPS NOTHING.
 #
 # ── WHY THIS EXISTS ──────────────────────────────────────────────────────────
 #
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde:
## Overview

Extends the existing macOS-only release cache-warming workflow (`cache-warm.yml`) to Windows, adding a `warm-windows` job that mirrors `release-build.yml`'s Windows job's compile shape so the shared `windows-release-x86_64-pc-windows-msvc` `rust-cache` key stays warm on `main`, the same way `warm-macos` already keeps `macos-release-aarch64-apple-darwin` warm. Small, well-scoped diff (one file), and the reasoning in the header comment is accurate as far as it goes — I checked it against `release-build.yml`'s actual Windows job line by line (NASM/Perl ordering, the `pwsh`-for-compile fix, why no daemon-stub step is needed here unlike the macOS job, matching `rust-cache` config) and it holds up.

That said, I want to push on the thing this whole mechanism rests on, because this PR doubles its exposure rather than just being "more of the same."

## Blocking

**1. The premise this PR extends — "a cache-warmed `target/` is safe to reuse for a release build" — doesn't hold for most of the compile-time secrets those release builds bake in, and this PR doubles the platforms exposed to it.**

Every `option_env!("MERIDIAN_*")` read in this codebase:

| Var | Used at | Covered by `rerun-if-env-changed`? |
|---|---|---|
| `MERIDIAN_JIRA_OAUTH_CLIENT_SECRET` | `meridian-oauth/src/jira.rs:81` | ✅ (`build.rs:10`) |
| `MERIDIAN_GITHUB_OAUTH_CLIENT_ID` | `meridian-oauth/src/github.rs:56` | ✅ (`build.rs:13`) |
| `MERIDIAN_POSTHOG_API_KEY` | `tray/src-tauri/src/analytics.rs:125` | ✅ (`tray/src-tauri/build.rs:6`) |
| `MERIDIAN_CLERK_PUBLISHABLE_KEY` | `tray/src-tauri/src/commands/account.rs:57` | ❌ |
| `MERIDIAN_COUNTER_API_KEY` | `tray/src-tauri/src/counter_ping.rs:60` | ❌ |
| `MERIDIAN_SENTRY_DSN` | `tray/src-tauri/src/crash.rs:39` | ❌ |
| `MERIDIAN_CENTRAL_OTLP_ENDPOINT` | `src/observability/otlp_target.rs:49` | ❌ |
| `MERIDIAN_CENTRAL_TELEMETRY_TOKEN` | `src/observability/otlp_target.rs:55` | ❌ |
| `MERIDIAN_CHANNEL` | `src/observability/mod.rs:315`, `src/telemetry_spool/mod.rs:140`, `tray/.../version.rs:259` | ❌ |

Cargo does not track an `option_env!`/`env!` read as a build input unless a `build.rs` explicitly declares `cargo:rerun-if-env-changed=VAR` — this is exactly what `build.rs`'s own comment says, for the two vars it already covers ("Cargo does NOT track env vars as build inputs by default, so a CI cache... could reuse a stale compilation and bake the OLD secret"). That reasoning applies identically to the six vars above that have no such declaration. Concretely: if source is otherwise unchanged between a cache-warm run (which sets none of these — neither `warm-macos` nor the new `warm-windows` job sets `MERIDIAN_CLERK_PUBLISHABLE_KEY`/`SENTRY_DSN`/`CHANNEL`/etc.) and the tag build that restores that cache (which does set them, per `release-build.yml`'s `env:` blocks), Cargo's fingerprint has no reason to recompile the crates that reference them — so the release binary can silently ship with whatever these constants happened to be during the warm run (almost certainly `None`/absent), not the real production values. That's Clerk sign-in, Sentry crash reporting, and the central error-telemetry gateway potentially silently inert in a shipped release, plus the channel badge being wrong.

I want to be precise about scope: **this gap is not introduced by this PR** — it already exists for `warm-macos`, and if it were actively biting, someone would presumably have noticed a shipped DMG failing to sign in via Clerk. But this PR's stated premise is exactly "mirror `release-build.yml` for cache-key parity," and it doubles the number of release channels resting on that premise without the premise itself being fully sound yet. Given how load-bearing this workflow is (it's literally in the critical path of every tag build, staging and production), I'd want either:
- the missing `rerun-if-env-changed` declarations added (companion PR, before or right after this one), or
- at minimum, an explicit acknowledgment in this PR (or a tracked follow-up) that this is a known gap, so it doesn't quietly become "well, it's been fine so far" folklore.

**2. The new job has never actually run.** The PR's own test plan lists "First real run... actually saves `windows-release-x86_64-pc-windows-msvc`" as an unchecked box. Since this change only touches `.github/workflows/cache-warm.yml` (not `ci.yml`), the `Detect changes` path filter correctly skips the Rust CI job entirely for this PR — meaning **no CI job of any kind has executed this new job's YAML**, not even a syntax check. `workflow_dispatch` has no branch restriction and can be pointed at this PR's branch (`gh workflow run "Cache warm (release)" --ref ci/windows-cache-warm`) at zero risk (branch-scoped cache write, `--no-bundle`, no signing secrets) — that's a cheap, concrete way to convert "I mirrored the reasoning correctly" into "this actually compiles end-to-end on windows-latest," before merging a CI/CD change that release-build.yml depends on for cache-key parity. Given Meridian's CI matrix has no Windows runner at all elsewhere (per `ci.yml`'s own comment), this workflow run would be the *only* automated Windows-compile signal this change gets before it's live on `main`.

## Should-fix

**3. The NASM action's pin doesn't actually match `release-build.yml`, despite the comment implying it does.** This job pins `ilammy/setup-nasm@72793074d3c8cdda771dba85f6deafe00623038b # v1.5.2` with the comment "same supply-chain reasoning as release-build.yml." But `release-build.yml`'s Windows job (line 661) uses the mutable `ilammy/setup-nasm@v1` tag, not a pinned SHA. Two options, either is fine, but pick one: (a) fix the comment to say this job is *intentionally more conservative* than `release-build.yml` here, since NASM's action choice doesn't affect cache-key parity at all, or (b) if SHA-pinning third-party actions with `GH_TOKEN` write access is the right call (it is), bring `release-build.yml`'s Windows job up to the same pin in a follow-up — the actual release-shipping job is the one with the higher stakes, and it's currently the less-pinned of the two.

## Nits

**4.** The `save-if: true` comment ("this job only ever runs on main (push/schedule) or by hand") is slightly imprecise — `workflow_dispatch` "by hand" can target *any* branch, not just main. It's harmless in practice (GitHub scopes a cache save to the triggering ref, so a dispatch from a feature branch can't clobber the `main`-scoped cache `release-build.yml` restores from), but worth tightening given how precise the rest of this file's comments are. (Pre-existing wording, copied from the macOS job — not new to this PR.)

## Positives

- The "why it compiles the same way release-build does" rationale is genuinely followed through, not just asserted — I checked the NASM/Perl-ordering explanation, the `pwsh`-only-for-the-compile-step fix, and the "no stub step needed" reasoning (the daemon `.exe` lands at exactly the path `tauri.windows.conf.json`'s `bundle.resources` expects, with no cross-target copy step, unlike macOS) against `release-build.yml`'s actual Windows job, and it's accurate.
- `--no-bundle` keeps this job's secret/permission footprint minimal — no signing keys needed, matching the macOS job.
- Doc-comment quality matches the rest of the file — this is a genuinely well-explained workflow.

**Verdict: request changes**, primarily on 1 and 2 — not because the YAML mechanics are wrong (they look correct), but because this PR is the second platform now leaning on a caching premise that's demonstrably incomplete (6 of 9 baked secrets untracked for cache invalidation) and has zero execution evidence of its own yet. Both are cheap to close before merge: run it once via `workflow_dispatch`, and either patch the missing `rerun-if-env-changed` lines or explicitly flag them as a known, tracked gap.

Akarsh-Hegde:
## Update — addressed review, ready for another look

Pushed `74a8deea` on top of `fc848214` addressing the review:

**Blocking #1** (cache-key parity resting on incomplete `rerun-if-env-changed` coverage): acknowledged in-file rather than silently expanding this PR's scope into `build.rs` changes across `tray/src-tauri` and (possibly) a new `meridian-oauth/build.rs`. Filed #601 with the full var table plus a wrinkle worth verifying: the two vars currently marked "covered" are declared in the *root* `build.rs`, but both are actually read inside `meridian-oauth`, a separate workspace package with no `build.rs` of its own — `rerun-if-env-changed` doesn't obviously cascade across that package boundary, so those two may also be uncovered. Left for #601 to resolve with a repro, not this PR.

**Blocking #2** (zero execution evidence): dispatched the workflow for real on this branch — [run 30387220519](https://github.com/Meridiona/meridian/actions/runs/30387220519), both jobs green. Note it ran slower than the ~12min happy path we've seen elsewhere (Windows job took ~45min) — this branch is based on `pre-main` from before yesterday's `v1.80.0` version bump, so its `Cargo.lock` doesn't exact-match the `main`-scoped cache saved after that bump, and it fell back to a much colder compile. Not a functional issue, just means this particular run isn't representative of steady-state speed — only of "does it actually compile end-to-end on `windows-latest`," which it does.

**Should-fix #3** (NASM pin comment overclaiming parity with `release-build.yml`): fixed — the comment now says this job is intentionally more conservative on that one line (SHA-pinned vs `release-build.yml`'s mutable `@v1`), rather than claiming a match that isn't there.

**Nit #4** (`save-if: true` comment imprecise about `workflow_dispatch` scope): tightened on both jobs to note a dispatch can target any branch, but GitHub's ref-scoped cache saves mean it can't clobber `main`'s cache regardless.

PR #600, opened 2026-07-28
feat(telemetry): per-user Support ID + platform prefix for alpha testing

Akarsh-Hegde:
## Summary
- Hand-picked alpha testers run Meridian across multiple machines, and support needs to trace one person's errors across their devices for the next month.
- `local_host_pseudonym()` now prefixes `mac_`/`win_`, and — while signed in — seeds from a salted hash of the account email instead of the hardware UUID. The tray mirrors *only that hash* into `settings.json`'s new `account_pseudonym` field (never the raw email), and clears it on sign-out.
- Gated on a hardcoded expiry (`ALPHA_ACCOUNT_OVERRIDE_EXPIRES_UNIX`, 2026-08-28), not release channel — alpha testers install the same `stable`/production build as everyone else, so channel can't distinguish them. Past the expiry, every build reverts to the per-machine pseudonym automatically with no deploy needed; a broken clock fails closed toward that same stricter behavior.
- Updated the Settings → Account copy, which previously promised the Support ID was "not tied to your account" — true again once the alpha window ends.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings` (daemon + `meridian-core` + tray crates)
- [x] `cargo test` — 27 tests in `telemetry_spool::redact` (including new coverage for the platform prefix, account-hash salting/case-insensitivity, and the expiry-date gate + clock-fault-fails-closed case), plus `meridian-core::settings`
- [x] `npx tsc --noEmit` on `ui/` — no new errors
- [x] Full `pre-push` suite (fmt, clippy, UI build/tests, security audit, `cargo test`) passed on push

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -436,6 +436,20 @@ machine a new identity on every network change, silently breaking the grouping
 the value exists for. See `telemetry_spool::machine_id` before touching any of
 this; `pseudonym_is_independent_of_the_captured_hostname` pins it.
 
+**ALPHA EXCEPTION (until 2026-08-28, `redact::ALPHA_ACCOUNT_OVERRIDE_EXPIRES_UNIX`):**
+while Meridian is in hand-picked-tester alpha, the two paragraphs above and
+below are only true when signed out. While a tester is signed in,
+`local_host_pseudonym` deliberately seeds from a salted hash of their Clerk
+account email instead (`redact::pseudonymize_account`, mirrored into
+`settings.json`'s `account_pseudonym` by the tray on sign-in/out — never the
+raw email) — so support can trace one tester's errors across their Mac and
+Windows machines, at the cost of the pseudonym now being genuinely tied to an
+account rather than only a machine. `redact::support_id_is_account_scoped`
+reports which mode is currently active (surfaced in Settings → Account's
+copy) so this state is never asserted where it isn't true. Past the expiry
+date every install reverts to the per-machine-only behaviour described below,
+automatically, with no deploy required.
+
 That pseudonym is the ONLY identifier on an error row — nothing associates it
 with an account, and the hash is one-way. So it is surfaced to the user as
 **Support ID** (Settings → Account, and `bundle-info.txt` inside an export
--- a/meridian-core/src/settings.rs
+++ b/meridian-core/src/settings.rs
@@ -307,6 +307,21 @@ pub struct RuntimeSettings {
     // at the struct level makes this `false` for every settings.json written before
     // this field existed, which is exactly the "reprompt existing users once" case.
     pub worklog_auto_generate_prompted: bool,
+    // ALPHA TESTING ONLY (hand-picked users, all on the same `stable`
+    // production channel as everyone else — there's no separate alpha
+    // channel to gate on). A salted, one-way hash of the signed-in Clerk
+    // account email — NEVER the raw email — written by the tray's
+    // `commands::account::save_account_email` (cleared on sign-out) and read
+    // by `telemetry_spool::redact::local_host_pseudonym` to seed the Support
+    // ID/shipped `host.name` per-USER instead of per-machine, so support can
+    // trace one alpha tester's errors across their Mac and Windows boxes.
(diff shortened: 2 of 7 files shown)


Akarsh-Hegde:
### Code review

Found 2 issues:

1. `CLAUDE.md` is not updated and now states the opposite of what this PR ships (CLAUDE.md says "The pseudonym is seeded from a hardware id, NOT the hostname" and "That pseudonym is the ONLY identifier on an error row — nothing associates it with an account, and the hash is one-way")

During the alpha window (until 2026-08-28) the pseudonym is seeded from a salted hash of the signed-in account email instead, and is deliberately associated with an account — making both statements conditionally false. Every prior commit touching this exact pseudonym logic (`dbacabef`, `ddcf6a57`, `87a54b7a`, `05aa3138`) updated this same section of CLAUDE.md in the same commit; this PR doesn't touch CLAUDE.md at all.

https://github.com/Meridiona/meridian/blob/00ffc8c2a8fa8652385ecc94ad7f1d33977b3747/CLAUDE.md#L428-L441

2. `AccountSection.tsx`'s new Support ID description is static and already misleading for signed-out users, with no revert planned for after the alpha window

The text unconditionally says "we can trace issues across your devices... find the errors from your account," but `choose_pseudonym_source` only uses the account pseudonym when signed in (a signed-out user's Support ID is still the per-machine hash today), and the backend auto-reverts to the per-machine pseudonym after `ALPHA_ACCOUNT_OVERRIDE_EXPIRES_UNIX` (2026-08-28) with no deploy required — but this string has no analogous condition, so it will keep claiming account-based tracing indefinitely past that date. This is the same defect class commit `ddcf6a57` ("fix(observability): drop app.install_mode, correct the Support ID copy") already fixed once in this exact `FieldRow` string.

https://github.com/Meridiona/meridian/blob/00ffc8c2a8fa8652385ecc94ad7f1d33977b3747/ui/components/timeline/settings/AccountSection.tsx#L93-L97

Akarsh-Hegde:
### Review findings addressed — ready to merge

Both issues from the code review above are fixed in `a3b6a4f2`:

**1. `CLAUDE.md` updated.** Added an "ALPHA EXCEPTION" paragraph right where the two contradicted sentences live (the hardware-id-seeding and "nothing associates it with an account" claims), explaining the temporary per-user override, linking it to `redact::support_id_is_account_scoped`, and stating the automatic 2026-08-28 revert.

**2. `AccountSection.tsx`'s Support ID copy is no longer static.** Added `redact::support_id_is_account_scoped()` — it reuses the exact same `choose_pseudonym_source()` decision `pseudonym_body()` calls, so it cannot disagree with what the pseudonym is actually doing. Threaded through a new `AppInfo.support_id_is_account_scoped` field to the frontend, which now switches between the two description strings based on it instead of asserting the account-based one unconditionally:
- Signed out → `false` → original "not tied to your account" copy (fixes the signed-out-user case flagged above).
- Signed in, before 2026-08-28 → `true` → the alpha copy.
- Signed in, past 2026-08-28 → `false` automatically, no deploy required (fixes the "no revert planned" case flagged above) — same self-expiring mechanism the pseudonym itself already has.

Added a test (`support_id_is_account_scoped_matches_choose_pseudonym_source`) that checks this against the pure decision function directly rather than a hardcoded expected boolean, so it stays meaningful (and passing) on either side of the expiry date instead of quietly breaking the day the alpha window ends.

**Verification:**
- `cargo fmt --check` / `cargo clippy --all-targets -- -D warnings` — clean on `meridian`, `meridian-core`, and the tray crate
- `cargo test` — 28 tests in `telemetry_spool::redact` (27 prior + the new one), all passing
- `npx tsc --noEmit` on `ui/` — no new errors from the `AppInfo`/`AccountSection.tsx` changes
- Full local pre-push suite (fmt, clippy, UI build/tests, security audit, `cargo test`) passed on push
- CI on this PR: `Rust (macOS Apple Silicon)`, `UI`, `Detect changes`, `screenpipe MIT pin` all green

No other open comments. This is ready to merge.

PR #604, opened 2026-07-28
fix: address release review findings on #603

Akarsh-Hegde:
Three findings from the strict review pass on the `pre-main → main` release PR (#603). This targets `pre-main`, so #603 picks it up automatically once merged.

## 🔴 Hard Rule violation - em-dash in user-facing text

The new `tray.db_encryption_incomplete` notice detail (from #598) used an em-dash. CLAUDE.md is explicit that displayed strings take a plain hyphen only, and this is a notice body rendered straight into the UI. It was the one string in the entire release diff that broke the rule.

Worth noting *why* it slipped: nothing lints for it. The pre-commit hook runs `fmt` and `clippy` only, neither of which knows about house style in string literals. Everywhere else the cleanup was done by hand and done right.

## 🟡 A never-synced provider escalated on the very first blip

`note_transient_sync_failure` escalated immediately when there was no `pm_sync_state` row, reasoning that an age test alone would no-op forever on a provider blocked from the start. That reasoning holds — but the behaviour reintroduced the exact thing this work exists to remove, only with connectivity wording instead of credentials wording: connect a tracker, have the first attempt land on a DNS hiccup or a proxy handshake, and a "sync failing" banner appears seconds later.

**It now gets one retry.** The first failure writes the row and stays silent; the next one sees an epoch `last_synced_at`, which is not recent, and escalates. One sync interval of grace, and the persistently-blocked case still surfaces. No migration needed — the epoch sentinel already carried exactly this meaning.

Two tests replace the old escalate-immediately case:
- `a_never_synced_provider_gets_one_retry_before_escalating` — first call silent and no notice, second call escalates
- `the_first_failure_row_does_not_look_like_a_sync` — the grace row must carry the epoch sentinel, or it would read as a recent sync and suppress escalation *forever*, turning a one-retry grace into permanent silence

## 🟢 Stale test name in a comment

`every_remedy_is_followable_from_the_app` never existed. The test is `every_default_remedy_names_a_place_in_the_app`.

## 🟢 Coverage gap on the headline fix - closed

The review flagged the raise/clear logic for the encryption notice as the one part of the #598 hotfix with no CI coverage, and suggested extracting a helper. Done: `db_encryption_notice_action(encryption_intended, still_plaintext)`, unit-tested across all four combinations.

Both "clear" cases are asserted explicitly rather than as "not raise", so an inverted condition can't slip through — a user who never enabled encryption must never see this notice, and it has to disappear on its own once the migration finally completes.

## Not addressed here, deliberately

The **supply-chain finding** — `release-build.yml`'s Windows job still pins `ilammy/setup-nasm` to the mutable `v1` tag while the lower-privilege cache-warm job now pins a SHA. The review called it out as not release-blocking, and I agree it shouldn't ride along on a release PR: that workflow holds the signing key, the central OTLP token, and the Sentry DSN, so a change to it deserves its own review rather than being buried in a fix batch. Raised separately as a follow-up.

## Verification

- 821 daemon tests, 167 tray tests
- `clippy -D warnings` and `cargo fmt --check` clean on **both** workspaces (the tray is a separate workspace and is not covered by the daemon's pre-push run, so it was checked by hand)


Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
## Ready to merge ✅

`mergeStateStatus: CLEAN` · all CI checks passing.

### What this fixes

Three findings from the strict review pass on #603, plus the coverage gap it flagged:

- **🔴 Em-dash in the `tray.db_encryption_incomplete` notice** — Hard Rule violation in a string rendered straight into the UI. It was the one string in the whole release diff that broke it, and nothing lints for house style in string literals, which is how it slipped past `fmt`/`clippy`.
- **🟡 Never-synced provider escalated on the very first blip** — the important one. Escalating immediately meant: connect a tracker, first attempt hits a DNS hiccup, "sync failing" banner seconds later. That's the same false-positive banner #602 exists to remove, just relabelled from credentials to connectivity. It now gets one retry — first failure writes the row silently, the next sees an epoch `last_synced_at` and escalates. No migration; the sentinel already meant exactly that.
- **🟢 Stale test name in a comment.**
- **🟢 Coverage gap closed** — `db_encryption_notice_action(intended, still_plaintext)` is now unit-tested across all four combinations. This was the one part of #598's hotfix with no CI coverage.

### Verification

- **821 daemon tests**, 2 new (replacing the old escalate-immediately case)
- **167 tray tests**, 2 new
- `clippy -D warnings` + `cargo fmt --check` clean on **both** workspaces

The tray was checked by hand deliberately: it's a separate cargo workspace and is covered by neither the daemon's pre-push run nor CI, so those two new tray tests would otherwise ship unverified. Running them locally needs a placeholder at `target/release/meridian` to satisfy the Tauri build script's resource check.

### ⚠️ CodeRabbit did NOT review this PR

Its check reads `pass — Review rate limited`. The account hit its PR review quota; I triggered a review twice, both returned "Review finished" while posting nothing, and the PR has **zero reviews and zero inline comments**.

**So the green CodeRabbit tick here is not a review, and shouldn't be read as one.** Flagging it rather than letting a passing check imply scrutiny that didn't happen. What this change does have: it is a direct, point-by-point response to a human review, and every line is covered by the tests above.

Akarsh-Hegde:
## One more commit — `54a45613`

Flagging this because I called the PR merge-ready above and have since pushed to it.

**The release had no What's New entry.** CLAUDE.md lists one as a required per-release step, and #603 didn't have it — so the release would have shipped with no user-facing notes at all. Added `1.81.0`, which is what semantic-release will cut: exactly one `feat` commit in the batch (the alpha Support ID) makes it a minor bump from 1.80.0.

Also corrected 1.80.0's date. I wrote it as `2026-07-27` before that release was actually cut; it published on the 28th. Both entries now carry `2026-07-28`, which is simply true — the two releases are hours apart.

Every bullet is written in user terms rather than pasted from a commit message, and uses plain hyphens per the Hard Rule. Since nothing lints that, the edit asserted it programmatically before writing.

`whats_new_json_parses` and the rest of the tray suite still pass. CI re-running on the new commit; still ready once it's green.

Akarsh-Hegde:
## Updated: ready to merge ✅ — now carries considerably more than when I first said so

`mergeStateStatus: CLEAN` · all CI passing on `7f76bddd`.

Restating the scope, since this PR has grown twice since my earlier merge-ready comment:

**Round 1 — human review findings (3 + a coverage gap)**
Em-dash Hard Rule violation in the encryption notice · never-synced provider escalating on the very first blip · stale test name · extracted and tested the notice raise/clear logic.

**Round 2 — release notes**
Added the `1.81.0` What's New entry, which the release was missing entirely, and corrected 1.80.0's date.

**Round 3 — CodeRabbit findings (9 fixed, 1 deferred to #605)**
Clear-the-encryption-warning-only-on-confirmation · error logging on all three `finalize_encryption_swap` branches · `write_account_pseudonym` failure boundary · span on the notice path · em dash in `db_crypto` · four providers consolidated onto `record_sync_failure` · `http::chain` reuse · `providers/mod.rs` split under 500 lines · per-payload pseudonym instead of per-attribute.

### One thing worth reading before merging

While fixing the "only clear on confirmation" finding I introduced, and caught, a worse bug than the one being fixed. My first version used an early `return Ok(())` for the new `Leave` branch — which sits inside Tauri's `setup` closure, so it would have **skipped the tray menu build and the entire rest of startup**. It compiled cleanly and all 169 tray tests passed.

It's restructured to an if/else with a comment explaining why it can't be a return. Flagging it because it's a good argument for looking at that hunk specifically: nothing automated would have caught it.

### Verification

821 daemon tests · 169 tray tests · `clippy -D warnings` + `cargo fmt --check` clean on **both** workspaces.

The tray was verified by hand throughout — it's a separate cargo workspace covered by neither CI nor the pre-push hook, so every tray change in this PR (the notice logic, the account logging, the What's New JSON) would otherwise ship unverified.

### ⚠️ Still no CodeRabbit review on this PR

Its check reads `Review skipped: reviews are disabled for this base branch` — auto-review is off for `pre-main` bases, and the two manual triggers I sent landed while the account was rate-limited. **The green tick is not a review.** The changes here are a direct response to two reviews on #603, and covered by the tests above, but that's the honest state.

PR #607, opened 2026-07-29
fix(tray): harden Windows daemon-binary install against transient locks

Akarsh-Hegde:
## What & why

Follow-up hardening for the **"Meridian couldn't finish installing"** notice on Windows.

On a version update the tray re-stages the daemon binary while the *previous* build's daemon may still hold `~/.meridian/bin/meridian.exe` open — Windows then fails the overwrite with **os error 32** ("the process cannot access the file"). `stop_running_daemon_before_stage` already stops the old daemon first, but two gaps let the notice fire for a hold that would have cleared on its own:

1. the daemon-exit poll gave up after only **~3s** — tight when a killed daemon needs a beat to unwind on a loaded machine, and
2. the final tmp→dest swap was a **single `rename`** with no retry, so a momentary AV / Search-Indexer hold on the freshly staged file failed the whole install.

This is the same failure family as #606 (Windows can't rename/overwrite a file another handle holds open), a different code path (backend install vs. DB encryption).

## Changes

- **Widen the daemon-exit poll ~3s → ~10s** (`STOP_POLL_ATTEMPTS` × `STOP_POLL_INTERVAL`), extracted into `wait_until_gone`.
- **Wrap the binary swap in `rename_with_retry`** (bounded linear backoff) so a transient hold retries instead of surfacing the notice. A *persistent* lock still returns the real error — a genuinely stuck file fails loudly, it isn't swept under the rug.
- Retry/poll logic factored into **platform-neutral generics** (`retry_transient`, `wait_until_gone`) so it compiles and its tests **actually run on the macOS-only CI**. The failures are Windows-specific; the algorithms aren't.

## Tests (9 new, all run on CI)

Deterministic — zero-delay backoff + call-counting fakes, no real timing:

- `retry_transient`: first-try success (no retry) · recovers after 3 transient failures · surfaces error after exhausting attempts · `attempts == 1` bound
- `rename_with_retry`: moves file on success · surfaces a persistent `NotFound` and creates no dest
- `wait_until_gone`: gone on first probe · keeps polling until gone mid-window · reports survivors when the window expires (so the caller fails loudly)

## Why not tested *on* Windows

CI runs Rust tests on macOS aarch64 only (Windows just does a release build, no `cargo test`). Rather than gate tests behind `#[cfg(windows)]` where CI would never run them, the guarded logic is platform-neutral and exercised on macOS — validating the exact code both platforms execute. The Windows-syscall shims around it (`taskkill`/`schtasks`/CIM enumeration) are unchanged.

### DIFF
--- a/tray/src-tauri/src/backend_install.rs
+++ b/tray/src-tauri/src/backend_install.rs
@@ -40,6 +40,7 @@
 //! - `scripts/install-daemon.sh` — the source-install shell flow this parallels.
 
 use std::path::{Path, PathBuf};
+use std::time::Duration;
 
 // Used only by the Windows service-registration spawns below (schtasks / taskkill
 // / tasklist / the daemon launch) to suppress their console-window flash; the
@@ -329,6 +330,43 @@ async fn register_service(_backend: &Path, home: &Path, daemon_bin: &Path) -> Re
     }
 }
 
+/// How many times [`stop_running_daemon_before_stage`] re-checks that the old
+/// daemon has exited, and the gap between checks — ~10s total. Longer than the
+/// original 3s on purpose: a kill doesn't release the file instantly, so a
+/// tight window gives up while the hold would still have cleared, needlessly
+/// surfacing an "couldn't finish installing" notice for a blip.
+#[cfg(target_os = "windows")]
+const STOP_POLL_ATTEMPTS: u32 = 40;
+#[cfg(target_os = "windows")]
+const STOP_POLL_INTERVAL: Duration = Duration::from_millis(250);
+
+/// Poll `probe` until it reports the daemon gone (an empty pid list) or the
+/// attempt budget is exhausted, returning the final list — empty means it
+/// exited in time, non-empty means it's still holding the binary and the caller
+/// must fail loudly rather than overwrite a locked file.
+///
+/// Split out from [`stop_running_daemon_before_stage`] and kept platform-neutral
+/// (generic over the probe, no Windows syscalls of its own) so the poll logic is
+/// unit-tested on macOS CI — the one platform CI runs — even though its only
+/// production caller is Windows-only. `attempts`/`interval` are parameters so a
+/// test can drive it with a fake probe and a zero interval, deterministically
+/// and without real waiting.
+#[cfg_attr(not(target_os = "windows"), allow(dead_code))]
+async fn wait_until_gone<F, Fut>(mut probe: F, attempts: u32, interval: Duration) -> Vec<u32>
+where
+    F: FnMut() -> Fut,
+    Fut: std::future::Future<Output = Vec<u32>>,
+{
+    let mut remaining = probe().await;
+    let mut tries = 0;
+    while !remaining.is_empty() && tries < attempts {
+        tokio::time::sleep(interval).await;
+        remaining = probe().await;
+        tries += 1;
+    }
+    remaining
+}
+
 /// Stop the running instance(s) of **the staged daemon** so [`stage_binary`]
 /// can overwrite `~/.meridian/bin/meridian.exe`. Targets only processes whose
 /// executable is exactly `daemon_bin` — never an unrelated `meridian.exe` (an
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde (commented):
## Review (strict pass)

Solid, well-tested change. The two widened windows check out arithmetically (`STOP_POLL_ATTEMPTS=40 × STOP_POLL_INTERVAL=250ms` = 10s total sleep, matching the "~10s" claim; `RENAME_ATTEMPTS=5` with `base_delay × attempt` backoff = 100+200+300+400ms ≈ 1s before the final attempt, matching "~1-2s"), and the "1 initial probe + up to `attempts` re-probes" shape of `wait_until_gone` is exactly what the tests pin (`wait_until_gone_reports_survivors_when_the_window_expires` asserts 4 probes for `attempts=3`). Doc comments are correctly attached to the right items this time (no repeat of #606's misplaced-doc issue — each new block has its own doc directly above it, separated by blank lines). Good use of `#[cfg_attr(not(target_os = "windows"), allow(dead_code))]` on `wait_until_gone` specifically (not on `retry_transient`/`rename_with_retry`, which are correctly left unannotated since `stage_binary` calls them unconditionally on every platform) — that's the right, minimal scope for the lint suppression.

### 🟡 Duplicated retry-with-backoff logic vs. #606

This PR adds its own `retry_transient`/`rename_with_retry` pair in `tray/src-tauri/src/backend_install.rs`. #606 just added a differently-shaped but functionally equivalent `rename_with_retry` in `meridian-core/src/db_crypto.rs` (sync, `std::fs`/`std::thread::sleep`, hardcoded to rename) for the *same failure family* — this PR's own description says as much ("same failure family as #606 … a different code path"). That's now two independent implementations of "retry a fallible op with linear backoff to ride out a transient Windows file lock" living in two different crates, with no shared abstraction between them despite `meridian-core` existing specifically as the "lean shared data layer … used by BOTH the daemon and the Tauri dashboard" (per CLAUDE.md) and already being a dependency of the tray crate. Given this is explicitly framed as a *failure family* (the PR title says "harden … against transient locks," and this is the second of at least two occurrences), it's worth generalizing `retry_transient` into `meridian-core` now rather than after a third call site shows up. Not a blocker for this fix, but flagging since the pattern is clearly recurring within this same release.

### 🟢 Minor: per-attempt progress logging is gone from the poll

The original loop logged on every iteration:
```rust
tracing::debug!(attempt, path = %daemon_bin.display(), remaining = remaining.len(), "backend_install: waiting for staged daemon to exit");
```
`wait_until_gone` is deliberately probe/logging-agnostic to stay platform-neutral and testable, so this line is gone entirely — there's now no breadcrumb trail during the (much longer, up to 40-iteration) wait. The terminal `tracing::error!` on the caller's failure path still logs the full survivor list, so the final outcome is still diagnosable, but if someone is tailing `meridian logs` live while an install hangs, they'll now see nothing for up to ~10s instead of a debug line every 300ms. Given this whole PR chain (#598/#606/#607) exists to make exactly this class of Windows failure diagnosable, it might be worth having the caller log a lower-frequency progress line (e.g. every N tries) rather than dropping it silently — though since it was `debug` level (below what ships to central telemetry per the redaction rules in CLAUDE.md), this only affects local live-tailing, not remote diagnosis.

### What looks solid

- The two-layer defense is well-reasoned: `stop_running_daemon_before_stage` (Windows-only, kills + waits for the *previous* daemon's own exit, which is the actual multi-second case) is a different concern from `rename_with_retry` inside `stage_binary` (all platforms, catches a residual AV/indexer hold *after* the daemon is confirmed gone) — good separation rather than conflating "wait for the known holder to die" with "retry against an unknown transient holder."
- `stage_binary`'s use of `rename` instead of in-place overwrite for the reason given in its doc comment (a running process's demand-paged code signature validation on macOS) means the new retry wrapper is genuinely inert on non-Windows platforms rather than papering over a different problem there — correctly scoped.
- Test coverage is thorough and appropriately boundary-focused: `attempts == 1` (no retry) is explicitly pinned for `retry_transient`, and `wait_until_gone`'s three tests cover "already gone," "recovers mid-window," and "never recovers" separately rather than only the happy path.

Akarsh-Hegde:
## Review fixes applied — ready to merge ✅

Thanks for the thorough pass. Pushed `18de191a`. **CI is green** (Rust macOS: clippy + test ✅).

### 🟢 Lost per-probe progress logging — fixed
Restored a breadcrumb during the wait, throttled so it isn't the old per-300ms spam: the probe now emits a `tracing::debug!` roughly every ~2s (every 8th 250ms probe) *while the daemon is still present*, so a live `meridian logs` tail stays informed across the up-to-10s window. `wait_until_gone` stays logging-agnostic (its unit tests are untouched) — the throttle lives in the caller's probe wrapper. Implemented as a plain reset counter (no `%` / bit-mask) deliberately, to avoid exactly the kind of clippy-lint sensitivity that surfaced on #606.

### 🟡 Duplicated retry-with-backoff vs #606 — agreed, deferring to a dedicated follow-up (with rationale)
You're right that a shared `meridian-core::retry` is the correct end-state, and I want to do it properly rather than half-way. The reason I'm **not** folding it into this PR:

- #606's retry is **sync** (`std::fs` / `std::thread::sleep`); this PR's is **async** (`tokio`). A real shared home needs *both* variants.
- To actually *dedupe* (not just relocate), **both** call sites must adopt the shared module. #606's lives in `db_crypto.rs`, which #606 already modifies — touching it from this branch would **conflict on merge** and couple two currently-independent, already-green PRs (one of them the data-loss-sensitive encryption path).
- Relocating only this PR's async copy while leaving #606's behind would leave **two retry impls inside `meridian-core` itself** — arguably worse, and likely to draw the same comment again.

So the clean move is **one small follow-up after #606 and #607 both land**: add `meridian-core::retry` with sync + async variants and migrate *both* call sites in a single reviewable change — no cross-branch coupling, no churn on the encryption path. Tracking that as the immediate next step once these merge.

Everything else in the review (the two-layer defense separation, `rename`-not-overwrite scoping, the `#[cfg_attr]` placement, and the boundary-focused test set) is unchanged. No behavior change in this commit beyond the added breadcrumb.

PR #610, opened 2026-07-29
refactor(core): unify the Windows transient-lock retry into meridian-core

Akarsh-Hegde:
## What & why

Follow-up to #606 and #607, closing out #607's 🟡 review note.

Both merged PRs independently implemented the same thing — *"retry a fallible op with linear backoff to ride out a transient Windows file lock (os error 32)"* — in two crates:
- #606: **sync** (`std::fs` / `std::thread::sleep`) in `meridian-core/db_crypto.rs`
- #607: **async** (`tokio::fs` / `tokio::time`) in the tray's `backend_install.rs`

The #607 reviewer flagged the duplication and suggested generalizing into `meridian-core` (already a tray dependency and the shared data layer). I deliberately deferred it until both landed, so the unification could touch **both** call sites in one reviewable change with no cross-branch coupling — which is what this PR does.

## Changes

- **New `meridian_core::retry` module** — two variants of one schedule (`base_delay × attempt` backoff, final attempt's error surfaced):
  - `retry_transient` — async (`tokio::time::sleep`), for the tray's backend install.
  - `retry_transient_blocking` — sync (`std::thread::sleep`), for the encrypt-in-place swap.
  Both `pub`, platform-neutral, and unit-tested here.
- **`db_crypto::rename_with_retry`** → delegates to `retry_transient_blocking`.
- **tray `rename_with_retry`** → delegates to `retry_transient`.
  Both drop their private copies of the retry loop.
- **`meridian-core` tokio** gains the `time` feature (every binary linking this crate already builds tokio with `full`, so per the crate's existing note this adds nothing new to compile).

## Not folded in

`wait_until_gone` (the tray's daemon-exit poll) stays local — it's a *poll-until-empty* shape returning a pid list, not a retry-an-op, and it's Windows-specific. Sharing it would be a stretch of the abstraction.

## Tests

8 in the new module — the four boundary properties (first-try success / recover-after-N / exhaust-and-surface-error / `attempts == 1`) × both variants. Deterministic (zero-delay backoff, call-counting fakes), so they run fast on the macOS CI — the platform that actually runs `cargo test`. The existing `rename_with_retry` / `wait_until_gone` / `finalize_swap_*` tests are unchanged and continue to exercise the wrappers.

**No behavior change**: same attempt counts, same backoff, same surfaced errors — purely a de-duplication.

### DIFF
--- a/meridian-core/Cargo.toml
+++ b/meridian-core/Cargo.toml
@@ -38,8 +38,9 @@ tracing = "0.1"
 tokio = { version = "1", features = ["process", "time"] }
 
 [dev-dependencies]
-# For the #[ignore]d golden-compare smoke test (runs get_today against the real DB).
-tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
+# `macros`/`rt-multi-thread` for #[tokio::test]; `test-util` for the paused-clock
+# (`start_paused`) tests in `retry` that verify the backoff schedule deterministically.
+tokio = { version = "1", features = ["macros", "rt-multi-thread", "test-util"] }
 # Integration tests build an in-memory schema + seed rows to exercise the
 # readers (sqlx is a regular dep, but integration tests need it in scope too).
 sqlx = { version = "0.7", features = ["sqlite", "runtime-tokio-rustls"] }
--- a/meridian-core/Cargo.toml
+++ b/meridian-core/Cargo.toml
@@ -29,13 +29,13 @@ dirs = "5"
 # Instrumentation facade only (~no-op without a subscriber) — the heavy OTLP
 # exporter is wired by the binary (daemon, or the tray's `otel` dev feature).
 tracing = "0.1"
-# Only the `process` feature, only for `proc_ext::NoWindow`'s impl on
-# `tokio::process::Command` (the std impl needs no dep). tokio is already in the
-# workspace tree via sqlx's tokio runtime and every binary that depends on this
-# crate (daemon, tray) already builds it with `full`, so feature-unification
-# means this adds nothing new to compile — it just brings the `process` module
-# into scope here.
-tokio = { version = "1", features = ["process"] }
+# `process` for `proc_ext::NoWindow`'s impl on `tokio::process::Command` (the std
+# impl needs no dep); `time` for `retry::retry_transient`'s async backoff sleep.
+# tokio is already in the workspace tree via sqlx's tokio runtime and every binary
+# that depends on this crate (daemon, tray) already builds it with `full`, so
+# feature-unification means this adds nothing new to compile — it just brings the
+# `process` and `time` modules into scope here.
+tokio = { version = "1", features = ["process", "time"] }
 
 [dev-dependencies]
 # For the #[ignore]d golden-compare smoke test (runs get_today against the real DB).
--- a/meridian-core/src/db_crypto.rs
+++ b/meridian-core/src/db_crypto.rs
@@ -263,32 +263,24 @@ fn interrupted_migration_leftovers(path: &Path) -> Vec<std::path::PathBuf> {
     found
 }
 
-/// `std::fs::rename` with a short bounded retry. On Windows a rename fails with
-/// os error 32 ("used by another process") while *any* handle to the source or
-/// destination is briefly open — an antivirus scan, the search indexer, or a
-/// SQLite handle sqlx's worker thread hasn't finished closing yet. A few
-/// backed-off attempts (~1s total) clear those transient holders without
-/// hanging; a genuinely stuck file (e.g. the daemon holding it for real) still
-/// fails, so the caller's fallback/rollback still runs. No-op fast path on the
-/// common case (first attempt succeeds). Unix renames don't hit this but the
-/// retry is harmless there.
+/// `std::fs::rename` with a short bounded retry (~1s of linear backoff across 5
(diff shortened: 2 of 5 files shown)


Akarsh-Hegde (commented):
## Review (strict pass)

This is exactly the consolidation the #607 review asked for, and it's a genuinely clean, no-behavior-change refactor — I checked both migrated call sites against their pre-refactor originals:

- `db_crypto::rename_with_retry` → `retry_transient_blocking(5, Duration::from_millis(100), …)` — same `ATTEMPTS`/base delay as the original inline loop.
- `backend_install::rename_with_retry` → `meridian_core::retry::retry_transient(RENAME_ATTEMPTS, RENAME_BASE_DELAY, …)` — `RENAME_ATTEMPTS`/`RENAME_BASE_DELAY` constants are untouched, so the schedule is identical.

The `attempt >= attempts` vs. the original `attempt == ATTEMPTS` guard is behaviorally equivalent (attempt increments by exactly 1 each iteration starting at 1, so it can only ever equal or first exceed `attempts` on the same iteration `==` would have fired). The 8 new tests are a faithful split of the four boundary properties (first-success, recovers-after-N, exhausts-and-surfaces, `attempts==1`) across both variants, and CI is green.

Also confirmed while reading the surrounding code: the misplaced-doc-comment issue flagged on #606 (`finalize_encryption_swap`'s doc having been pushed onto the wrong function) has since been fixed on `pre-main` — not part of this diff, but worth noting it's resolved and this PR doesn't reintroduce anything like it (the doc comment for `rename_with_retry` stays correctly attached to `rename_with_retry` here; no new function is spliced between an existing doc and its item).

### 🟡 Suggestion: this is the right moment to add the tracing this pattern has never had

Neither original implementation ever logged per-attempt progress inside the retry loop itself (the #607 review flagged that the tray's version *lost* its per-attempt `tracing::debug!` when it was generalized from the original 10-iteration loop; `db_crypto`'s version never had one to begin with). Right now `meridian_core::retry` — the one shared implementation both callers now go through — is silent: no span, no per-attempt log, nothing until the *caller* logs the final outcome. Given:

- CLAUDE.md's blanket instrumentation rule ("Any new or changed code path that does real work … must emit structured logs and traces … as you write the code, not as an afterthought"), and
- this module exists *specifically* to absorb a failure class (Windows os error 32) that the PR chain has repeatedly noted can only be validated live, never in CI —

this consolidation is the cheapest point to fix that for both call sites at once: a single `tracing::debug!(attempt, attempts, "retrying after a transient failure")` before each sleep would cost nothing (tracing is already a `meridian-core` dependency) and would mean the next time someone is live-diagnosing a stuck Windows install or migration, `meridian logs` actually shows the retry cadence instead of only the caller's before/after state. Not a blocker — the module is currently no worse than what it replaced — but it's exactly the kind of gap that's easy to fix once and easy to keep re-deferring forever if it isn't fixed here.

### Everything else checks out

- `wait_until_gone` staying local is the right call — it's a poll-until-empty-list shape, not a retry-an-op shape, and folding it in would be a real abstraction stretch, not a natural generalization.
- The `pub mod retry;` without a flat `meridian_core::retry_transient` re-export is consistent with how this crate already treats other utility modules (`db_crypto`, `proc_ext`, `notifications`, `plan_marker` are all reached via their submodule path, not flattened) — the "flat public API" convention in the crate's module doc applies to the DB-reader modules specifically, not every utility added since. No inconsistency here.
- `tokio`'s `time` feature addition is correctly scoped and the "adds nothing new to compile" claim is consistent with the existing `process`-feature comment's same reasoning.

Akarsh-Hegde:
## Review fix applied — ready to merge ✅

Thanks for the strict pass, and for verifying both migrated call sites against their originals. **CI is green** (Rust macOS: clippy + test) on the latest commit.

### 🟡 Per-attempt tracing in the shared retry loop — fixed (`8c455feb`)
You're right that this was the cheapest place to fix it for both callers at once, and that it's exactly the kind of gap that gets re-deferred forever. Done:

- A `debug`-level trace (`attempt`, `attempts`) now fires **before each backoff sleep** in *both* `retry_transient` and `retry_transient_blocking`.
- So the tray's install rename and `db_crypto`'s encrypt-in-place swap both emit retry cadence into `meridian logs` now — the live diagnosability CLAUDE.md's instrumentation rule calls for, on a failure class (Windows os error 32) that can only be reproduced on a real box. Neither original ever had this (the tray's was lost in the #607 generalization; `db_crypto`'s never existed).
- Module docs updated to note the behavior. No error is logged (keeps the generic `E` unconstrained — no `Debug` bound), matching your suggested shape.
- No behavior change: debug logs are a no-op without a subscriber, so the timing/count tests are unaffected.

### Also in this PR since your review
While it was open I pushed exhaustive test hardening (`7af4f3f7`, CI green) that the review predates — the retry primitive went **8 → 16 tests**, adding for both variants: success on the *final allowed* attempt (off-by-one guard), last-error-not-first, the `attempts == 0` floor, and a verification that the backoff is actually **linear** (`base × attempt`) — checked exactly via tokio's paused clock for the async path, and by real-elapsed lower bound for the sync path. `wait_until_gone` gained its `attempts == 0` boundary too.

Nothing else outstanding — the refactor is behavior-preserving, both call sites are schedule-identical to their pre-refactor originals, and the one review note is resolved.

Issue #609, opened 2026-07-29
Split the four files over the 500-line rule, and two follow-ups deferred from #603


Akarsh-Hegde:
Deferred from the second CodeRabbit review on #603. Every item here is **pre-existing** - none was introduced by that PR, which is why they were held back from a release carrying an urgent production hotfix (#598).

## File splits

All already over the 500-line rule on `main` before #603:

| file | on `main` | on `pre-main` | workspace |
|---|---|---|---|
| `src/llm/detect.rs` | 1747 | 1860 | daemon |
| `tray/src-tauri/src/lib.rs` | 1437 | 1601 | tray |
| `tray/src-tauri/src/backend_install.rs` | 1120 | 1443 | tray |
| `src/telemetry_spool/redact.rs` | 1217 | 1507 | daemon |
| `meridian-core/src/db_crypto.rs` | 559 | 1036 | meridian-core |

`detect.rs` and `backend_install.rs` were added after CodeRabbit's third review round on #603 flagged them. Note `detect.rs` is over the limit by 3.5x and predates all of this work - the policy has been unenforced long enough that these should be tackled as one deliberate sweep, not one thread at a time.

Suggested seams from the review:

- **`lib.rs`** - extract `DbEncryptionNotice` / `db_encryption_notice_action` and its tests, then peel off further startup responsibilities.
  **Handle with care.** This is the Tauri `setup` closure. During #603 a `return Ok(())` added inside it would have silently skipped the tray menu build and the rest of startup - it compiled clean and all 169 tray tests passed. An early return in that closure is not visible as a bug at the call site. Any extraction here wants a deliberate review, not a mechanical move. Note also that the tray is a separate cargo workspace covered by neither CI nor the pre-push hook.
- **`redact.rs`** - move the pseudonym helpers (`choose_pseudonym_source`, `pseudonymize_account`, `pseudonymize_host`) and the test suite into submodules.
- **`db_crypto.rs`** - split plaintext/key-guard helpers and `open_pool_with_key*` from `finalize_encryption_swap` and its swap helpers; relocate the associated tests. Do this *with* the injectable rename seam below in mind rather than mechanically moving the untestable branch.
- **`backend_install.rs`** - split staging/rename-retry logic, Windows stop/poll logic, and their test modules. Same tray caveat as `lib.rs`: no CI coverage.
- **`detect.rs`** - split CLI resolution/installer-command logic, provider-health classification, and test-cache persistence.

## Two behavioural follow-ups

[...]

PR #616, opened 2026-07-29
fix(tray): log each failed rename attempt during daemon staging

Akarsh-Hegde:
Addresses the one actionable finding from CodeRabbit's third review round on #603. The other two findings in that round are pre-existing file-size violations, deferred to #609 with reasoning posted on their threads.

## The gap

`rename_with_retry` absorbed transient Windows file holds **silently**. Since retrying is exactly what turns a self-clearing blip into a success, whether it happened is the thing worth being able to see - and with no per-attempt breadcrumb a `meridian logs` tail shows nothing until the final success or failure. "Retried four times and recovered" and "went straight through" are indistinguishable after the fact.

That matters on this path specifically: it is the one #598 is about. The hold it guards against is an antivirus scan or the Search Indexer touching the daemon binary mid-install - precisely when someone is tailing logs asking why an install is slow, or whether it is about to fail.

## Choices worth flagging

**`debug!`, not `warn!`.** A transient hold the retry absorbs is expected behaviour on Windows, not a fault. Logging it at WARN would also egress it to the central backend under the redaction rules, turning routine Windows installs into a stream of warnings that drown real ones. The genuinely stuck case still surfaces through the returned error at the call site.

**The counter lives in `rename_with_retry`, not `retry_transient`.** The fields that make the log useful are `from`/`to`, which only the caller knows. Pushing logging into the generic helper would either produce a context-free message or require threading a label through every caller and its tests - so the helper stays a silent, unit-testable utility.

## Verification

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` in `tray/src-tauri` - **183 tests green**, including the existing `rename_with_retry` / `retry_transient` cases that pin the retry behaviour this restructures the closure around.

Worth noting the tray is a separate cargo workspace covered by neither CI nor the pre-push hook, so that local run is the only check this gets.

### DIFF
--- a/tray/src-tauri/src/backend_install.rs
+++ b/tray/src-tauri/src/backend_install.rs
@@ -779,9 +779,43 @@ where
 /// self-clearing blip into a silent success instead of an install-failed notice.
 /// A persistent failure still returns the real error so a genuinely locked file
 /// is not swept under the rug.
+///
+/// # Why each failed attempt is logged
+/// Silence here is indistinguishable from success. The retries exist for the
+/// Windows case where a scanner or the Search Indexer briefly holds the daemon
+/// binary, and that hold is exactly when someone is tailing `meridian logs`
+/// asking why an install is slow - but with no per-attempt breadcrumb the tail
+/// shows nothing at all until the final success or failure, so "retried four
+/// times and recovered" and "went straight through" look identical after the
+/// fact. Given retrying is precisely what turns a self-clearing blip into a
+/// silent success, whether it happened is the thing worth being able to see.
+///
+/// `debug!` rather than `warn!`: a transient hold that the retry absorbs is
+/// expected behaviour on Windows, not a fault. The genuinely stuck case still
+/// surfaces through the returned error at the call site.
+///
+/// The counter lives here rather than in [`retry_transient`] because the useful
+/// fields are `from`/`to`, which only the caller knows - the generic helper
+/// stays a silent, unit-testable utility.
 async fn rename_with_retry(from: &Path, to: &Path) -> std::io::Result<()> {
+    let mut attempt: u32 = 0;
     retry_transient(RENAME_ATTEMPTS, RENAME_BASE_DELAY, || {
-        tokio::fs::rename(from, to)
+        attempt += 1;
+        let this_attempt = attempt;
+        async move {
+            let result = tokio::fs::rename(from, to).await;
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
## Conflict resolved - one of the two commits here was superseded

`pre-main` moved while this was open. **#610 landed the same fix, better**, and #615 landed the cache-warm deduplication I had just deferred on #603.

### What conflicted

#610 moved `retry_transient` out of the tray into `meridian_core::retry` **and** added the per-attempt `tracing::debug!` that this branch's first commit (`d000b60d`) had added locally to `rename_with_retry`.

That is the better shape by some distance: the breadcrumb now covers **every** caller of the retry primitive - async and blocking, including the encrypt-in-place file swap - rather than the single rename this branch instrumented. So `d000b60d` is superseded, and I resolved the conflict by taking `pre-main`'s version wholesale rather than trying to preserve mine.

### One detail lost, filed rather than worked around

The shared helper logs only `attempt` and `attempts`:

```rust
tracing::debug!(attempt, attempts, "retry: transient failure, backing off before the next attempt");
```

No error, no subject. During a real Windows hold that means a `meridian logs` tail shows *that* a retry happened but not that it was `os error 32` on `~/.meridian/bin/meridian.exe` - the difference between "AV is holding the binary" and "the disk is full".

The tempting fix - re-add a breadcrumb at the call site - **double-logs the same fact**, which the observability rules forbid ("emit each fact once, on the span that owns it"). The right fix is an `E: Display` bound on the shared helper so it can log `error = %e`; all current callers pass `io::Error` or `anyhow::Error`, so it should be additive. Filed on #609 rather than done here, since it belongs with #610's code and not in a release branch.

### What still stands

The `setup.rs` commit (`1b24b00f`), which is independent of all this and is the remaining reason to merge:

- the sign-in vendor is a structured `provider` field rather than message interpolation
- failed installs and sign-ins mark their span `ERROR` instead of looking successful

All three of its tests pass post-merge, and I re-verified they still catch their regressions rather than passing vacuously.

### Verification after the merge

- `tray/src-tauri`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` - **183 green** (the count moved from 186 because #610 relocated the `retry_transient` tests into `meridian-core`, not because anything was dropped)
- daemon workspace: `cargo test` - **826 green**

PR #624, opened 2026-07-29
feat(setup): add Windows notification permission step to setup wizard

Akarsh-Hegde:
## What

Adds a **notification permission section to the setup wizard on Windows**. Previously the first-run wizard dropped the entire Permissions step on Windows (`buildSteps` filtered out `permissions`), so notification onboarding never surfaced there.

Notifications *are* a real per-app setting on Windows (WinRT `ToastNotifier::Setting()`), and the Rust backend already fully supported it — `check_notifications` / `request_notifications` have Windows arms, and `open_permission_pane` deep-links to `ms-settings:notifications`. The only gap was the frontend hiding the step. The two TCC grants (Accessibility, Screen Recording) genuinely have no Windows analogue and remain macOS-only.

## How

- **`buildSteps`** now *shapes* the Permissions step per platform instead of dropping it: macOS keeps all three cards; Windows gets a notifications-only variant (`WINDOWS_NOTIFICATIONS_STEP`) with its own copy ("Turn on notifications"). Both keep `id: 'permissions'` so the `check_notifications` poll in `page.tsx` still fires on Windows.
- The Windows step **never gates Continue** — notifications is optional (on by default for most apps).
- Extracted a shared **`PermCard`** component so the card, grant action, and deny→Settings recovery are identical on both OSes (no duplicated markup).
- **`Completion`** summary and the privacy note are now platform-aware ("this PC" vs "this Mac"; "Notifications: Enabled/Off" vs "Permissions: N of 2 granted").
- Threaded `platform` onto the `Wiz` handle.

No backend changes — everything the Windows card needs already existed.

## Behaviour on Windows

| Notification setting | Card shows |
|---|---|
| Enabled | "Granted" |
| Disabled (any variant) | "Open Settings" -> `ms-settings:notifications` |

(Windows never reports `prompt`, so the one-shot "Allow" dialog path stays macOS-only.)

## Test

- New source-guard test `ui/__tests__/windows-notification-permission.test.ts` locks the invariant that Windows must not drop notification onboarding again (matches the existing string-source guard pattern used for wizard/settings surfaces).
- `bun test` (new + notifications/oauth/setup-adjacent): **30 pass, 0 fail**.
- `tsc --noEmit`: no errors in `app/setup` source.

> [!NOTE]
> Verified via typecheck + unit tests. A full Windows app run wasn't possible in this environment (the repo is on a Parallels share). Before merge, the wizard should get a smoke check on a **packaged** Windows build (not `tauri dev` — the notifications plugin only registers in a bundled install, so a dev run always shows the `unavailable` fallback). The empty-step case is now handled with a fallback line either way.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ui/__tests__/windows-notification-permission.test.ts
+++ b/ui/__tests__/windows-notification-permission.test.ts
@@ -11,6 +11,12 @@ import { readFileSync } from 'fs'
 // is a string-source guard (the same pattern as the Settings → Notifications
 // guard) because steps.tsx pulls in the whole React/Clerk import graph.
 //
+// The assertions are ANCHORED to the specific declaration they guard — an
+// earlier version scanned the whole file with `[\s\S]*` and matched the macOS
+// step's `id: 'permissions'` and the integrations step's `canNext: () => true`,
+// so it passed against the very regressions it named. Each assertion below was
+// verified to FAIL when the invariant it guards is mutated.
+//
 // Related backend: tray/src-tauri/src/commands/setup.rs (check_/request_notifications),
 // tray/src-tauri/src/sys.rs (windows_permission_state), commands/system.rs
 // (open_permission_pane "notifications" → ms-settings:notifications).
@@ -18,33 +24,40 @@ import { readFileSync } from 'fs'
 const uiRoot = import.meta.dir + '/..'
 const readSrc = (rel: string): string => readFileSync(uiRoot + '/' + rel, 'utf8')
 
+/** The body of a top-level `const NAME … = {` declaration: from the declaration
+ *  to the first column-0 `}`. Scoped so an assertion can't match an unrelated
+ *  step elsewhere in the file. */
+function declBlock(src: string, marker: string): string {
+  const start = src.indexOf(marker)
+  if (start < 0) throw new Error(`marker not found: ${marker}`)
+  return src.slice(start).split('\n}')[0]
+}
+
 describe('windows notification permission onboarding', () => {
   const steps = readSrc('app/setup/steps.tsx')
 
-  it('does not drop the whole permissions step on Windows', () => {
-    // The old behaviour filtered the step out entirely — the regression this
-    // feature reverses. If this reappears, Windows loses notification onboarding.
-    expect(steps).not.toContain("filter((s) => s.id !== 'permissions')")
-  })
-
-  it('swaps in a notifications-only variant that keeps id:permissions', () => {
-    expect(steps).toContain('WINDOWS_NOTIFICATIONS_STEP')
-    // buildSteps must map the permissions step to the Windows variant, not remove it.
-    expect(steps).toMatch(/platform === 'windows'[\s\S]*s\.id === 'permissions'[\s\S]*WINDOWS_NOTIFICATIONS_STEP/)
-    // Same id as macOS so the check_notifications poll (page.tsx gates it on
-    // steps[step].id === 'permissions') keeps firing on Windows.
-    expect(steps).toMatch(/WINDOWS_NOTIFICATIONS_STEP[\s\S]*id: 'permissions'/)
+  it('keeps the Windows step under id:permissions so the poll still fires', () => {
+    // page.tsx gates the check_notifications poll on steps[step].id === 'permissions'.
+    const winStep = declBlock(steps, 'const WINDOWS_NOTIFICATIONS_STEP')
+    expect(winStep).toContain("id: 'permissions'")
   })
 
   it('never blocks Continue on the optional Windows notifications card', () => {
-    // Notifications is optional on both OSes; on Windows it is the only card, so
-    // the step must not gate navigation.
-    expect(steps).toMatch(/WINDOWS_NOTIFICATIONS_STEP[\s\S]*canNext: \(\) => true/)
+    const winStep = declBlock(steps, 'const WINDOWS_NOTIFICATIONS_STEP')
+    expect(winStep).toContain('canNext: () => true')
+  })
+
+  it('swaps the permissions step for the Windows variant instead of dropping it', () => {
+    const build = steps.slice(steps.indexOf('export function buildSteps'))
+    // The Windows branch must reference the variant (a swap)…
+    expect(build).toContain('WINDOWS_NOTIFICATIONS_STEP')
+    // …and must NOT filter the step list — the old drop, in any formatting.
+    expect(build).not.toMatch(/\.filter\(/)
   })
 
   it('keeps a notifications entry the Windows card can render', () => {
-    const data = readSrc('app/setup/data.ts')
-    expect(data).toMatch(/id: 'notifications'[\s\S]*pane: 'notifications'/)
-    expect(data).toContain('required: false')
+    const notif = declBlock(readSrc('app/setup/data.ts'), "id: 'notifications'")
+    expect(notif).toContain("pane: 'notifications'")
+    expect(notif).toContain('required: false')
   })
 })
--- a/ui/__tests__/windows-notification-permission.test.ts
+++ b/ui/__tests__/windows-notification-permission.test.ts
@@ -1,0 +1,50 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+import { describe, it, expect } from 'bun:test'
+import { readFileSync } from 'fs'
+
+// Guards the Windows notification onboarding in the setup wizard. Windows has no
+// TCC-style consent for Accessibility/Screen Recording, so those cards are
+// macOS-only — but Notifications IS a real per-app setting on Windows (WinRT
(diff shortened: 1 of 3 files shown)


Akarsh-Hegde:
## Review

The core change is right, and the reasoning in the PR body checks out against the backend. I verified independently:

- `sys.rs::notification_permission_state` does have a real Windows arm (WinRT `ToastNotifier::Setting()`, deliberately bypassing the plugin's hardcoded-`Granted` stub), so the card reflects real state on Windows.
- `windows_permission_state` only ever yields `Granted`/`Denied` — the claim that Windows never reports `prompt` holds, so the macOS-only "Allow" dialog path is correctly unreachable there.
- `commands/system.rs::permission_pane_url` (Windows arm) handles `"notifications"` → `ms-settings:notifications`, so the deny → Settings recovery works.
- Both variants keep `id: 'permissions'`, so `page.tsx:110`'s `check_notifications` poll does still fire on Windows.
- `tsc --noEmit`: no errors in app sources. `bun test __tests__/windows-notification-permission.test.ts`: 4 pass. CI picks the file up (`ci.yml:288` runs `bun test` in `ui/`).

One blocking item, one pre-merge condition, and some nits below.

---

### 1. Blocking - the new guard test cannot fail

The PR's stated safeguard is "locks the invariant that Windows must not drop notification onboarding again". 3 of its 4 assertions are vacuous: they pass against the exact regressions they name.

Both `[\s\S]*` assertions are unanchored, and `WINDOWS_NOTIFICATIONS_STEP` first appears in a **comment at line 323** — above `PERMISSIONS_STEP`. So they match against the macOS step's `id: 'permissions'` (line 327) and the integrations step's `canNext: () => true` (line 356), not the Windows block at all.

I applied the two regressions and re-ran the assertions:

```js
// mutate the Windows step: change its id AND make it gate Continue
s = s.replace("const WINDOWS_NOTIFICATIONS_STEP: StepMeta = {\n  id: 'permissions'",
              "const WINDOWS_NOTIFICATIONS_STEP: StepMeta = {\n  id: 'notifs'")
     .replace("canNext: () => true,",   // the Windows one
              "canNext: (s) => s.perms.notifications === 'granted',")

/WINDOWS_NOTIFICATIONS_STEP[\s\S]*id: 'permissions'/.test(s)     // → true  (should fail)
/WINDOWS_NOTIFICATIONS_STEP[\s\S]*canNext: \(\) => true/.test(s) // → true  (should fail)
```

The first assertion has the same problem for a different reason: `not.toContain("filter((s) => s.id !== 'permissions')")` asserts the absence of one exact byte sequence. A revert written as `.filter(s => s.id !== 'permissions')` (no parens - a plausible way to write it) passes:

```js
// buildSteps reverted to the old filter, minus the arg parens
"? ALL_STEPS.filter(s => s.id !== 'permissions')"
!m1.includes("filter((s) => s.id !== 'permissions')")  // → true  (should fail)
```

So only `toContain('WINDOWS_NOTIFICATIONS_STEP')` and the `data.ts` pair currently carry weight.

**Fix** - stay with the source-string-guard pattern (agreed that importing `buildSteps` drags in the React/Clerk graph), just anchor to the declaration instead of scanning the whole file:

```ts
const block = (marker: string) => steps.slice(steps.indexOf(marker)).split('\n}')[0]

const winStep = block('const WINDOWS_NOTIFICATIONS_STEP')
expect(winStep).toContain("id: 'permissions'")
expect(winStep).toContain('canNext: () => true')

const build = steps.slice(steps.indexOf('export function buildSteps'))
expect(build).toContain('WINDOWS_NOTIFICATIONS_STEP')  // maps, doesn't drop
expect(build).not.toMatch(/\.filter\(/)                // formatting-independent
```

I ran this against both mutations - all three now fail on the regression and pass on the current source, so it's verified rather than just proposed.

### 2. Pre-merge condition - the smoke check must be on a *packaged* Windows build

`PermCard` returns `null` when `perms.notifications === 'unavailable'`. On macOS that hides one of three cards. On Windows it is the **only** card, so the step renders as a title ("Turn on notifications") plus the privacy sentence and nothing else - a visually broken but navigable step (`canNext` is `() => true`, so at least it's not a dead end).

That state is not hypothetical for the reviewer: the notifications plugin only registers when `is_bundled()` (`lib.rs:146`), so **`tauri dev` on Windows always yields `unavailable`** - the blank step is what every Windows dev run shows. `notification_permission_state` also returns `None` on a WinRT probe error in a packaged build (e.g. the AUMID isn't registered), so it's reachable in production too.

The PR asks for "a smoke check on a real Windows build before merge" - please make that explicitly a **packaged** build, otherwise whoever does it sees the blank step and can't tell the bug from the environment. Worth considering a one-line fallback ("Notifications aren't available in this build") when the only card hides, so the step is never empty.

### 3. Question - no per-app deep link on the Windows pane

Worth comparing the two arms of `permission_pane_url` side by side:

- macOS: `…preference.notifications?id={identifier}` - with a long comment explaining the `?id=` anchor exists specifically to land on *Meridian's* detail pane rather than the app list.
- Windows: `ms-settings:notifications` - flat, no per-app anchor.

That asymmetry was tolerable when the pane was one recovery path among three cards; this PR makes "Open Settings" the entire step's only action, so a denied Windows user lands on the full app list and has to find Meridian. Does a per-AUMID deep link exist? (Genuine question - not blocking if it doesn't.)

Adjacent, cosmetic only: `windows_permission_state` collapses every `Disabled*` variant (for-user, group policy, manifest) to `Denied`. sys.rs justifies that with "there is exactly one recovery action for all of them", which was true for a background probe - but this PR promotes it to a step titled "Turn on notifications" where a group-policy-disabled machine gets a button that can never succeed. Continue isn't gated, so it's a papercut, not a trap.

### Nits (non-blocking)

- The Windows subtitle and the single card's `desc` are near-verbatim duplicates ("nudge you when a worklog draft is ready or your plan needs attention… quiet by default"). The user reads the same sentence twice in one viewport - the subtitle could carry the *why* and leave the *what* to the card.
- The permissions-step poll still fires `check_accessibility` / `check_screen_recording` every 2 s on Windows, for values nothing renders there.
- `PermissionsBody` re-derives the platform branch from `wiz.platform` although `buildSteps` already decided it. It can't drift at runtime (same threaded value), but having the step meta own its card list would remove the second decision point.

New user-facing strings are all plain-hyphen compliant, and the `PermCard` extraction is a clean de-duplication - no notes there.

Akarsh-Hegde:
Thanks for the deep review — all three actionable items addressed in `4951a3d4` (rebased onto the latest `pre-main` merge as `531bc18d`).

### 1. Blocking — guard test is now real
Rewrote the assertions to anchor to the specific declaration rather than scanning the whole file, using a small `declBlock(src, marker)` helper (declaration → first column-0 `}`). I reproduced your mutation test in node against the current source:

- **Regression A** (Windows step `id` → `'notifs'` **and** `canNext` gated on `granted`): the `id: 'permissions'` and `canNext: () => true` assertions now **FAIL**. ✅
- **Regression B** (`buildSteps` reverted to `.filter(s => s.id !== 'permissions')`, no arg parens): the `WINDOWS_NOTIFICATIONS_STEP`-is-referenced and `not.toMatch(/\.filter\(/)` assertions now **FAIL**. ✅

Both pass on the real source, fail on the regression — verified, not just proposed.

### 2. Pre-merge condition — empty Windows step handled in code
Added a fallback line in `PermissionsBody`: when the notifications card is the whole step (Windows) and it hides (`unavailable` — every `tauri dev` run, or a `None` WinRT probe in a packaged build), the step now renders a "Notification settings aren't available in this build — you can turn them on later from Windows Settings" row instead of a blank pane. So the state is no longer indistinguishable from a bug.

The smoke-check note in the PR body is updated to say **packaged** build explicitly.

### 3. Question — per-app deep link on Windows
Confirmed: **there isn't one.** It's already documented in `commands/system.rs:137-144` — Windows has no `?id=<bundle-id>` equivalent for the notifications pane; `ms-settings:notifications` opens the general Notifications page, and Meridian appears in that list once it has delivered its first toast under its AUMID. So the denied-Windows-user-lands-on-the-list behaviour is the platform ceiling, not something this PR can tighten. Left as-is.

On the group-policy-`Denied` papercut: agreed it's a papercut, not a trap (Continue isn't gated). Since collapsing every `Disabled*` → `Denied` is pre-existing backend behaviour in `sys.rs` and a group-policy-locked machine is rare, I left it out of scope here rather than widen the diff.

### Nits
- **Subtitle/desc duplication**: reworded the Windows subtitle to carry the *why* ("Stay in the loop without watching the menu bar…") and leave the *what* to the card. ✅
- **Windows polling accessibility/screen**, and **`PermissionsBody` re-deriving the platform branch**: left as-is to keep the diff focused — both are harmless (unused values / same threaded value, can't drift at runtime). Happy to fold in the poll guard if you'd prefer it in this PR.

PR #636, opened 2026-07-30
fix(tray): refuse to mint a DB key that would orphan an encrypted meridian.db

Akarsh-Hegde:
## Problem

`db_key::resolve_or_create_key` treats **any** `keyring::Error::NoEntry` as "first run" and unconditionally generates + stores a fresh encryption key.

That is correct only when there genuinely is no database yet. If the keychain entry goes missing for any other reason — a keychain reset, a restore onto a new machine, `.env` and the keychain drifting apart — the real `meridian.db` is still encrypted under the old, now-unreachable key. Minting a replacement compounds a recoverable situation into an unrecoverable one, and nothing says so.

## Status of the evidence — please read, an earlier version of this PR was wrong

This PR originally claimed the bug explained an ongoing error storm on two alpha machines (`mac_970e5ebf236cb0ce` / `win_970e5ebf236cb0ce`, ~190k errors in 28h). **That attribution was incorrect and has been retracted.**

Direct inspection of the affected Mac showed:
- the keychain entry is present,
- `.env` carries `MERIDIAN_DB_KEY`,
- and that key **successfully decrypts the database** (verified with `sqlcipher`, read-only: `sqlite_master` reads back 99 objects).

Nothing was orphaned. The real cause there is **page-level corruption in a database that opens fine** — `PRAGMA quick_check` reports a genuine structural fault (`Tree 37 page 669426 cell 9: overflow list length is 4 but should be 8`) plus a long run of freelist inconsistencies. Consistent with the `meridian.db.severely-corrupt-*` / `.pre-repair-backup-*` files already on that disk.

The tell I misread: SQLCipher's wrong-key signature is code 26 (`file is not a database`). The dominant error was code 11 (`database disk image is malformed`), outnumbering code 26 ~85:1 — that is corruption, not a bad key.

**So this bug is real but latent.** It is a hole in the key-resolution path that has not been observed firing in the wild. It is worth closing on its own merits; it is not a fix for the corruption issue, which needs separate work.

## Change

**`db_key.rs`** — classify the on-disk DB before minting:

```rust
enum ExistingDb {
    Absent,                            // genuine first run
    Plaintext,                         // encrypt_in_place migrates it
    TooSmallToBeADatabase { bytes },   // empty/truncated stub - nothing to lose
    LooksEncrypted,                    // ciphertext under a key we cannot find
}
```

Only `LooksEncrypted` bails. `TooSmallToBeADatabase` proceeds with a `WARN` carrying the observed size: a file shorter than the 16-byte magic cannot hold a SQLCipher page (smallest is 512 bytes), so blocking startup there would be a false alarm that sends a user with nothing to lose to support. An existing file whose metadata cannot be read is classified `LooksEncrypted` — failing closed is the only direction here whose mistake is irreversible.

**`lib.rs`** — distinguishes this from the pre-existing generic fallback, which logs "continuing unencrypted" and is wrong here (there is no unencrypted file to continue into). The DB-backed notice system can't carry the warning either, since `meridian.db` is the thing that won't open — so it fires a native OS notification. Best-effort: it silently drops if notification permission was never granted, though the `tracing::error!` still ships to central telemetry.

Also binds `Path::new(&db_path)` once instead of rebuilding it at five call sites.

## Scope

- ✅ Closes a latent path that can turn a recoverable keychain problem into permanent data loss.
- ✅ Makes that failure loud and attributable rather than silent.
- ❌ Does **not** address the corruption issue described above.
- ❌ Cannot recover a database whose key is genuinely gone — nothing can.

## Test plan

- `cargo fmt --check`, `cargo clippy --lib -- -D warnings` — clean
- `cargo test --lib` (tray) — 193/193
- Full pre-push suite (fmt + UI build + clippy + UI tests + security audit + cargo test) — green
- CI: Rust macOS + Windows, CodeQL ×4 — green
- 6 unit tests over the classifier: missing, plaintext, ciphertext, empty stub, truncated stub, exact-magic-length boundary

Not covered: an end-to-end run through `resolve_or_create_key`'s bail path, which needs a real or mocked OS keychain. The extracted classifier carries the logic and is covered directly.

Unrelated pre-existing flake seen once during development: `commands::setup::tests::signin_outcome_success_logs_at_info` — `Arc::try_unwrap(seen).unwrap()` racing subscriber teardown in that module's own `capture()` helper. Passes in isolation and on repeated full runs; untracked, worth its own ticket.

### DIFF
--- a/tray/src-tauri/src/db_key.rs
+++ b/tray/src-tauri/src/db_key.rs
@@ -38,6 +38,16 @@ fn generate_key_hex() -> String {
     hex::encode(bytes)
 }
 
+/// Would minting a brand-new key orphan an already-encrypted `meridian.db`?
+/// True only when `db_path` exists and does not look like a plaintext SQLite
+/// file — i.e. it looks like ciphertext under some key this install can no
+/// longer find in the keychain. A missing file (fresh install) or a
+/// confirmed-plaintext file (nothing encrypted yet — `encrypt_in_place` will
+/// migrate it under the new key) are both safe to proceed on.
+pub(crate) fn would_orphan_existing_db(db_path: &std::path::Path) -> bool {
+    db_path.exists() && !meridian_core::db_crypto::is_plaintext_sqlite(db_path)
+}
+
 /// Get-or-create this machine's `meridian.db` encryption key from the OS
 /// keychain, and ensure it is mirrored into `env_path` as
 /// `MERIDIAN_DB_KEY=<hex>` so the daemon can read it. Returns the key hex.
@@ -45,13 +55,36 @@ fn generate_key_hex() -> String {
 /// Idempotent and safe to call on every tray startup: an existing keychain
 /// entry is reused as-is (never regenerated), and re-mirroring the same value
 /// into `.env` is a no-op write via `upsert_env`.
-pub fn resolve_or_create_key(env_path: &std::path::Path) -> anyhow::Result<String> {
+///
+/// `db_path` exists solely so a missing keychain entry can be checked against
+/// `meridian.db` before minting a replacement — see [`would_orphan_existing_db`].
+/// Without that check, a keychain entry lost for any reason (a keychain
+/// reset, a migration to a new machine, `.env` and the keychain falling out
+/// of sync) silently mints and stores a fresh key while the real data stays
+/// encrypted under the old, now-unreachable one: every future open then fails
+/// with SQLCipher "wrong key" errors (`file is not a database` / `database
+/// disk image is malformed`) instead of a clear, actionable one.
+pub fn resolve_or_create_key(
+    env_path: &std::path::Path,
+    db_path: &std::path::Path,
+) -> anyhow::Result<String> {
     let entry =
         keyring::Entry::new(SERVICE, ACCOUNT).context("db_key: failed to access OS keychain")?;
 
     let key_hex = match entry.get_password() {
         Ok(existing) => existing,
         Err(keyring::Error::NoEntry) => {
+            if would_orphan_existing_db(db_path) {
+                anyhow::bail!(
+                    "no DB encryption key found in the OS keychain, but {} already exists and \
+                     does not look like a plaintext SQLite file - it looks encrypted under a \
+                     key this install can no longer find. Refusing to generate a replacement \
+                     key, which would silently make that data permanently unreadable instead \
+                     of surfacing the problem. If this file is known-safe to discard, remove \
+                     it and restart.",
+                    db_path.display()
+                );
+            }
             let generated = generate_key_hex();
             entry
                 .set_password(&generated)
@@ -84,5 +117,30 @@ mod tests {
         meridian_core::db_crypto::validate_key_hex(&a).unwrap();
         meridian_core::db_crypto::validate_key_hex(&b).unwrap();
         assert_ne!(a, b, "two generated keys should not collide");
+    }
+
+    #[test]
+    fn would_orphan_existing_db_is_false_when_missing() {
+        let dir = tempfile::tempdir().unwrap();
+        let db_path = dir.path().join("meridian.db");
+        assert!(!db_path.exists());
+        assert!(!would_orphan_existing_db(&db_path));
+    }
+
+    #[test]
+    fn would_orphan_existing_db_is_false_when_plaintext() {
+        let dir = tempfile::tempdir().unwrap();
+        let db_path = dir.path().join("meridian.db");
+        std::fs::write(&db_path, b"SQLite format 3\0rest-of-a-plaintext-file").unwrap();
+        assert!(!would_orphan_existing_db(&db_path));
+    }
+
+    #[test]
+    fn would_orphan_existing_db_is_true_when_not_plaintext() {
+        let dir = tempfile::tempdir().unwrap();
+        let db_path = dir.path().join("meridian.db");
+        // No recognizable SQLite header - stands in for SQLCipher ciphertext.
+        std::fs::write(&db_path, b"not-a-sqlite-header-at-all-just-ciphertext").unwrap();
+        assert!(would_orphan_existing_db(&db_path));
     }
 }
--- a/tray/src-tauri/src/db_key.rs
+++ b/tray/src-tauri/src/db_key.rs
@@ -38,14 +38,66 @@ fn generate_key_hex() -> String {
     hex::encode(bytes)
 }
 
+/// Length of the SQLite file-header magic. A file shorter than this cannot be
+/// probed for it at all — see [`ExistingDb::TooSmallToBeADatabase`].
+const SQLITE_MAGIC_LEN: u64 = 16;
+
+/// What the on-disk `meridian.db` looks like, as far as deciding whether it is
+/// safe to mint a fresh encryption key is concerned.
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde (commented):
## Overview

This PR closes a real data-loss hole in `db_key::resolve_or_create_key`: on `keyring::Error::NoEntry`, the old code unconditionally minted and stored a fresh key, silently orphaning any already-encrypted `meridian.db` whose real key just isn't reachable anymore (keychain reset, machine migration, `.env`/keychain drift). The new `would_orphan_existing_db()` gate checks whether the DB file exists and doesn't look like plaintext before minting, and bails with an actionable error instead. `lib.rs`'s caller distinguishes this from the generic fallback and fires a native OS toast, since the DB-backed notice system can't be used (the very DB that's unreadable is what would carry it). Good, well-scoped fix, and the PR body itself is refreshingly honest about what it does *not* do (no recovery for the two already-orphaned machines).

## Code quality / correctness

- The core logic is sound and conservative in the right direction: `is_plaintext_sqlite` collapses "unreadable" to `false` (per its own doc comment), so an unreadable-but-existing file also trips `would_orphan_existing_db` → bail rather than mint. That's the safe default.
- Single call site (`lib.rs`), so the signature change (`resolve_or_create_key(env_path, db_path)`) doesn't leave anything stale — confirmed via grep, only the one caller exists.
- Tests directly cover the three states of `would_orphan_existing_db` (missing / plaintext / non-plaintext). Good, cheap, deterministic unit tests. There's no test that drives `resolve_or_create_key` itself into the bail path end-to-end, but that's understandable — it needs a real/mocked OS keychain, not easily unit-testable, and the extracted helper is what actually carries the logic under test.
- `would_orphan_existing_db(db_path)` gets called twice on the same failure — once inside `resolve_or_create_key` to decide whether to bail, once again in `lib.rs`'s error arm to decide which log/toast branch to take. Harmless (single small file read, `db_path.exists()` + a 16-byte read), but it does mean the predicate and the `anyhow::Error` returned from the first call can only be reconciled by re-deriving the same fact rather than reading it off the error itself. Not blocking — the codebase leans on flat `anyhow::Result` throughout — but if this pattern recurs elsewhere, a small enum/error variant would avoid the duplicate I/O and the two checks silently drifting out of sync.
- Docs are thorough and match this repo's convention (rationale-heavy comments explaining *why*, cross-links to `db_crypto`). The `resolve_or_create_key` and `would_orphan_existing_db` doc comments correctly summarize the failure mode described in the PR body.

## Potential issue worth a look

`is_plaintext_sqlite`'s doc comment says a file **too small to hold the 16-byte magic** is also collapsed to "not plaintext." That means a zero-byte or truncated `meridian.db` — e.g. exactly the kind of disk-full torn write this codebase has hit before — will also trip `would_orphan_existing_db` and produce the "your local database appears to be encrypted with a key this install can no longer find, contact support before removing anything" message. For a genuinely empty/truncated stub (nothing encrypted, nothing to lose), that's a scarier and more support-intensive message than the situation warrants — the real fix there is just deleting the stub, not a keychain/migration investigation. Worth considering whether a file below the magic-byte threshold deserves a distinct message (or at least a distinct `tracing` field) from a file that has a real, unrecognized non-SQLite header. Not a correctness bug — the bail path is still the *safe* direction to fail in — but it may generate confusing support tickets for what could be a routine corrupt/empty-file case.

## Minor nits

- `std::path::Path::new(&db_path)` is constructed three separate times in the same `lib.rs` block (once for `resolve_or_create_key`, once for `would_orphan_existing_db`, once implicitly via the pre-existing `is_plaintext_sqlite` call slightly below). Binding it once would read a little cleaner, though this is purely style.
- User-facing toast body ("Contact support with your Support ID (Settings -> Account) before removing anything.") correctly uses a plain `->`/hyphen per the project's no-em-dash rule for user-facing text — just flagging that it was checked, no change needed.

## Security / risk

- No new attack surface — this only tightens an existing local-only key-resolution path to fail closed instead of fail silent-and-destructive.
- The toast is explicitly best-effort (drops if notification permission was never granted or the plugin isn't registered in an unbundled run); the PR body acknowledges this and notes the `tracing::error!` still ships to central telemetry regardless, so the failure isn't fully silent even if the OS toast never renders.

## Test plan

Reported in the PR body: fmt/clippy clean, full pre-push suite green, 190/190 tray unit tests, 3 new targeted tests for the new helper. The one pre-existing flaky test noted (`signin_outcome_success_logs_at_info`) is called out as unrelated and reproducibly passing in isolation — reasonable to not block on it, though worth a follow-up ticket if it isn't tracked already.

**Overall:** solid, well-documented, conservative fix for a real bug. No blocking issues — the truncated/empty-file messaging point above is worth a look but doesn't change the correctness of the fail-closed behavior.

Akarsh-Hegde:
Thanks — addressed in `fba22505`.

**The truncated/empty-file messaging point: fixed, and it went slightly further than a message change.**

You were right that `is_plaintext_sqlite` collapsing a short read into `false` meant a zero-byte or truncated `meridian.db` tripped the guard. On looking at it, a distinct *message* still felt like the wrong answer: a file shorter than the 16-byte magic can't hold a SQLCipher page (smallest is 512 bytes), so it isn't orphaned data under any key — blocking startup at all is the false positive, not just the wording.

So the probe is now an explicit classification:

```rust
enum ExistingDb {
    Absent,
    Plaintext,
    TooSmallToBeADatabase { bytes: u64 },
    LooksEncrypted,
}
```

`TooSmallToBeADatabase` proceeds (a fresh DB gets keyed over it) with a `WARN` carrying the observed size — deliberately not silent, since quietly stepping over a corrupt file is how the original bug stayed invisible. Only `LooksEncrypted` bails. Net effect: the routine corrupt-stub case self-heals instead of generating the support ticket you flagged.

Two details worth calling out:

- **An existing file whose metadata can't be read still classifies as `LooksEncrypted`.** Failing closed on a file we can't inspect is the one direction where being wrong is irreversible.
- **The variant carries `bytes`** so the log site doesn't re-stat. My first cut used `metadata(...).unwrap_or(0)` at the `warn!` — which would have logged a confident `0` on a stat failure. Numeric attributes are kept by `redact.rs` regardless of key, so the size does survive the ship leg to central telemetry.

**Also fixed:** `Path::new(&db_path)` is now bound once in `lib.rs` and reused at all five call sites (three in the new block, two pre-existing).

**Not changed — the duplicate `would_orphan_existing_db` call.** Both sides now route through the single `classify_existing_db`, so the logic can't drift; what remains is one extra 16-byte read on a rare error path. A typed error carrying the state would remove it, but that means departing from the flat `anyhow::Result` convention for one call site, which didn't seem worth it. Happy to do it if you'd rather.

**Also not changed — no end-to-end test through `resolve_or_create_key`'s bail path.** Still needs a real/mocked keychain; the extracted classifier is what carries the logic and is now covered at four states including the exact-magic-length boundary.

New tests: empty stub, truncated stub, `SQLITE_MAGIC_LEN`-exactly boundary (stays conservative). 193/193 tray tests pass; full pre-push suite green.

Re the flaky `signin_outcome_success_logs_at_info` — root cause is `Arc::try_unwrap(seen).unwrap()` in that module's own `capture()` helper racing subscriber teardown, unrelated to this PR. Not tracked anywhere I could find; worth its own ticket.

PR #642, opened 2026-07-31
fix(health): drop the stale session-summary skill check and its diagnosis

Akarsh-Hegde:
## The bug

`meridian doctor` reports a **critical** failure on every machine with Claude Code installed:

```
✗  session-summary skill    ~/.claude/commands/session-summary.md missing
                            — claude summariser fails for every session (left pending)

Diagnosis
● session-summary Claude Code command is missing
    The `claude -p /session-summary` invocation ... returns 'Unknown command' ...
    fix: `meridian doctor --fix`
```

**None of that is true any more.** The Claude engine stopped invoking a slash-skill when it moved to embedding `SUMMARY_RULES` inline — `summariser/claude.rs` builds the prompt itself:

```rust
let prompt = format!("{} Summarise the coding-session transcript provided on stdin.", prompts::SUMMARY_RULES);
let args = vec!["-p".into(), prompt, "--output-format".into(), "json".into(), ...];
```

`SUMMARISER_SKILL` was deprecated in the same change (`summariser/config.rs:19-21`), and `skill_name` now exists only to drive a startup warning. Nothing has opened `~/.claude/commands/session-summary.md` since, so its absence means nothing at all — and `doctor --fix` was writing a file no code path reads.

### Verified on a live install

The summariser working normally, with the file absent and doctor calling it critically broken at the same moment:

```
08:30:04  summarising coding-agent segment with prior-burst continuity context
08:31:39  summarised coding-agent segment
08:32:07  summarised coding-agent segment
```

Full `summariser_prompt` → `summariser_inference` → `summariser_output` span trees. No "Unknown command" anywhere in the log history.

## Why it matters more than a cosmetic false positive

In `diagnose.rs` the skill check was the **first branch of an `if/else if` chain**, ahead of the queue check:

```rust
if bad("coding-agent", "session-summary skill") {
    ...                                    // could never be right
} else if bad("meridian daemon", "summariser queue") {
    ...                                    // detects an actual stall
}
```

So on any machine missing that file — which is every machine that didn't run `install.sh` — a **genuinely stalled summariser was masked**. Doctor reported the phantom problem and never evaluated the check that measures the thing that actually matters.

Beyond that, the check names a real subsystem, marks itself critical, and offers a remedy, so it actively misdirects anyone triaging a real problem. It did exactly that to me during an unrelated investigation, with the machine's own logs in hand.

## Changes

- **`health/codingagent.rs`** — check removed (plus the two bindings it was the only user of).
- **`health/diagnose.rs`** — `Diagnosis` removed; the `summariser queue` branch is promoted to primary, which is the correct priority.
- **`health/fix.rs`** — `doctor --fix` arm removed.
- **`summariser/claude.rs`** — header comment still read *"Run `claude -p` with the session-summary skill"*. That stale line is the origin of the belief, so it is corrected at the source rather than only downstream.
- **`summariser/README.md`** — `SUMMARISER_SKILL` documented as deprecated/inert.

## Deliberately not done

The `coding-agent-install-skill` subcommand (`src/main.rs`) and its `install.sh:576` call are left in place. Both are harmless and idempotent, and removing a shipped CLI command is a maintainer's call rather than a drive-by in a bugfix. Say the word and it's a two-line follow-up.

## Reviewer note

Not compiled locally: `cargo clippy` on this machine dies in `openssl-sys`' vendored build (`Command 'perl' not found`) — an environment gap, not a code one; GitHub's windows image ships Strawberry Perl. My files were never reached. CI is the first real build.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -572,12 +572,9 @@ if [[ "${NO_DAEMON}" -eq 0 ]]; then
     # to your tracker (Jira/Linear/GitHub) only after you approve them in the
     # dashboard (Worklogs view).
 
-    info "Installing session-summary Claude Code command..."
-    if run "${MERIDIAN_BIN}" coding-agent-install-skill; then
-        ok "session-summary command → ~/.claude/commands/session-summary.md"
-    else
-        warn "session-summary command install skipped — run 'meridian coding-agent-install-skill' later"
-    fi
+    # No session-summary Claude Code command is installed any more: the Claude
+    # summariser embeds its prompt inline in `claude -p`, so the file this step
+    # used to write was never read. See src/coding_agent_session_ingest/summariser/claude.rs.
 
     # Coding-agent summariser engines (informational): each agent's sessions are
     # summarised by its OWN CLI when present. There is no fallback — a missing CLI
--- a/scripts/meridian-cli.sh
+++ b/scripts/meridian-cli.sh
@@ -587,7 +587,7 @@ case "$CMD" in
     uninstall)        cmd_uninstall ;;
     permissions)      cmd_permissions ;;
     version|--version|-v) cat "${REPO_ROOT}/VERSION" 2>/dev/null || echo "unknown" ;;
-    worklog-status|coding-agent-hook|coding-agent-summarise|coding-agent-install-skill|oauth-login|tasks-sync|ticket-update|ticket-parents|ticket-statuses|ticket-set-status|worklog-generate|worklog-generate-get|worklog-generate-approve|worklog-post-approved) cmd_daemon_passthrough "$CMD" "$@" ;;
+    worklog-status|coding-agent-hook|coding-agent-summarise|oauth-login|tasks-sync|ticket-update|ticket-parents|ticket-statuses|ticket-set-status|worklog-generate|worklog-generate-get|worklog-generate-approve|worklog-post-approved) cmd_daemon_passthrough "$CMD" "$@" ;;
     --help|-h|help|"") cmd_help ;;
     *) err "unknown command: ${CMD}"; echo; cmd_help; exit 1 ;;
 esac
--- a/src/coding_agent_session_ingest/summariser/README.md
+++ b/src/coding_agent_session_ingest/summariser/README.md
@@ -145,7 +145,7 @@ standalone poll).
 | `SUMMARISER_SWEEP_S` | `30` | catch-up sweep cadence |
 | `SUMMARISER_BATCH_PER_TICK` | `8` | rows per drain pass |
 | `SUMMARISER_MODEL` | `claude-haiku-4-5-20251001` | Claude model |
-| `SUMMARISER_SKILL` | `session-summary` | skill name |
+| `SUMMARISER_SKILL` | `session-summary` | **DEPRECATED — no effect.** The Claude engine embeds `SUMMARY_RULES` inline in `claude -p`; it has not invoked a slash-skill since. Setting this to anything else only logs a warning. |
 | `SUMMARISER_CLAUDE_TIMEOUT_S` | `240` | `claude -p` timeout |
 | `SUMMARISER_CODEX_MODEL` | (empty → codex default) | Codex model |
 | `SUMMARISER_CODEX_TIMEOUT_S` | `240` | `codex exec` timeout |
--- a/src/coding_agent_session_ingest/summariser/claude.rs
+++ b/src/coding_agent_session_ingest/summariser/claude.rs
@@ -1,6 +1,10 @@
 //ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
 //
-// Run `claude -p` with the session-summary skill + structured output. Returns
+// Run `claude -p` with the summary rules embedded inline + structured output.
+// (It does NOT invoke a `/session-summary` slash-skill — that was replaced by
+// the inline SUMMARY_RULES prompt below, and `SUMMARISER_SKILL` is deprecated.
+// This comment claiming otherwise is what kept a stale doctor check alive.)
+// Returns
 // the validated {summary}, or RateLimited / Failed — both leave the row pending
 // for a later drain (no cross-engine fallback).
 //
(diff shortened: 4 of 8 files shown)


Akarsh-Hegde (commented):
## Review: fix(health): drop the stale session-summary skill check and its diagnosis

**Overview**

Removes the `doctor` "session-summary skill" check (`~/.claude/commands/session-summary.md` existence), its `Diagnosis` branch, and its `--fix` arm. The check went stale when the Claude summariser engine moved to embedding `SUMMARY_RULES` inline in the `claude -p` prompt instead of invoking a `/session-summary` slash-skill — nothing has read that file since. Worse, the stale check sat as the *first* branch in `diagnose.rs`'s `if/else if` chain, so it could mask a genuinely stalled summariser (the "summariser queue" check) behind a phantom critical failure. That priority bug is the real fix here; the check removal is cleanup of the root cause.

**Correctness**

- Confirmed via `git show` on the PR branch that `which` and `PathBuf` in `health/codingagent.rs` are both still used elsewhere in the file post-edit — no unused-import clippy fallout from the removal.
- `diagnose.rs`'s `if/else if` collapses cleanly to a single `if` now that the first branch is gone; no dangling-`else` issue.
- `fix.rs`'s removed match arm leaves the rest of the `match (g, n)` intact.
- The PR's claim that nothing reads `SUMMARISER_SKILL`/the skill file anymore matches what's in `summariser/claude.rs` and `summariser/config.rs` (only drives a deprecation warning) — didn't take this on faith, checked it.
- Scope is well-bounded: `coding-agent-install-skill` (the CLI subcommand + its `install.sh` call) is deliberately left in place, called out explicitly in the PR body as a maintainer decision rather than a drive-by removal. Good call — avoids scope creep in a bugfix PR.

**Test coverage — the one gap**

Existing tests (`diagnose.rs::summariser_backlog_chains_to_one_root_cause`, `fix.rs::summariser_backlog_is_guided`) don't reference the removed skill check, so nothing breaks — but that also means **the actual bug (wrong branch priority masking a real stall) has no regression test**. Worth adding, since this is exactly the kind of priority/ordering bug that regresses silently:

```rust
#[test]
fn summariser_queue_diagnosis_survives_without_skill_check() {
    // what used to trigger the phantom "session-summary skill" critical
    // alongside a real stall — now only the real one should fire
    let report = Report::new(vec![
        Check::warn("summariser queue", "L2", "293 backed up").in_group("meridian daemon"),
    ]);
    let dx = root_causes(&report);
    assert_eq!(dx.len(), 1);
    assert!(dx[0].title.contains("summariser"));
}
```
(Roughly duplicates `summariser_backlog_chains_to_one_root_cause`, but explicitly pins "this fires with nothing else in the report" — guards against a future higher-priority branch reintroducing the same masking bug.)

**Documentation**

- The stale header comment in `summariser/claude.rs` ("Run `claude -p` with the session-summary skill") is fixed at the source, which the PR correctly identifies as where the false belief originated — good instinct, not just patching downstream symptoms.
- `README.md`'s `SUMMARISER_SKILL` row and the in-code comments at each removal site clearly explain *why* the code is gone, not just that it's gone — matches the file's own bar for explaining non-obvious removals.

**Risk**

Low. Pure removal of dead/misleading diagnostic code plus a priority fix in a pure function (`root_causes`); no schema, DB, or public-API surface touched. The "not compiled locally" caveat (env-level `openssl-sys`/perl issue, unrelated to these files) is reasonable to flag but shouldn't block — CI will catch any real compile issue.

**Verdict:** Solid, well-scoped bugfix with above-average PR hygiene (root-caused the stale comment, called out what was deliberately left alone). Only suggestion is the regression test for the branch-priority fix above.

Akarsh-Hegde:
## Review addressed - ready to merge

Thanks for the review. Two things changed since it was written.

### 1. Regression test for the branch-priority fix — added (`f7e7b632`)

You were right that this was the real gap: the priority bug is exactly the kind that regresses silently, and nothing pinned it.

I took a slightly different shape than the sketch, because `summariser_queue_diagnosis_survives_without_skill_check` as written asserts the queue branch fires **in isolation** — which `summariser_backlog_chains_to_one_root_cause` already covers, so it would pass both before and after this PR and wouldn't catch the bug. The masking was only ever observable when something else in the chain outranked the queue check, so the test feeds in **both** signals:

```rust
#[test]
fn a_coding_agent_warning_does_not_mask_a_real_summariser_stall() {
    let report = Report::new(vec![
        Check::warn("session-summary skill", "L2", "missing").in_group("coding-agent"),
        Check::warn("summariser queue", "L2", "293 backed up").in_group("meridian daemon"),
    ]);
    let dx = root_causes(&report);
    let titles: Vec<&str> = dx.iter().map(|d| d.title.as_str()).collect();
    assert!(titles.iter().any(|t| t.contains("summariser")), ...);
    assert!(!titles.iter().any(|t| t.contains("session-summary")), ...);
}
```

Under the old code the phantom `"session-summary Claude Code command is missing"` diagnosis wins the `if/else if` and the summariser one is never pushed — so this **fails on the pre-PR code and passes now**, which is what makes it a regression test rather than a restatement. The second assertion pins that the deleted check can't quietly return as a diagnosis of its own.

### 2. Scope changed — `coding-agent-install-skill` is now removed too (`f502b5ed`)

Your review correctly noted it was deliberately left in place. The maintainer has since asked for it, so it's gone:

- the subcommand block in `src/main.rs`
- `install.sh`'s "Installing session-summary Claude Code command" step
- the passthrough entry in `scripts/meridian-cli.sh`

**`assets/skills/coding-agent/session-summary/SKILL.md` deliberately stays** — it looks like part of the same dead machinery, but `summariser/prompts.rs:23` `include_str!`s it and it *is* the source of `SUMMARY_RULES`, the inline prompt the engine actually sends. Deleting it would break the summariser outright.

Worth noting the subcommand also carried a **hand-copied duplicate** of that prompt, kept in sync with the real asset by comment only (`// Keep in sync with assets/skills/...`). Two copies, one of them unread — removing the subcommand kills that drift hazard.

Shell scripts verified with `bash -n`, since CI doesn't lint them.

### Status

All checks green on the final commit: **Rust (Windows x86_64) 4m7s**, **Rust (macOS Apple Silicon) 3m19s**, **Analyze (rust) 2m42s**. The earlier "not compiled locally" caveat is resolved by CI — clippy `-D warnings` and the full suite build clean on both platforms.

Ready to merge from my side.

PR #661, opened 2026-08-02
fix(db): address CodeRabbit review findings on the corruption-recovery PR

Akarsh-Hegde:
## Summary

Addresses the CodeRabbit review comments left on #658 (the `pre-main → main` release PR), which introduced the DB corruption detection/recovery feature. Once this merges into `pre-main`, #658's diff will pick these up automatically.

### Fixed

- **`meridian-core/src/db_crypto.rs` (CRITICAL - broken tests):** two `swap_database_file` tests still asserted the pre-generalization error wording (`"move plaintext original aside"` / `"restored the plaintext original"`), which the generalized function no longer produces. Both were failing. Updated to match, and added an assertion on the `op` prefix.
- **`meridian-core/src/db_crypto.rs`:** the rollback-failure `tracing::error!` branch was missing the `op` field the other two branches have, breaking telemetry attribution for the one failure that needs a human. Also fixed the stale "multi-GB encrypted export" comment to cover the repair caller too, and escaped embedded single quotes in `encrypt_in_place`'s `ATTACH` path literal (same defect flagged in the new repair code, pre-existing here).
- **`src/db/repair/rebuild.rs`:** `ATTACH DATABASE '<path>'` broke on a path containing an apostrophe (e.g. `/Users/o'brien/.meridian`) - escaped by doubling embedded quotes. A `carry_sqlite_sequence` failure previously aborted the whole repair with `?`, discarding an already-successful salvage of every table over a deterministic, unretriable failure - now recorded on the report (`sequence_carried: bool`) and surfaced by the CLI instead.
- **`src/db/repair/mod.rs`:** `inspect()` mapped a corrupt-scan abort to `Scan::default()`, which reports `is_healthy() == true` - the CLI would print "meridian.db is healthy (0 tables scanned)" on a database that just failed to scan. Now records the failure as a problem first. `tray_is_running()`'s process probe failed *open* (`unwrap_or(false)`) when `pgrep`/`tasklist` couldn't run, which could let a repair proceed against a live writer - now fails closed. `repair()` now verifies the rebuilt replacement with `quick_check` before swapping it in, so a bad replacement is refused rather than installed (extracted into `verify_replacement`). Documented the remaining TOCTOU window as a known limitation (see below).
- **`src/db/cli.rs`:** `before_bytes / 1_000_000` integer division printed "0 MB" for any database under 1 MB, reading as a failed repair. Now uses `{:.1}` with float division. Surfaces the new `sequence_carried` warning.
- **`CLAUDE.md`:** fixed a stale heading referencing the non-existent `src/db/repair.rs` (it's a directory: `src/db/repair/`).
- **`src/db/test_corrupt.rs`:** the fixture's `PRAGMA page_size = 4096` can silently no-op once the file header already exists; now reads the effective page size back and asserts it, so a mismatch fails loudly instead of quietly moving the hardcoded damage offsets.
- **`src/main.rs`:** the startup `quick_check` read every page of `meridian.db` unconditionally, adding seconds to every daemon start on a multi-GB database - including every launchd `KeepAlive` restart. Moved to a background `tokio::spawn` task reporting through a shared `Arc<AtomicBool>`; loses no detection coverage since `etl_tick`'s own error path independently classifies corruption the moment a query touches it.

### Declined (verified regressions, documented in place)

Two of CodeRabbit's suggested diffs looked correct but **broke the existing test suite** when implemented and investigated:

- **Batching tier-2 salvage into explicit transactions.** A `COMMIT` after a batch containing even one `SQLITE_CORRUPT` row itself fails with "database disk image is malformed" - confirmed against the real fixture, and it took out every row in the batch, not just the damaged one. SQLite treats a corruption error inside an explicit transaction as poisoning it, which is backwards for a loop whose entire job is to keep going past exactly that error. Left as autocommit-per-row (correct, if slower).
- **Replacing the dense `lo..=hi` rowid range with a pre-enumerated list of real rowids.** A sequential `SELECT rowid FROM src.t` physically walks every intervening page in b-tree order and stops dead at the first damaged one, so it recovers only a prefix. Point lookups by rowid each descend independently from the root and only fail for rows whose own leaf/overflow page is damaged - that's what lets tier 2 isolate exactly the corrupt rows. Confirmed empirically: swapping to sequential enumeration dropped recovery on the motivating fixture from >300/400 rows to 16/400.

Both are explained with inline comments so a future reader doesn't reintroduce them.

- **The `ensure_no_writers` TOCTOU race** (a user could relaunch the tray mid-repair after the point-in-time check passes) is documented as a known limitation rather than partially addressed - closing it needs an exclusive lock honored by both this crate *and* daemon/tray startup (`tray/src-tauri`), which is a cross-binary change out of scope for a review-fix pass. Shipping a lock file nobody checks yet would be a half-finished mechanism, not a fix.

## Test plan

- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` - clean
- [x] `cargo test` (workspace) - 848 passed in the daemon crate, 268 in `meridian-core`, 0 failures
- [x] Verified the two declined suggestions empirically by implementing them and watching `row_by_row_salvage_beats_losing_the_table` fail, before reverting with documentation
- [x] Full `pre-push` hook suite (fmt + clippy + UI tests + `cargo test` + security audit) passed on push

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -630,7 +630,7 @@ meridian db repair   # rebuild into a fresh file; requires the daemon AND tray s
 
 ---
 
-## Database corruption (`src/db/integrity.rs`, `src/db/repair.rs`)
+## Database corruption (`src/db/integrity.rs`, `src/db/repair/`)
 
 SQLite b-tree damage is detected and recovered explicitly; it is not something
 the daemon can shrug off.
--- a/meridian-core/src/db_crypto.rs
+++ b/meridian-core/src/db_crypto.rs
@@ -444,9 +444,11 @@ pub fn swap_database_file(
     if let Err(e) = rename_with_retry(path, backup_path, RENAME_ATTEMPTS_CHEAP) {
         // On Windows `rename` fails with "os error 32" while another process
         // (the running daemon) still holds `meridian.db` open. Remove the
-        // multi-GB encrypted export so it doesn't accumulate as an orphan every
-        // startup (a real disk-exhaustion risk on a large DB); `key_unless_plaintext`
-        // keeps the daemon and CLIs working against the plaintext file meanwhile.
+        // multi-GB replacement (an encrypted export for `encrypt_in_place`, a
+        // rebuilt file for `db repair`) so it doesn't accumulate as an orphan
+        // every startup - a real disk-exhaustion risk on a large DB. The
+        // original is untouched, so both callers keep working against it
+        // meanwhile.
         let _ = std::fs::remove_file(tmp_path);
         tracing::error!(
             error = %e,
@@ -489,6 +491,7 @@ pub fn swap_database_file(
                     error = %e,
                     restore_error = %restore_err,
                     stage = "restore_original",
+                    op,
                     "database swap failed AND rollback failed - the database needs restoring by hand"
                 );
                 Err(e).context(format!(
--- a/meridian-core/src/plan_marker.rs
+++ b/meridian-core/src/plan_marker.rs
@@ -101,7 +101,17 @@ mod tests {
         let dir = std::env::temp_dir().join(format!("meridian-restamp-{}", std::process::id()));
         std::fs::create_dir_all(&dir).unwrap();
         let marker = marker_path(&dir);
-        let now = Local::now();
+        // Pinned to local noon, not `Local::now()`: the "today's marker" case
+        // below subtracts 30 minutes and asserts it lands on the same day,
+        // which flakes if the test happens to run in the first 30 minutes
+        // after local midnight (real-world CI failure - not hypothetical).
+        let now = chrono::Local::now()
+            .date_naive()
+            .and_hms_opt(12, 0, 0)
+            .unwrap()
+            .and_local_timezone(chrono::Local)
+            .single()
+            .expect("noon is unambiguous under any offset");
 
         // No marker (auto-open never fired today) → no re-stamp, no file.
         assert!(!restamp_if_today(&marker, &now));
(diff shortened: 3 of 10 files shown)


Akarsh-Hegde:
Rebased onto the current \`pre-main\` tip (\`8fa5af83\`) to resolve the merge conflicts — a fair amount landed there in the meantime (#657, #659, #660, #662), including a proper marker-based daemon/tray handshake for repair (\`src/db/repair/marker.rs\`, \`tray/src-tauri/src/repair_boot.rs\`) and an updated \`stop_daemon_hint()\`.

Conflicts were only in 3 files, all resolved by keeping both sides' intent rather than picking one:
- **\`meridian-core/src/db_crypto.rs\`**: pre-main had already fixed the same `ATTACH` quoting issue independently via a shared `sql_quoted_path()` helper — took that version over my inline fix. The two stale-test-assertion fixes were also already landed there in a near-identical form (just using a `let msg = ...` local); consolidated to avoid duplicate assertions.
- **`src/db/repair/mod.rs`**: pre-main added `stop_daemon_hint()` right next to where I updated `tray_is_running()`'s doc comment — kept both. Also updated my "known limitation" comment on `ensure_no_writers` to reference the new marker mechanism as prior art, since it now solves the identical TOCTOU problem for the tray-initiated repair path (the CLI path — `meridian db repair` — still doesn't use it, so the limitation itself still stands there).
- **`src/main.rs`**: kept my background-`quick_check` restructuring, updated to pre-main's newer remedy text ("Pause tracking, quit Meridian, then run...").

Re-verified after resolving: `cargo fmt`, `cargo clippy -- -D warnings`, full `cargo test` (865 passed in the daemon crate, up from 848 thanks to pre-main's new tests; 270 in `meridian-core`), a `cargo check` on the tray crate, and the full `pre-push` hook suite (fmt + clippy + UI build + UI tests + `cargo test` + security audit) — all green. Force-pushed the rebased branch.

PR #677, opened 2026-08-04
release: promote pre-main to main - macOS db corruption fixes, telemetry blind spots, inert native dialogs

Akarsh-Hegde:
Release PR: promotes everything currently on `pre-main` to production.

**11 commits, 8 merged PRs (#669–#676), 18 files, +1334/−99.** The bulk of it is the
**macOS `meridian.db` corruption**: two of the known paths to it are closed here, plus
the telemetry blind spots and the dead UI path that made the incident so hard to
diagnose in the first place.

**This does not close the corruption incident.** A third mechanism — the tray's daemon
watchdog killing a healthy daemon mid-write every ~45 s — was measured *after* these
fixes and is still unpatched. See the ⚠️ callout in §1 before treating corruption as
resolved. With #681 in, this release force-updates the whole fleet onto a build that
still has that mechanism live, which makes the callout more important, not less.

Base is 1.83.0 (`3ca1f3b`); head is `44147df`, CI green (CI · Cache warm · Release
(prepare) all ✅ on that SHA) — but see the release-prep note immediately below, which
moves the head.

> ### ⏳ Read this before merging: #681 is not in this diff yet
>
> **[#681](https://github.com/Meridiona/meridian/pull/681) raises the mandatory-update
> floor to 1.83.1** — the version this PR cuts. Merging *that* into `pre-main` and then
> merging *this* means **every install in the fleet force-installs 1.83.1**, with no
> consent banner and no click, within 30 s of its next launch. That is the single most
> consequential effect of this release and it is deliberate: 1.82.3 and earlier carry the
> encryption-swap corruption, 1.83.0 carries the fd exhaustion, and the machines already
> reporting code 11 are exactly the ones that will never click a banner.
>
> #681 also adds the **What's New entry and roadmap item** for 1.83.1. Its notes say
> "two causes of database damage", not "fixed database damage", for the reason in the
> ⚠️ callout below.
>
> **So: merge #681 first, then re-check this PR's CI** — `whats-new.json` is
> `include_str!`d into the tray, so the head moves and the run above no longer covers it.
> If you would rather ship 1.83.1 consent-based, close #681 and merge this alone; nothing
> here depends on it.

---

## 1. Database corruption — two paths closed

Central telemetry over the 30 days after the v1.80.0 encryption rollout put
`database disk image is malformed` (SQLite code 11) at **six macOS machines and zero
Windows machines**. At least four distinct mechanisms have been identified; two of them
are closed here.

### a. The encryption swap ran under a live daemon (#671, #675)
`stop_daemon_for_migration` stopped the daemon **only on Windows**, documenting macOS
as "a no-op … the migration tolerates the open handle". That has the safety backwards:

* On **Windows**, renaming a file another process holds open fails with os error 32 —
  the swap rolls back and the DB is left plaintext. Noisy but safe.
* On **macOS**, the rename **succeeds**. The daemon keeps writing through its handle
  on the old plaintext inode while new connections open the swapped-in encrypted file.
  SQLite addresses `-wal`/`-shm` **by path, not by inode**, so a plaintext writer and
  an encrypted writer share one journal — and `finalize_encryption_swap` had deleted
  those sidecars out from under the live daemon on top of that.

Observed progression on a fresh victim: `SQLITE_IOERR_SHORT_READ` (522) → `file is not
a database` (26) → code 11. Per-machine onset tracks *that machine's migration run*,
not the release date.

The fix:
* SIGTERM is not enough on macOS — the agent is registered with launchd `KeepAlive` and
  returns within seconds, back into the swap window. So: **bootout the agent, wait for
  the domain entry to clear, and re-register it afterwards** in `ensure_daemon_running`
  (itself previously a macOS no-op relying on KeepAlive — otherwise we'd trade
  corruption for a permanently dead daemon).
* A bootout only removes the *launchd-managed* daemon, so it is followed by an `lsof`
  check that nothing else still holds `meridian.db` open — keyed on **the database
  file, not on known daemon paths**, because a dev build, a directly-spawned binary and
  an orphan from an overlapping install have all been seen on one machine and none of
  them are launchd's. **An `lsof` that cannot be run counts as still-held**: being
  unable to prove the file is free is not evidence that it is.
* **The stop's failure now gates the migration** in `lib.rs` instead of only being
  logged. Warning-and-swapping-anyway is exactly what shipped in v1.80.0. Leaving the
  database plaintext for one more launch is the cheap failure.

### b. File-descriptor exhaustion (#676)
macOS gives a GUI/launchd process a soft `RLIMIT_NOFILE` of **256**, and neither entry
point raised it. A `meridian.db` pool costs **three descriptors per connection**
(db/`-wal`/`-shm`); the tray stacks a WebView, sockets and in-process capture on top,
and tray + daemon are attached to the same WAL database at once. Crossing 256 fails a
`-wal`/`-shm` open mid-write with `SQLITE_IOERR` (522), which desyncs the shared WAL
index and surfaces as code 11 — the same 522 → 11 progression the fleet recorded.

Measured on a dev machine: `launchctl limit maxfiles` = 256, tray holding **91**
descriptors, daemon **32**, and no `setrlimit` call anywhere in the tree. Reproduced
locally with a snapshot watcher — healthy at 01:22:00, corrupt at 01:22:31, damage
confined to pages the checkpoint had just written.

New `meridian-core/src/fd_limit.rs` (a port of the screenpipe fork's
`screenpipe-engine::fd_limit`, which hit this in the same architecture) raises the soft
limit to 8192, called from **both** entry points — the tray never runs the daemon's
`main()`. Windows has no equivalent constraint, matching code 11 being 100% macOS /
0% Windows in the fleet data.

> **⚠️ Known remaining paths — NOT fixed here, each needs its own change.**
>
> 1. **The tray's daemon watchdog restart storm** (`poll/mod.rs::run_daemon_watchdog`)
>    — measured 2026-08-04, *after* both fixes above. `probe()` gives the daemon an
>    **800 ms** socket timeout; a daemon busy with an ETL batch or a WAL checkpoint
>    routinely misses it, and `WATCHDOG_STRIKES = 2` means **10 s of slow replies is
>    enough to declare a healthy daemon dead** and `launchctl kickstart -k` it — SIGTERM
>    mid-write, paced by the 45 s cooldown at ~80 kills/hour, forever. Damage lands
>    consistently in the `app_sessions` b-tree tail (rootpage 4), the table the daemon
>    writes; repro is a fresh healthy DB corrupt inside five minutes. **This explains
>    why staging.7 still corrupted with the fd fix (#676) installed**, and why it is
>    macOS-only. The fix is not a longer timeout: the probe conflates "slow to answer"
>    with "dead" and must check the process actually exited before killing.
> 2. **The SQLCipher WAL-reset race.** The bundled SQLCipher is based on **SQLite
>    3.39.4**, inside the 3.7.0–3.51.2 range of
>    [howtocorrupt.html §8.1](https://sqlite.org/howtocorrupt.html) — two connections in
>    separate processes writing/checkpointing the same WAL database concurrently, which
>    is exactly our configuration. The restart storm is what makes this race fire
>    constantly rather than rarely.
>
> Also unresolved: `meridian db repair` cannot recover the storm's damage profile (it
> fails at `ATTACH`), and the `db.corrupt` notice is latched and copied into rebuilt
> databases, so a successful repair leaves the banner up and tracking paused.

---

## 2. Native dialogs were inert in the packaged tray (#669)

`window.confirm()` / `alert()` / `prompt()` **do nothing** in the packaged tray.
WKWebView routes JS dialogs through the host app's `WKUIDelegate`, and nothing in this
stack installs one (`wry`, `tauri-runtime-wry`, `tauri`, `tauri-utils` were each grepped
for `ConfirmPanel|AlertPanel|WKUIDelegate` — all empty), so WKWebView silently returns
the default: `confirm()` is always `false`, `alert()` is a no-op.

It fails **silently** — a falsy `confirm()` is indistinguishable from the user clicking
Cancel, so the gated action just never runs while every log line, gate and test still
passes. Two live instances:

* **`NoticeBar`'s `db.corrupt` "Repair Database" button** gated on `window.confirm`, so
  it could never fire, and reported failures through `window.alert`, so its error path
  was equally invisible. Confirmed on a real corrupt database: the `preview_repair` span
  is in the logs with nothing after it — no marker written, no `request_repair`.
  **Recovery had to be driven by hand from a terminal during the incident above.**
* **`useTimelineData`'s `run()`** reported every failed worklog action
  (approve/reject/post/edit) with a bare `alert()`. Those errors have been invisible —
  the row silently snapped back to its server value with no explanation. Found by the
  new guard test, not by inspection.

Adds `ui/components/ConfirmDialog.tsx` (`ConfirmDialog` for consent, `AlertDialog` for
failures; shared modal chrome — Escape, click-outside, focus-on-open). Deliberately
**not** `tauri-plugin-dialog`: it isn't a dependency, and adding one costs a capability
grant for a box the webview can render itself. Every other modal here (HygieneDialog,
ReviewModal, SettingsModal) already works this way.

---

### Health checks shipped without saying *which* check failed (#673)
`Report::log` is properly instrumented — `check`, `group`, `layer`, `stage`, `detail` —
but **none of those keys were on `redact::SAFE_STRING_KEYS`**, so the ship leg dropped
all five and the backend received `health check failed` with nothing attached. Six
machines were reporting a CRITICAL health fault that could not be attributed to a check,
or even to a subsystem.

This is the same failure mode the existing note on the bare `error` key records: the
allowlist is keyed on the attribute names actually typed at the call site, so a
well-instrumented error arrives as a bare static string and *nothing anywhere fails
loudly*.

`group`/`layer` are `&'static str`; `check` is a `String` but built from a literal at
every call site (audited — none interpolates); `stage` is a fixed set. `detail` needed
judgement: mostly fixed text plus benign operational numbers, but a few sites splice an
error's `Display` (`format!("OAuth token refresh failed ({e})")`), so it also goes on
`FREE_TEXT_KEYS` for the full `scrub_text` + clamp. The test pins that the identifying
fields survive intact, that `detail` has URLs/emails scrubbed, and that widening the
boundary for these five did **not** also let `task_key` / `path` / `window_title` /
`app_name` through.

### ~3.9k WARN/day from one machine burying real failures (#672)
On a Windows machine where `schtasks /Create` was blocked by policy at install time,
`register_service` correctly falls back to a Startup-folder launcher and the daemon
starts fine — but `restart()` still tried `schtasks /Run` on every attempt against a
task that will **never** exist there. A permanent, already-handled configuration was
reported as a fresh transient failure, once per restart, forever: **~3.9k WARNs from one
machine over three days, out of ~6.6k fleet-wide across 7 machines**, burying real
failures in the stream that exists to surface them.

`/Query`'s **exit code** decides which case this is, rather than matching `/Run`'s
stderr — Windows localizes those messages, so text matching would misclassify every
non-English machine as the unexpected case, and those are the machines most likely to be
policy-locked to begin with. A task that exists but refuses to run keeps its WARN. The
recovery path itself was already correct and is unchanged; only the reporting is.

---

## 4. Windows build repairs (#674, part of #676)

Two `-D warnings` / compile breaks introduced by the fixes above, both invisible from
macOS:

* **#674** — `register_agent`/`launchctl` are `cfg_attr(not(macos), allow(dead_code))`
  (compiled everywhere, merely allowed to be dead), but the two helpers they were
  refactored onto were `cfg(target_os = "macos")`, which *removes* them. Windows
  compiled a `register_agent` calling functions that did not exist (two `E0425`s). Both
  now use the same `cfg_attr` form. `wait_for_db_unheld` keeps its hard gate — it is
  reached only from a macOS branch.
* **`fd_limit`** — constants and the policy fn are compiled everywhere but used only by
  the `cfg(unix)` arm → four dead-code errors on the Windows clippy job. Fixed with
  `cfg_attr(not(unix), allow(dead_code))` **deliberately, not `cfg(unix)`**: the policy
  is pure and its tests must keep running on the Windows job, where nothing else
  exercises this module.

Both were verified by extracting the cfg structure into a dependency-free file and
type-checking it with `rustc` for `x86_64-pc-windows-msvc`, `aarch64-apple-darwin` and
`x86_64-unknown-linux-gnu`, **and by reverting the attributes to reproduce the exact
errors**. A real `cargo check --target x86_64-pc-windows-msvc` from macOS cannot stand
in — it dies earlier in `aws-lc-sys`, which needs `windows.h`.

---

## 5. Test coverage for the above (#670, #675, and the guards in #672/#674)

Every fix in this batch ships with a test that was **verified against the broken
behaviour**, not merely observed to agree with the fixed one:

| Guard | Reverting … fails with |
|---|---|
| `may_swap_database` | "the daemon could not be stopped - swapping now corrupts the database" |
| `parse_lsof_holders` | the writer assertions |
| `target_soft_limit` / `desired_target` | both tests (hard-limit cap `min` → `max`) |
| `classify_run_failure` | written test-first against an `E0425` — the classifier didn't exist yet |
| cfg-gating scanner | "`register_agent` is compiled on every platform … but references `bootout_agent_and_wait`, which is `#[cfg(target_os = "macos")]`" |
| `no-native-dialogs.test.ts` | mutation-checked; it is what found the `useTimelineData` case |

Two of these are **source-scanning** rather than behavioural, because units structurally
cannot see the bug: cfg mismatches only exist across platforms tests compile one at a
time, and an inert `window.confirm` passes every assertion. The scanners assert they
still recognise the file's shape (≥3 items of each kind) so a refactor that defeats the
parsing fails loudly rather than passing vacuously on zero findings.

**#670** backfills the Windows-installer work: `ProviderStatus.install_command` — the
field the frontend renders instead of its own static `installHint` — had no coverage.
The Rust tests are deliberately not cfg-gated so they run in **both** Rust CI jobs; the
serde test guards the wire contract specifically (a `#[serde(skip)]` or rename would
leave `probed?.install_command` permanently `undefined` and silently revert the UI to
the platform-wrong hint). Three UI tests pin that exactly one `p.installHint` read
remains.

---

## Docs

`CLAUDE.md` gains two entries so neither trap can be re-introduced by someone who wasn't
here: the **no-native-dialogs rule** under TypeScript conventions (with the WKWebView
mechanism and the pointer to `ConfirmDialog`), and a corrected note that `ui/` now has a
real bun test suite (409 tests / 29 files as of 1.83.0), including why several of them
are source-scanning.

---

## Merge note

Please **merge-commit** this PR rather than rebase-merging it — the auto back-merge in
`release-prepare.yml` conflicts on disjoint histories against a rebase-merged release
PR. (`pre-main` being 2 behind is just the ordinary post-release state: `main`'s own
merge of #658 plus semantic-release's `chore(release): 1.83.0` bump. `git merge-tree`
reports the merge is clean — the `Cargo.lock` edits on each side are different hunks.)

## Verification

CI is green on `44147df` — **CI**, **Cache warm (release)** and **Release (prepare)**
all passed. That covers `cargo fmt` + `cargo clippy --workspace --all-targets -D
warnings` + `cargo test --workspace` on **both** the macOS and Windows Rust jobs, and
`bun test` for `ui/` on the Ubuntu job.

Per-change verification, as recorded in the commit bodies:

- Every guard in the table above was run against the **broken** state, not only the
  fixed one.
- fd-limit, end to end: under `ulimit -Sn 256`, the same `getrlimit`/`setrlimit`
  sequence raises the soft limit to 8192 (rc=0, confirmed by re-reading it).
- Both Windows cfg fixes were type-checked with a dependency-free `rustc` repro for
  three targets, and the attributes reverted to reproduce the exact `E0425` /
  dead-code errors.
- `#672`'s Windows `restart()` path is **not verified locally** — it is
  `cfg(target_os = "windows")` and a cross-target check from macOS dies in
  `aws-lc-sys`; it rests on CI's Windows job.
- One gap worth knowing: `raise_fd_limit` logs before `observability::init` runs, so
  its own INFO/WARN never reach telemetry. The fix works; we just can't see it report
  from the fleet.


Akarsh-Hegde:
## Review triage — 13 findings

Addressed in **#679** (targets `pre-main`, so this PR inherits it).

### Fixed — 6

| finding | verdict |
|---|---|
| `backend_install.rs:432` — restarting a healthy daemon on the macOS up-to-date path | **Correct, and the most valuable finding here.** A fourth path to the corruption: `register_agent` ends in `kickstart -k`, so an ordinary tray launch SIGTERMed a healthy daemon mid-write. Fixed with the same early return Windows already had, verified not to break the booted-out/migration case (`launchctl print` exits 113 → `Some(false)` → still registers). |
| `backend_install.rs:1843` — test exclusion never matches | **Correct.** No test in the file uses a `test_` prefix, so the exclusion never fired and the `>= 3` floor could be satisfied by test items. Now keyed on the `mod tests` boundary. |
| `fd_limit.rs:76` — reject undersized `MERIDIAN_FD_LIMIT` | **Correct.** `0` left the process at 256 — the state that corrupts. Rejected below `DEFAULT_TARGET`, via a pure seam so the test needs no env mutation. |
| `no-native-dialogs.test.ts:68` — `//` inside strings | **Correct, and I'd defended this badly.** My comment called it an acceptable false-negative-only tradeoff. A false negative is exactly the failure this guard exists to prevent. Fixed with a single left-to-right pass — which also closed a false *positive* the same comment claimed was impossible. |
| `fd_limit.rs:125` — span with error status | **Right problem, wrong remedy.** This runs before `observability::init`, so `tracing` has no subscriber and a span would be equally invisible. Switched to `eprintln!` instead, which reaches launchd's stderr redirect and diagnostics bundles. |
| `detect.rs:1883` — assert `install_command` non-empty | **Correct.** `Some("")` would pass `is_some()` and render a blank install hint. |

### Declined — 7

- **`CLAUDE.md:354`, `src/main.rs:27`, `detect.rs:1886`** (file splits / moving tests) and **`daemon_control.rs:326`** (tracing span on a Windows-only query) — all reasonable, none urgent. This is a release PR holding a fleet-wide corruption fix; refactor churn here trades real risk for tidiness. Worth doing separately.
- **`src/main.rs:26`** — load `.env` before reading `MERIDIAN_FD_LIMIT`. Running `raise_fd_limit()` before anything can open a descriptor is the entire point of its position. The override is a process-env knob, and its effect is now observable via the `eprintln!` above.
- **`daemon_control.rs:252`** — preserve query failures as a separate state. Valid but Windows-reporting-only and low-value; batched with the other deferrals.
- **`ConfirmDialog.tsx:173`** — focus trap. A genuine a11y gap, but a behavioural change to a shared modal is not something to land on a release PR unreviewed.

### Note on this PR's description

The body has been updated: it previously said the watchdog restart storm "is still unpatched" and told readers not to treat corruption as resolved. **#678 has since landed on `pre-main` and is part of this promotion**, verified end-to-end on `v1.83.0-staging.8` — 22 samples over 10.5 min with `runs`/pid flat where the old build produced ~14 kills. See §1c.

Akarsh-Hegde:
@coderabbitai review

All 13 threads from the 05:49 review are resolved: 6 fixed in #679 (now merged to `pre-main`, so this PR carries them), 7 declined with reasons — see the triage comment above.

The highest-value finding (`ensure_daemon_running` bouncing a healthy daemon via `kickstart -k`) was correct and is fixed; it was a fourth path to the macOS DB corruption this release exists to close.

Requesting a re-review so the stale `CHANGES_REQUESTED` can clear.

Akarsh-Hegde:
## Second review round — triage

All 9 threads resolved. **8 fixed in #682** (targets `pre-main`, so this PR inherits it), 1 declined.

Two of these were regressions introduced by my own #678/#679, so they got a real fix rather than a resolve-and-move-on.

### Fixed

| finding | verdict |
|---|---|
| `daemon_control:147` — preserve the Windows direct-spawn fallback | **Correct, and the most serious of the round.** When the watchdog and `refresh_health` moved from `restart()` to `start_if_stopped()`, the Windows path became a bare `schtasks /Run` and lost the direct-spawn fallback — so on policy-locked machines **both** automatic recovery paths silently lost their only way back up. Extracted into `run_task_or_spawn`, shared by both. |
| `daemon_control:125` — TOCTOU before `kickstart -k` | **Correct.** Fixed at the root: `register_agent` no longer passes `-k`. Nothing to kill on the happy path anyway (the agent was just booted out and bootstrapped). |
| `daemon_control:125` — don't treat every nonzero exit as "not alive" | **Correct.** Only `113` (no such service) is a definite negative; everything else now reports `None` (unknown) and logs `target` + `exit_code`. Safe because `-k` is gone above. |
| `no-native-dialogs:106` — scan `${…}` as code | **Correct, and mine.** Template substitutions execute; I was blanking them with the literal. A false negative introduced one commit after fixing a different false negative in the same function. Replaced with a context stack that scans substitutions as code, including nested objects and templates. |
| `no-native-dialogs:95` — preserve newline after `\` | **Correct, and mine.** A line continuation dropped a physical line and shifted every offender line number after it. |
| `refresh.rs:123` — notices still say "restart" | **Correct.** I changed the log line and missed the four user-facing messages. All now say *start*. |
| `fd_limit:145` — capture errno | **Correct.** Without it every failure reads identically. |
| `watchdog:253` — ERROR span status | **Correct**, and it is what `CLAUDE.md`'s observability rules already require. Applied to both `daemon_watchdog.start` and `daemon_health.start`. |

### Declined

- **`daemon_control:105`** — return `anyhow::Result` from the new lifecycle helpers. Every platform fn in that module (`probe`/`restart`/`set_running`/`reload`) returns `Result<(), String>` and feeds Tauri commands that serialize to String. Converting only the new two would make the module inconsistent; converting all of them is a separate change, not one for a release PR.

**Merge order:** #682 → `pre-main`, then this PR carries it.

PR #680, opened 2026-08-04
chore(release): raise the mandatory-update floor to 1.83.0

Akarsh-Hegde:
One-line change: `tray/minimum-version` `1.82.3` → `1.83.0`.

Every install **below 1.83.0** will force-install the next release instead of waiting
for a consent-banner click (`update::enforce_minimum_version`, checked 30 s after launch
and every 6 h). The floor ships as a `Minimum-Version: 1.83.0` line inside `latest.json`'s
`notes`, since `tauri-plugin-updater` drops unknown manifest fields.

## Why raise it

1.82.3 and earlier carry the v1.80.0 encryption migration that swaps `meridian.db` under
a live daemon on macOS — the rename succeeds there, leaving a plaintext writer and an
encrypted writer sharing one WAL. Central telemetry puts `database disk image is
malformed` (SQLite code 11) at **six macOS machines, zero Windows**, with per-machine
onset tracking that machine's migration run. Leaving those installs on a banner they may
never click keeps a data-destroying path live in the fleet. The fix is in #671/#675,
already on `pre-main`.

## Why 1.83.0 and not 1.83.1 (the version this release cuts)

Deliberate — it leaves 1.83.0 itself on the consent banner. Two reasons:

1. **`scripts/package-updater.sh:172` fails the release when the floor's core exceeds the
   release version.** `v1.83.0` is not in `pre-main`'s history (the post-release
   back-merge never ran — it is still 2 behind `main`), so the staging channel can still
   compute `1.83.0-staging.N`, which a `1.83.1` floor would exceed and fail loudly. A
   `1.83.0` floor passes on both channels with no back-merge needed.
2. **Forcing 1.83.0 installs would not settle it anyway.** The tray's daemon watchdog
   (`poll/mod.rs::run_daemon_watchdog`) kills a *healthy* daemon mid-WAL-write roughly
   every 45 s — an 800 ms socket probe plus `WATCHDOG_STRIKES = 2` means 10 s of slow
   replies condemns a daemon busy with an ETL batch — and that is still unpatched. It was
   measured on staging.7 *with* the fd fix (#676) installed, so the corruption it drives
   survives the update.

Raising the floor past 1.83.0 is the right move once the watchdog fix lands and `main` is
back-merged into `pre-main`.

## Scope and verification

- Only `tray/minimum-version` changes. No code, no behaviour change beyond the floor
  itself; the enforcement path (`update.rs:243`, `current_version < minimum`) is untouched.
- The value is a plain `X.Y.Z`, which the packager's regex requires.
- Full pre-push suite green locally: `cargo fmt` · `cargo clippy --workspace` · `cargo
  test --workspace` · UI build · UI tests · security audit.
- Flows into the release PR #677 automatically, since that PR tracks `pre-main`.
- Reverting to consent-based later means emptying or removing this file — the floor ships
  with **every** release while it has content.


Akarsh-Hegde:
Superseded by #681, which sets the floor to 1.83.1 (the version this release cuts) instead of 1.83.0.

The reasoning in this PR for stopping at 1.83.0 was over-cautious about the wrong path. `scripts/package-updater.sh:172` fails a release only when the floor's core is **strictly greater** than the version carrying it, so a 1.83.1 floor on a 1.83.1 release passes — and the stable channel derives 1.83.1 from this batch by construction (all `fix`/`test`, no `feat`). The staging concern is real but does not block: staging is `workflow_dispatch`-only, so no merge triggers it.

1.83.0 as the floor would have left the 1.83.0 installs — the ones actually reporting corruption — on the consent banner, which defeats the point.

PR #691, opened 2026-08-05
fix(telemetry): spool prune was stalling the daemon's runtime for 30-100s

Akarsh-Hegde:
## The investigation

Follow-up to #686's open question: why do the daemon's 60 s ETL ticks stretch to 89-164 s? Answered by catching a stall live: a monitor probed the daemon's socket every 2 s and ran `sample` (thread stacks) the instant a probe missed. The stalled thread, pinned at 74% CPU:

```
shipper::run_tick
  -> telemetry_spool::retention::prune_by_age
      -> std::fs::ReadDir::next -> __getdirentries64   (376 samples)
      -> std::fs::metadata      -> stat                (1,958 samples)
```

`sent/` sits at a **~264k-file steady state** (one file per flush x 7-day retention; measured 263,787), and every **30 s** shipper tick re-scanned all of it - blocking `read_dir` + per-file `stat`, inline on a tokio runtime worker. During the walks the daemon's poll timers fired 30-100 s late and its socket greeting starved: the alive-but-slow-to-greet false negative behind the status flapping (#686) and, pre-#678, the watchdog restart storms. One cause, three symptoms.

## The fix

1. **The whole filesystem sweep moves to the blocking pool** as one `retention::run_housekeeping` unit - the runtime never blocks on it again.
2. **The age prune is rationed to every 60th tick** (~30 min at the default interval; tick 0 still prunes so a long-stopped daemon catches up). A 7-day cutoff does not need twice-a-minute precision.

## Tests (TDD, proven failing first)

- `prune_runs_on_the_first_tick_then_every_60th` + `run_housekeeping_prunes_only_when_due` - E0425 before the fns existed; the latter asserts on `sent/` only because `pending/` is subject to the env-configured cap that parallel tests mutate (found as a real 1-in-N flake during development)
- `housekeeping_stays_off_the_async_runtime` - source-scan: the async shipper must not call any fs helper directly; needles assembled at runtime so the test cannot match its own source

`cargo test --workspace` green (23 suites), clippy `-D warnings` clean.

## Verification after it ships

On staging: ETL tick gaps should pin to ~60 s (was 4-10 stretched ticks/hour), and the popover/badge should hold steady even before #686's second opinion - #686 stays correct regardless (any future slow-to-greet moment must still not read as an outage).

## Follow-up deliberately not here

`sent/` holding a quarter-million files is itself worth fixing (day-sharded subdirs or batched spool files) - this PR makes the scan harmless, not the file count.

### DIFF
--- a/src/telemetry_spool/retention.rs
+++ b/src/telemetry_spool/retention.rs
@@ -8,8 +8,10 @@
 //!
 //! # Who calls this
 //! `shipper::run_tick()`, every `MERIDIAN_TELEMETRY_SHIP_INTERVAL_S` (default
-//! 30s), before the ship-target check: orphan sweep → age-based prune (both
-//! `pending/` and `sent/`) → pending-size cap → (only then) ship attempt.
+//! 30s), before the ship-target check — as ONE [`run_housekeeping`] unit on
+//! `spawn_blocking`, never inline on the runtime: orphan sweep → age-based
+//! prune (both `pending/` and `sent/`, rationed to every 60th tick — see
+//! [`prune_due`]) → pending-size cap → (only then) ship attempt.
 //!
 //! # Related
 //! - `writer.rs` — the `<signal>-<unix_micros>-<seq>.otlp` filename scheme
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde (commented):
Reviewed. This is a solid fix, and it closes a real open question: the "why do ETL ticks stretch to 89-164s" follow-up from #686 was flagged as unowned, and this nails it with a live thread-sample rather than a guess. Verified rather than taken on faith:

**Built and tested the branch directly:**
- `cargo test --lib telemetry_spool::` → 64 passed, including both new `retention` tests and the `housekeeping_stays_off_the_async_runtime` source-scan.
- `cargo clippy --all-targets -- -D warnings` → clean.
- `cargo test --lib` (full root package) → 878 passed, 0 failed.
- Reverted the fix by hand (restored the direct `prune_by_age`/`sweep_tmp_orphans` calls in `run_tick`) and re-ran `housekeeping_stays_off_the_async_runtime` — it correctly fails with "prune_by_age called directly from the async shipper - route it through retention::run_housekeeping inside spawn_blocking", confirming the TDD claim rather than just trusting it.

**Design read as correct:**
- `run_housekeeping`'s composition preserves the original per-step error propagation exactly (sequential `?`, same short-circuit-on-first-failure behavior as before) - this is a relocation, not a semantic change to error handling.
- Only gating the two `prune_by_age` calls on `prune_due(tick)` while running `sweep_tmp_orphans`/`enforce_pending_cap` every tick is the right split: those two only touch the bounded `pending/` dir, so they're cheap regardless of cadence; the `sent/` walk is the one that scales with the 264k-file steady state, so it's the one worth rationing.
- `TICK_COUNT` as an `AtomicU64` (rather than a plain counter) is the right call given `drain_once()` is a second entry point into `run_tick()` alongside the main loop - worth having even though the common case is single-threaded.
- `tick.is_multiple_of(60)` math checks out against its own test (`prune_due(0)` true, 1..59 false, 60 true, 61 false, 120 true), and moving the whole sweep onto `spawn_blocking` is exactly the right primitive for blocking `readdir`/`stat` syscalls.

**One nit, not blocking:** `shipper.rs`'s top-of-file module comment (lines 5-18) is untouched by this diff and is now stale in a way that's easy to trip over later - it still describes step 2 as running "UNCONDITIONALLY" every tick with no mention of the 60-tick rationing or the move to `spawn_blocking`. It's not exactly wrong (unconditional-relative-to-ship-target is still true), but a reader skimming just that header - the most likely thing to be read first - would get exactly the stale picture this PR fixes. `retention.rs`'s own module doc was updated correctly to describe the new cadence; this one wasn't. Worth a one-line update while it's fresh.

Nothing else stood out. Good instrumentation trail (thread sampling → root cause → targeted fix → regression tests that fail on the old behavior), and the follow-up note about the 264k-file count itself being worth addressing separately is the right scope call for this PR.

Akarsh-Hegde:
Addressed the review nit in 87aa8d2e - the module header now leads with the `run_housekeeping` + `spawn_blocking` unit and the every-60th-tick rationing, and keeps the property the old wording was guarding (retention runs independent of the ship-target check, so a Canonical install still prunes). Steps 4-8 unchanged. Tests + clippy green after the edit.

PR #693, opened 2026-08-05
feat(tray): auto-repair a damaged meridian.db at boot

Akarsh-Hegde:
## Why

The 2026-08 corruption incident left machines that 1.83.2 could not heal: the corruption *cause* was fixed, but existing damage persisted, and the repair offer depends on a chain the damage itself breaks - the `db.corrupt` notice is a row in the corrupt database, so hard-down machines (page-1 damage, daemon crash-looping on `file is not a database`) showed a dead app with no repair path at all. As of this writing 2 fleet machines are hard-down and 2 are latched, all waiting on hands-on recovery.

## What

The tray now probes `meridian.db` at the one provably-safe boot moment (key resolved, no pool, no capture engine - the exact spot the existing button-driven `repair_boot` already runs) and acts unattended:

- **Damaged** (opens, fails `quick_check`): writes the existing repair marker, stops the daemon best-effort (same division of labour as `commands::repair`), runs the existing salvage. Original preserved as `corrupt-backup-<ts>` as always.
- **Unopenable** (keyed open fails with SQLITE_NOTADB/code 26 - page-1 damage): nothing can read the file under any tool, so it is set aside as `unopenable-backup-<ts>` (sidecars follow) and a fresh database starts, reported via a loud `db.reset` warning notice naming the kept file.

Guard rails:
- No resolved key + unopenable file = **left alone**: indistinguishable from a healthy encrypted file during a keychain hiccup.
- A second unopenable verdict within 24h of a fresh start is **refused** - a machine with failing hardware accumulates one set-aside file, not an empty database per launch.
- Inconclusive probes (locks, timeouts, unknown errors) read as healthy; repair on uncertainty would stop the daemon over a transient.
- `quick_check` capped at 30s so a huge database cannot hold the tray icon hostage.

Also adds `integrity::is_unopenable_error` (code-26 family) beside the existing code-11 helpers.

## Fleet effect

Shipping this behind a `minimum-version` floor bump heals the hard-down and latched machines on their post-update relaunch with zero user action (their trays are alive - that is how they took 1.83.2). The two machines with dead updaters still need manual outreach; nothing shippable can reach them.

## Testing

- 8 new tests in `repair_boot` (probe classification against real damaged files, set-aside + sidecars, the 24h loop guard, the no-key guard) + 1 in `integrity`
- `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt` all green
- The Damaged auto path deliberately has no end-to-end test: it would `launchctl stop` the developer's real daemon

## Out of scope (follow-ups)

- Rate-limiting the `after_connect` error site (95% of the 1.83.2 central error stream) - touches the observability pipeline, better as its own reviewed change
- The release + floor bump itself, per the release process

### DIFF
--- a/tray/src-tauri/src/lib.rs
+++ b/tray/src-tauri/src/lib.rs
@@ -473,7 +473,18 @@ pub fn run() {
                     // bound to the inode that was moved aside, reading stale
                     // data forever without erroring.
                     let repair_outcome =
-                        repair_boot::run_if_requested(db_path_ref, db_key_hex.as_deref());
+                        repair_boot::run_if_requested(db_path_ref, db_key_hex.as_deref())
+                            // No repair was requested - probe the file anyway
+                            // and heal it unattended if it is damaged. This is
+                            // the path that recovers a machine whose corrupt
+                            // database prevented the repair offer from ever
+                            // being shown (see `repair_boot`'s module header).
+                            .or_else(|| {
+                                repair_boot::run_auto_if_needed(
+                                    db_path_ref,
+                                    db_key_hex.as_deref(),
+                                )
+                            });
 
                     // Prepare the meridian.db pool ONCE at startup and share it with
                     // commands via managed state (no migrations — the daemon owns the
@@ -552,17 +563,48 @@ pub fn run() {
                         if let Some(p) = capture_pool.clone() {
                             // Own the strings before the task takes them: the
                             // notice borrows its fields, and `outcome` cannot
-                            // outlive this scope.
-                            let (repaired, title, detail) = match outcome {
-                                repair_boot::Outcome::Repaired { summary } => {
-                                    (true, "Database repaired", summary)
-                                }
-                                repair_boot::Outcome::Failed { error } => {
-                                    (false, "Database repair failed", error)
-                                }
+                            // outlive this scope. `fault_cleared` decides both
+                            // halves below: whether the db.corrupt banner is
+                            // dropped, and which notice replaces it. A fresh
+                            // start clears the fault too - the damaged file is
+                            // gone from the canonical path - but reports under
+                            // its own id/severity so data loss is never
+                            // dressed up as a clean repair.
+                            let (fault_cleared, id, severity, title, detail, remedy) = match outcome
+                            {
+                                repair_boot::Outcome::Repaired { summary } => (
+                                    true,
+                                    "db.repaired",
+                                    "info",
+                                    "Database repaired",
+                                    summary,
+                                    None,
+                                ),
+                                repair_boot::Outcome::Failed { error } => (
+                                    false,
+                                    "db.corrupt",
+                                    "error",
+                                    "Database repair failed",
+                                    error,
+                                    Some(
+                                        "Your data is unchanged. Contact support with your Support ID.",
+                                    ),
+                                ),
+                                repair_boot::Outcome::FreshStart { backup } => (
+                                    true,
+                                    "db.reset",
+                                    "warning",
+                                    "Database was reset",
+                                    format!(
+                                        "Your database could not be opened or repaired, so tracking restarted with a new one. The old file was kept at {backup}."
+                                    ),
+                                    Some(
+                                        "Contact support with your Support ID if you want to attempt recovery of the old file.",
+                                    ),
+                                ),
                             };
                             tauri::async_runtime::spawn(async move {
-                                if repaired {
+                                if fault_cleared {
                                     // The fault is gone; drop the banner that
                                     // sent the user here in the first place.
                                     //
@@ -585,15 +627,15 @@ pub fn run() {
                                 let _ = meridian::notices::raise_typed(
                                     &p,
                                     meridian::notices::Notice {
-                                        id: if repaired { "db.repaired" } else { "db.corrupt" },
-                                        severity: if repaired { "info" } else { "error" },
+                                        id,
+                                        severity,
                                         title,
                                         detail: &detail,
-                                        remedy: (!repaired).then_some(
-                                            "Your data is unchanged. Contact support with your Support ID.",
-                                        ),
-                                        event_key: if repaired { "db.repaired" } else { "db.corrupt" },
-                                        deep_link: (!repaired).then_some("/logs"),
+                                        remedy,
+                                        // event_key mirrors id so each outcome's
+                                        // OS toast dedupes independently.
+                                        event_key: id,
+                                        deep_link: (!fault_cleared).then_some("/logs"),
                                     },
                                 )
                                 .await;
--- a/tray/src-tauri/src/lib.rs
+++ b/tray/src-tauri/src/lib.rs
@@ -604,41 +604,75 @@ pub fn run() {
                                 ),
                             };
                             tauri::async_runtime::spawn(async move {
-                                if fault_cleared {
-                                    // The fault is gone; drop the banner that
-                                    // sent the user here in the first place.
-                                    //
-                                    // `clear_typed`, NOT the plain `clear` (which
-                                    // hardcodes `event_key = "system.fault"`): the
-                                    // daemon raised this with `event_key:
-                                    // "db.corrupt"` (see `DB_CORRUPT_NOTICE` in
-                                    // `src/main.rs`), and `notices::raise_typed`
-                                    // dedupes its OS toast on `<event_key>:<id>`.
-                                    // Clearing with the wrong event_key deletes the
-                                    // banner row (that DELETE matches by id alone)
-                                    // but retracts the wrong toast dedup key, so a
-                                    // future corruption would silently never toast
-                                    // again — deduped against a delivery that was
-                                    // never actually retracted.
-                                    let _ =
-                                        meridian::notices::clear_typed(&p, "db.corrupt", "db.corrupt")
-                                            .await;
+                                // RETRIED, not fire-once: this notice can outrun
+                                // its own database. After a fresh start the
+                                // canonical path is empty until the daemon
+                                // relaunches and migrates - which on a machine
+                                // leaving the watchdog's "giving up" window can
+                                // be most of an hour - so a single write through
+                                // the lazy pool fails silently, and the user who
+                                // just lost their database is exactly the user
+                                // who must see this message. The loop is idle
+                                // and bounded; it ends on the first success.
+                                // (A repair-failed notice against a still-
+                                // unopenable file can never land - the retries
+                                // then just run out, and the tracing::error!
+                                // at the failure site remains the record.)
+                                let deadline = std::time::Instant::now()
+                                    + std::time::Duration::from_secs(2 * 60 * 60);
+                                loop {
+                                    let mut ok = true;
+                                    if fault_cleared {
+                                        // The fault is gone; drop the banner that
+                                        // sent the user here in the first place.
+                                        //
+                                        // `clear_typed`, NOT the plain `clear` (which
+                                        // hardcodes `event_key = "system.fault"`): the
+                                        // daemon raised this with `event_key:
+                                        // "db.corrupt"` (see `DB_CORRUPT_NOTICE` in
+                                        // `src/main.rs`), and `notices::raise_typed`
+                                        // dedupes its OS toast on `<event_key>:<id>`.
+                                        // Clearing with the wrong event_key deletes the
+                                        // banner row (that DELETE matches by id alone)
+                                        // but retracts the wrong toast dedup key, so a
+                                        // future corruption would silently never toast
+                                        // again — deduped against a delivery that was
+                                        // never actually retracted.
+                                        ok &= meridian::notices::clear_typed(
+                                            &p,
+                                            "db.corrupt",
+                                            "db.corrupt",
+                                        )
+                                        .await
+                                        .is_ok();
+                                    }
+                                    ok &= meridian::notices::raise_typed(
+                                        &p,
+                                        meridian::notices::Notice {
+                                            id,
+                                            severity,
+                                            title,
+                                            detail: &detail,
+                                            remedy,
+                                            // event_key mirrors id so each outcome's
+                                            // OS toast dedupes independently.
+                                            event_key: id,
+                                            deep_link: (!fault_cleared).then_some("/logs"),
+                                        },
+                                    )
+                                    .await
+                                    .is_ok();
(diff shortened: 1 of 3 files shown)


tray/src-tauri/src/lib.rs
```diff
@@ -585,15 +627,15 @@ pub fn run() {
                                 let _ = meridian::notices::raise_typed(
```
Akarsh-Hegde: **[1 - should fix] The FreshStart and guard-refusal notices are one-shot writes against a database that cannot receive them yet - the "loud db.reset warning" is likely lost on exactly the machines this PR exists to heal.**

Sequence for FreshStart: the file was just moved aside, so the canonical path is empty until the daemon relaunches and runs migrations. This spawned task fires immediately with a single `let _ = raise_typed(...)` through the lazy pool (10s acquire timeout). On the current hard-down fleet machines the daemon is in the watchdog's "too many start attempts this hour - giving up" window, so the fresh database may not exist for up to an hour - the write fails once, silently, and the user who just lost their database sees nothing.

The guard-refusal case (`Outcome::Failed` from the 24h breaker) is strictly worse: the database at the canonical path is the *unopenable* one, so this write can never succeed no matter the timing.

`Outcome::Repaired` is fine - the rebuilt DB is healthy and present.

Suggestion: persist the outcome (managed state is enough; a small file if it must survive a tray restart) and retry the raise from the poll loop until it succeeds - the poll loop already knows when the DB becomes reachable. Cheap alternative that covers most of it: retry with backoff for a few minutes instead of one attempt.

tray/src-tauri/src/repair_boot.rs
```diff
@@ -43,29 +64,143 @@ use std::path::Path;
...
+            // re-checks the marker at startup). The marker is what actually
+            // keeps it out - same division of labour as `commands::repair`.
+            if let Err(e) =
+                tauri::async_runtime::block_on(crate::commands::daemon_control::set_running(false))
+            {
+                tracing::warn!(error = %e, "could not stop the daemon before the automatic repair - the marker will hold it off instead");
+            }
+            Some(execute_pending_repair(db_path, key_hex))
+        }
+        Probe::Unopenable { error } => {
```
Akarsh-Hegde, line 187: **[2 - should fix] The Unopenable path renames the file with no marker and no daemon stop - asymmetric with the Damaged path, and unsafe if anything can still open the file.**

The Damaged path writes the marker first and best-effort stops the daemon, with a comment explaining a KeepAlive relaunch must find the marker and stand down. This path renames immediately with neither.

For a genuinely page-1-dead file the daemon is crash-looping and holds nothing, so it mostly works out. But the probe's verdict is only as good as the tray's key: a tray-side key desync (keychain holds a wrong key while the daemon's `.env` mirror holds the right one - e.g. a restored `~/.meridian` or a keychain restored out of step) makes a *healthy, live* database read as Unopenable. Rename then leaves the daemon writing into the backup inode until its next restart: the "preserved with every damaged byte intact" contract is broken and new writes are silently lost.

Two cheap hardenings, ideally both:
- Probe the daemon first (`daemon_control::status()` exists and is nearly free). If the daemon is up and healthy, someone can open this file - that is a key problem, not corruption; log loudly and leave it alone, same spirit as the no-key guard.
- Mirror the Damaged path's ordering around the rename: marker down, `set_running(false)`, quiesce, rename, marker cleared - so nothing can be mid-write when the file moves.

tray/src-tauri/src/repair_boot.rs
```diff
@@ -43,29 +64,143 @@ use std::path::Path;
 /// repair either proceeds or gives up.
 const DAEMON_QUIESCE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(20);
 
+/// Ceiling on the startup probe's `quick_check`. The check reads every page,
+/// so a very large database could otherwise hold the whole `setup` closure -
+/// and with it the tray icon - hostage. On timeout the probe reports healthy
+/// (fail open): a database too large to check in this budget is being written
+/// by a live daemon that would have latched `db.corrupt` itself if it were
+/// damaged.
+const PROBE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
```
Akarsh-Hegde, line 73: **[3 - consider] Every launch now pays an open + full-page `quick_check` before the tray icon appears.**

`run_if_requested` was deliberately "one stat" on the hot path; `run_auto_if_needed` is a full scan of every page on every boot - seconds on a multi-GB database (the pre-repair 2.7GB specimen would sit in the several-second range), 30s worst case by design, all inside the sync `setup` closure.

An alternative that keeps healthy boots at ~zero cost: gate the probe on a failed daemon socket probe. The hard-down profile this PR targets is precisely "daemon cannot start", so probe-the-daemon-first (one socket roundtrip) and only probe the file when the daemon is unreachable. The latched profile (daemon up, `db.corrupt` raised) still heals through the existing banner + button, which works there because the DB is readable.

If the every-boot scan is a deliberate choice, worth stating the measured cost in the module header - future readers will assume boot got slower for a reason.

tray/src-tauri/src/repair_boot.rs
```diff
@@ -95,6 +230,119 @@ pub fn run_if_requested(db_path: &Path, key_hex: Option<&str>) -> Option<Outcome
...
+/// corruption signatures trigger action.
+async fn probe_database(db_path: &Path, key_hex: Option<&str>) -> Probe {
+    if !db_path.exists() {
+        return Probe::Healthy; // fresh install - the daemon will create it
+    }
+    let uri = format!("sqlite://{}", db_path.display());
+    let pool = match meridian_core::db_crypto::open_pool_with_key_readonly(&uri, key_hex).await {
+        Ok(pool) => pool,
+        Err(e) => {
+            return if meridian::db::integrity::is_unopenable_error(&e) {
```
Akarsh-Hegde: **[4 - minor] This arm routes the whole corrupt family to Unopenable, skipping a repair that could work.**

`is_unopenable_error` deliberately includes `is_corrupt_error` (code 11), and the query-error arms below order plain-corrupt before unopenable for exactly that reason - code 11 means the file ATTACHes and is repairable. This open-error arm has no such ordering, so a code-11 open failure would classify Unopenable and go straight to set-aside + fresh start instead of a salvage attempt.

In practice the open path runs no page reads (`PRAGMA key` is deferred, corruption surfaces at `quick_check`), so this is latent rather than live - but it is a one-line `is_corrupt_error` check to make the two sites consistent, and the `Probe::Unopenable` doc comment ("code 26 / code 11") currently bakes the conflation in.

tray/src-tauri/src/repair_boot.rs
```diff
@@ -95,6 +230,119 @@ pub fn run_if_requested(db_path: &Path, key_hex: Option<&str>) -> Option<Outcome
...
+            backup.display()
+        )
+    })?;
+    // Stale sidecars must follow the main file: a leftover -wal beside a fresh
+    // database would be offered to SQLite as that database's log.
+    for suffix in ["-wal", "-shm"] {
+        let sidecar = PathBuf::from(format!("{}{}", db_path.display(), suffix));
+        if sidecar.exists() {
+            let kept = PathBuf::from(format!("{}{}", backup.display(), suffix));
+            let _ = std::fs::rename(&sidecar, &kept);
```
Akarsh-Hegde: **[6 - minor] The sidecar rename failure is swallowed directly under the comment explaining why a leftover `-wal` is dangerous.**

If this rename fails, the exact hazard the comment describes comes true: a stale (possibly old-key-encrypted) `-wal` sits beside the fresh database SQLite will create at the canonical path. Since the main file already moved, the WAL's salvage value is tied to the backup anyway - on rename failure, fall back to `remove_file`, and `tracing::warn!` whichever fallback ran.

tray/src-tauri/src/repair_boot.rs
```diff
@@ -95,6 +230,119 @@ pub fn run_if_requested(db_path: &Path, key_hex: Option<&str>) -> Option<Outcome
...
+        }
+    }
+    Ok(backup)
+}
+
+/// The most recent `unopenable-backup` set aside within [`FRESH_START_GUARD`],
+/// if any - the loop breaker that keeps a machine with failing hardware from
+/// minting a new empty database every launch.
+fn recent_unopenable_backup(db_path: &Path) -> Option<PathBuf> {
```
Akarsh-Hegde, line 398: **[5 - nit] Doc says "The most recent" backup; `find_map` returns the first match in directory order.**

Behaviour is still correct for the guard (any backup inside the window blocks), but the doc overpromises - either say "any backup set aside within the window" or actually take the max mtime.

Akarsh-Hegde (commented):
Reviewed against the live fleet state (central OO, 2026-08-05): 2 hard-down machines crash-looping on code 26, 2 latched, plus the watchdog's give-up window active on the hard-down ones. The design is the right one - probe the file directly because the notice chain is severed by the damage itself - and the guard rails (no-key guard, 24h loop breaker, fail-open on inconclusive probes) are exactly the ones this needs. Tests genuinely exercise real damaged files rather than mocks.

Six findings, ordered by severity. **1 and 2 are worth fixing before this ships behind a floor bump**; the rest are judgment calls or nits.

1. **The loud notice this PR promises is likely lost on the exact machines it heals** (lib.rs:627) - FreshStart/Failed outcomes write once through a pool whose database doesn't exist yet or can't be opened, and the error is swallowed.
2. **The Unopenable path renames without the marker/stop the Damaged path uses** (repair_boot.rs:156) - a daemon that CAN open the file (key desync) would keep writing into the backup.
3. Every boot now pays a full `quick_check` before the tray icon (repair_boot.rs:73) - consider gating on a failed daemon probe.
4. The open-error arm classifies code-11 as Unopenable (repair_boot.rs:251) - latent misroute, one-line fix.
5. `recent_unopenable_backup` doc says "most recent", code returns first match (repair_boot.rs:326) - doc nit.
6. Sidecar rename failure is silently swallowed right under the comment explaining why it's dangerous (repair_boot.rs:317).

Not findings, verified explicitly: marker-first ordering matches `commands::repair`; `launchctl stop` + KeepAlive means the daemon comes back on its own after the marker clears; the no-key guard correctly covers the encrypted-file-without-key case because `resolve_or_create_key` refuses to mint a replacement key when the entry is lost; user-facing strings are plain-hyphen clean; file headers intact.

tray/src-tauri/src/lib.rs
Akarsh-Hegde: Fixed in 7d0bbb65. The spawned task now retries the clear+raise pair every 30s for up to 2h - enough to cover the watchdog's give-up window plus the daemon's relaunch and migration. It ends on first success; on final failure it logs an ERROR naming the notice id. You're right that the guard-refusal notice against a still-unopenable file can never land - the retries run out and the tracing::error! at the failure site remains the record for that case.

tray/src-tauri/src/repair_boot.rs
Akarsh-Hegde: Fixed in 7d0bbb65, with the first of your two hardenings doing double duty: run_auto_if_needed is now gated on the daemon being unreachable (one daemon_already_running probe before anything else). A live daemon = someone can open the file = key desync, not corruption - the probe never runs, the file is never touched, and the stranded-writes scenario is structurally impossible. The set-aside also now writes the repair marker before the rename and clears it after, mirroring the Damaged path's handshake against a KeepAlive relaunch mid-rename. I did not add set_running there - the gate has just proven there is nothing to stop, and calling it would be a no-op with a misleading comment.

tray/src-tauri/src/repair_boot.rs
Akarsh-Hegde: Fixed in 7d0bbb65 - took your alternative wholesale, and it turned out to also be the safety fix for comment 2. Healthy boots now pay one socket probe and no file scan; the quick_check only runs when no daemon answers, which is precisely the hard-down profile. The latched profile heals through the existing banner+button as you noted. The trade-off (a boot that races the daemon's own start pays one wasted scan of a healthy file, bounded by PROBE_TIMEOUT) is documented on run_auto_if_needed.

tray/src-tauri/src/repair_boot.rs
Akarsh-Hegde: Fixed in 7d0bbb65 - the open-error site now orders is_corrupt_error before is_unopenable_error, matching the query-error arms, with a comment noting it is latent today (lazy open) but must stay correct if that changes. The Probe::Unopenable doc now says code 26 only and points code 11 at Damaged.

tray/src-tauri/src/repair_boot.rs
Akarsh-Hegde: Fixed in 7d0bbb65 - on rename failure the sidecar is deleted instead (its salvage value is tied to the backup once the main file has moved), with a warn! naming which fallback ran, and a second warn! if even the delete fails.

tray/src-tauri/src/repair_boot.rs
Akarsh-Hegde: Fixed in 7d0bbb65 - doc now says any backup inside the window blocks, first hit in directory order. Kept find_map since the guard only needs existence, not recency.

Akarsh-Hegde:
All six review findings addressed in 7d0bbb65 - inline replies on each thread.

The structural change worth calling out: findings 2 and 3 collapsed into one mechanism. `run_auto_if_needed` is now **gated on the daemon being unreachable**, which simultaneously (a) keeps healthy boots at one socket probe with no file scan, and (b) makes the dangerous rename structurally impossible - a live daemon proves the file opens, so an Unopenable verdict under the tray's key can only mean a tray-side key problem, and the file is left alone. The set-aside additionally runs under the repair marker now, closing the KeepAlive-relaunch race.

Finding 1 (lost notices): the outcome notice retries every 30s for up to 2h, covering the watchdog give-up window plus daemon relaunch + migration; gives up with a logged ERROR.

Findings 4/5/6: open-error arms ordered corrupt-before-unopenable to match the query arms, sidecar rename failure falls back to delete-with-warn, docs corrected.

Gates: `cargo test --workspace` (all 23 suites), `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt` - all green. The 8 repair_boot tests now exercise `probe_and_heal` (the body behind the gate) so they cannot be perturbed by a real daemon running on the dev machine.

PR #695, opened 2026-08-07
fix(coding-agent): apply the SQLCipher key in the one-shot CLI pools

Akarsh-Hegde:
## What

`open_meridian_pool` and the SessionEnd hook built their pools with a raw
`SqlitePool::connect_with`, which **never applies the SQLCipher key**. On a packaged
(encrypted) install every query failed with SQLite code 26, `file is not a database`.

Found while diagnosing a real machine: the daemon was healthy but 330 coding-agent
sessions were backed up with no way to drain them.

## Two casualties, both silent

| CLI | Effect |
|---|---|
| `meridian coding-agent-summarise` | Died on its first read, reporting only `ensure column: check summary_source column`. This is the **only** documented way to drain the historical backlog - `drain()` deliberately sweeps yesterday+today and leaves older days to an operator. |
| `meridian coding-agent-hook` | Failed to open the DB and sealed nothing. It exits 0 on every path by design, so nothing surfaced - sessions fell through to the indexer's 1 h idle backstop instead of sealing at SessionEnd. |

## Fix

Both route through `meridian_core::db_crypto::open_pool_with_key`, the same opener the
daemon uses. Its `key_unless_plaintext` guard drops the key for a plaintext file, so
dev/source installs are unchanged. `create_if_missing` stays false - the daemon owns
creation and migrations.

- Split the env read from the open (`open_meridian_pool_at`) so key handling is testable
  without mutating process-global env vars.
- `hook.rs` carried a duplicate private copy of `meridian_db_path` - removed.
- CLI error paths now print anyhow's full context chain (`{e:#}`). Plain `{e}` renders only
  the outermost context, which is exactly what reduced this to an unactionable one-liner
  and threw away the real SQLite error.

## Tests

Both written first and **proven red**, per the repo's TDD-for-bug-fixes convention:

- `opens_an_encrypted_db` - seeds an encrypted DB and reads it back through the CLI opener.
  Failed pre-fix with `SqliteError { code: 26, message: "file is not a database" }`, which
  is the error the `eprintln!` had been swallowing.
- `one_shot_clis_do_not_hand_build_a_pool` - source-scan guard covering the regression
  **class** in both one-shot CLIs (a unit test cannot reach the hook: it reads stdin and the
  real `MERIDIAN_DB`). Verified red against a planted regression, and scoped to the
  non-test portion of each file so it cannot match its own assertion strings.

`cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, and
`cargo test --workspace` (1518 passed, 0 failed) all green.

## Note

This does not by itself drain the existing backlog - it makes the command that drains it
work. Two follow-ups from the same investigation are coming as separate PRs: `meridian
doctor` misreporting on packaged installs, and backoff for the `missing_provider` worklog
warning.

### DIFF
--- a/src/coding_agent_session_ingest/hook.rs
+++ b/src/coding_agent_session_ingest/hook.rs
@@ -64,7 +64,16 @@ pub async fn run_hook() {
     let pool = match super::open_meridian_pool().await {
         Ok(p) => p,
         Err(e) => {
-            tracing::error!(error = %e, "coding-agent-hook: failed to open db");
+            // `{e:#}` (the full anyhow context chain), not `%e` — plain Display
+            // renders only the outermost context ("failed to open SQLite at
+            // …"), dropping the cause that actually identifies the fault
+            // (`code: 26, file is not a database`). This hook has no terminal
+            // and exits 0 regardless, so the shipped record is the ONLY evidence
+            // it ever ran; stripping the cause from it defeats the purpose.
+            // `error` is on both SAFE_STRING_KEYS and FREE_TEXT_KEYS, so the
+            // chain is path/URL/token-scrubbed on the ship leg.
+            let detail = format!("{e:#}");
+            tracing::error!(error = %detail, "coding-agent-hook: failed to open db");
             return;
         }
     };
--- a/src/coding_agent_session_ingest/hook.rs
+++ b/src/coding_agent_session_ingest/hook.rs
@@ -12,12 +12,9 @@
 
 use std::io::Read;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 
 use chrono::Utc;
 use serde_json::Value;
-use sqlx::sqlite::SqliteConnectOptions;
-use sqlx::SqlitePool;
 
 use super::indexer::register_session;
 
@@ -58,23 +55,16 @@ pub async fn run_hook() {
         }
     };
 
-    let db_path = meridian_db_path();
-    let uri = format!("sqlite://{}", db_path.display());
-    let opts = match SqliteConnectOptions::from_str(&uri) {
-        Ok(o) => o.create_if_missing(false),
-        Err(e) => {
-            tracing::error!(uri, error = %e, "coding-agent-hook: bad db uri");
-            return;
-        }
-    };
-    let pool = match SqlitePool::connect_with(opts).await {
+    // The shared keyed opener, NOT a hand-built pool. This used to be a raw
+    // `SqlitePool::connect_with`, which never applies the SQLCipher key, so on a
+    // packaged (encrypted) install every query failed with "file is not a
+    // database". The hook exits 0 on every path by design (it must never block
+    // Claude), so that failure was completely silent: SessionEnd sealed nothing
+    // and sessions fell through to the indexer's 1 h idle backstop instead.
+    let pool = match super::open_meridian_pool().await {
         Ok(p) => p,
         Err(e) => {
-            tracing::error!(
-                db_path = %db_path.display(),
-                error = %e,
-                "coding-agent-hook: failed to open db"
-            );
+            tracing::error!(error = %e, "coding-agent-hook: failed to open db");
             return;
         }
     };
@@ -131,12 +121,6 @@ fn within_accepted_roots(resolved: &Path) -> bool {
             .map(|r| resolved.starts_with(r))
             .unwrap_or(false)
     })
-}
-
-fn meridian_db_path() -> PathBuf {
-    let raw =
-        std::env::var("MERIDIAN_DB").unwrap_or_else(|_| "~/.meridian/meridian.db".to_string());
-    expand(&raw)
 }
 
 fn expand(p: &str) -> PathBuf {
--- a/src/coding_agent_session_ingest/mod.rs
+++ b/src/coding_agent_session_ingest/mod.rs
@@ -18,7 +18,6 @@ pub mod sources;
 pub mod summariser;
 
 use std::path::PathBuf;
-use std::str::FromStr;
 
 pub use segment::{
     iso_utc, norm_iso, parse_iso, parse_session_segments, Segment, SegmentParams, SessionMeta,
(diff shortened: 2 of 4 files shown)


Akarsh-Hegde (commented):
## Review — the fix is right; one test-guard hole to close before merge

I traced the whole path and the core change holds up:

- `open_pool_with_key` is the same opener `db::meridian::setup_db` uses, and `key_unless_plaintext` genuinely no-ops the key against a plaintext file, so the "dev/source unchanged" claim is sound.
- The key actually reaches these subcommands: `main.rs` step 1/1a loads the repo `.env` *and* the canonical `~/.meridian/.env` **before** the `coding-agent-hook` / `coding-agent-summarise` dispatch at L72/L83, so `MERIDIAN_DB_KEY` is populated even when Claude Code invokes the hook from an arbitrary project cwd. That was the thing most likely to make this fix a no-op in the field, and it doesn't.
- `libsqlite3-sys` carries `bundled-sqlcipher-vendored-openssl` unconditionally in both `Cargo.toml`s, so `opens_an_encrypted_db` is a real test and not one that trivially passes against a plaintext file.
- No orphaned imports, tilde expansion preserved via `mod.rs::meridian_db_path`, `create_if_missing(false)` semantics unchanged.
- Bonus that's worth stating: `meridian doctor --fix` shells out to `coding-agent-summarise` (`src/health/fix.rs:72`), so this unblocks that path too.

Below, one item I'd fix before merge (test quality, not correctness) and a few smaller ones.

---

### 1. The source-scan guard misses the most likely regression — in *both* halves

`one_shot_clis_do_not_hand_build_a_pool` has two parts and the same regression slips past each.

**(a) The delegation assertion is self-satisfying.** The loop carefully slices at `#[cfg(test)]` to avoid matching its own strings, then the final assert throws that away:

```rust
assert!(
    include_str!("mod.rs").contains("db_crypto::open_pool_with_key("),
```

That's a fresh whole-file read, and the test module immediately above calls `meridian_core::db_crypto::open_pool_with_key(&uri, Some(TEST_KEY), true, &[])` in its seed step. The substring is present no matter what production code does — the assert passes even if `open_meridian_pool_at` stops delegating entirely.

**(b) The ban list doesn't cover the shape a regression would actually take.** It matches `SqlitePool::connect_with`, `SqlitePool::connect(` and `connect_lazy_with`, but not `SqlitePoolOptions::new()…​.connect_with(opts)` — which is precisely the pattern someone lands on by copying the daemon's own pool construction (`db_crypto.rs`, `db/meridian.rs:563`) and forgetting the key.

So for the single most probable future regression, neither half of the guard fires. Both fixes are small: reuse the sliced `src` for the delegation assert, and ban bare `connect_with(` / `connect_lazy_with(` / `SqlitePool::connect(` — neither file legitimately calls any of them post-fix, so a bare-substring ban is safe.

(Trivial, same function: `src.split("#[cfg(test)]").next().unwrap_or(src)` — `next()` on a `Split` is always `Some`, so the `unwrap_or` is dead.)

### 2. The outermost error path is still silent — on exactly the install this bug lives on

`main.rs:105`:

```rust
Err(e) => tracing::error!(error = %e, "coding-agent-summarise: failed to open db"),
```

Both `fmt` layers in `observability::init` are `cfg!(debug_assertions).then(...)` (`observability/mod.rs:165`, `:173`), and the spool is the sole persisted sink. In a **release/packaged build** — the only place the unkeyed pool ever failed — that line prints nothing at all, and `main` then returns `Ok(())`. An operator running the documented backlog-drain command against a bad/missing key sees an empty terminal and exit 0.

That's the same silent-failure class the PR is fixing, one frame further out, and it's the frame an operator actually stands in. Worth an `eprintln!("summarise: open db: {e:#}")` next to the tracing call (and arguably a non-zero exit).

### 3. `%e` throws away the chain the PR just went to the trouble of recovering

`hook.rs`'s new `tracing::error!(error = %e, …)` and `main.rs:105` both use `%e` — anyhow's `Display`, which renders only the outermost `.context(...)`. Here that's `failed to open SQLite at sqlite://…`, so the path survives but the real cause (`code: 26, file is not a database`) is dropped from the shipped record too. Same defect the `{e:#}` change fixes for the `eprintln!`s.

Worth doing at these two sites specifically (not a codebase-wide restyle — `redact.rs` explicitly documents `error = %e` as the convention at 203 sites). `%format_args!("{e:#}")` is awkward with temporary lifetimes, so:

```rust
let detail = format!("{e:#}");
tracing::error!(error = %detail, "coding-agent-hook: failed to open db");
```

No egress concern: `error` is on both `SAFE_STRING_KEYS` and `FREE_TEXT_KEYS`, so the chain gets home-dir-path-scrubbed, URL/email/token-scrubbed and clamped on the ship leg.

### 4. Missing test for the half the PR's own compatibility claim rests on

`opens_an_encrypted_db` covers the encrypted path. The claim that dev/source installs are unchanged rests entirely on `key_unless_plaintext` — and nothing here exercises it. One more `#[tokio::test]`: seed a *plaintext* DB, open through `open_meridian_pool_at` with `Some(TEST_KEY)`, assert the read succeeds. That's the exact real-world configuration (dev machine, key in env from a leftover `.env`, file not yet encrypted) that would regress silently.

Minor: the test passes a bare path while `open_meridian_pool` builds `sqlite://{path}`. `sqlite_uri_to_path` handles both and is separately tested, so this is fine — but passing the `sqlite://` form would exercise the production string shape.

### 5. Note, not an objection: this restores a third writer to `meridian.db`

On encrypted (packaged) installs the hook has been failing to open and therefore writing nothing. After this it does an `UPSERT` per SessionEnd, against a DB with an open, unexplained corruption investigation currently pointing at concurrent writers around WAL checkpoints. Restoring intended behaviour isn't a defect, and the keyed opener actually improves it — the hand-built pool had no `busy_timeout`, this one gets `busy_timeout=5000` from `BASE_PRAGMAS`. Flagging it only so the change in writer topology is on the record if corruption reports shift after this ships.

### 6. Same bug class, out of scope

`src/bin/cat_smoke.rs:47` opens `~/.meridian/meridian.db` with a raw read-only `SqlitePool::connect_with` and no key — broken on any encrypted install. Different severity (dev-only binary, and it has `.context("failed to open meridian.db")` plus a non-zero exit, so it fails loudly rather than silently). Not for this PR; worth a note wherever the follow-ups are tracked.

---

**Verdict:** approve on the fix itself. #1 is the one I'd want closed before merge — as written the guard would let the regression back in, which is the whole reason it exists. #2–#4 are cheap and in the same spirit as the change.

Akarsh-Hegde:
Write exhaustive unit tests so that there is not regression

Akarsh-Hegde:

All four are fixed in `8542989d`. CI green (13/13), `MERGEABLE`.

Also answers the "write exhaustive unit tests so there is no regression" comment: the two guard holes below were the reason the existing tests weren't that, and every test in this PR is now verified red against a planted regression rather than just observed green.

### 1. The source-scan guard missed the regression in both halves - **confirmed empirically before fixing**

I planted a live `SqlitePoolOptions::new()...connect_with(opts)` pool in `hook.rs` and ran the guard as written:

```
test one_shot_clis_do_not_hand_build_a_pool ... ok
```

It passed with exactly the bug it exists to catch sitting in the file. Both halves fixed:

- **Delegation assert** was a fresh whole-file `include_str!`, and `opens_an_encrypted_db` calls `db_crypto::open_pool_with_key` to seed its fixture - so the substring was present regardless of production code. Now scans the production half, like the ban loop already did. Verified red by pointing `open_meridian_pool_at` at `open_pool_with_key_lazy`.
- **Ban list** now covers bare `connect_with(` / `connect_lazy_with(` / `SqlitePool::connect(`. Re-ran the same plant:
  ```
  test one_shot_clis_do_not_hand_build_a_pool ... FAILED
  Offending lines: ["        .connect_with(sqlx::sqlite::SqliteConnectOptions::new())"]
  ```

(The dead `unwrap_or` is gone too - extracted as a `production_half` helper used by both checks.)

### 2. Silent outermost error path

`main.rs` now `eprintln!`s and **exits non-zero**. Checked your point about `doctor --fix`: `fix.rs:210` is `cmd.status().map(|s| s.success())`, so the non-zero exit makes it report the failure instead of a false pass - strictly better, not a behaviour break.

### 3. `%e` dropping the chain

Fixed at both sites via `let detail = format!("{e:#}")`, scoped to these two only - `error = %e` stays the convention elsewhere per `redact.rs`.

### 4. Missing plaintext-compat test

Added `opens_a_plaintext_db_even_with_a_key_configured` - seeds a plaintext DB, opens it through the CLI opener with a key set, asserts a bounded successful read. That was the half of the "dev/source unchanged" claim nothing exercised. The encrypted test now also uses the production `sqlite://` string shape.

### On your other points

- **#5 (writer topology)** - agreed and on the record. Worth adding that the keyed opener *improves* it: the hand-built pool had no `busy_timeout`, this one inherits `busy_timeout=5000` from `BASE_PRAGMAS`.
- **#6 (`cat_smoke.rs`)** - left out as you suggested. Noted as a follow-up.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test --workspace` (1519 passed, 0 failed) all green.

Not merging - that's yours.

PR #696, opened 2026-08-07
fix(doctor): stop reporting false faults on packaged installs

Akarsh-Hegde:
## What

`health::platform::repo_root()` walks up from `current_exe()` looking for a `Cargo.toml`.
On a DMG install the binary is `~/.meridian/bin/meridian` with no repo above it, so it is
permanently `None` - and **five checks folded that into "broken"**.

Measured on a healthy 1.84.0-staging.2 machine, `meridian doctor` printed
**`✗ critical issues`** plus a **fabricated root-cause diagnosis**. That is worse than no
check: a permanent false CRITICAL trains users to ignore the report, and the two genuine
faults on that machine sit in the same output.

One root cause, five symptoms - hence one PR.

## The five

| Check | Was | Now |
|---|---|---|
| **daemon binary** | `CRITICAL not installed` - probed only the two `install.sh` symlinks (`meridian-daemon`), never the tray's staged `~/.meridian/bin/meridian` | probes all three |
| **config (.env)** | checked only `<repo>/.env` | either that **or** the canonical `~/.meridian/.env`, mirroring what `main.rs` actually loads (step 1 walk + step 1a unconditional) |
| **settings file** | warned that `~/.meridian/settings.json` "is not read by the daemon" | checks for a real `MERIDIAN_SETTINGS_PATH` split-brain |
| **mcp built** | `not built`, remedy naming a directory the user does not have | `Info: not applicable` when there is no repo |
| **openobserve** | `export disabled - telemetry not collected` | distinguishes packaged (ships to the central gateway) from dev-without-credentials |

### On `settings file` - the premise was obsolete, not mis-implemented

`settings_json_path()` resolves the canonical `~/.meridian/settings.json` **first**; a
repo/cwd copy is only a fallback when the canonical is absent. So the layout the check
flagged as broken **is the correct one**, and the summary escalated it to *"UI settings
aren't reaching the daemon"* on machines where they demonstrably were. (Doctor even
contradicted itself, printing `✓ poll interval - from settings.json` two lines below.)

The one real split-brain left is `MERIDIAN_SETTINGS_PATH` shadowing the dashboard's file,
so that is what it now checks - with `diagnose.rs` and the `--fix` hint updated to match.

### Why `daemon binary` is not gated on install mode

"Is THIS PROCESS packaged" (`is_canonical_install`) is a different question from "is a
daemon installed on this system". `cargo run -- doctor` on a machine that also has the DMG
installed must still find the staged binary - which is the one launchd is actually running.
Probing all candidates is correct in both directions.

`openobserve` is the only site that genuinely needs install mode, which is why
`install_mode::is_canonical_install` widens from `pub(super)` to `pub`.

## Verification

Run against the **live packaged install**, both with and without a repo above the binary
(confirmed safe first: the only DB open in `health/` is
`db_crypto::open_pool_with_key_readonly`, so doctor cannot migrate the DB it inspects):

```
before:  ✓ 30 ok  · 5 info  ⊘ 6 warn  ✗ 2 fail   + 2 diagnoses (one fabricated)
after:   ✓ 34 ok  · 5 info  ⊘ 3 warn  ✗ 1 fail   + 1 diagnosis
```

The remaining fail is a genuine, unrelated `capture_coverage` finding.

## Tests

- `daemon_binary_candidates_cover_both_install_types` - pins all three paths, including the
  platform executable suffix (the staged name must track `backend_install::DAEMON_FILE`).
- The settings decision is split into a pure `settings_verdict`, so all three branches are
  covered without touching `$HOME` or the process-global `MERIDIAN_SETTINGS_PATH`.

`cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test --workspace`
(1520 passed, 0 failed) all green.

## Out of scope

The `meridian logs` duplicate-line question and the `worklog-status --day` reporting
discrepancy came out of the same investigation but are separate findings, not part of this.

### DIFF
--- a/src/health/contracts.rs
+++ b/src/health/contracts.rs
@@ -63,17 +63,24 @@ fn settings_contract() -> Check {
     let resolved = meridian_core::settings::settings_json_path();
     let ui_exists = ui_settings.is_file();
 
-    // Two spellings of the same file (a symlinked home, a doubled separator, a
-    // `/./`) are not a split-brain — warning there would tell the user to unset
-    // a variable that is causing no harm. Only meaningful when both resolve on
-    // disk; `canonicalize` fails on a path that does not exist, in which case
-    // fall through to the textual comparison below.
-    if let (Ok(a), Ok(b)) = (resolved.canonicalize(), ui_settings.canonicalize()) {
-        if a == b {
-            return Check::ok("settings file", "config", "daemon + dashboard agree");
-        }
+    if same_resolved_file(&resolved, &ui_settings) {
+        return Check::ok("settings file", "config", "daemon + dashboard agree");
     }
     settings_verdict(&resolved, &ui_settings, ui_exists)
+}
+
+/// Whether two paths name the same file on disk.
+///
+/// Two spellings of the same file (a symlinked home, a doubled separator, a
+/// `/./`) are not a split-brain — warning there would tell the user to unset a
+/// variable that is causing no harm. Only answerable when both resolve on disk;
+/// `canonicalize` fails on a path that does not exist, and `false` then lets the
+/// caller fall through to the textual comparison.
+fn same_resolved_file(a: &std::path::Path, b: &std::path::Path) -> bool {
+    match (a.canonicalize(), b.canonicalize()) {
+        (Ok(a), Ok(b)) => a == b,
+        _ => false,
+    }
 }
 
 /// The decision half of [`settings_contract`], split out so every branch is
@@ -131,7 +138,7 @@ fn dead_poll_env() -> Check {
 
 #[cfg(test)]
 mod tests {
-    use super::{expand, settings_verdict};
+    use super::{expand, same_resolved_file, settings_verdict};
     use std::path::Path;
 
     /// The canonical, correct configuration: the daemon resolves exactly the
@@ -172,6 +179,63 @@ mod tests {
             false,
         );
         assert_eq!(c.severity, crate::health::Severity::Info, "{c:?}");
+    }
+
+    /// Two spellings of the SAME file must not read as a split-brain.
+    ///
+    /// `settings_verdict` compares paths textually, so a symlinked home (common
+    /// on macOS with `/tmp` → `/private/tmp`, and on any setup that symlinks the
+    /// home directory) would warn falsely and tell the user to unset a variable
+    /// that is causing no harm. The wrapper canonicalizes when both paths exist;
+    /// this exercises that with a real symlink, which is the only way to reach it.
+    #[cfg(unix)]
+    #[test]
+    fn two_spellings_of_the_same_settings_file_are_not_a_split_brain() {
+        use crate::health::Severity;
+
+        let dir = tempfile::tempdir().unwrap();
+        let real = dir.path().join("settings.json");
+        std::fs::write(&real, "{}").unwrap();
+
+        let link_dir = dir.path().join("linked");
+        std::os::unix::fs::symlink(dir.path(), &link_dir).unwrap();
+        let via_link = link_dir.join("settings.json");
+
+        // Textually different, same inode.
+        assert_ne!(real, via_link);
+        assert_eq!(
+            real.canonicalize().unwrap(),
+            via_link.canonicalize().unwrap()
+        );
+
+        // The raw verdict cannot tell — that is exactly why the wrapper
+        // canonicalizes first.
+        assert_eq!(
+            settings_verdict(&real, &via_link, true).severity,
+            Severity::Warn,
+            "precondition: the textual comparison should disagree here"
+        );
+
+        // The canonicalizing path must agree.
+        assert!(
+            same_resolved_file(&real, &via_link),
+            "a symlinked spelling of the same file was treated as a split-brain"
+        );
+    }
+
+    /// ...and genuinely different files must still be distinguishable, so the
+    /// canonicalize step cannot be "simplified" into always-equal.
+    #[test]
+    fn genuinely_different_settings_files_are_not_collapsed() {
+        let dir = tempfile::tempdir().unwrap();
+        let a = dir.path().join("a.json");
+        let b = dir.path().join("b.json");
+        std::fs::write(&a, "{}").unwrap();
+        std::fs::write(&b, "{}").unwrap();
+        assert!(
+            !same_resolved_file(&a, &b),
+            "two distinct files were reported as the same"
+        );
     }
 
     #[test]
--- a/src/health/contracts.rs
+++ b/src/health/contracts.rs
@@ -2,11 +2,11 @@
 //
 // Cross-process config contracts: values multiple processes must agree on, where
 // a silent mismatch masquerades as an AI/runtime fault. Verified against the
-// code: the UI reads MERIDIAN_DB_PATH (not MERIDIAN_DB), and the daemon reads
-// <repo>/settings.json (not the UI's ~/.meridian/settings.json).
+// code: the UI reads MERIDIAN_DB_PATH (not MERIDIAN_DB), and the daemon and the
+// dashboard must resolve the same settings.json (canonically
+// ~/.meridian/settings.json, overridable via MERIDIAN_SETTINGS_PATH).
 
 use crate::config::Config;
-use crate::health::platform::repo_root;
 use crate::health::Check;
 use std::path::PathBuf;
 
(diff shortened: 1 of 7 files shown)


Akarsh-Hegde (commented):
## Review — right diagnosis, right scope; two places the PR commits its own sin inverted

The premise checks out end to end. I verified the parts a wrong answer would have hidden:

- `settings_json_path()` really does resolve the canonical `~/.meridian/settings.json` **first** (`meridian-core/src/settings.rs`), with the cwd copy only as a fallback — so the old check's premise was genuinely obsolete, not just noisy, and `MERIDIAN_SETTINGS_PATH` really is the only remaining split-brain.
- `diagnose.rs`'s `bad` closure is `severity >= Severity::Warn`, and `Info < Warn`, so the new `Info` branch correctly stops firing the "UI settings aren't reaching the daemon" diagnosis rather than merely downgrading its glyph. Worth stating explicitly since that fabricated diagnosis is half the reported damage.
- `is_otlp_configured()` already early-returns `false` for canonical installs, so the old message really was structurally wrong there, not just pessimistic.
- The `Check::info` move has precedent right there in `platform.rs` — `service_manifest_check`'s `Unknown` arm, with the comment "a health check that asserts a problem it has not actually looked for trains the user to ignore it." This PR is that comment applied consistently, which is the right call.

Grouping five symptoms of one root cause into one PR is correct here; splitting them would have made each look like a cosmetic downgrade.

Two findings below are the same shape, and it's worth naming it: **in two places this PR reproduces the exact failure mode it exists to fix, inverted.**

---

### 1. `openobserve` now asserts a good state it never checked

The new packaged arm returns a fixed string — "errors ship to the central gateway" — without looking at whether they do. `resolve_otlp_target()` → `resolve_central_target()` returns `None` when `!settings.error_reporting_enabled` (`otlp_target.rs:198`), and the endpoint/token are release-injected `option_env!` constants, so a source-built binary is inert by construction.

Two real machines get told something false:

- A user who **opted out** of error reporting (it's opt-out, default on — so this is a deliberate choice someone made) is told their errors are shipping.
- A **locally-built binary staged into `~/.meridian/bin`** — which `is_canonical_install()` classifies as canonical, since it compares `current_exe()` against exactly that path — reports central shipping from a build where `option_env!` is unset and nothing can ever be sent.

That is `service_manifest_check`'s comment inverted: a check asserting a *good* state it has not looked for erodes trust the same way asserting a bad one does, and it's harder to catch because nobody investigates a green line.

The function that knows the answer is already there, and the fix is a two-branch match rather than a new probe:

```rust
if crate::observability::is_canonical_install() {
    return vec![match crate::observability::resolve_otlp_target() {
        Some(_) => Check::info("openobserve", "obs",
            "packaged install — captured locally; errors ship to the central gateway"),
        None => Check::info("openobserve", "obs",
            "packaged install — captured locally; error reporting is off (or this build has no baked endpoint)"),
    }];
}
```

Cost is fine: `resolve_otlp_target()` loads settings via `load_runtime_settings()`, which is one `read_to_string` + `serde_json::from_str` — and the arm this replaces was already going to call `is_otlp_configured()`, which loads the same file.

### 2. The `DAEMON_FILE` mirror now has a third copy, guarded by a test that can't fail

`src/observability/install_mode.rs` already carries `staged_daemon_file_name()` — the sanctioned single mirror of the tray's `backend_install::DAEMON_FILE`, complete with a doc comment about the exact drift it's there to prevent: it was hardcoded `"meridian"`, so on Windows `is_canonical_install()` never matched and **central error reporting was silently inert on every Windows install**.

`daemon_binary_candidates()` hand-rolls that same path again from `home()` + `format!("meridian{}", EXE_SUFFIX)`. And the new test asserts against a copy it recomputes with the identical expression:

```rust
let staged = home().join(".meridian/bin").join(format!("meridian{}", std::env::consts::EXE_SUFFIX));
assert!(candidates.contains(&staged), ...);
```

Both sides derive from the same expression, so if the production copy drifts from `DAEMON_FILE` the test drifts with it and stays green — which is precisely the failure this file already suffered once, at a cost the doc comment spells out. The test's own docstring says the path "must track `backend_install::DAEMON_FILE`", and that is the one thing it does not check.

The PR is already widening `install_mode::is_canonical_install` to `pub` for exactly this kind of reuse. Widen `staged_daemon_file_name()` alongside it (or better, expose a `staged_daemon_path()` and have both call sites use it) — then `daemon_binary_candidates()` and `is_canonical_install()` agree by construction rather than by two people editing two files, and the test becomes a real assertion. Right now they don't even share a root: candidates build from `paths::home_dir_or_cwd()`, `is_canonical_install()` from `paths::meridian_dir()`.

Same severity class as the bug this PR exists to kill — a drifted path here means a false CRITICAL. I'd fix it with #1.

### 3. `mcp built` infers install type from a repo probe

The PR's own rationale for *not* gating `daemon binary` on install mode is exactly right: "is THIS PROCESS packaged" ≠ "is a daemon installed here." Two functions later, `mcp_service` does the conflation it just rejected — `repo_root() == None` becomes "not applicable on a packaged install," when what was actually observed is "no `Cargo.toml` above `current_exe()`." That's also true of a `cargo install`ed binary, a release binary copied into `~/bin`, or a `.app`-embedded one.

The downgrade to `Info` is right; the wording overclaims. Either report what was observed ("no source checkout found — the MCP server is built from one") or, since `is_canonical_install()` is now `pub`, actually ask.

### 4. `daemon binary` reports the first candidate found, not the one launchd runs

Staged-first ordering is the correct default — keep it. But `.find(|p| is_exec(p))` names a path, and on a mixed machine that path may not be the binary launchd is configured to execute. This repo has a documented failure mode from exactly that skew (DMG + local builds contending for the staged path, producing a "migration N missing" crash-loop), and it isn't hypothetical — on my machine:

```
/Users/…/.local/bin/meridian-daemon -> /Users/…/meridian/.claude/worktrees/dashboard-style-cleanup-phase2/target/debug/meridian
```

a stale debug build in a worktree for a long-finished branch. After this change doctor reports the staged binary and that symlink becomes invisible; before it, the reverse. Neither is wrong, but neither answers "which binary is launchd running." Doctor already reads the plist for `service_manifest_check` — reporting the `ProgramArguments` path beside the found candidate, and flagging a mismatch, would turn this check into the one that actually catches the crash-loop. Follow-up, not this PR.

(The same two-symlink blind spot exists in `scripts/meridian-cli.sh::_daemon_bin`, but that wrapper only exists in source checkouts where those symlinks also exist — no victim, mentioned only so it doesn't look overlooked.)

### 5. Two small ones

- **`settings_verdict` compares paths textually.** `MERIDIAN_SETTINGS_PATH` set to the same file by a different spelling (symlinked home, a doubled separator, `/./`) warns falsely and tells the user to unset a variable that isn't causing a problem. Cheap hardening: when both paths exist, fall back to comparing `canonicalize()`.
- **The `(_, true)` `.env` arm hides the repo copy.** When both files exist it reports only `~/.meridian/.env`, but step 1 loads the repo copy and step 1a's `from_path` does *not* override — so for overlapping keys the repo file is the one that wins. A dev chasing a stale key is told the wrong file is in play. Report both when both exist.

---

**Verdict: approve.** #1 and #2 are worth fixing before merge — both are small, and both are the PR's own thesis pointed the other way. #3–#5 are fine as follow-ups. The before/after numbers and the "one root cause, five symptoms" framing hold up against the code.

Akarsh-Hegde:

All five are fixed in `29eef603`. CI green (13/13), `MERGEABLE`.

Your framing was right and worth keeping: **two of these were this PR committing its own sin inverted.** Both are fixed, not deferred.

### 1. `openobserve` asserted a good state it never checked

Fixed as suggested - the packaged arm now asks `resolve_otlp_target()` and reports the two cases separately. Your cost analysis held: the arm it replaces already called `is_otlp_configured()`, which loads the same `settings.json`, so this is free.

The reasoning is now in the code, because it's the non-obvious part: a green line nobody investigates is worse than a red one, and the two machines it would have lied to (opted-out user, locally-built binary staged into `~/.meridian/bin`) are both real.

### 2. The `DAEMON_FILE` mirror - and a test that couldn't fail

This was the sharper finding. Fixed in three parts:

- **`install_mode::staged_daemon_path()`** is now the single source. `is_canonical_install()` and `daemon_binary_candidates()` both use it, so they agree by construction rather than by two people editing two files. They previously didn't even share a root (`meridian_dir()` vs `home_dir_or_cwd()`) - you were right to call that out.
- **A test that actually pins the mirror.** You were correct that asserting against a recomputed copy is worthless. But so is asserting against the shared value - it just moves the tautology. So the new test reads the *tray's* source across the crate boundary and asserts on `DAEMON_FILE` itself. Verified red against a planted tray-side rename:
  ```
  test staged_daemon_file_name_matches_the_trays_daemon_file_constant ... FAILED
  the tray's DAEMON_FILE values changed to ["meridian.exe", "meridian-daemon"];
  staged_daemon_file_name() must be updated to match
  ```
  It checks both cfg arms, so it fails on whichever platform CI runs.
- The health test now asserts against the shared value **and** pins staged-first ordering.

### 3. `mcp built` inferred install type from a repo probe

Fair - that's the conflation this PR explicitly rejects for `daemon binary`, two functions later. Now reports what was observed: `no source checkout found - the MCP server is built from one`.

### 4-5. The two small ones

- `settings_verdict` compares `canonicalize()` when both paths exist, so a symlinked home or a doubled separator no longer warns falsely.
- The `.env` check reports **both** files when both exist, and says the repo copy wins for overlapping keys - since step 1a's `from_path` doesn't override. Your point: a dev chasing a stale key was being sent to the wrong file.

### Re-verified on the live packaged install

Both with and without a repo above the binary, since I changed how the staged path is derived:

```
daemon binary   ✓  /Users/…/.meridian/bin/meridian
config (.env)   ✓  ~/.meridian/.env present
settings file   ✓  daemon + dashboard agree
mcp built       ·  no source checkout found — the MCP server is built from one
```

### Left as follow-ups, per your note

**#4 (which binary launchd actually runs)** - agreed this is the one that would catch the crash-loop class, and your `.local/bin/meridian-daemon` → stale worktree debug build is a good example of why. Out of scope here; it needs the plist's `ProgramArguments` read and a mismatch check.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test --workspace` (1521 passed, 0 failed) all green.

Not merging - that's yours.

Akarsh-Hegde:
## Regression tests added - still ready to merge

`c87e4412`. CI green (13/13), `MERGEABLE`, merged up to date with `pre-main`.

**Heads-up: this touches production code, on a PR I'd already called ready.** `daemon_binary_candidates` was the only check here reachable by a test - the other four read the filesystem, `current_exe()` or the network inline, so four of the five behaviours this PR fixes had no coverage at all. Closing that needed seams, not just tests. I've re-verified the live install afterwards (below).

### Seams extracted

Mirroring the `settings_verdict` shape this PR already established - thin wrapper resolves the real inputs, pure fn decides:

- `env_verdict(repo_env, home_env)`
- `mcp_verdict(built: Option<bool>)` - `None` = no source checkout
- `offline_verdict(canonical, central_resolves, dev_configured)` - `Some` short-circuits, `None` means "go probe the dev endpoint"
- `same_resolved_file` - split out of `settings_contract` so the canonicalize step is directly testable

`Diagnosis` gains `#[derive(Debug)]` so assertion messages can print it.

### Tests - every one verified RED against a planted regression

| Test | Planted regression it catches |
|---|---|
| `env_verdict_covers_every_combination` | packaged case reverting to `warn` |
| `mcp_verdict_is_info_without_a_checkout_and_never_claims_an_install_type` | no-checkout going back to a warn with an unreachable remedy |
| `a_packaged_install_reports_shipping_only_when_a_target_resolves` | asserting shipping unconditionally |
| `the_dev_path_is_reported_separately_and_never_claims_telemetry_is_dark` | (same seam; also pins that a configured dev install still falls through to the live probe) |
| `an_info_settings_check_raises_no_diagnosis` + `a_warning_settings_check_still_raises_the_diagnosis` | loosening `bad` from `severity >= Warn` to `!= Ok` |
| `two_spellings_of_the_same_settings_file_are_not_a_split_brain` (real symlink) + `genuinely_different_settings_files_are_not_collapsed` | collapsing `canonicalize` to never-equal |

Two details worth stating:

- The `.env` test asserts not just that both files are named when both exist, but that the **precedence is stated** - step 1a's `from_path` does not override, so the repo copy wins, and a message naming only the canonical path sends a dev to the wrong file.
- The diagnose pair covers **both directions** of your "half the reported damage" point: an Info check must raise nothing, and a genuine split-brain must still raise the diagnosis. The second is what stops the fix being a silent deletion.

### One trap worth recording

Two plants initially reported no failures and proved **nothing**: `-D warnings` killed the build on a now-unused parameter before any test could run, so "0 failing tests" looked like the test passing. The `openobserve` one especially looked fine at a glance. Both were redone keeping the symbol referenced. Same trap as the guard-test holes earlier in this PR - a check that cannot execute is indistinguishable from a check that succeeds.

### Live re-verification after the refactor

```
daemon binary   ✓  /Users/…/.meridian/bin/meridian
config (.env)   ✓  ~/.meridian/.env present
settings file   ✓  daemon + dashboard agree
openobserve     ✓  reachable
```

Identical to before the seam extraction.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test --workspace` (1532 passed, 0 failed).

Not merging - that's yours.

PR #697, opened 2026-08-07
fix(worklog): throttle the unconfigured-provider warning

Akarsh-Hegde:
## What

`missing_provider` deliberately leaves an approved worklog in place when its provider is
not configured on this daemon - correct, so it posts if that tracker is later connected.
But **nothing ever clears the row**, so the 60 s approved-sweep re-encounters it forever
and warned on every single pass.

Measured on a real install: one worklog drafted against **Linear** on a Jira-only daemon
(`pm_providers = ["jira"]`) was the **only** warning the daemon emitted - **224 times in
the last 3000 log lines**, 1440 a day, indefinitely.

## Why this is more than noise

The local telemetry spool is the daemon's **sole** log sink, and **WARN+ is exactly the
severity that egresses** to the central error pipeline. One benign, permanently-stuck row
crowds out real errors on both - locally in `meridian logs`, and remotely in the central
gateway that error reporting depends on.

## Fix

Warn once per row, then at most every 6 h. Suppressed repeats drop to DEBUG, so the
condition stays visible in a local sweep without dominating it. The span still records
`waiting` on every pass, so per-sweep traces are unchanged.

The ledger is in-memory per process, matching the summariser's `MAX_ROW_ATTEMPTS` ledger -
a restart re-warns once, which is the right behaviour: a fresh log should state the
condition rather than inherit a silence.

Placed in a new `warn_throttle` module rather than added to `post.rs`, which is already
past the repo's 500-line cap. `post.rs` shrinks 692 → 663.

## Tests

Five cases over an **injected clock**, so no test sleeps:

- the first occurrence always warns (throttling is not silencing)
- every sweep inside the window is suppressed (359 simulated sweeps)
- the warning returns at the boundary
- rows throttle independently - one noisy row must not mask another's first report
- resurfacing resets the window (guards the off-by-one in which timestamp gets stored)

`cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test --workspace`
(1521 passed, 0 failed) all green.

## Note

This silences the symptom, not the cause: MER-192 is still an approved worklog that will
never post until Linear is connected. Surfacing that to the user (rather than only in logs)
is a separate UX question, deliberately not bundled here.

Third and last of the three fixes from the same machine investigation - after #695
(SQLCipher key in the one-shot CLI pools) and #696 (doctor's false faults on packaged
installs). The three are independent and can merge in any order.

### DIFF
--- a/src/observability/filter.rs
+++ b/src/observability/filter.rs
@@ -198,6 +198,46 @@ mod tests {
         );
     }
 
+    /// A throttled site's "quieter repeat" must be INFO, not DEBUG.
+    ///
+    /// This pins a coupling nothing else expresses, and getting it wrong is
+    /// invisible at the call site. `pm_worklog::post` throttles a permanent
+    /// warning and emits the suppressed repeats at a lower level so the
+    /// condition stays readable in `meridian logs` without egressing (only
+    /// WARN+ ships). That was first written as `debug!` — and at the production
+    /// default this filter grants debug ONLY to `etl`, `intelligence` and
+    /// `embedder`, so `meridian::pm_worklog::post` fell under `meridian=info`
+    /// and the repeats were dropped **before the spool**. Not quieter: gone.
+    ///
+    /// So: INFO from that target must survive the default filter, and DEBUG
+    /// must not. If someone adds a debug directive covering `pm_worklog` this
+    /// fails and they can drop the `debug!`-is-invisible workaround; if someone
+    /// removes `meridian=info` it fails too.
+    #[test]
+    fn a_throttled_repeat_at_info_survives_the_default_filter_but_debug_does_not() {
+        const TARGET: &str = "meridian::pm_worklog::post";
+        let filter = build_default_filter("INFO");
+
+        let passed = targets_passing(&filter, || {
+            tracing::info!(target: "meridian::pm_worklog::post", "throttled repeat");
+        });
+        assert!(
+            passed.iter().any(|t| t == TARGET),
+            "INFO from {TARGET} was dropped by the default filter, so a throttled \
+             repeat would be invisible even locally: {passed:?}"
+        );
+
+        let passed = targets_passing(&filter, || {
+            tracing::debug!(target: "meridian::pm_worklog::post", "throttled repeat");
+        });
+        assert!(
+            !passed.iter().any(|t| t == TARGET),
+            "DEBUG from {TARGET} now passes the default filter. That is fine, but \
+             `post.rs` emits throttled repeats at INFO specifically BECAUSE debug \
+             was being dropped here — revisit that comment: {passed:?}"
+        );
+    }
+
     /// EVERY entry in `CAPTURE_TARGETS` must be covered at EVERY log level.
     ///
     /// Deliberately iterates the constant rather than naming crates: an earlier
--- a/src/pm_worklog/mod.rs
+++ b/src/pm_worklog/mod.rs
@@ -27,6 +27,7 @@ pub mod post;
 pub mod post_comment;
 pub mod status;
 pub mod trello;
+mod warn_throttle;
 
 pub use config::PmWorklogConfig;
 pub use generate::{approve, generate, ApproveResult};
--- a/src/pm_worklog/post.rs
+++ b/src/pm_worklog/post.rs
@@ -27,6 +27,7 @@
 
 use anyhow::{Context, Result};
 use sqlx::SqlitePool;
+use std::time::Duration;
 use tokio::sync::watch;
 
 use crate::config::{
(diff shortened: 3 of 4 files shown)


Akarsh-Hegde (commented):
## Review — correct fix, honest tests; one stated rationale doesn't hold, and two siblings left unthrottled

The diagnosis is right and the measurement makes the case: a row that is deliberately never cleared, met by a 60 s sweep, is a permanent warn generator, and WARN+ is precisely what egresses to the central pipeline. The helper itself is good work — the injected clock is the right seam, and the five cases cover the boundary, independence, and the reset-off-by-one rather than just the happy path. Splitting it out of a 692-line `post.rs` instead of growing it further is also right.

Findings below; none block.

---

### 1. The DEBUG downgrade is invisible at the default filter, so the stated rationale doesn't hold

The PR body, the commit message, and the doc comment on `missing_provider` all say the same thing: suppressed repeats "drop to DEBUG, so the condition stays visible in a local `meridian logs` sweep without dominating it."

It isn't visible. `RuntimeSettings::default().log_level` is `"INFO"` (`meridian-core/src/settings.rs:337`), and `build_default_filter`'s INFO arm is:

```
meridian=info,meridian::etl=debug,meridian::intelligence=debug,meridian::embedder=debug,sqlx=warn
```

`meridian::pm_worklog::post` is not among the debug-level overrides, so it falls under `meridian=info` and the new `tracing::debug!` is dropped by the `EnvFilter` **before the spool** — the filter decides what is captured at all, ahead of severity and redaction. At default config the suppressed repeats aren't quieter, they're gone; the line only exists for someone who has already set `log_level=DEBUG`, i.e. someone who didn't need the breadcrumb.

Cheapest fix that makes the sentence true: emit the suppressed repeats at `info!`. INFO passes the default filter (so it *is* in a local `meridian logs` sweep, which is the whole stated point) and INFO does **not** egress (only WARN+ does), so it costs the central pipeline nothing. That's exactly the property the PR is arguing for. Alternatively keep `debug!` and drop the claim from all three places — but then the throttle is a plain silencer between warns, which is worth saying out loud rather than implying otherwise.

### 2. Two sibling sites in the same file, same sweep, same permanent shape — still unthrottled

`run_post_loop` (line 572-590) calls three things on the same 60 s tick, and two of the others have the identical structure this PR just fixed:

- **`unpost_stale`, line 531** — `warn!("unpost skipped — unknown provider, leaving stale entry in place")` followed by `continue`. An unknown provider string never becomes known and the pending marker is never cleared, so that row warns every 60 s for the life of the process. This is the cleanest match: it's per-row (`p.id`), so it drops straight into the existing `WarnThrottle` map. The function's own doc comment at line 501 already groups it with this PR's branch — "same forgiving posture as `missing_provider` below."
- **`process_approved_proposals`, line 408** — `warn!("approved proposals waiting but no PM provider configured — skipping")`, then an early return that leaves the queue intact. Permanent until the user connects a tracker. Not per-row, so it needs a separate static or a sentinel key rather than the row map.

The helper is `pub(super)` and already in the module, so this is a few lines each. Not calling it a defect — the PR fixed the site it measured — but "one root cause, several symptoms" was the framing on #696, and the same argument applies here: fixing one of three instances of a shape the surrounding comments already name as a family invites the next one to be rediscovered from logs. Worth either applying it or noting why these two are different.

### 3. Coverage stops at the module boundary

All five tests exercise `WarnThrottle` in isolation, and they do it honestly — no complaint about their quality. But the bug lived in `post.rs`, in the unconditional `warn!`, and nothing pins that `missing_provider` consults the throttle at all: delete the `if should_warn` branch and restore the original unconditional warn, and all five stay green.

This is a "worth adding," not a "fix before merge" — unlike a guard that claims to cover something and doesn't, these tests simply don't reach that far. A capture-subscriber assertion around `missing_provider` (warn once, then not) would close it, and would also have caught #1 above, since it would show what actually survives the filter.

### 4. Question, not a defect: is WARN-forever the intended steady state?

The 360× reduction already achieves the goal — the residual isn't crowding anything out. But a permanently-stuck row still emits 4 WARNs/day to the central error channel indefinitely for a condition that is user configuration, not a fault. If the answer to #1 is `info!` for repeats, then the first-occurrence WARN reaches central once and the condition stays local thereafter, which seems like the shape you actually want.

### 5. One-liner

`.unwrap_or(true)` on the poisoned lock is permanent: once poisoned, every subsequent sweep reverts to the exact unthrottled warn this PR removes, silently. `.unwrap_or_else(|e| e.into_inner())` keeps the throttle working and costs one word. The panic is essentially unreachable in a `HashMap` insert, so take it or leave it — but "losing the throttle is far better than losing the signal" trades away the fix permanently rather than for one pass.

---

**Verdict: approve.** #1 is the one I'd change before merge — it's a one-word fix and the claim is currently in the PR body, the commit message, and a doc comment. #2 is a few lines while the context is loaded. #3–#5 are follow-ups.

Akarsh-Hegde:

All five are fixed in `a83f7ddf`. CI green (13/13), `MERGEABLE`.

### 1. The stated rationale didn't hold - you were right, and I'd asserted it in three places

Confirmed exactly as you described. `build_default_filter`'s INFO arm is `meridian=info` with debug overrides only for `etl`, `intelligence` and `embedder`; `meridian::pm_worklog::post` isn't among them, so the `debug!` was dropped by the `EnvFilter` **before the spool**. At default config the suppressed repeats weren't quieter - they were gone, and the claim in the PR body, the commit message and the doc comment was false.

Took your suggested fix: repeats now emit at **INFO**. It passes the default filter (so it genuinely is in a local `meridian logs` sweep) and INFO doesn't egress, so the central pipeline costs nothing.

**That also answers #4.** With repeats at INFO, a permanently-stuck row reaches the central error channel **once per 6 h at most**, not 4x/day forever - and after the first warn the condition stays local. That is the shape you described, so I've taken it rather than leaving the question open.

### 2. The two siblings

Both throttled - the "one root cause, several symptoms" argument applies here as it did on #696.

- **`unpost_stale`'s unknown-provider arm** - same permanent shape exactly as you said.
- **`process_approved_proposals`'s no-tracker branch** - daemon-wide, so it uses a sentinel key.

One deliberate deviation: `unpost_stale` got its **own** map rather than reusing the worklog one. The ids come from unrelated tables, so a shared map would let one site silence the other purely because it warned first - a collision that would look exactly like the throttle working.

### 3. Coverage stopping at the module boundary

Correct - all five `WarnThrottle` tests passed with `missing_provider`'s throttle branch deleted. Added a capture-layer test that drives the real function and asserts `WARN, INFO, INFO, INFO, INFO`. Verified red against a planted bypass:

```
test missing_provider_warns_once_then_throttles_to_info ... FAILED
expected one WARN then INFO repeats;
got [Level(Warn), Level(Warn), Level(Warn), Level(Warn), Level(Warn)]
```

As you predicted, it also pins the **level**, so finding #1 can't regress silently.

### 5. The poisoned-lock one-liner

Taken - `.unwrap_or_else(|e| e.into_inner())`. Your reasoning is the deciding point and is now in the code: poisoning is permanent, so `.unwrap_or(true)` traded the fix away *forever* rather than for one pass, which is the opposite of the intent I'd written next to it.

### Also

The interval const is now `PERMANENT_CONDITION_WARN_INTERVAL`, since three sites share it.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test --workspace` (1522 passed, 0 failed) all green.

Not merging - that's yours.

Akarsh-Hegde:
## Regression tests added - still ready to merge

`19716e38`. CI green (13/13), `MERGEABLE`, merged up to date with `pre-main`.

The five `WarnThrottle` tests only exercised the helper in isolation - the bug lived in `post.rs`. Six new tests close that, and **every one was verified RED against a planted regression** before being committed.

| Test | Planted regression it catches |
|---|---|
| `unpost_stale_throttles_the_unknown_provider_warning` | bypass that site's throttle → 4 WARNs |
| `process_approved_proposals_throttles_the_no_tracker_warning` | bypass that site's throttle → 4 WARNs |
| `the_per_site_throttles_do_not_share_a_map` | point `unpost_stale` at `MISSING_PROVIDER_WARNED` → `[Info]` instead of `[Warn]` |
| `a_poisoned_throttle_still_throttles` | restore `.unwrap_or(true)` → 3 WARNs |
| `a_throttled_repeat_at_info_survives_the_default_filter_but_debug_does_not` | add `meridian::pm_worklog=debug` to the default filter |

Both sibling sites are driven through the **real functions** against an in-memory pool with a seeded row (the pattern already used in `ledger.rs`/`generate.rs`/`escalate.rs`), not through the helper.

### The two worth calling out

**Cross-map independence** pins the one place this PR deliberately departs from your review. A `pm_worklogs` id and a pending-unpost id can collide numerically, so a single shared map would let one site silence the other purely because it warned first - and that bug is invisible, because a missing warning looks exactly like the throttle working correctly.

**The filter coupling** is the one that would have caught finding #1 in the first place. Nothing connected the level chosen in `post.rs` to the directives in `observability/filter.rs`, so `debug!` looked fine at the call site while being dropped before the spool. The test asserts INFO survives the default filter and DEBUG does not, and its failure message says what to do if someone changes that deliberately.

Per your note about `tracing` caching callsite interest per process, I ran it **standalone and under `cargo test --workspace`** - passes both.

### One implementation note

The capture layer is now target-scoped to `meridian::pm_worklog::post`. Tests that build an in-memory pool run `sqlx::migrate!` inside the capture window, and sqlx's own DEBUG/INFO events were landing in the assertions - which read as a throttle failure when nothing was wrong.

`cargo fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test --workspace` (1530 passed, 0 failed).

Not merging - that's yours.

PR #704, opened 2026-08-07
fix: keep the error cause when a tray command or the OTLP ship leg fails

Akarsh-Hegde:
## Why

Two 1.83.2 machines reported the dashboard's **Sync now** failing. What the UI showed was `tasks: today presence` (and on the second machine `tasks: detect pm_tasks.status_category`) - the reader's own `.context()` string and nothing else. The same truncated value was all that would have reached central OpenObserve.

`anyhow::Error`'s `Display` renders **only the outermost `.context(...)`**, and both halves of the command error boundary used it:

```rust
.map_err(|e| { tracing::warn!(error = %e, "…"); e.to_string() })
```

Every `meridian-core` reader wraps each query in `.context(...)`, so this discarded the entire cause on every command. The real fault was recovered only by luck - `get_today`'s reader happens *not* to wrap that query, so its untouched sqlx error leaked `(code: 11) database disk image is malformed` on an adjacent log line.

The ship leg had the same defect independently: `format!("send OTLP request: {e}")` on a `reqwest::Error` keeps `"error sending request for url (…)"` and drops the `source()` that says whether it was DNS, TLS, connect-refused, or timeout.

## What changed

- **`crate::cmd_err!`** (`commands.rs`) - formats with `{:#}` (full source chain) **once** and uses that single value for both the log and the string returned to the frontend, so a user quoting their on-screen error can be matched to their shipped rows. Converted all **43** sites carrying the lossy shape.
- **Source-scanning guard** - the hand-rolled shape reads as correct and fails silently (the command still errors, just without the cause), so nothing at the call site would reveal a regression. Verified by planting one back: it fails with `notices.rs:52`.
- **`tasks-sync` timeout arm** now logs `bin` / `cwd` / `timeout_s`, matching its sibling arms. `kill_on_drop` reaps the child and its stderr with it, so that one line is the *entire* record of a timeout - and a bare `"tasks-sync timed out"` cannot distinguish a slow tracker sync from a binary that never got past opening an unreadable database. `SYNC_TIMEOUT` replaces three independent `30` literals.
- **`with_causes()`** (`telemetry_spool/mod.rs`) - renders an error's `source()` chain for the ship leg.

## Egress safety

Unchanged in kind. `error` is already on **both** `SAFE_STRING_KEYS` and `FREE_TEXT_KEYS`, so a shipped chain is URL/email/token scrubbed and clamped to 2000 chars - far above anything these produce. Widening `{}` to `{:#}` adds causes, not new leak surface.

## Tests

TDD - the failing test was written first and reproduced the field string byte-for-byte (`root cause was dropped: tasks: today presence`).

- `cmd_err_keeps_the_root_cause` - the field chain, both layers present
- `plain_display_still_drops_the_root_cause` - pins *why* the macro exists
- `logged_and_returned_values_match` - on-screen string == shipped string
- `extra_fields_pass_through`
- `no_command_hand_rolls_the_lossy_error_shape` - source scan
- `with_causes_keeps_the_root_cause` / `plain_display_drops_the_root_cause`

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (23 binaries, 0 failures; tray 226 → 231) all pass.

## Scope note

This does **not** fix the underlying failure on those machines - it makes it diagnosable. See the investigation notes on the ticket: in the supplied bundle all 23 `malformed` errors come from the **tray** process and **none** from the daemon, whose startup `quick_check` raised no `db.corrupt` notice - so the file is readable by one process and not the other, which points away from b-tree damage.

Pushed with `--no-verify` at the maintainer's direction: the pre-push audit flagged pre-existing `build_export_bundle` code that this branch does not modify. Every other gate ran and passed.

### DIFF
--- a/src/telemetry_spool/mod.rs
+++ b/src/telemetry_spool/mod.rs
@@ -198,6 +198,26 @@ impl std::fmt::Display for ShipError {
     }
 }
 
+/// Render an error together with its whole `source()` chain, as
+/// `"outer: inner: root"`.
+///
+/// `reqwest::Error`'s own `Display` stops at `"error sending request for url
+/// (https://…)"` and keeps the part that actually identifies the fault — DNS
+/// failure vs. TLS rejection vs. connection refused vs. timeout — in
+/// `source()`. A field report from 1.83.2 showed the ship leg failing that way
+/// for three days straight across every tick, and because the cause had been
+/// dropped it was impossible to tell from the diagnostics bundle whether the
+/// gateway was down, the machine was offline, or TLS was being intercepted.
+///
+/// `anyhow::Error`'s alternate formatter already walks a source chain, so the
+/// conversion borrows it rather than re-implementing the walk.
+fn with_causes<E>(e: E) -> String
+where
+    E: std::error::Error + Send + Sync + 'static,
+{
+    format!("{:#}", anyhow::Error::from(e))
+}
+
 /// POST one OTLP payload to `endpoint`, classifying any failure so the caller can
 /// quarantine a permanently-rejected payload without stalling the whole queue.
 ///
@@ -219,7 +239,7 @@ pub async fn ship_one(
         .body(bytes)
         .send()
         .await
-        .map_err(|e| ShipError::Retryable(format!("send OTLP request: {e}")))?;
+        .map_err(|e| ShipError::Retryable(format!("send OTLP request: {}", with_causes(e))))?;
 
     let status = resp.status();
     if status.is_success() {
@@ -244,6 +264,56 @@ mod tests {
     // Rust runs tests in parallel threads by default, so any two tests here
     // that set them concurrently would race. Serialize this module's tests.
     static ENV_LOCK: Mutex<()> = Mutex::new(());
+
+    /// A two-level error shaped like the `reqwest` failure the ship leg hits:
+    /// a generic outer message plus the cause that actually names the fault.
+    #[derive(Debug)]
+    struct Outer(Inner);
+    #[derive(Debug)]
+    struct Inner;
+
+    impl std::fmt::Display for Outer {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(
+                f,
+                "error sending request for url (https://telemetry.example/v1/logs)"
+            )
+        }
+    }
+    impl std::fmt::Display for Inner {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            write!(f, "dns error: failed to lookup address information")
+        }
+    }
+    impl std::error::Error for Outer {
+        fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+            Some(&self.0)
+        }
+    }
+    impl std::error::Error for Inner {}
+
+    /// Regression guard for the 1.83.2 field incident: the ship leg warned
+    /// `send OTLP request: error sending request for url (…)` on every tick for
+    /// days, and the cause — which is the only part that says whether to fix
+    /// DNS, TLS, the network, or the gateway — was silently discarded.
+    #[test]
+    fn with_causes_keeps_the_root_cause() {
+        let rendered = super::with_causes(Outer(Inner));
+        assert!(
+            rendered.contains("dns error"),
+            "root cause was dropped: {rendered}"
+        );
+        assert!(
+            rendered.contains("error sending request"),
+            "outer message was dropped: {rendered}"
+        );
+    }
+
+    /// Pins why the helper exists: plain `Display` is NOT equivalent.
+    #[test]
+    fn plain_display_drops_the_root_cause() {
+        assert!(!format!("{}", Outer(Inner)).contains("dns error"));
+    }
 
     #[test]
     fn build_export_bundle_archives_pending_and_sent_otlp_files() {
--- a/tray/src-tauri/src/commands.rs
+++ b/tray/src-tauri/src/commands.rs
@@ -113,6 +113,23 @@ pub use worklogs::*;
 /// `tracing::warn!` takes them: `crate::cmd_err!(e, id = body.id, "edit_worklog
 /// failed")`.
 ///
+/// The default level is `warn`. Sites where the failure is fatal to the
+/// operation take the explicit form, which is otherwise identical:
+///
+/// ```ignore
+/// crate::cmd_err!(level: error, e, "could not write the repair marker")
+/// ```
+///
+/// # Scope — this helps `anyhow` errors ONLY
+/// `{:#}` walks a source chain because **`anyhow`'s** `Display` impl checks
+/// `f.alternate()`. `std::io::Error`, `reqwest::Error` and `JoinError` do not:
+/// for those, `format!("{:#}", e)` is byte-identical to `format!("{}", e)` and
+/// this macro adds a log line but recovers nothing. Such a site needs the
+/// error promoted first (`anyhow::Error::from(e)`, as
+/// `meridian::telemetry_spool::with_causes` does for the OTLP ship leg) —
+/// routing it through `cmd_err!` unchanged would look fixed while still
+/// dropping `source()`. `cmd_err_is_a_no_op_for_non_anyhow_errors` pins this.
+///
 /// # Why this exists
 /// `anyhow::Error`'s `Display` (`{}`, and therefore `.to_string()` and
 /// `tracing`'s `%e`) renders **only the outermost `.context(...)`** — the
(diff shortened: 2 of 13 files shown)


Akarsh-Hegde (commented):
Reviewed the full diff plus the surrounding files on the branch. The core change is right and the reasoning in the PR body checks out — I verified the egress claim independently rather than taking it on trust: `error` is on **both** `SAFE_STRING_KEYS` (redact.rs:114) and `FREE_TEXT_KEYS` (redact.rs:192), and `scrub_text` → `scrub_paths` rewrites `/Users/<name>` → `/Users/<user>`, so a widened chain that splices a home-dir path still can't leak the OS account name. `{}` → `{:#}` adds causes, not leak surface.

`logged_and_returned_values_match` is the invariant that actually matters here and it's good that it's pinned. `plain_display_still_drops_the_root_cause` as a canary that would tell you when the macro can be retired is a nice touch.

Nothing below blocks the merge. Item 1 is three lines and worth doing in this PR; 2 and 3 are follow-ups.

Note: CodeRabbit is skipped on `pre-main` (auto-review is off for non-default base branches), so this is the only review on the PR.

---

## 1. Three sites the PR converted *around* but not *in* — `dashboard.rs:418, 445, 464`

`dismiss_day_task`, `merge_day_task` and `restore_day_task` each got their mutation's `map_err` converted to `cmd_err!`, and each leaves the `get_day_tasks` refresh on the very next statement as:

```rust
meridian_core::day_tasks::get_day_tasks(pool, &body.day)
    .await
    .map_err(|e| e.to_string())
```

That second call is the one whose payload the UI renders, and it has *no* log at all plus a truncated string to the frontend — strictly worse than the shape this PR set out to remove. To be fair about it: on a damaged file the mutation above almost certainly fails first, so I'm not claiming this is what bit the two 1.83.2 machines. It's the same failure class, silently unlogged, sitting one line below a site the PR deliberately fixed. Cheap to close.

## 2. `repair.rs:59-62` and `repair.rs:91-94` invert `logged_and_returned_values_match`

```rust
.map_err(|e| {
    tracing::warn!(error = %e, "repair preview failed");   // truncated → ships to central OO
    format!("{e:#}")                                        // full chain → on screen only
})?;
```

The on-screen string carries the chain; the shipped one doesn't. That's the exact inversion of the invariant the new test pins, and it's on the **database-corruption path** — the one this PR exists to make diagnosable from a support bundle. `request_repair` at :91 has the same split, and the `set_running` warn at :99 is `%e`-only with no returned string at all, so that log *is* the whole record.

These can't just become `cmd_err!` as written: the macro hardcodes `warn!`, and :91 deliberately uses `error!`. A level parameter (`cmd_err!(level: error, e, "…")`) or a sibling `cmd_err_fatal!` would make it mechanical — otherwise these two stay hand-rolled and drift.

## 3. The guard is narrower than the class it protects — and CI proves it

Not a hypothesis: CI is green on this branch *while* the scanned directory still contains

- `cli_exec.rs:113-114` — `tracing::warn!(bin = %bin, error = %e, …)` + `format!("spawn error: {e}")`
- `integrations.rs:499-500` and `626-627` — `warn!(error = %e, …)` + `format!("could not write .env: {e}")`

Same defect, different spelling, invisible to the scan. Contributing factors, in rough order of how likely each is to matter:

- The return half must be **exactly** `e.to_string()`; any `format!("…: {e}")` wrapper passes.
- The binding must be named `e` (`|err|` passes).
- `error = %e` must sit on the **same physical line** as `tracing::warn!(`. This one is self-demonstrating: `cargo fmt` split your own new timeout arm in `tasks.rs` across five lines. Convert a site with 3+ fields, regress it later, and the scan won't see it.
- `read_dir` doesn't recurse. No subdirs under `commands/` today, but the 500-line rule pushes toward `commands/foo/mod.rs` splits (cf. `readers/today/`), at which point the scan goes partially blind without failing.
- `windows(3)` while only indices 0 and 1 are read — `windows(2)` is equivalent and also catches a match on a file's last two lines.

Suggest widening to the class rather than the string: any `map_err` closure that logs `error = %<binding>` and returns a `Display`-rendered string. Same test, matched against the joined source rather than line pairs.

There are ~22 bare `.map_err(|e| e.to_string())` sites across tray commands beyond the three in item 1 (`app_icons`, `custom_llm`, `daemon`, `diagnostics`, `llm_lab`, `pause`, `statuses:122,179`, `system`, `version`). Not this PR's job, but "all 43 sites" is accurate for one shape and not for the class — worth saying so in the commit trailer or a follow-up ticket so the next person doesn't read the guard as complete coverage.

## Nits

- `tasks.rs:37` — the `[`SYNC_TIMEOUT`]` intra-doc link points from a `pub fn`'s docs at a private const, which trips rustdoc's `private_intra_doc_links` and renders as a broken link. Won't break the build (RUSTDOCFLAGS is separate from the `-D warnings` in `.cargo/config.toml`); backtick it as plain code or make the const `pub(crate)` and it's fine.
- `SYNC_TIMEOUT` replacing the three `30` literals is the right call, and the timeout arm's new `bin`/`cwd` fields genuinely separate "slow tracker" from "binary died opening a corrupt DB". No notes.

Akarsh-Hegde:
Addressed in `f8cb1411`. Items 1, 2 and the nit are done; item 3 is done in part, and the part I did **not** do is the interesting half - the review's own examples turned out to disprove the widening.

## 1. The three refresh sites - fixed

`dismiss/merge/restore_day_task` now route the `get_day_tasks` refresh through `cmd_err!` with a distinct message each (`"<cmd>: task-list refresh failed"`), so the call whose payload the UI actually renders is no longer the unlogged one.

## 2. `repair.rs` - fixed, with one deliberate exception

`cmd_err!` gained a `level:` arm, so `:60` (warn) and `:92` (error) both convert and the inversion is gone. `level_arm_logs_and_returns_the_same_value` pins the new arm against exactly the drift you found - the existing test only covered the default arm, which is how the `error!` path drifted unnoticed in the first place.

The third site, the `set_running` warn at `:100`, I left alone on purpose: `daemon_control::set_running` returns `Result<(), String>`. There is no `source()` to walk, so `{:#}` renders identically and `cmd_err!` would add the appearance of a fix and recover nothing. Commented in place so it does not get re-flagged.

## 3. Guard - hardened, not widened

Hardened, all four of your mechanics:

- brace-matched closure body with whitespace collapsed, so it no longer cares how `cargo fmt` breaks a site across lines (this replaces the `windows(3)`/same-physical-line coupling entirely)
- binding-name agnostic - `|err|` is caught
- recurses into `commands/` subdirectories, before the 500-line rule creates the first `commands/<domain>/mod.rs`

I checked it rather than assuming: planting a reformatted offender back into `notices.rs` fails the test with `["notices.rs:51"]`, which is precisely the shape the old line-pair match walked past.

**On widening to the class - I think this one is a trap, and your three examples are the evidence.** `{:#}` walks a source chain only because *`anyhow`'s* `Display` impl reads `f.alternate()`. `std::io::Error`, `reqwest::Error` and `JoinError` do not, so for those `format!("{:#}", e)` is byte-identical to `format!("{}", e)`. Verified rather than reasoned:

```
std {}   = error sending request for url (https://example)
std {:#} = error sending request for url (https://example)   <- source() dropped
```

All three sites you cited are `std::io::Error` - `cli_exec.rs:113` from `.spawn()`, and `integrations.rs:499`/`:626` because `strip_env_keys` and `upsert_env` are both `-> std::io::Result<()>`. A scan cannot see types, so a widened guard would have forced all three through `cmd_err!` and gone green while `source()` was still being dropped - a passing test asserting coverage it does not have, which is worse than the narrow guard. They need `anyhow::Error::from(e)` first, the way the ship leg does in this PR. `cmd_err_is_a_no_op_for_non_anyhow_errors` now pins that boundary, and it is written to **fail** if `{:#}` ever starts walking a plain chain - at which point widening becomes correct and the test says so.

## Nit - fixed

`SYNC_TIMEOUT` backticked rather than made `pub(crate)`; it has one consumer and widening visibility to satisfy a doc link seemed the worse trade.

## On "all 43 sites"

You are right that it is accurate for one shape and not for the class, and the commit trailer now says so explicitly: **16** bare `.map_err(|e| e.to_string())` sites remain (`app_icons`, `custom_llm`, `daemon`, `diagnostics`, `llm_lab`, `pause`, `statuses`, `system`, `version`) - no log at all, a different defect, each needing a per-site error-type judgement of the kind above. The guard does not claim them.

## Gates

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace` all pass; tray suite 231 -> 236. Full pre-push suite green, including the security audit - **this push did not use `--no-verify`** (the earlier waiver was for the pre-existing `build_export_bundle` finding on the telemetry commit, which this one does not touch).

Ready for a merge decision - leaving the merge itself to you.

PR #706, opened 2026-08-09
feat(onboarding): first-run walkthrough, daily summary parity, and a productionisation pass

adityaharishch:
## What changed

Two things, in one branch: the **feature work** (a first-run walkthrough, the daily-summary/worklog draft rebuild, the AI provider picker redesign, personal-task escalation) and a **productionisation pass** over it - removing redundancy, closing observability and test gaps, and fixing one user-visible regression the audit turned up.

## The regression worth reviewing first

`PlanView`'s Refresh chip could never show `⚠ Sync failed` again. `syncTasks` never rejects by contract, `refreshPlan` resolves `null` rather than throwing, so the `.catch(() => setSyncError(true))` was dead code and `syncError` was pinned to `false`. Introduced by this branch - the previous code called `bridgeMutate`, which did reject.

Fixed at the contract rather than the call site: `syncTasks` now resolves an outcome while still never rejecting. A promise that cannot reject has to carry its outcome in the value, or the information is gone. A failed sync also logs now - a first import is a real round trip to Jira and previously left no trace anywhere.

## Observability

- `push_health_update` discarded its emit result. Its own comment claimed a failure was "a stale banner for one tick"; discarding the error is what made that unfalsifiable.
- `resize_setup_window` was uninstrumented, so a platform-specific resize failure surfaced only as "the wizard looks wrong on my machine".
- `day` and `endpoint_id` are now pinned by test as deliberately **not** on `SAFE_STRING_KEYS`. `endpoint_id` reads as opaque but slugifies the provider name the user typed, so an internal service name would egress verbatim.

## Design tokens

`--color-state-proposal` is a theme-independent **semantic state colour** ("a proposal exists") that had spread across the UI as a stand-in for the brand accent - 173 uses on buttons, borders and links that ignore the user's theme entirely.

Reclassified per site: fills carrying white content take `--btn-primary-bg` (which `ink` overrides back to dark violet), everything else takes `--t-accent`. **Four ternary fills would otherwise have rendered white on `#DCCFFB` in `ink`** - my first pass missed them because the property detector mis-parsed ternaries. A new test now pins both rules; one documented fixed-brand pill is deliberately left alone.

## Tests

`cargo test --workspace` 1604 passing, `bun test` 700 passing, `tsc --noEmit` clean.

New coverage where it was thinnest:
- `day_draft_states` had **zero** tests while its near-twin `stale_drafts` had nine over byte-identical staleness SQL. Now eight, covering the edge cases they share by construction.
- auto-generate's tracker-gate removal was guarded only by grepping its own source for a string. The sweep now returns its counts, so a functional test proves it walks its task list with no tracker connected - without invoking the LLM.
- guards for the `syncTasks` contract, the connect warm-start call sites, `cached_test_result`, and the token rules above.

## Tooling

- `ui typecheck` passes for the first time and now runs in CI. Getting there needed `@types/bun` (never installed, so `bun:test` did not resolve across 35 test files) plus a fixture typed looser than the payload it stands in for.
- The `lint` script is **deleted**: Next 16 removed `next lint`, so it parsed "lint" as a directory and could only fail, and there is no ESLint config or dependency here. Real ESLint is worth adding as its own change.
- `TESTING.md` gains a Windows section. Windows has shipped and has a CI leg, but the manual checklist never followed it there, and `tests/install/` is launchd-shaped with no Windows equivalent.

## Merge notes

`pre-main` (198 commits) is merged in. Two resolutions worth a look:

- **Migration collision.** Both branches added a `077`. `pre-main`'s has shipped, so this branch's two renumber (`077_worklog_stale` → `079`, `078_day_summary_standup` → `080`). Without it sqlx fails every coding-agent ingest test on a duplicate `_sqlx_migrations.version`, and a fresh DB cannot migrate at all. The rename needed one `--no-verify` commit, approved by the author: the pre-commit guard protects migrations that have been *applied*, and these only ever existed on this never-pushed branch. The CI guard diffs against the PR base, where they do not exist.
- **`sticky` vs `runtime_health`.** `pre-main`'s new `runtime_health` supersedes this branch's `note_live_call_success` + failure-TTL pair, but did not know about `sticky` - the flag marking an *asserted* verdict (the dev-only Disconnect button). Without it a live call succeeding would overturn the state that button exists to hold: the exact bug `sticky` was added for, reached by a new path. `most_recent_outcome` now honours it, with a test carrying over the coverage deleted alongside the old function.

## How this was tested

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test --workspace` - all clean.
- `bun test`, `npm run typecheck`, `npm run build` in `ui/` - all clean.
- Not verified by hand: the `⚠ Sync failed` chip needs a tracker connected and a failing sync; Windows behaviour is left to the `rust-windows` CI leg.


Akarsh-Hegde (commented):
Reviewed the full diff at max effort - 150 files, +17.6k/-2.7k, feature work and a productionisation pass in one branch. The engineering standard here is high and the PR body is unusually honest about its own trade-offs. Three findings below block, in my view; the rest are worth fixing but not gating.

I want to flag the shape of the risk before the specifics: this branch mixes a first-run walkthrough, a draft rebuild, a provider-picker redesign, a 173-site token reclassification, two renumbered migrations and a 198-commit merge. Every blocker below sits in the seams between those, not inside any one of them. That is an argument about how the branch is packaged, not about the work.

---

# Blockers

## 1. The new health gate in `complete_inner` is self-latching - one failure blacks out every AI feature

`src/llm/resolver.rs:392`

The gate refuses the call when `in_use_provider_health` reports `ok == false`. The reasoning in the comment is right - one funnel, so no future caller can forget it. The problem is what can clear the verdict once it is set.

`most_recent_outcome` applies `OBSERVATION_MAX_AGE_HOURS` (6h) to **runtime observations only**. A `ProviderTestResult` has no expiry at all. And the only thing that records a success is `record_success` at `resolver.rs:394` - two lines *below* the `return Err(refusal)`. So the gate prevents the call that would produce the evidence that clears the gate.

Two concrete states:

- **A failed runtime observation** blacks out hourly folds, day summaries, worklog drafts, plan drafts and classification for the **full 6 hours**, because no call can be attempted to clear it early.
- **A failed manual Test Connection** blacks them out **permanently**. `persist_test_result` has exactly one non-test caller (`detect.rs:551`), so nothing but the user pressing Test Connection again ever writes over it. A user who tests while on a flaky VPN silently loses all AI features until they happen to press that button a second time.

This is not hypothetical, and that is why I am calling it a blocker rather than a design question. **A machine running current `pre-main` was in precisely this state yesterday:**

```json
"custom": { "outcome": { "status": "failed",
            "message": "custom provider returned an empty answer" },
            "observed_at": "2026-08-08T15:33:44Z" }
```

Under `pre-main` that record was cosmetic - `in_use_provider_health` had one consumer, the dashboard banner, and the next hourly call ran normally and cleared it. This PR promotes that same record into a hard gate on every prose call. The identical incident, on this branch, is a six-hour silent blackout.

Worth saying: the `rate_limited` exclusion is correct and well-argued, and it is what keeps the *current* state of that machine working. The gap is only on `ok == false`.

The fix I would argue for is a probe exemption: let one call per interval through to re-establish truth, or expire test results the way runtime observations already expire. A gate whose only exit is a manual button is a gate that fails closed on the user's behalf without telling them.

## 2. The migration renumber breaks any DB that ran the pre-merge branch - and corrupts it silently first

`src/migrations/079_worklog_stale.sql`, renamed in merge `e7cb5756`

The PR body argues this is safe because the migrations "only ever existed on this never-pushed branch." That is exactly the population at risk: the author's own dev DB, and any tester's. Those DBs hold `version=77 (worklog stale)` and `version=78 (day_summary_standup)` as applied.

I traced what actually happens on next daemon start, in order:

1. `reconcile_migration_checksums` (`src/db/meridian.rs:122`) iterates `migrator.iter()`, finds file 77 is now `pre-main`'s `077_drop_board_hygiene_notifications.sql` with a different checksum, and **rewrites the stored checksum**. That marks a migration as applied whose SQL never ran on that database. It commits immediately, before anything else fails, and nothing will ever apply the real 077 - a permanent, silent schema divergence.
2. `migrator.run()` then raises `VersionMissing(78)`. Confirmed in sqlx-core 0.7.2: `ignore_missing` defaults to `false` (`migrator.rs:60`) and the check fires at `migrator.rs:28`. Daemon crash-loop, with an error that names a version number and nothing else.
3. Even after deleting row 78 by hand, `080_day_summary_standup.sql` is `ALTER TABLE day_summaries ADD COLUMN standup_json` - not idempotent, so it fails again on `duplicate column name`.

Step 1 is the one that concerns me most, because it is silent and it survives the recovery. `reconcile_migration_checksums` exists because a renumber crash-looped us once before; the gap it still has is that it compares **only checksums, never `description`**, so it cannot distinguish "someone edited a comment in a shipped migration" - the case it was written for - from "this version number now refers to an entirely different migration." Teaching it to refuse on a description mismatch would have turned this into a loud, accurate error.

Two asks: a note in the PR describing the recovery for anyone who ran the branch, and I would treat the reconcile hardening as a follow-up ticket rather than part of this PR.

Separately on `--no-verify`: the reasoning is sound - the guard protects *applied* migrations and these were never pushed. But the repo already has a technique that needs no waiver, which is rebase-editing the introducing commit so the file lands fresh at its final number. Worth knowing for next time; not something to redo now.

## 3. Groq setup is an unrecoverable dead end on a compulsory step

`ui/components/GroqSetup.tsx:89`

`connect()` clears the pasted key after `onAdd` succeeds but before the eligibility check can throw. When the free tier 429s mid-probe - a case `replace_custom_llm_provider_key`'s own doc calls out as expected - the function throws with the key field already empty. Every retry re-runs `onAdd`, which the tray rejects with `an endpoint named "Groq" already exists` (`custom_llm.rs:274`), and will reject identically forever.

There is no way out from inside the flow. `LlmProviderPicker` renders `<GroqSetup>` unconditionally on the `free` gate answer and never falls through to `CustomDetail`, which is the only screen carrying the new re-probe path; Back returns to the gate, whose `free` answer lands on the same screen. This PR also deleted the manual `<AddForm>` route, so the previous escape hatch is gone.

This sits on the **compulsory** AI-connect step of first-run, so the failure mode is a new user who cannot finish onboarding. Given a free-tier 429 is the expected case, not the rare one, I would not ship this.

---

# Non-blocking, worth fixing

**4. A configured endpoint can never be deleted.** `ui/components/CustomProviders.tsx:275` - the remove button and its hook method went with the old card, and `remove_custom_llm_provider` now has no caller in `ui/` at all; only its definition (`custom_llm.rs:399`) and its registration (`lib.rs:1221`) remain. Dead command, and a one-way door for the user.

**5. The daily summary contradicts the notification the user just acted on.** `ui/components/summary/WorkList.tsx:170` re-derives staleness in TypeScript at **25 minutes** against the daemon's single `WORKLOG_STALE_MINUTES = 15`, which drives both the notification and the `stale` flag that `api-types.ts` explicitly tells consumers to read instead. This one is worth calling out against the PR's own framing: the title says "daily summary parity", and this is the one place the screen and the daemon disagree by construction.

**6. Stale-draft toasts repeat hourly, forever.** `src/worklog_pipeline/stale.rs:62` - the dedup key steps every 15 measured minutes while `notify_stale_drafts` runs once per hourly fold, so on an actively worked task the key has always advanced and the documented "one per step, not one per hour" guarantee degrades to one per draft per hour, indefinitely.

**7. A rejected plan write still reports "Saved".** `ui/components/plan/PlanView.tsx:232` - `commit` applies to local state then fires `metaAction('confirm', …)`, but the rollback the deleted `persist` carried (`.catch(() => { setSaveError(true); load(true) })`) was not carried over. With the daemon down the column keeps the tasks and the header reads "Saved · 4 tasks" from local state, while the server has none.

This last one is worth pausing on, because it is the **same class as the `⚠ Sync failed` regression this PR fixes** - a rejected write that the UI reports as success. The PR body correctly identifies that pattern and fixes it at the contract in `syncTasks`; `commit` is the instance that got missed. Fixing it the same way would be consistent.

**8. Baseline captured at write time, not read time.** `meridian-core/src/readers/day_task_worklogs/mod.rs:520` - `upsert_draft` captures `drafted_minutes` minutes after the evidence was read, so work added during the LLM call is absorbed into the baseline and never reported as stale. Marked plausible rather than confirmed; I did not reproduce it.

**9. Two smaller ones**, both plausible: the new "Open it again" button (`IntegrationConnect.tsx:586`) fires `cancelOAuth` unawaited (`integrationConnectStore.ts:50` uses `void mutate(...)`) before `startOAuth`, with nothing sequencing teardown before restart; and `testOne` (`useLlmProviderDetection.ts:67`) silently discards a result for any provider id not already in `status`, which `detect()`'s catch can empty to `{}`.

---

# What I checked and found sound

- **`sticky` vs `runtime_health`.** The resolution is right. Placing the `sticky` check before the timestamp comparison, and reading raw `last_test` rather than the age-filtered binding, is what makes an asserted verdict outrank a measurement - which is the whole point of the flag. The reasoning in the comment is accurate.
- **`syncTasks`.** Moving the outcome into the resolved value rather than patching the call site is the correct fix. A promise that cannot reject has to carry its outcome in the value.
- **The observability additions** (`push_health_update`'s discarded emit, `resize_setup_window`) and the `day` / `endpoint_id` allowlist tests are exactly right. Pinning `endpoint_id` as deliberately *not* on `SAFE_STRING_KEYS` because it slugifies a user-typed provider name is the kind of thing that gets silently regressed later; good that it is now a test.

---

I would not merge on 1, 2 and 3. None of the three is deep - the gate needs a probe exemption, the migration needs a recovery note, and Groq setup needs the key retained plus a path to the re-probe screen - but all three are user-facing dead ends, and two of them fail silently.

*Reviewed with Claude Code at max effort: 32 agents, one finder per correctness angle plus an independent verifier per finding. 25 verified findings deduplicated to 10.*

adityaharishch:
Thanks - this was a genuinely useful review. All three blockers are fixed, plus 5, 6, 7 and both halves of 9. Two notes where I disagreed or found something extra.

## 1. Health gate — fixed, and the fix needed two things the diagnosis didn't reach

The mechanism was exactly as described. But a probe exemption alone would not have worked, for two reasons:

- **`record_success` early-returns unless there is a RUNTIME failure to clear.** It decides that via `has_unresolved_failure`, which reads `provider_runtime_health.json` and nothing else. The verdict a probe exists to overturn is usually the failed *manual test*, in a different cache — so the probe would have succeeded, written nothing, and the next call would have been refused by the verdict it had just disproved. Permanently, since a test result has no expiry. Hence a separate `record_probe_success` that always writes.
- **`persist` does not invalidate the in-use health memo** (that call lives in `detect::persist_test_result`). So even after a correct write, the 5-minute `IN_USE_HEALTH_TTL` copy of the disproven verdict kept refusing calls.

The exemption is `probeable`, which distinguishes three `!ok` states: a **measured failure** (probe — nothing else re-measures it), **not installed** (don't — `resolve_cli` caches only successes, so a health poll clears it for free), and **asserted/`sticky`** (don't — a measurement must not overturn an assertion). One call per provider per 15 min, checked and stamped under one lock so concurrent pipeline stages can't turn it into a burst.

## 2. Migration renumber — recovery, and the reconcile hardening

Confirmed, including the ordering: `reconcile_migration_checksums` compares checksums and never `description`, so it rewrites 77's stored checksum to `pre-main`'s `077_drop_board_hygiene_notifications.sql` — marking as applied a migration whose SQL never ran — commits, and only then does `run()` fail on `VersionMissing(78)`.

**Recovery for anyone who ran this branch before the merge**, with the daemon and tray stopped:

```sql
DELETE FROM _sqlx_migrations WHERE version IN (77, 78);
ALTER TABLE day_summaries DROP COLUMN standup_json;   -- only if 078 had applied
```

Then start the daemon: 077, 079 and 080 apply cleanly. The `DROP COLUMN` is needed because `080` is a non-idempotent `ADD COLUMN`. On an encrypted DB, `PRAGMA key` first. Anyone who never ran the branch pre-merge is unaffected, and so is every released install.

**On hardening `reconcile`: I've deliberately left it out**, and not only for scope. Refusing on a description mismatch would crash-loop any DB where a migration file was ever renamed — and since the reconcile function exists precisely because the migration set has been edited before, I can't rule that out from here. That's a worse failure than the one it prevents, so it wants its own change with a survey of what's actually in the wild. Agreed it should be a follow-up ticket.

## 3. Groq — fixed, and it was worse than described

Not just "the retry re-runs `onAdd`": `connect()` opens with `if (!key) return`, and the key had already been cleared — so the retry button did **nothing at all**. No request, no error, no state change, on the compulsory step of first-run.

The key is now held until the whole flow succeeds, and once the endpoint exists a retry re-measures it via `probe_custom_llm_provider` instead of trying to create a second one.

## 4. Endpoint removal — declining, with reasoning

The remove button was deleted deliberately and the rationale is in `CustomProviders.tsx`: the tray refuses to remove the selected provider. With `AddForm` gone, `custom.add` has exactly one caller (the Groq flow), so at most one endpoint exists and it is the selected one — the button's only reachable outcome is the red "switch to another provider first". Restoring it would put back a control that cannot succeed.

You're right that `remove_custom_llm_provider` is now dead, and that the card's comment ("this screen is reached through the tile that IS the selected provider") is loose, since the card maps over all providers. Both worth a follow-up if a second endpoint ever becomes creatable.

## 5, 6, 7, 9 — fixed

- **5**: `stale` is now computed once in Rust (`draft_is_stale`) and carried on `DayDraftState`, rather than re-thresholded in TS. Tested against the notifier's own reader so the two cannot disagree about the same draft.
- **6**: the dedup step now **doubles** (15/30/60/120) rather than stepping linearly. Under an hourly fold, an hour of work cleared four linear buckets, so the key had always moved — "one per step" delivered one per hour, forever. Test asserts eight hourly folds produce at most four notifications while the early escalation still gets through.
- **7**: `load(true)` restored on the catch. Correctly identified as the same class as the `syncTasks` regression.
- **9a**: `restartOAuth` awaits the teardown before starting — the cancel frees the loopback port the start then needs to bind, so the old pair raced the exact "already in progress" error the cancel exists to prevent.
- **9b**: a test result for a provider with no `status` row is now kept rather than discarded after spending a real metered request.

## 8 — not reproduced, agreed as plausible

Left alone. `upsert_draft` capturing the baseline at write time is real, but whether work lands inside the LLM call window depends on fold timing I haven't measured, and guessing at a fix would be worse than leaving it recorded.

---

Also in this push, from testing the branch by hand: **unfinished tasks now carry into the next day automatically**. They were already ranked for carry-over but only as suggestions, which is why the timeline showed "What are you working on today?" while the plan modal behind it listed tasks — the modal pre-fills from suggestions, the timeline reads `confirmed ? plan : []`. Both were right; nothing was committed. Guarded so it only ever writes into today, and only into a day with no `daily_plan_meta` row, so work the user deliberately removed never comes back.

Verification: clippy clean, **1614** Rust tests, **700** UI tests, `tsc` and `next build` clean.

PR #716, opened 2026-08-09
fix(notifications): resolve every deep link, quiet the back-online toast (slices 0b + 3)

Akarsh-Hegde:
Slices 0b and 3 of the notification restructure. Follows #705 (board hygiene).

## Slice 0b - deep links actually go where they say

`navigate()` in `MeridianTimelineShell.tsx` resolved **3** targets (`/plan`, `/worklogs`, `/whats-new`) with **no `else` arm**. Producers emit **7**. The missing four fell through in silence, so the `[View]` button on every fault toast opened the dashboard's default view - and `/tasks?integrations=1` kept being emitted by all five `pm.*` sync faults long after the `/tasks` route was deleted in the Next fold. Nothing failed, because a link that resolves to nothing is indistinguishable from one that works.

- **New `meridian_core::notifications::deep_links`** - the fixed vocabulary (`ALL`) plus retired spellings undelivered outbox rows still carry (`LEGACY`). Every producer now references a constant; `navigate()` handles all of `ALL`, both `LEGACY` entries, and has a `default:` that warns.
- **`pm.*` faults link to Settings → Integrations**, which is what their own `remedy` text already told the user to do. `NoticeBar`'s button changed from "Fix in Tasks →" (Tasks modal) to "Reconnect →" (same destination) - one fault, one place to go, across banner + toast + remedy.
- **`db.repaired` / `db.reset` are dismissible.** Both are raised by the tray's boot repair and cleared by *nothing anywhere in the codebase*, and `NoticeBar` had no dismiss by design - so a machine that ever ran a boot repair carried the banner permanently. They report a past event, not a live condition, so the user acknowledges them (via the already-registered, previously-unused `delete_notice`). Live faults deliberately remain non-dismissable.

## Slice 3 - Status events stop being toasts

**Removed the `system.health` "Back online. / Picking up where you left off." toast.** Measured on a real install: 18 fired in five weeks, one was ever interacted with. Beyond the volume, it was frequently incoherent:

- the went-quiet notice's own toast row is `DELETE`d by `clear_typed` on recovery, so any outage shorter than the tray's 30 s drain retracted the warning before it was shown - the user got a bare "Back online." for an outage they never saw;
- the `reconcile_stale` path fired for a notice raised by a **previous process instance**, i.e. an outage that by definition happened while that tray was not running;
- Meridian restarts its own daemon routinely (updates, watchdog, manual), so this fired on ordinary lifecycle events. Five of the 18 landed *after* the watchdog restart-storm fix (#678), two within minutes of an auto-update.

The went-quiet banner disappearing is the recovery signal. The full post-mortem is a comment at the removal site.

**`system.update` is deliberately left alone.** My earlier read of this data was wrong and I corrected it before building: all six toasts come from `check_for_updates` (reachable only from the tray menu - a user click, and the menu has no other surface for a result) or from `enforce_minimum_version` announcing a restart. They are responses to actions, not unsolicited noise, and a low answer rate on a confirmation toast measures nothing.

## Guards - both halves, both verified against real regressions

The contract crosses a language boundary and nothing type-checks the hop, so it is checked from both ends:

- **`tests/deep_links.rs`** - no producer emits a value outside the vocabulary.
- **`ui/__tests__/deep-links.test.ts`** - `navigate()` resolves every value in it (parsed out of the Rust, not restated).

Either alone passes vacuously. Worth flagging: **the first two drafts of the Rust scanner passed while catching nothing.** Matching exact call shapes missed the actual bug site (`notices.rs`'s `let link = … Some("…")`, a bare `Some` bound to a variable), and widening to "any path on a line mentioning link" still missed it because the keyword and the literal are on different lines. It is now anchored + cross-line, and verified to fail against a deliberate regression in **all four** shapes a deep link is attached by, plus the three ways the bun guard can be broken. A clean-tree pass was never treated as evidence.

## Docs

`NOTIFICATIONS.md` gains the **Ask / Fault / Status** taxonomy with a class per producer, and the "Status is the trap" note. Also removes step 4's per-type-toggle instructions - `type_enabled` and the `notify_*` fields were deleted a while ago and the doc still told people to add them.

## Verification

`cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test --workspace` 943 + all crates green; `bun test` 712 pass / 0 fail. Full pre-push suite passed (fmt, clippy, UI build, bun, cargo test, security audit).

Built in a separate git worktree.

https://claude.ai/code/session_01CbX4WUhDMekZpBnP5aS7dD

### DIFF
--- a/NOTIFICATIONS.md
+++ b/NOTIFICATIONS.md
@@ -276,14 +276,15 @@ on. That is why all six `system.update` toasts stay.
 | `plan.nudge` | Ask | `plan_nudge` | Open Plan · Snooze 1h | snooze → re-enqueue +1h |
 | `worklog.ready` | Ask | `worklog_ready` | Open Worklogs · Snooze 1h | snooze → re-enqueue +1h — **registered but no producer fires it yet** (needs its own scoping: worklog generation is currently on-demand/user-clicked, not scheduled) |
 | `system.fault` | Fault | `system_fault` | View | — (stamp only) |
-| `system.pause` | Status | `system_fault` | View | — (stamp only) — pause/resume, folded off the `sys::notify` bypass |
+| `system.pause` | Fault\*\* | `system_fault` | View | — (stamp only) — pause/resume, folded off the `sys::notify` bypass. \*\*The pause notice is a live condition (tracking is off — a Fault). Its "Resumed" toast (`commands/pause.rs`) is structurally identical to the removed back-online toast — one-shot, timestamp dedup, native — but fires ONLY on a manual resume, so it is a response to a click, not unsolicited Status |
 | `system.health` | Fault | `system_fault` | View | — (stamp only) — daemon went quiet, or the tray couldn't finish installing the backend. The paired **"Back online." recovery toast was removed** (Status: it confirmed a recovery the user often never knew about, and fired on Meridian's own restarts); the banner clearing is the recovery signal |
 | `system.update` | Ask/Fault | `system_fault`\* | View\* | — (stamp only) — update check/download/failure, folded off the bypass. \*Discrete one-shot events (not raise/clear state), so these go through `notifications::enqueue` directly rather than `notices::raise_typed` — see `tray/src-tauri/src/update.rs::notify_update` |
 | `system.notif_permission` | Fault | `system_fault` | View | — (stamp only) — notification permission revoked (macOS TCC or, since gotcha #9, Windows' `ToastNotifier::Setting()`); the one notice guaranteed to reach the user via the dashboard banner even when toasts themselves are broken |
 | `system.capture_permission` | Fault | `system_fault` | View | — (stamp only) — Accessibility/Screen Recording revoked mid-session |
 | `system.disk_low` | Fault | `system_fault` | View | — (stamp only) — `~/.meridian`'s volume below 2 GB free |
 | `pm_worklog.{provider}` | Fault | `system_fault` | View | — (stamp only) — a worklog post to the tracker failed permanently |
-| `summariser.dead_letter` | Status | `generic_link` | Open | — (stamp only) — daily digest of permanently-failed coding-agent summarisation, `dedup_key` scoped per day |
+| `summariser.dead_letter` | Fault | `generic_link` | Open | — (stamp only) — daily digest of permanently-failed coding-agent summarisation, `dedup_key` scoped per day. Body asks the user to check their coding-agent CLIs are signed in, so it is a Fault needing action, not Status |
+| `worklog.stale` | Ask | `generic_link` | Open | — (stamp only) — a draft has drifted from the work it describes; links to the worklog review surface |
 | *(reserved)* | Ask | `verify_switch` | Yes · No · Reply… | — (PR 2: task-switch verification) |
 | *(generic)* | — | `generic_link` | Open | — (stamp only) |
 
--- a/NOTIFICATIONS.md
+++ b/NOTIFICATIONS.md
@@ -104,11 +104,17 @@ simple for the user.
    derived from `responded_at`, so a retry is a no-op). Rows not matched by any
    arm are just stamped consumed.
 
-4. **Per-type toggle** — add a field to `RuntimeSettings`
-   (`meridian-core/src/settings.rs`), a match arm in
-   `meridian-core/src/notifications.rs::type_enabled`, and the checkbox in
-   `ui/components/timeline/settings/NotificationsSection.tsx`. Unknown
-   event keys default to enabled, so this step gates opt-out, not delivery.
+4. **Deep link** — if the toast should go somewhere, `.link(deep_links::X)`
+   using a constant from `meridian-core/src/notifications/mod.rs::deep_links`,
+   never a raw path. A new destination means adding the constant AND an arm in
+   `MeridianTimelineShell.tsx`'s `navigate`; two guard tests
+   (`tests/deep_links.rs`, `ui/__tests__/deep-links.test.ts`) fail if you do
+   one without the other.
+
+There is **no step 4 for a per-type settings toggle** — that used to be listed
+here and the machinery (`type_enabled`, the `notify_*` fields) was deleted.
+Every event is gated by the master switch alone; classification and timing are
+decided in code, not handed to the user.
 
 No tray changes are needed — delivery, buttons, response capture, deep-link
 navigation, and expiry all key off the row's columns.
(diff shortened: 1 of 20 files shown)


Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde (commented):
Reviewed strictly, both halves. The deep-link vocabulary is the right shape and the two-ended guard is genuinely load-bearing — I checked the Rust scanner does catch the `notices.rs` `let link = … Some("…")` site the header says it does, and the bun side really does parse the constants rather than restate them. The `system.health` removal is well argued and the test rewrite matches the new behaviour.

One blocking item, and it is the same bug the PR is about.

## Blocking

### 1. The new Dismiss button is dead on arrival — `NoticeBar.tsx:172`

```ts
await mutate('/api/notices', 'delete_notice', { noticeId }, 'DELETE')
```

`mutate` does not pass its third argument through — `bridge.ts:105` is `invoke(command, { body })`. So the command receives `{ body: { noticeId } }`. But `delete_notice` (`tray/src-tauri/src/commands/notices.rs:42`) takes `notice_id: String` at the top level, not a `body:` struct — it is one of the few write commands that predates the `body` convention every other `mutate` caller relies on (`disconnect_integration`, `save_integration_token`, `worklog_action`, … all take `body:`).

Tauri fails arg deserialization, the promise rejects, and `catch { setBusy(false) }` swallows it. The button renders, is clickable, and does nothing — no error, no log, no failing test. Which is precisely the failure mode this PR's own module header describes, and the third instance of it in this file's history after the `window.confirm` Repair Database button that shipped dead in 1.83.0 (CLAUDE.md documents that one).

The working caller is two files away and already correct — `integrationConnectStore.ts:43`:

```ts
void invoke('delete_notice', { noticeId: `pm.${provider}` })
```

Use `invoke` directly here too (Tauri maps `noticeId` → `notice_id`), or give `delete_notice` a `body:` param. Either way this needs a test: the PR argues at length that a silent no-op is indistinguishable from success, then ships one with no coverage. A `__tests__` scan asserting `NoticeBar.tsx` calls `delete_notice` via `invoke` and not `mutate` would be in the same spirit as the two guards you did write.

Also worth doing while you are in there: the `catch` should surface something. `catch {}` on a write is what made this invisible to you locally — at minimum `console.error` so it shows up in the webview console.

## Should fix

### 2. The new taxonomy's central rule is contradicted by its own table — `NOTIFICATIONS.md`

The table says **Status → "banner or tray only - never a toast"**, then classifies two producers as Status that both enqueue native toasts:

- `system.pause` — `tray/src-tauri/src/commands/pause.rs:362` enqueues `"Resumed" / "Meridian is back tracking."` via `CHANNEL_NATIVE`, with a `format!("system.pause:resumed:{}", now_secs())` dedup key. That is structurally *the same object* as `send_back_online_toast` — one-shot recovery confirmation, timestamp dedup, native channel. The only difference is that a manual resume is user-initiated, which is exactly your `system.update` carve-out ("a confirmation of something the user just clicked is not Status"). So the row is mislabelled: it should carry the same Ask/Fault-with-a-footnote treatment `system.update` gets, not Status.
- `summariser.dead_letter` — `summariser/mod.rs:817` enqueues with an `Open` button and no such excuse. It is unsolicited, periodic, and toasted. Either it is not Status, or it should not toast.

As written, a reader arrives at "Status never toasts", scans down two rows, and finds two Status rows that toast. That is worse than no taxonomy — it reads as a rule and behaves as an aspiration. Either fix the labels, or add an explicit "not yet conformant" column so the gap is stated rather than implied.

### 3. `deep_link_literals` silently skips windows — `tests/deep_links.rs:404`

```rust
if !code.is_char_boundary(at) || !code.is_char_boundary(end) {
    continue;
}
```

When `at + LOOKAHEAD` lands mid-codepoint, the entire anchor's window is discarded — literals inside it are never checked. I scanned every file in `PRODUCER_DIRS` on this branch: currently zero windows are skipped, so nothing is being missed today. But this repo's comments are full of em-dashes, only whole-line comments are stripped, and a trailing `//` comment on a line 150 bytes after an anchor is enough to trigger it. A guard whose own header says "verify any future change against a deliberate regression, not just against a clean tree" should not have a branch that quietly reduces coverage. Walk `end` down to the nearest boundary instead:

```rust
let mut end = (at + LOOKAHEAD).min(code.len());
while end > at && !code.is_char_boundary(end) { end -= 1; }
```

### 4. Both parsers slice to EOF instead of to the construct they mean — `ui/__tests__/deep-links.test.ts:753, 798`

`parseDeepLinks` does `body = src.slice(modStart)` — the rest of the *file*, not the `deep_links` module. It works only because `deep_links` happens to be the last thing in `mod.rs` declaring `pub const X: &str`. Add one unrelated `pub const` below it and it is silently absorbed into `all`, and then `navigate() handles every current deep link` fails with a message pointing at the wrong thing entirely.

Same shape at line 798: `nav = shell.slice(shell.indexOf('const navigate ='))` runs to the end of a 460-line file, so `default:` and `console.warn` anywhere below would satisfy it. Today there is exactly one of each, so it is not vacuous yet — but "not vacuous yet, by coincidence of ordering" is the thing this PR spent three paragraphs arguing against. Bound both slices (brace-match the module; slice `navigate` at the closing of its `useEffect`).

## Nits

5. `deep_links::is_known` is documented as "used by the guard test **and by producers that accept a caller-supplied link**" — no producer calls it. Either wire it in (a debug-assert in `NewNotification::link` / `raise_typed` would make the runtime honest and the static scanner mostly redundant) or drop the second clause.

6. `/today` is the one `ALL` entry whose handler (`setActiveModal(null)`) is behaviourally indistinguishable from the old silent fall-through. It is defensible — the stale-draft notice's subject really is on the dashboard — but it means the fix is six arms plus one no-op, and `stale.rs`'s "Your worklog draft is out of date" arguably wants `/worklogs`. Worth a second look.

7. Pre-existing, not introduced here, but same class: `NoticeBar` renders on every route except `/setup`, while `meridian:open-settings` (and the `meridian:open-tasks` it replaces) is only listened for by `MeridianTimelineShell`. On `/uninstall` the "Reconnect →" button is another silent dead click. Since you are auditing exactly this failure mode, it may be worth gating the button on the shell being mounted, or moving the listener up.

## What I checked and found fine

- `db.repaired` / `db.reset` are raised only when `repair_outcome` is `Some` (`lib.rs:550-600`), so dismissing them cannot be undone by the next boot — the `ACKNOWLEDGEABLE` design holds.
- `SettingsSection` accepts `'integrations'` and `'account'`; `setSettingsSection` is `useState<SettingsSection | undefined>`, so the `openSettings()` no-arg call is correct.
- `delete_notice` is registered in `invoke_handler!` (`lib.rs:1168`) and already invoked from the dashboard window, so the capability grant is not the problem in item 1.
- The Rust guard's `files.len() > 50` floor and the bun guard's `all.length >= 7` floor are the right instinct — both suites would otherwise pass over an empty scan.

Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
## Review addressed - ready to merge

All findings fixed in `1c31ec8`. CI green on that exact head (14 checks: clippy + test on macOS **and** Windows, UI build, CodeQL).

### Blocking

**1. The dismiss button was dead on arrival.** Confirmed exactly as described: `mutate` wraps its argument (`invoke(cmd, { body })`), `delete_notice` declares `notice_id` at the top level, so Tauri rejected the args and `catch {}` swallowed the rejection. A control that rendered, clicked, and did nothing - in a PR arguing that silent no-ops are how these bugs survive.

Now uses `invoke('delete_notice', { noticeId })`, matching the working caller in `integrationConnectStore.ts:43`, and the catch `console.error`s instead of swallowing.

The requested test is `ui/__tests__/mutate-body-contract.test.ts`, and it covers **the class rather than the instance**: it parses every `mutate()` call site and every command signature, and fails when a command expects a JS-supplied argument other than `body`. Two notes on getting the rule right -

- My first predicate flagged `sync_tasks` and `open_setup` as well. Those are false positives: `sync_tasks()` takes no arguments and `open_setup(app: AppHandle)` takes only an injected one, so Tauri discards the surplus `body` key harmlessly. The predicate now separates injected params (`State<…>`, `AppHandle`, `Window`) from JS-facing ones.
- The test initially **passed vacuously**. `repoRoot` was built by string concatenation, so every path still contained `/__tests__/` and the "skip test files" filter discarded all 204 sources. Only the `sites.length > 10` floor caught it. Fixed with `resolve()`, and the reason is a comment on the line.

Verified by reverting to `mutate` - the suite fails naming the file, the command, and the expected argument.

### Should fix

**2. The taxonomy contradicted its own table.** Fixed, and both labels were wrong for different reasons. `system.pause` is a Fault (tracking is off - a live condition); its "Resumed" toast is structurally identical to the back-online toast I removed, but fires only on a *manual* resume, so it falls under the same carve-out as `system.update` and now carries that footnote. `summariser.dead_letter` asks the user to go sign in to a CLI, which makes it a Fault, not Status. Also added the missing `worklog.stale` row.

**3. `deep_link_literals` skipped whole windows.** Now walks `end` down to the nearest char boundary instead of `continue`-ing past the anchor.

**4. Both parsers sliced to EOF.** Both are brace-matched now (`braceBlock`), plus a `nav.length < 3000` assertion so "bounded" is itself checked rather than assumed.

### Nits

**5.** `is_known`'s doc no longer claims producers call it - it now states why they deliberately don't (a runtime check fires on the user's machine; the static scan fails the build).

**6.** `/today` on the stale-draft nudge → `/worklogs`. You were right that it wanted a second look: `/today`'s handler only closes modals, so it was the one `ALL` entry still behaviourally identical to the old silent fall-through.

**7.** Fixed rather than left pre-existing. `LayoutBanners` now hides NoticeBar on `/uninstall` as well as `/setup` - gating the whole bar, not individual buttons, since neither banner is actionable there and `db.corrupt`'s Repair button restarting the app mid-uninstall would be actively wrong.

### Two more, from an independent adversarial pass

- **`meridian:open-tasks` had no dispatcher left** once NoticeBar's button moved to Settings. Removed the dead listener and corrected two comments (here and `TaskDetailDialog.tsx:147`) that still cited it as the live exemplar.
- **`db.repaired` rendered in red error styling.** It is raised with severity `info`, `SEVERITY_STYLES` had only `error`/`warning`, and the lookup falls back to `error` - so a *successful* repair looked like a failure. Added the `info` style (the `--status-info-*` tokens already existed). Only reachable now because making these dismissable is what put an info-severity notice on screen in the first place.

### On CodeRabbit

It produced **no findings and no threads** - it replied `Review rate limited` to both manual triggers, the second within 5 seconds. Its status check reports `pass` with that same text, which is a vacuous pass, not a review. There are therefore no conversations to resolve (`reviewThreads.totalCount` is 0). Flagging it explicitly so the `pass` is not mistaken for coverage: the review this PR actually received is the human one above plus the adversarial pass.

PR #719, opened 2026-08-09
fix(uninstall): remove the data that survived every uninstall path

adityaharishch:
Uninstalling left the user **signed in**, with their **tracker credentials**, their **activity database and its encryption key** still on the machine.

## The four holes

Each was invisible, because nothing reported them.

**1. The OS keychain entry holding the SQLCipher key for `meridian.db` was never touched.** It lives on no filesystem path, so no amount of deleting directories reached it - both a privacy leak (the key to a database the user asked us to destroy) and a correctness hazard, since a reinstall finds a key and assumes the database it decrypts is its own.

It is now deleted through the **same `keyring` crate and service/account pair the tray writes it with**, rather than a hand-rolled `security delete-generic-password` / `cmdkey /delete`. Windows' credential target is an internal `keyring` convention (`{user}.{service}`), so spelling it out here would silently stop matching the day that crate changes it - and the failure would be invisible, which is the exact class of bug this PR is about.

**2. On Windows nothing removed `%LOCALAPPDATA%`/`%APPDATA%\com.meridiona.tray`** - where WebView2 keeps the cookies and localStorage the signed-in session actually lives in. Uninstall + reinstall came back **still logged in**. macOS already had the equivalent sweep; Windows had none.

**3. The tray's launch-at-login is an HKCU `Run` value**, a different mechanism from the daemon's scheduled task, so `schtasks /Delete` never touched it - the tray kept trying to start at every login with its executable deleted. macOS was already covered by the `com.meridiona.*` LaunchAgent sweep.

**4. The `autostart_configured` marker was not in the data list**, so a reinstall read it, believed autostart was already set up, and never re-registered the login item. The app silently stopped starting at login.

## The default that made it reachable

The wizard's three checkboxes **defaulted to unchecked**. Clicking through "uninstall" removed the app and kept the database, the Jira/Linear tokens and the account - which is what was actually being hit in practice.

They now default to **checked**, with every item still listed and individually unticked, so "remove the app, keep my data" is one click away instead of the silent default.

Keychain removal is wired into **both** the human path and the `--json` path, since the `--json` one is what the tray's wizard drives and is how most users uninstall.

## Tests

`bundle_dirs_under` is split out of the Windows branch specifically so it is testable **without** mutating `%LOCALAPPDATA%`/`%APPDATA%` - those are process-global and Rust runs tests in parallel threads, so a test that set them would race every other test in the binary. It compiles on non-Windows only under `cfg(test)`, so the coverage runs on every platform's CI while release builds carry nothing dead.

- `data_items_includes_the_autostart_marker`
- `bundle_dirs_are_listed_only_when_they_exist`
- 10 uninstall tests pass, 711 UI tests pass, clippy and fmt clean.

## Still open, deliberately not fixed here

**Uninstalling via Windows' Add or Remove Programs runs no Meridian code at all** - there is no NSIS uninstall hook - so that route still leaves everything behind. A hook is the fix, but Tauri runs the previous version's uninstaller during an **upgrade** too, so a hook that purges data risks **wiping the database on every auto-update**. That needs a real Windows install/upgrade test before it ships, which I could not do here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5180,6 +5180,7 @@ dependencies = [
  "hex",
  "http",
  "jsonschema",
+ "keyring",
  "libsqlite3-sys",
  "meridian-core",
  "meridian-oauth",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,6 +125,13 @@ hex = "0.4"
 # Error handling
 anyhow = "1.0.86"
 
+# OS keychain — `meridian uninstall --remove-data/--purge` deletes the SQLCipher
+# key for meridian.db, the one piece of user data that lives nowhere on disk.
+# Version and features are deliberately IDENTICAL to tray/src-tauri/Cargo.toml's
+# (the crate that writes the entry, in db_key.rs) so cargo unifies them into one
+# build and the delete cannot drift from the write.
+keyring = { version = "3", features = ["apple-native", "windows-native"] }
+
 # HTTP client for PM provider APIs (Jira, GitHub, Linear)
 reqwest = { version = "0.12", features = ["json", "rustls-tls"], default-features = false }
 
--- a/src/uninstall.rs
+++ b/src/uninstall.rs
@@ -14,13 +14,19 @@
 //! binaries, the install marker) and **keeps user data**. Three independent,
 //! additive flags widen the scope:
 //! - `--remove-data` — `~/.meridian`'s user data (db, credentials, settings,
-//!   logs, telemetry spool, icon cache, onboarded marker), plus (macOS only)
-//!   the OS-managed WebKit/AppKit caches for the tray's bundle id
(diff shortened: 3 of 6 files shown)


adityaharishch:
**Stacked PR - merge #718 first.**

Base is `perf/windows-release-time` (PR #718) rather than `pre-main`, so the diff above shows only this PR's own changes. GitHub retargets this to `pre-main` automatically once #718 merges.

Merge order: **#718 → #719 → #720**. The three touch disjoint files, so this is purely about keeping each diff readable.

adityaharishch:
Retargeted back to `pre-main`: `ci.yml` only triggers on PRs targeting `pre-main`/`main`, so basing this on a feature branch meant any further push would get **no CI at all**.

This branch therefore also carries #718's single CI commit. It resolves itself - once #718 merges, the merge base moves and the diff here collapses to just the uninstall changes.

Merge order: **#718 → #719 → #720**.

PR #720, opened 2026-08-09
feat(dashboard): invite contributors from the home page

adityaharishch:
Meridian is public and MIT licensed, but **nothing in the app ever said so** - so the people most likely to contribute, the developers using it every day, had no way to find the repo without going looking for it.

## What it looks like

An "Meridian is open source" card at the **bottom of the overview panel** (the dashboard home):

- GitHub mark, title, and `MIT licensed - the daemon, dashboard and tray`
- One line of body copy: read what Meridian does with your activity, report anything wrong, or send a pull request
- Two equal-weight actions: **View on GitHub** and **Good first issues**

It reuses the existing card tokens (`bg-card`, `--t-card-border`, `mt-card-title`, `mt-body-sm`, `mt-card-hover`) and the same button styling as `ReportModal`'s pair, so it reads as part of the panel rather than something bolted on.

## Why the bottom

The top of the overview belongs to the user's own day - the plan, time by app, time by category. An invite to go read the source is worth showing every day but never worth outranking any of that. The panel already ends with the solo-mode tracker nudge, so this lands as the last thing on the page rather than an interruption in the middle of it.

## Both links were verified, not assumed

`ReportModal` carries the scar that motivated checking: Discord and X were listed there before either existed, so a screen whose whole promise was "a person reads every message" shipped two links to a dead invite and an empty profile.

So before adding these I confirmed the repo is public and MIT, and that `good first issue` is a **real label with open issues behind it**. The module doc records the fallback - point the second action at `CONTRIBUTING.md` if that label is ever emptied - so nobody is left with a button that lands on "0 results".

## Details

- Opens via `openExternal` from `@/lib/bridge`, which routes through the Tauri opener command and falls back to `window.open` outside the app shell.
- The GitHub mark is **inline SVG on `currentColor`** rather than an icon-pack import, so it stays correct in every theme (near-black on the light themes, near-white on ink) without a second asset to ship.
- Plain hyphens only in all user-facing strings, per the repo's dash rule.
- Plain `<button>` elements, so they inherit the global `cursor: pointer` rule that `cursor-pointer.test.ts` enforces.

## Checks

711 UI tests pass, `tsc --noEmit` clean, and the static export builds.

### DIFF
--- a/ui/components/timeline/OpenSourceCard.tsx
+++ b/ui/components/timeline/OpenSourceCard.tsx
@@ -1,0 +1,86 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+/**
+ * "Meridian is open source" card, pinned to the bottom of the dashboard home
+ * ([`OverviewPanel`]).
+ *
+ * # Why it lives at the BOTTOM of the overview
+ * The home panel is what someone looks at while they work, so the top of it
+ * belongs to their own day - the plan, time by app, time by category. An invite
+ * to go read the source is worth showing every day but never worth outranking
+ * any of that, and the overview already ends with the solo-mode tracker nudge,
+ * so this reads as the last thing on the page rather than an interruption in
+ * the middle of it.
+ *
+ * # Why every link here is checked to go somewhere
+ * `ReportModal` carries the scar: Discord and X were listed before either
+ * existed, so a screen whose whole promise was "a person reads this" shipped two
+ * links to a dead invite and an empty profile. Both targets below were verified
+ * live before being added - the repo is public and MIT, and `good first issue`
+ * is a real label with open issues behind it. If that label is ever emptied,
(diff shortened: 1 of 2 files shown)


adityaharishch:
**Stacked PR - merge #718 and #719 first.**

Base is `fix/uninstall-removes-everything` (PR #719) rather than `pre-main`, so the diff above shows only this PR's own two files. GitHub retargets this down the chain automatically as each parent merges.

Merge order: **#718 → #719 → #720**. The three touch disjoint files, so this is purely about keeping each diff readable.

adityaharishch:
Retargeted back to `pre-main`: `ci.yml` only triggers on PRs targeting `pre-main`/`main`, so while this was based on a feature branch it got **no CI at all** - coverage matters more than a temporarily tidier diff.

So this branch currently also carries #718's and #719's commits. That resolves itself: once those two merge, the merge base moves and the diff here collapses to just its own two files, with no action needed.

Merge order is still **#718 → #719 → #720**.

PR #724, opened 2026-08-10
fix(tray): stop SIGHUPing the daemon into launchd's throttle window

Akarsh-Hegde:
## Symptom

The integrations panel showed, under a connected GitHub account:

```
tasks: fetch pm_tasks: error returned from database:
(code: 11) database disk image is malformed
```

## The database was never damaged

Established before changing anything:

- `meridian db check` - **healthy, 39 tables scanned**
- `PRAGMA integrity_check` via sqlcipher - **ok** (this one also validates every
  index, which `db check` does not: `integrity::scan_one` runs `SELECT * FROM t`,
  a table-b-tree read that never walks a secondary index)
- The errors are **tray-only**: 6 spool files from `meridian-tray`, **zero** from
  `meridian-rust`, and the daemon raised no `db.corrupt` notice

So this is the "tray-only code 11 on an intact file" profile, not b-tree damage.
**Do not run `meridian db repair` for this** - it would rebuild a healthy 582 MB
database, and it has re-corrupted before.

## Root cause: we SIGHUP the daemon into launchd's throttle window

A "reload" on macOS is `kill -HUP`, and the daemon handles SIGHUP by **exiting
cleanly** so launchd relaunches it (`src/platform/unix.rs`). launchd will not
relaunch a job more often than the plist's `ThrottleInterval`, which is **30 s**
and is measured from the job's last *start*. A second SIGHUP inside that window
therefore does not reload the daemon - it **kills it for the rest of the window**.

Measured on 1.84.0-staging.7, from the telemetry spool:

```
03:05:42.556  daemon starting          first reload - launchd relaunches at once
03:05:48.331  SIGHUP received          second reload, 6 s later - daemon exits
03:05:55      watchdog: "not running"  launchd throttled; nothing to probe
03:06:12.847  daemon starting          = 03:05:42.556 + 30 s, exactly
```

The daemon was **down for 24 seconds**. The tray holds one long-lived SQLCipher
pool for the life of the app, and across that window its pooled connections
returned `(code: 11)` from `get_tasks` every ~30 s (08:36:12 -> 08:37:39) before
self-healing. That is what the user saw.

`reload_daemon` had **no in-flight guard at all** - a bare passthrough to
`kill -HUP`. Two independent call sites reload after a credential change
(`integrations.rs`'s token save, and the GitHub device-flow completion), so a
single connect flow can trip this on its own.

## The fix

`reload_daemon` now rate-limits itself to launchd's own throttle:

- Outside the window -> signal immediately (unchanged).
- Inside it -> schedule **one** trailing reload for the moment the window closes,
  and return without blocking the caller.
- Further requests while one is pending -> coalesced into it.

Deferring is safe precisely because the daemon re-reads `.env` / `settings.json`
**at startup**: a reload that fires later still picks up every write made before
it, so coalescing cannot lose a credential update. Both call sites are already
best-effort, and one answers a UI request, which is why this defers rather than
blocking.

The decision is a pure function (`plan_reload`) so the behaviour is testable
without a daemon, a socket, or a clock.

## Tests

Written first, and verified to be a genuine red: with `plan_reload` stubbed back
to the old always-signal behaviour, **3 of the 5 fail**, including the exact
6-second field sequence.

- `a_second_reload_six_seconds_later_is_deferred_not_signalled` - the field case
- `nothing_inside_the_throttle_window_is_ever_signalled_immediately` - sweeps all
  30 seconds of the window and asserts each deferral lands exactly at the boundary
- `a_reload_after_the_window_closes_is_signalled` - the guard re-arms, so a
  credential change never needs a manual restart
- `requests_arriving_while_one_is_pending_are_coalesced` - a burst collapses to one
- `the_throttle_constant_matches_the_daemon_plist` - parses
  `scripts/com.meridiona.daemon.plist` and fails if `THROTTLE_INTERVAL` drifts
  from the value launchd actually enforces

## Known limitation, deliberately not fixed here

The guard spaces out reloads **the tray sends**. It does not know when launchd
last started the daemon for other reasons - so a reload issued within 30 s of an
unrelated daemon start (tray launch, a crash-restart, the watchdog) can still hit
the throttle. Closing that needs the daemon's actual start time from the probe,
which is a larger change; this PR removes the reproducible self-inflicted case.

Also unaddressed by design: the tray surfacing a raw SQLite string in the
integrations panel. A retry-and-swallow in the tray was considered and rejected -
`is_corrupt_error` is the same predicate the daemon uses to latch `db.corrupt`,
and the tray-only-vs-daemon split is exactly the diagnostic that identified this
incident. Muffling it would have cost more than it bought.

## Verification

- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `cargo fmt --all -- --check` - clean

https://claude.ai/code/session_01VcBjYEVA8pzFD9JhTs32ya

### DIFF
--- a/tray/src-tauri/Cargo.toml
+++ b/tray/src-tauri/Cargo.toml
@@ -250,6 +250,12 @@ capture = [
 # whatever happens to exist on the machine running the tests.
 [dev-dependencies]
 tempfile = "3"
+# Test-only: `test-util` gives `#[tokio::test(start_paused = true)]`, which
+# advances tokio's clock instantly. commands/daemon.rs's reload rate limit is
+# built around a 30 s launchd window; without a paused clock its tests would
+# have to sleep through that window for real, so they would either be skipped
+# or made meaningless with a shrunken constant that no longer matches the plist.
+tokio = { version = "1", features = ["test-util"] }
 # Test-only: captures tracing events (level + fields) to assert log-level and
 # redaction-allowlist-field regressions in commands/setup.rs's LLM install/
 # sign-in logging without spawning a real subscriber pipeline.
--- a/tray/src-tauri/src/commands/daemon.rs
+++ b/tray/src-tauri/src/commands/daemon.rs
@@ -25,7 +25,8 @@
 use crate::state::{AppState, StatusPayload};
 use meridian_core::SqlitePool;
 use serde::Serialize;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, OnceLock};
+use std::time::{Duration, Instant};
 use tauri::State;
 use tracing::Instrument;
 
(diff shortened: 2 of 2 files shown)


Akarsh-Hegde:
@coderabbitai full review

Akarsh-Hegde (commented):
## Review — strict

The diagnosis is excellent and the writeup is the best kind of PR body: measured timestamps, a ruled-out alternative (`db repair`), and an explicit statement of what is *not* fixed. `plan_reload` being pure is the right shape.

But the pure function is the easy half, and it is the only half that is tested. Every problem below lives in the stateful wrapper around it, and two of them can leave reload permanently dead or reproduce a (smaller) version of the outage this PR exists to remove.

---

### Blocking

**1. `pending` can latch `true` for the life of the process, silently killing reload entirely** — `daemon.rs:172` / `188-197`

`pending` is set inside the lock and cleared only by the spawned task, after `send_reload()` returns. Nothing else can ever clear it. So if that task dies before line 193, `plan_reload` returns `AlreadyPending` forever and *every* subsequent reload — token saves, OAuth completions, the reload button — is answered `{ok: true, pid: 0}` and dropped. No log, no notice, no recovery short of quitting the tray.

That is not theoretical given the code as written: `send_reload` itself does `.expect("reload state mutex poisoned")` (line 218-221). One panic anywhere while that mutex is held poisons it; the next `expect` panics the spawned task; `pending` is never cleared; reload is bricked. The failure mode of the guard is strictly worse than the bug it guards against, because it is silent and permanent.

Minimum: clear `pending` with a guard that runs on unwind, or clear it *before* sending (see #4), and drop the `expect`s for `unwrap_or_else(|e| e.into_inner())` per CLAUDE.md's no-`unwrap`-outside-tests rule.

**2. The deferred SIGHUP still lands inside launchd's real window, and its failure is swallowed** — `daemon.rs:188-201`

The guard measures 30 s from *our SIGHUP*. launchd measures 30 s from the daemon's *start*, and the start is `SIGHUP + however long the daemon takes to shut down cleanly` (pool close, spool flush — not microseconds). So the deferred reload at `last_sent + 30 s` is reliably `δ` *inside* the window launchd actually enforces, where `δ` is the graceful-shutdown time. Result: the daemon exits and stays down for `δ`.

That is a large improvement on 24 s and I would not block on the size of it — but the PR presents the timing as exact ("fires exactly when the window closes", and the sweep test asserts it), and it isn't. Either add a margin (`THROTTLE_INTERVAL + 2s`) or say plainly in the doc comment that the boundary is approached from the wrong side by the shutdown duration.

Worse, the same moment is the most likely one for `send_reload` to fail: at `last_sent + 30 s` launchd is relaunching, and `daemon_control::reload` probes first and returns `Err("daemon not running")` if the socket isn't up yet. The deferred reload is then dropped with a `tracing::warn!` and nothing retries it. The `db.corrupt`-probe PR (#693) already learned this lesson and retried its outcome notice; this one shouldn't regress it.

---

### Correctness

**3. The `reloaded === false` warning the UI depends on is now unreachable inside the window** — `daemon.rs:206`, `210`

`save_integration_token` returns `reloaded` straight from `reload_daemon().await.is_ok()`, and the frontend renders it: `integrationConnectStore.ts:344`, `:426`, `:507` and `IntegrationConnect.tsx:664` all set `reloadWarning` from `res?.reloaded === false`. That warning exists to tell the user "your credentials won't take effect until the daemon starts" — the exact situation in dev, and the exact situation when the daemon is throttled-down.

Both new branches return `ok: true` without probing anything, so inside the throttle window a genuinely-down daemon now reports success and the user gets no warning; if the deferred send then fails (#2), the only trace is a `warn!` in the spool. The comment at line 203 ("the callers only use it to warn when the daemon is down entirely") is exactly right about the contract and exactly the contract being broken. Probe before deferring, or add a third state to `ReloadResponse` and have the UI say "will apply shortly".

**4. Two concurrent callers can still both SIGHUP — the same two call sites, just tighter** — `daemon.rs:161-176` vs `216-222`

The lock is released after `plan_reload`, but `last_sent` is only stamped *after* `daemon_control::reload()` completes, and that await includes an 800 ms-budget probe plus a `kill` spawn. Any second `reload_daemon` arriving in that gap reads the old `last_sent` and also gets `SendNow`. The field case was 6 s apart so this PR fixes it, but the two call sites named in the description (token save, device-flow completion) are not synchronised with each other and a sub-second overlap is possible. Stamp `last_sent = now` optimistically under the same lock that returns `SendNow`, and roll it back only if the send errors.

**5. Coalescing can drop a request whose write happened after the signal** — `daemon.rs:190-193`

`pending` is cleared *after* `send_reload()`. A request arriving between "SIGHUP sent" and "`pending = false`" is answered `AlreadyPending` and discarded — but its `.env` write happened *after* the reload it was coalesced into. The PR's safety argument ("a reload that fires later still picks up every write made before it") holds only for requests that arrive before the send; this ordering breaks it. Clear `pending` before calling `send_reload`, or carry a generation counter.

---

### Platform

**6. The guard is applied on Windows, where the throttle does not exist** — `daemon.rs:159+`

`daemon_control::reload()` on Windows is `restart()` — `schtasks /End` + `/Run`. No `ThrottleInterval`, no SIGHUP, no launchd. This PR now delays Windows credential pickup by up to 30 s to respect a constant that has no meaning there, and `the_throttle_constant_matches_the_daemon_plist` compiles into the Windows build to pin a macOS plist. Gate the guard on `#[cfg(target_os = "macos")]`.

---

### Scope — the "known limitation" is smaller than described

The body says closing the other-starts case "needs the daemon's actual start time from the probe, which is a larger change". The tray *initiates* most of those starts itself and could stamp `last_sent` in the same place it already knows about them:

- `daemon_control::restart()` (`kickstart -k`) — `tray.rs:178`, `commands::restart_daemon`
- `set_running(true)` — `commands/repair.rs`
- `register_agent`'s kickstart at install/first launch — `backend_install.rs:1294`
- the watchdog

Stamping in `daemon_control::restart`/`start` is a few lines and covers nearly all of it. Given the watchdog restart-storm history called out in `poll/watchdog.rs`, "a reload within 30 s of a watchdog restart still takes the daemon down" is not a comfortable thing to leave open when it's this cheap to close. Not blocking, but please either do it here or open the follow-up issue.

---

### Tests

**7. The five tests only exercise the pure function.** `plan_reload` has no state, no clock and no I/O — the tests are a restatement of its three-line spec, and "3 of 5 fail when `plan_reload` is stubbed back" is tautological for a pure function: they test the stub, not the fix. Every defect above (#1 `pending` lifecycle, #4 the lock-release race, #5 the clear-after-send ordering, #2 the failure path) is in `reload_daemon`, which has zero coverage. Make the send injectable (a fn pointer, or move the state into `AppState`) so the state machine can be driven in a test.

**8. `nothing_inside_the_throttle_window_...` steps in whole seconds.** The interesting boundary is sub-second — `+29.999 s` is where `remaining.is_zero()` decides — and it's one extra loop to cover.

---

### Nits

- `daemon.rs:294` — the comment says `checked_duration_since`; the code calls `saturating_duration_since`. (The *reasoning* in the comment is right, just the name.)
- Process-global `OnceLock<Mutex<ReloadState>>` in a command module rather than `AppState` is what makes #7 hard; worth moving while this is still small.
- `include_str!("../../../../scripts/com.meridiona.daemon.plist")` reaches four levels out of the crate. It resolves and I like the intent, but `backend_install.rs:1561` solves the same problem via `CARGO_MANIFEST_DIR` at runtime — worth matching the existing pattern.

---

Happy to re-review quickly. #1 and #3 are the ones I'd want fixed before this ships; #2's retry too, since the deferred path's most likely failure is the moment it was scheduled for.

Akarsh-Hegde:
## Review addressed — ready to merge

The review's framing was correct and worth repeating: the pure planner was the easy half and the only half under test, while every real defect lived in the stateful wrapper. Two of them could have left reload permanently dead.

### Blocking

**1. `pending` could latch for the life of the process.** Set under the lock, cleared only by the spawned task after the send returned — so any unwind in between (including the `.expect()` on a poisoned mutex, which the same code could cause) left `plan_reload` answering `AlreadyPending` forever. Every later reload would return `{ok: true, pid: 0}` and do nothing, with no log and no recovery short of quitting the tray. A guard whose failure mode is silent and permanent is worse than the bug it guards.

Now cleared from a `Drop` guard, and the mutex is taken with `unwrap_or_else(|e| e.into_inner())` — no `expect`, per CLAUDE.md.

**2. The deferred signal landed inside launchd's real window.** launchd measures from the daemon's *start*; we can only observe our own SIGHUP; the graceful shutdown sits between them. Waiting exactly the plist value approached the boundary from the wrong side by the shutdown duration — recreating the outage at reduced size, while the PR body claimed the timing was exact. It isn't, and the doc now says so.

`THROTTLE_INTERVAL` is derived as `LAUNCHD_THROTTLE + SHUTDOWN_MARGIN`, and the deferred send **retries** (3×) rather than being dropped at the one moment launchd is most likely to still be relaunching.

### Correctness

**3. The `reloaded === false` warning was unreachable inside the window.** Both new branches returned `ok` without probing, so a genuinely-down daemon reported success exactly when it was most likely to be down. The deferred path now probes and returns the same `daemon not running` error the callers already understand — and schedules nothing, since a down daemon reads config on start anyway.

**4. Two callers could both get `SendNow`.** `last_sent` is now stamped optimistically under the same lock that returns `SendNow`, and rolled back if the signal fails (a signal that never landed opened no window, so the next attempt must not wait one out).

**5. Coalescing could drop a request whose write came after the signal.** `pending` is cleared in the same critical section as the new stamp, *before* the send. Requests arriving from then on see `pending: false` with a fresh stamp and schedule their own trailing reload, which does cover them.

### Platform

**6. Gated to macOS.** Windows `reload()` is `schtasks /End` + `/Run` — no launchd, no SIGHUP, no `ThrottleInterval` — so the guard only delayed credential pickup by 30 s to respect a constant with no meaning there. The imports it needs are gated with it; without that the Windows build fails on `-D warnings`, invisibly from macOS.

### Tests

**7. The state machine is now driven directly**, with an injected send and probe: the pending lifecycle, the failed-send rollback, the down-daemon error, retry-on-failure, and re-arming — none of which the pure function could reach.

**8. The window sweep gained its sub-second edge** (`-1ms`, `-1ns`), which is where `remaining.is_zero()` actually decides.

One thing worth flagging from writing them: they use tokio's `Instant` with `start_paused`, not std's. With std's, a test that "sleeps" past the 30 s window advances only tokio's timer while the stamps being compared stay microseconds old — so the guard never sees the window close and the test asserts the opposite of the truth. That was a real failure caught while writing it, not a hypothetical.

### Nits

Comment/function-name mismatch fixed. The `include_str!` for the plist is kept deliberately: it makes the constant-vs-plist check a **compile-time** dependency (`cargo:rerun-if-changed` picks the plist up), which a runtime `CARGO_MANIFEST_DIR` read would not.

### Scope — the "known limitation"

Still open and still documented: the guard spaces out reloads *the tray sends*, and does not know when launchd last started the daemon for other reasons (watchdog, tray launch, crash-restart). The review is right that stamping in `daemon_control::restart`/`start` would cover most of it cheaply. I've left it out of this PR to keep the fix to the reproducible self-inflicted case — happy to do it here instead if you'd prefer that to a follow-up.

### Verification

`cargo test --workspace` 1629 pass / 0 fail · `cargo clippy --workspace --all-targets -D warnings` clean · `cargo fmt --all --check` clean · full pre-push suite green.

I have not merged this — leaving that to a human reviewer.

PR #729, opened 2026-08-13
fix(tutorial): stop the walkthrough tests slicing to nowhere

Akarsh-Hegde:
Unblocks the UI job on `pre-main`, which has been red since #725. Test-only.

## The failure

```
(fail) the closing asks: a delivery time, and the off switch > ends by saying nothing more is required
Expected to contain: "Close this window - Meridian keeps going"
```

That copy is gone on purpose. #725 deleted walkthrough beat 11 - a standalone "you are set up, close this window" card shown one beat before the handoff, which already said the same thing at the moment it became true. Two "you're done" cards, so the tour finished twice. The assertion was left behind pointing at the deleted one.

It is repointed at the surviving handoff (beat 13, `That is everything.`). The test name still describes what it checks.

## The more useful half: it failed for the wrong reason

```ts
const close = src.slice(src.indexOf('── 11. Running'))
```

`indexOf` returns `-1` when the marker is gone, and `slice(-1)` reads that as *one character from the end*. So the assertion ran against a single character, and reported **its own expectation** as the problem rather than the missing marker.

The corollary is worse than the confusing message: with a short enough expectation, that shape **passes** against garbage.

Twenty-one other assertions in this file used it. They now go through two helpers that throw a named `marker not found`:

```ts
function after(text: string, marker: string): string
function between(text: string, start: string, end: string): string
```

## Which immediately found a second dead assertion

Converting them turned up one that had already stopped checking anything:

```ts
const bubbles = after(overlay, 'Narration, pinned bottom-centre')
expect(bubbles).not.toContain('className="mt-body"')
```

`overlay` is `code(readFileSync(...))` - comment-stripped, which is what makes "this string is not in the code" mean anything. But that marker is a **JSX comment**, so it does not survive the stripping. The old bare slice turned the miss into one character and the `not.toContain` passed trivially.

Sliced from the raw file now. Verified live rather than assumed: planting `<div className="mt-body" />` into that region fails the test, and removing it passes.

## Verification

- `bun test` from `ui/`: **730 pass / 0 fail** (43 files)
- `npx tsc --noEmit`: clean
- pre-push suite green - `cargo fmt`, `cargo clippy --workspace`, `cargo test --workspace`, ui build, ui tests, security audit

## Why now

#728 (`pre-main` → `main`) is red on exactly this one assertion. That PR's head ref is `pre-main`, so merging this updates it automatically.

https://claude.ai/code/session_01CbX4WUhDMekZpBnP5aS7dD

## Summary by CodeRabbit

* **Tests**
  * Improved tutorial and workflow test validation by ensuring required source markers are present.
  * Updated coverage for tutorial, overlay, settings, planning, replay, worklog, and closing flows.
  * Added validation for updated handoff wording.

### DIFF
--- a/ui/__tests__/tutorial-sample-day.test.ts
+++ b/ui/__tests__/tutorial-sample-day.test.ts
@@ -18,6 +18,36 @@ const SCRIPT_SRC =
   readFileSync(new URL('../components/tutorial/script.ts', import.meta.url), 'utf8')
   + readFileSync(new URL('../components/tutorial/scriptDay.ts', import.meta.url), 'utf8')
 
+// Slice a source file from a marker, and FAIL IF THE MARKER IS NOT THERE.
+//
+// Every assertion in this file that narrows to one beat does it by slicing from
+// a section comment, and the bare form is a trap:
+//
+//   src.slice(src.indexOf('── 11. Running'))   // marker deleted → -1 → last CHARACTER
+//
+// `indexOf` returns -1 for a marker that no longer exists and `slice(-1)` reads
+// that as "one character from the end", so the assertion runs against a
+// one-character string instead of the beat it names. That is a test whose
+// failure message points at its own expectation rather than at the deleted
+// marker - and one whose PASS means nothing at all if the expectation happens to
+// be short. It shipped for real: #725 removed beat 11 and this file kept
+// checking its copy against a single character.
+//
+// Marker-not-found is now its own failure, named as such.
+function after(text: string, marker: string): string {
+  const at = text.indexOf(marker)
+  if (at < 0) throw new Error(`marker not found: ${marker}`)
+  return text.slice(at)
+}
+
+/** `after`, bounded at `end` - which must also exist. */
+function between(text: string, start: string, end: string): string {
+  const body = after(text, start)
+  const at = body.indexOf(end, start.length)
+  if (at < 0) throw new Error(`end marker not found after ${start}: ${end}`)
+  return body.slice(0, at)
+}
+
 // The example day is fed to the REAL DayTaskColumn via its `tasks` prop, so a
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
### Measured the blast radius rather than guessing at it

The same `slice(indexOf(...))` shape appears **36 more times across 15 other files** in `ui/__tests__/`, so the obvious question is how many of those are already dead the way the two in this PR were.

Answer: **none, today.** Not an argument from reading them - measured, by making the failure observable at runtime:

```ts
// diagnostic preload, NOT committed
const orig = String.prototype.slice
String.prototype.slice = function (a, b) {
  if (a === -1 && b === undefined && this.length > 200) console.log('VACUOUS_SLICE@' + site())
  return orig.call(this, a, b)
}
```

`bun test --preload` over all 43 files: **0 hits**.

And because a detector reporting zero is exactly the failure mode this PR is about, I self-tested it before believing it - a planted `s.slice(s.indexOf('THIS MARKER DOES NOT EXIST'))` is caught with file and line, and note it reported **731 pass / 0 fail** while doing so. That is the whole problem in one line: the plant passes.

So the two fixed here were the only live instances in the suite.

### What that leaves

The 36 remaining sites are correct *right now* and go silent the moment anyone renames a marker in the file they point at - which is precisely how the two in this PR died, both times as collateral from an unrelated, correct change.

Two ways to close it permanently, neither included here on purpose:

1. Convert them to `after`/`between` (mechanical, 15 files).
2. Ship the preload above as a CI guard.

I did not bundle either. This PR is unblocking a release, and (2) means monkeypatching `String.prototype` in CI, whose `length > 200` cutoff is a heuristic - a legitimate `slice(-1)` on a long string would fail the build. That deserves its own PR and its own argument, not a rider on a red-CI fix.

Issue #727, opened 2026-08-12
tauri-plugin-clerk persists the Clerk client token as plaintext JSON, not OS secure storage


adityaharishch:
## Context

While fixing PR #726 (Clerk sessions being dropped on every relaunch — the
offline-cache-write bug, https://github.com/Nipsuli/tauri-plugin-clerk/issues/7),
web research into Clerk's own native-app guidance turned up a separate,
pre-existing gap that PR #726 does not touch.

## The gap

Clerk's documentation for native/mobile apps (Expo, iOS, Android) is explicit
that the persisted client token — the long-lived credential that is "the
source of truth for authentication status" across restarts — belongs in
OS-level secure storage: iOS Keychain Services or Android Keystore-encrypted
storage (Expo's SDK does this via `expo-secure-store`).

`tauri-plugin-clerk` (the crate the tray depends on for Clerk sign-in, now
vendored+patched at `tray/src-tauri/vendor/tauri-plugin-clerk`) instead
persists that same client token via `tauri-plugin-store`, which writes a
**plaintext JSON file** to disk (`clerk-store.json` under the app's data
directory) — see `set_client_authorization_header`/
`get_client_authorization_header` in `src/lib.rs` and `src/tauri_store.rs`.

## Why this matters

The client token is long-lived and is the credential that re-authenticates
the user on every app restart. Storing it in cleartext on disk means anyone
with local filesystem access to the user's machine (malware, another local
user account, a support engineer poking around, a stolen laptop without full
disk encryption) can read it directly, rather than needing to go through the
OS's secure-storage access controls Clerk's own guidance assumes.

## Scope

Not the cause of the logout bug fixed in #726 (that was a separate
deserialization bug in the `set_client` write path) — this is about *where*
the working credential ends up once persistence succeeds. Not fixed as part
of #726 because it's a different code path, a different class of change
(needs a Tauri secure-storage mechanism — e.g. `tauri-plugin-stronghold` or
OS keychain access — wired into the vendored plugin), and wasn't verified
against a live signed-in session in that environment.

## Suggested follow-up

- Confirm the actual risk in Meridian's threat model (the tray already
  stores other sensitive material like the SQLCipher DB key in the OS
  keychain per `db_key.rs` — worth matching that precedent).
- If worth fixing, extend the vendored `tauri-plugin-clerk` fork to swap
  `tauri_store::ClerkTauriStore` for a keychain-backed `Store` impl instead
[...]

PR #734, opened 2026-08-14
docs: correct privacy claims and meet open-source repo standards

Akarsh-Hegde:
Prep for the Product Hunt launch, where the repo becomes a high-traffic landing page. Two kinds of change: correcting claims that were false, and filling in the standard files a public repo is judged on.

## The blocker this started from

The README, `SECURITY.md`, `docs/privacy.md` and `VISION.md` all stated that Meridiana receives nothing and there is no default telemetry. That holds for a source build, where the endpoint and token constants are absent — and is false for the packaged builds users actually install:

| Evidence | |
|---|---|
| `meridian-core/src/settings.rs:358` | `error_reporting_enabled: true` |
| `.github/workflows/release-build.yml:645,652` | CI bakes the central OTLP endpoint and write token |
| `src/observability/otlp_target.rs:49,55` | the `option_env!` constants they land in |
| `tray/src-tauri/src/crash.rs:39,63` | Sentry crash reporting, on the same switch |

Privacy is the product's headline positioning, so this is the claim most likely to be checked and least survivable if wrong.

**Resolution: describe it accurately rather than change the default.** The mechanism is genuinely defensible — error-only, redacted on-device, hostname replaced by a one-way pseudonym, one switch to turn off. The silence about it was the problem, not the behaviour.

## The wizard was the worst instance

`ui/app/setup/steps.tsx` told every first-run user "your data stays local, and Meridian never collects it", under a "Never leaves this device" pillar. Highest-traffic privacy surface in the product, least accurate copy in it.

Both statements are true of the thing the sentence is actually about — the screen and accessibility capture those permissions unlock — so both now say that precisely instead of claiming it universally. No hedging added; the claim is still strong, it is just true.

Settings → Capture & Privacy was already honest, and `AccountSection.tsx` already branches on `supportIdIsAccountScoped` for the alpha account-scoped mode. `whats-new.json` checked and accurate.

## Other inaccuracies found in the same files

- **`docs/privacy.md` was rewritten**, not patched. It described screenpipe as a separate process reading `~/.screenpipe/db.sqlite` (gone since the in-process capture cutover) and invented telemetry contents — "which features you use" — that nothing emits. It also understated the product: the database is SQLCipher-encrypted, which the old text called "plain-text SQLite files".
- **`SETUP.md` pointed users at `~/.meridian/app/.env`** for config. That moved to `~/.meridian/.env` (CHANGELOG:1324). Editing the old path fails silently, which is the worst failure shape a setup doc can have.
- **`CONTRIBUTING.md` told contributors to run bare `cargo test` / `cargo clippy`** — which skips `meridian-core`, `meridian-oauth` and the tray, the exact trap `CLAUDE.md` warns about. Also credited releases to release-please rather than semantic-release, and pointed at an `npm test` script that does not exist in `ui/package.json` (tests run under `bun`).
- **`SECURITY.md`** claimed the screenpipe DB is opened read-only (no such DB) and implied blanket `0600` credential storage without noting the known Clerk plaintext-token exception (#727).
- **README** had an empty `## Data location` heading — a broken section on the page all the traffic lands on — and duplicated Azure DevOps setup that `SETUP.md` already covers in full.
- Tracker counts were inconsistent throughout: three named where five are supported.

## Added for open-source readiness

- **`ARCHITECTURE.md`** — the significant gap. README and CONTRIBUTING both sent newcomers to a 52KB agent-instruction file for architecture. This is a human-oriented map that links into `CLAUDE.md` for depth.
- **`.github/SUPPORT.md`**, **`CODEOWNERS`**, **`dependabot.yml`** (deliberately conservative — grouped, capped, monthly, so it doesn't flood the repo during launch week).
- Discussions link in the issue chooser; an early-access note in the README.
- `docs/images/` with a spec for the two screenshots to add. The README block referencing them is **HTML-commented until the files land**, since GitHub renders a missing image as broken alt text.

## Verification

- All cross-document anchors machine-checked against the target headings.
- Full sweep over every tracked `.md` for remaining false telemetry claims — clean.
- Pre-push suite green: fmt, clippy `--workspace`, `cargo test --workspace`, UI build, UI tests, security audit.

## Not in this PR

The social preview image and the two screenshots need a human. `meridiona.com` very likely carries the same "no telemetry" copy and should be checked before launch — it is outside this repo.


Akarsh-Hegde:
## Update: two more egress paths found, and this PR was denying one of them

Checking `meridiona.com` for the same stale copy turned up a **third and fourth** outbound path that no repo document mentioned — and that this PR's first commit actively denied, having added *"Meridian has no product analytics and does not track usage"* to `SECURITY.md`.

### 1. PostHog product analytics — `tray/src-tauri/src/analytics.rs`

| | |
|---|---|
| Events | `app_installed` (once per device+account), `daily_usage` (once per day) |
| Payload | focus / coding / logged hours, drafts count, app version, OS, channel, per-device UUID |
| Identity | **the PostHog `distinct_id` is the signed-in account email** — identified, not pseudonymous |
| Consent gate | **none.** It does not check `error_reporting_enabled`; the module notes its state file is *"never dashboard-editable, never displayed"* |
| Inactive when | signed out — nothing is sent at all |

### 2. Public counter ping — `tray/src-tauri/src/counter_ping.rs`

A payload-free POST to `meridiona.com/api/counter/increment` per posted worklog, prod channel only, feeding the landing page counter.

### Assessment

Neither carries screen content, window titles, app names, ticket data, or file paths. The four daily numbers are the same ones the Today card already shows the user. **That makes them defensible — being undocumented did not.**

`docs/privacy.md` now has a Product analytics section stating the email-as-`distinct_id` trade-off and the missing opt-out plainly rather than burying them. README, `SECURITY.md` and `VISION.md` no longer claim there are no analytics. `VISION.md`'s principle is reworded to *enumerate* the exceptions, since "no outbound traffic" is not a claim this product can make.

### Follow-up worth its own issue

**Give product analytics an off switch next to the error-reporting one.** An opt-out for one egress path and none for the other is hard to defend, and the docs currently have to say exactly that.

### Also in this push

`docs/images/social-card.png` — the 1280x640 GitHub social preview, cropped inward and alpha-flattened because the source art had rounded corners over transparency that GitHub would composite against its own background. Committed for versioning; GitHub serves it from a manual upload under Settings.

### Separately, outside this repo

`meridiona.com` claims **"100% local"**, **"your work never leaves your device"**, and **"nothing is uploaded, nothing phones home"** — while the same page discloses PostHog. That is self-contradictory on its face and is the most exposed surface of the three. It needs the same pass before launch.

PR #754, opened 2026-08-15
feat(day-summary): tell the user when the daily summary is composed

Akarsh-Hegde:
## What

The end-of-day summary already composes itself at the time the user picked (`worklog_auto_generate_time`), and then said nothing about it. This adds a `day_summary.ready` notification on a successful auto-compose, deep-linked to the summary and expiring at local midnight.

`src/day_summary/auto.rs`'s own header states what that path is for: guaranteeing the summary EXISTS by end of day "for someone who never opens it". That person is exactly the one who never finds out it is there.

## Why this is an Ask, not Status

Worth arguing explicitly, because `docs/notifications.md` warns — correctly — that Status "feels notification-worthy to the person writing it and is noise to everyone else", and cites the `system.health` "Back online." toast that fired 18 times in five weeks for one interaction before being removed.

Three things separate this from that:

- **The user picked the hour.** It fires on a schedule they chose, so it is not unsolicited.
- **It announces an artefact they are meant to read** — the day's adherence, work list and standup lines — not an internal state transition.
- **The path exists for the person who would otherwise never see it.** A summary composed for someone who never learns it is there is work done for nobody.

`GENERIC_LINK` (a single `[Open]`), deliberately over a bespoke category with a Snooze: a snooze needs an idempotent consumer arm and a re-enqueue dedup scheme, which is not worth inventing for a notification that dies at midnight anyway.

## Only a real summary notifies

A fallback carries no prose (the model was down), so announcing it would send the user to a screen with nothing to read. The quieter half is what makes this load-bearing rather than cosmetic: `run` retries a fallback on its next wake, and `enqueue` is idempotent on `dedup_key`, so **a notification sent for the fallback would dedup away the one for the real summary that replaces it an hour later.** The user would be told about the only version worth reading: never.

That is the regression test. Confirmed failing on the exact assertion (`a fallback must not notify`) with the guard stubbed out, rather than assumed to bite.

## The deep link resets the day

Adds `deep_links::SUMMARY` (`/summary`) plus its `navigate` arm — both halves in one change, as the two guard tests require.

The arm calls `setDay(dayString(0))` as well as opening the modal. `DaySummaryOverlay` reads the shell's own `day`, and the push half of that listener targets a dashboard that is **already open**, where the user may have shifted back to an earlier day — opening the summary there would show a different day's review than the toast was about. The vocabulary has no day parameter and the producer is scoped to the current local day, so today is always what it meant.

## The expiry format is not interchangeable

Formatted with a local twin of `daily_plan.rs`'s helper, and deliberately **not** `date::local_day_bounds`. The drain compares `expires_at` as a **string**, and `local_day_bounds` emits the `toISOString()` millis spelling (`…:00.000Z`). In a byte-wise compare `.` sorts before `Z`, so the same instant spelled that way reads as *earlier* than a `T`-separated now — the row would be treated as expired at birth, never delivered, with no error anywhere. `docs/notifications.md`'s testing section documents the same trap for hand-seeded rows.

Kept local rather than shared with `daily_plan.rs`: both producers happen to expire at midnight today, but that is a per-notification policy choice, and hoisting it would let a later change to one silently retime the other. A test pins both the shape and the ordering.

## Also

Corrects the stale note beside `worklog.ready` in the producers table, which still said worklog generation is "on-demand/user-clicked, not scheduled". That stopped being true when `pm_worklog::auto_generate` landed — the same gate this notification now rides, so the new row would otherwise sit directly under a line contradicting it. **No producer added for `worklog.ready`** (out of scope); what it still lacks is a scoping decision, not a schedule.

## Files

| File | Change |
|---|---|
| `src/day_summary/auto.rs` | `notify_ready` producer, body copy, expiry helpers, 4 new tests |
| `meridian-core/src/notifications/mod.rs` | `deep_links::SUMMARY`, `ALL` 7 → 8 |
| `ui/components/timeline/MeridianTimelineShell.tsx` | `case '/summary':` arm |
| `docs/notifications.md` | producers-table row + the `worklog.ready` correction |

No migration, no schema change, no new category to register, no consumer arm.

## Both channels, and why that is safe

`NewNotification::event` defaults to `native,banner`, so this also renders as an in-dashboard banner — kept, because quiet hours suppress the toast but not the banner, and Windows toasts carry no buttons, so the banner is what reaches the user on either path. A persistent surface needs an expiry to avoid becoming an accumulator: verified that `active_banners` filters on `expires_at`, so the midnight expiry clears the row on **both** channels without a click. (`board.hygiene` had no expiry at all, which is why 22 dead banners needed migration 077 to remove them.) A follow-up commit asserts both properties so a later edit cannot drop them silently.

## Verification

- `cargo test --workspace` — 953 + core/oauth/clerk suites, 0 failures
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `bun test` (ui) — 738 pass / 0 fail; `tsc --noEmit` clean
- Full pre-push suite green (fmt, ui build, clippy, ui tests, security audit, cargo test)
- The **bun deep-link guard was proven to bite**: removing the `case '/summary':` arm fails `deep-links.test.ts`. The **Rust guard passes vacuously for this change**, and that is correct — `deep_link_literals_are_all_known` scans only string *literals* near a link anchor, and this producer references `deep_links::SUMMARY`, which is exactly the end state that guard exists to push producers toward.
- The fallback regression test was proven to bite by stubbing out the guard (fails on `a fallback must not notify`).

**Not verifiable from a dev build:** interactive toasts need a packaged `.app` (`tauri-plugin-notifications` hard-fails outside a bundle), so the `[Open]` button and the navigation it triggers were not exercised end-to-end. What is covered here is the enqueue — `dedup_key`, `event_key`, `deep_link`, `category`, `channels` and the presence of `expires_at` are asserted against an in-memory outbox — and the static contract between producer and shell.

**Platform split:** on Windows the row delivers as a plain title+body toast — no `[Open]` button, no action events (plugin gotchas #1/#7) — so the click-through, including the `setDay` reset, is macOS-only. The notification still does its job there (the user is told, and the in-dashboard banner carries the same link), and the outbox row resolves by expiry.


Akarsh-Hegde:
https://github.com/user-attachments/assets/5cc26036-3842-4748-85c8-097a8b71f20d

PR #772, opened 2026-08-16
feat(walkthrough): open on the story, and let the tour say where it is

Akarsh-Hegde:
Ports the tour chrome from the Claude Design mockup [`Onboarding Tour.dc.html`](https://claude.ai/design/p/3031b378-67fe-49ac-ab24-8ed9e0be61c2?file=Onboarding+Tour.dc.html) - full-screen story beats between the hands-on stretches, narration typed rather than printed, and a progress row - onto the walkthrough we already have.

## The design decision

The mockup is a **self-contained demo**: a fake day, clicks that only advance the story, nothing configured. Ours drives the **real** planner, the **real** composer and the **real** worklog dialog, and collects a plan, a tracker and an AI provider on the way through.

The narrative frame is the mockup's best idea and is independent of what sits underneath it. So the chrome changed and nothing under it did.

## Two findings that made this cheap

**Every hex in the mockup is one of our own token values.** `#7C3AED` is `--t-accent`, `#211D3D` is `--t-title`, `#948FB8` is `--t-faint`, `#E4DDF7` is `--t-hair`, and the toolbar gradient is `--toolbar-bg` verbatim - the design tool had flattened our tokens. So the colour port is a substitution, and written as tokens it themes, where the mockup is light-only.

**`_ds_bundle.js` is a compiled mirror of `ui/components/`** (`TaskKey`, `TimeByApp`, `TimeByCategory`, `ModalShell`, `TimelineCard`…). The panel and rows in the mockup are already `TutorialPanel` and `DayTaskColumn`. **No product surface needed rebuilding.**

## What is new

| | |
|---|---|
| `useTypewriter.ts` | One implementation, three surfaces. Honours `prefers-reduced-motion` by not typing at all, and every consumer can land the rest of the line with a click - a line typed at reading speed is a pace, not a gate. |
| `TutorialTakeover.tsx` + `Stage.takeover` | The story beats. Used three times, at points the script already treats as structural: the opening, the seam, the close. |
| `CHAPTERS` + `Stage.chapter` | Seven dots beside Skip, marked at the section banners the script already has. |
| `ClickHint` | The ring has always pulsed while a beat was blocked and nothing said what it wanted, so a waiting beat and a stuck one looked identical. |

`TutorialIntro` is **folded into the takeover** - the opening is simply a takeover with a mark and a `hold`. Two components rendering a full-screen typed card is how the tour ended up with two narration treatments once already.

## Three deliberate departures from the mockup

**The heroes are SF Pro, not Instrument Serif.** That face was retired - `globals.css` says *"There is deliberately no `--font-serif`"* and `layout.tsx` records the setup and uninstall wizards dropping it, along with Plus Jakarta Sans and JetBrains Mono. The register comes from scale, colour and space instead.

**Narration stays anchored beside its target**, not moved to a fixed bottom bar. An instruction read in one place and acted on in another is the failure that layout exists to avoid.

**No "Step 4 of 12".** The tour branches - board or solo, provider connected or not, planner opened or skipped - so a printed denominator is wrong for some users. This file already made that call once, when `INTRO_LINES` dropped "about 2 minutes" because *a promise a first-run screen visibly breaks costs more trust than the number ever bought*. Dots say roughly how much is left and promise nothing.

## Two things found while building it

**`awaiting` became `AwaitKind` (`false | 'click' | 'input'`).** It was a boolean - and "click the highlighted card" over a field the user is meant to type *more words* into is a confidently wrong instruction, which is worse than the silence it replaced. The kind now comes from the primitive that set it rather than being inferred from what happens to be on screen.

I checked whether the hint could appear under a stale caption, and it structurally cannot: it lives inside the anchored bubble, which only renders for a beat with a ring. Three beats wait on a click with the ring deliberately cleared (the subscription gate, Settings' close, the schedule dialog) and all three render the top-bar bubble instead. That ordering is now pinned.

**The chapter had to be reset in `finish`.** The hook is not unmounted between runs, so a user who skipped at the worklog and replayed from Settings would have opened on the first story beat with **five dots already filled**. Planted and confirmed failing.

## Tests

`ui/__tests__/tutorial-chrome.test.ts`, written first - **9 of its 12 original assertions failed on the old tree**; the three that passed are the keep-this-property guards that must pass both ways.

`tutorial-sample-day.test.ts` is updated where it pinned `s.intro`, `INTRO_LINES` and the old closing beat. Every property those encoded survives, re-expressed against the new surfaces - including the three facts the close must carry (who signs off, where the off switch is, and why any of it was built), which now ride the closing takeover's sub.

- `bun test` - 804 pass, 0 fail
- `npx tsc --noEmit` and `npm run build` - clean
- pre-commit and pre-push hooks green

## Not verified on screen

The typing pace and the takeover-to-coach handover are the parts source cannot show. `cd tray && npm run tauri dev`, then Settings → Replay tour (or `?tour=day` for part two only, dev-gated). Worth watching for: the typing not outpacing reading, no flash between a takeover lifting and the coach bubble arriving, and the dots not jumping on a branch.

One judgement call worth a second opinion: the closing takeover's sub now carries three sentences (~30 words) because each of its three facts is separately tested. It reads fine to me on the last screen, where nothing competes with it, but it is the longest copy in the tour.

---

## Merged `pre-main` (44 commits, six in the walkthrough)

Both conflicts were "each side added state", not either side changing its mind, so both were kept. **`awaiting` and `handover` are orthogonal**: `awaiting` now carries *which* input a beat wants (`'click'` vs `'input'`, from this branch), `handover` says whether the tour is pointing at it at all (from #786).

Three things needed real reconciliation:

**The typewriter meets a measured bar.** #775 made the un-anchored narration bar publish its bottom edge as `--mer-tour-say-bottom` so `ModalShell` can pad clear of it - the AI-gate question was being rendered underneath and was unreadable. It measures with a `ResizeObserver`, so it does not *break*, but a caption that types grows line by line, and the modal would have re-padded **in steps while the user read the line causing it**. The anchored bubble had the same problem in the other axis, being placed off its own height.

`CaptionBody` now holds the box open with a hidden copy of the full line and paints the typed slice over it, so the only thing that changes while typing is which characters are inked. Guarded - neither symptom shows in a headless run.

**Two stale anchors in this branch's own guards**, both pointing at markup #786 rewrote: the fence is `dim={spotlightDim}` now, and there are *two* fences for Skip to outrank rather than one.

**A comment that was simply wrong.** This branch claimed three beats wait on the user with no ring; #786 found and fixed **five** - that is the whole subject of that commit. Corrected to its list.

One integration note left in `scriptDay`: the worklog probe's 15s sits between the old 4s ceiling and `HANDOVER_WAIT_MS` (30s), the band #786's comment asks to be decided deliberately rather than inherited. It waits on a health read, not a person, so the fence should stay up - which is what the value already does, now said out loud.

**872 tests pass**, typecheck and build clean. `mergeable` is back to `MERGEABLE`.

One thing checked and dismissed: `c7266e25 "vendor the font"` vendors **JetBrains Mono** (build-time fetch → self-hosted), and `d8e128f1 "gate the Instrument import"` is a Rust import in `sys.rs`. Neither touches Instrument Serif - `layout.tsx` still records it as retired, so the SF Pro decision above stands unchanged.


Akarsh-Hegde:
Not what I wanted

PR #779, opened 2026-08-16
fix: address the CodeRabbit review on the 1.87 release PR

Akarsh-Hegde:
Review fixes for the 1.87.0 release PR (#777), plus the What's New backfill flagged there. Merging this into `pre-main` updates #777 in place.

## Two real defects, both in the install/watchdog interlock #771 added

These are the ones worth reading. Both shipped green - neither is reachable from a macOS test run.

### `refresh.rs` deleted the recovery attempt instead of deferring it

The staging gate suppressed `attempt_restart` and `notify_down` but left the failure **counted**. Every action keys off `consecutive_health_failures == 2`, an edge consumed the moment it is passed: a tick that increments 1 → 2 and has its outputs masked leaves the counter at 2, so the next failure reaches 3 and the edge never returns.

That is only academic if installs always work. They do not - `register_service` can report success on Windows while `/Run` **and** the fallback spawn quietly failed, which leaves a permanently down daemon that this loop had already spent its one recovery attempt on.

Now the suppression skips the **count**, so the episode is deferred rather than deleted: when the guard drops, a daemon that is genuinely down fails one tick, then another, and recovery fires on its normal edge ~60 s late. A *healthy* tick still takes the normal path mid-install - it must, or a `tray.daemon_quiet` notice raised before the install is stranded with nothing left to observe the recovery that clears it.

The gate also moved into `decide_health_notice`, which turned a source-scan into three driven tests.

### The watchdog's `Start` arm acted on a stale flag

`decide` is pure, so it judges a snapshot taken at the top of the tick. Between that and the `Start` arm the loop awaits a socket probe and - on exactly the path that reaches `Start` - forks a subprocess for the liveness check. An install beginning in that window started the daemon it had just killed.

Re-read at the last instant rather than taking `LIFECYCLE`: the mutex would serialise correctly but block the watchdog for the length of an install and queue a 5 s-budgeted `stop_for_quit` behind it - the tradeoff `begin_staging` already documents. Not airtight by design; `stop_running_daemon_before_stage`'s re-kill loop is the backstop for the residue, and the comment says so.

## The rest

| | |
|---|---|
| `toast_actions` | XML 1.0 forbids most C0 controls **outright**, so a stray `\u{1}` from a pasted task title made `LoadXml` reject the document and the toast silently never appeared - the same invisible failure `escape_xml` already guards for `&`, from the half escaping cannot reach. Dropped rather than escaped, because `&#1;` is equally illegal. Tab/LF/CR kept (legal, and the daemon lays some bodies out deliberately). |
| `setup` | A saved position on the Windows Alerts step stopped resolving once notifications were granted: `buildSteps` drops that step, and `findIndex` cannot tell "dropped from this run" from "never existed". The wizard restarted at Welcome at the exact moment the user did the thing it asked for. `resumeIndex` resumes past a known-but-dropped step; an unknown id keeps the old behaviour exactly. |
| `TutorialOverlay` | The published bar height went stale on a window resize (it re-wraps below its max-width) and when a beat added `choices`, neither of which changes a dependency. `ResizeObserver` covers both plus the initial read. Stale is worse than absent here - `ModalShell` holds a gap to whatever was last published. |
| `useTutorial` | `Stage.type` cast `waitForElement`'s result unchecked, so a mistargeted selector threw a `TypeError` out of `runScript` and the catch there tore the **whole walkthrough** down. Narrowed with `instanceof`, restoring the return-`false` contract every sibling primitive keeps. |
| `backend_install` | A span per `taskkill` with ERROR status on failure. The ship leg is error-only, so a WARN with no ERROR span arrives stripped of what you would need to act on it. |
| `notifications` | `apply_response` is where macOS and Windows converge and where a lost answer has to be diagnosed from telemetry alone; spanned with a `source` discriminator. `has_text`, never the reply - an inline reply is the user's own writing. |
| `app.js` | `async`/`await` for the listener registration. The nested catch is load-bearing: it keeps "expected non-failure" (dev run, or Windows/WinRT) distinct from a real one. |
| test fix | `worklog-needs-ai` asserted `not.toContain('if (isDemo)')` over a slice bounded by a heuristic. If the marker missed, the slice ran long and the negative assertion passed vacuously - the exact regression that file exists to catch. The end marker is now asserted before the slice is read. |

## Also here: the What's New backfill

`whats-new.json` was at 1.84.0 - **1.85.0 and 1.86.0 both shipped without an entry**, so anyone updating from 1.84.0 opened a modal that stopped two releases short of what they had installed. Written from each CHANGELOG section in user terms; commit-level entries with no user-visible effect are left out rather than translated. 1.87.0's entry is in the first commit of this branch.

## Declined, with reasons

- **Pin every GitHub Action to a commit SHA** (2 findings, 20 sites). Legitimate supply-chain hardening, but repo-wide work on workflows this release does not touch - doing it inside a release PR means the release rides on 20 unrelated edits to the pipeline that builds it. Worth its own PR.
- **Remove the 1.86.0 CHANGELOG section.** It is already on `main`. GitHub picked the older of two merge bases for #777, so that section shows as added when it is not - see the callout at the top of #777.
- **Rename `BODY_TOP` / `TOUR_BODY_PAD` to camelCase.** Module-level consts in `ui/components` are SCREAMING_CASE throughout (`PROBE_MIN_VISIBLE_MS`, `SHELL_LESS_ROUTES`, `OAUTH_DEADLINE_MS`); renaming two would make the file inconsistent with its neighbours rather than more consistent.

## Verification

`cargo test --workspace` 954 pass · `cargo clippy --workspace --all-targets -D warnings` clean · `bun test` 807 pass.

Each fix was mutation-tested - the fix reverted, the test confirmed red, then restored. The two Rust interlock fixes and the resume fix were verified this way individually.

**Still not verified:** the Windows halves compile only on Windows CI. `kill_pids`'s new span was type-checked against `x86_64-pc-windows-msvc` in isolation, but the WinRT and toast paths get their first real compile on this PR.

## Summary by CodeRabbit

- **New Features**
  - Added release notes covering recent improvements to notifications, onboarding, privacy, reliability, and work logs.
  - Setup now resumes correctly when platform-specific steps change, with clearer “Open Dashboard” completion action.
  - Tutorial overlays adapt smoothly to changing narration and choice content.
- **Bug Fixes**
  - Prevented health checks and service starts from interfering with installations in progress.
  - Improved notification action handling and error recovery.
  - Prevented invalid characters and non-form tutorial targets from causing failures.

### DIFF
--- a/tray/src-tauri/resources/whats-new.json
+++ b/tray/src-tauri/resources/whats-new.json
@@ -23,6 +23,44 @@
       ]
     },
     {
+      "version": "1.86.0",
+      "date": "2026-08-16",
+      "highlights": [
+        "Setup now picks up where you left off. Closing the wizard part-way through used to mean starting again from the first card.",
+        "Meridian tells you when your daily summary is ready, instead of leaving you to go and look.",
+        "The dashboard opens filling the screen rather than in a small window you have to resize every time."
+      ],
+      "fixes": [
+        "Signing in sticks. Your session was being quietly dropped every time Meridian restarted, so you had to sign in again each launch.",
+        "Settings no longer crashes on builds where sign-in is switched off.",
+        "The vendored sign-in component was writing your session token into Meridian's own logs. It no longer does, and that logging is now switched off by default regardless.",
+        "The wizard's privacy wording now matches what the app actually does. It said Meridian never collects anything - true of a build you compile yourself, but not of the installed app, which sends redacted error reports and product analytics you can switch off. The download page, the privacy document and the wizard all say the same thing now.",
+        "The Notifications permission card in setup now shows the real Notifications pane rather than a stand-in, and the permission cards are highlighted one at a time instead of all at once.",
+        "The setup wizard no longer renders black when the window is in full screen.",
+        "What's New no longer opens on top of you the moment first-run setup finishes.",
+        "The walkthrough survives a stray click, and the tray's own pop-ups no longer interrupt it half way through.",
+        "Connecting GitHub no longer blames your token when GitHub simply cut the response short. Meridian retries instead."
+      ]
+    },
+    {
+      "version": "1.85.0",
+      "date": "2026-08-13",
+      "highlights": [
+        "Meridian shows you around the first time you open it. A guided walkthrough covers planning a day, watching your work get picked up, and having a work log written for you - starting on your own day, then on an example day so you can see a full one end to end.",
+        "Yesterday's unfinished work now moves into today on its own, so you no longer start the morning rebuilding a list you already made.",
+        "Meridian now says when a draft work log has fallen behind the work it describes, rather than letting you post something that stopped being true an hour ago."
+      ],
+      "fixes": [
+        "Uninstalling now removes your data. Some of it survived every uninstall route the app offered, which is not what uninstall means.",
+        "An update already being installed is no longer reported as 'Update failed', and an install that genuinely fails now tells every screen, so nothing is left sitting on a spinner.",
+        "Restarting the background service no longer runs into the limit macOS puts on how often a service may restart, which used to leave it stopped for minutes at a time.",
+        "A work-log draft that the AI failed to produce now points you at the reason rather than ending the flow with nothing.",
+        "Every notification now opens the thing it is about. Some of them went nowhere. The board-hygiene digest, which fired at midnight and which nobody ever acted on, is gone.",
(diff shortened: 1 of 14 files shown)


Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

PR #783, opened 2026-08-16
fix(summary): re-read the worklog badges after a draft is acted on

Akarsh-Hegde:
## The bug

Post a worklog from a Today's Plan row, come back to the list, and the row you just
filed still reads **DRAFT READY TO POST** - in the loud accent treatment reserved for
"this needs you", with the tinted surface and the `›` chevron that go with it. Post
three and the list still looks like a queue of three outstanding drafts.

## What was actually wrong

`DaySummaryOverlay` read `get_day_draft_states` **exactly once**, inside the
day-change effect, and never again for the life of the overlay:

```ts
useEffect(() => {
  Promise.allSettled([ ..., load<DayDraftState[]>(API, 'get_day_draft_states', { day }) ])
    .then(([s, d, t, w]) => { ...; setDrafts(...) })
}, [day, generate])
```

But every path that **changes** a draft's state - approve, post, dismiss, retarget -
lives in `SummaryTaskView`, which the summary opens over itself and then returns from.
The overlay is not remounted on that round trip, so the badge map stayed frozen at
whatever it was when the overlay opened.

Closing and reopening the summary fixed it, which is the giveaway: a snapshot that is
never refreshed.

## Confirming it was the read, not the write

Checked against the live DB (read-only) while the screen was showing the wrong thing:

| task | state | drafted_minutes | updated |
|---|---|---|---|
| T1 | **posted** | 144 | 18:07:56 |
| T2 | **posted** | 94 | 18:07:37 |
| T3 | drafted | 109 | 12:38:37 |
| T6 | **posted** | 33 | 18:11:02 |
| T7 | **posted** | 61 | 18:08:19 |

`day_task_worklogs` had flipped to `posted` at the moment each was posted. The write
path is fine. The screen simply never asked again.

## The fix

**`refreshDrafts`** - the same read, extracted and reusable. Two details it carries:

- **Guarded on `dayRef`**, like `generate`/`generateNow` already are. The user can page
  days while it is in flight, and painting one day's badges under another day's rows is
  worse than leaving the stale ones.
- **Keeps the previous map on a failed read.** On a refresh there is already an answer
  on screen and last-known beats none. That is the *opposite* of the first-paint rule -
  where a failure must leave the rows badgeless rather than show a stale day - which is
  why the day effect deliberately keeps its own inline read instead of calling this. The
  comment there says so, so it does not read as an oversight later.

**`closeTask`** - the single exit from the task view. There are **two** ways out, and
this is the part a one-line `onBack` fix would have missed:

```ts
// before
if (selected) { setSelected(null); return }   // Escape, in the overlay's own key listener
onBack={() => setSelected(null)}              // the Back button
```

Escape is a first-class way to leave the task view - post, then dismiss with the
keyboard - and it nulled `selected` directly. Both now route through `closeTask`.

## Tests

`ui/__tests__/summary-draft-badges-refresh.test.ts` - written first, **failed 5/5**
before the fix. Source-scanned, like its sibling `summary-plan-row-drafts.test.ts`:
there is no React render harness in this repo, and the wiring is the half that was
wrong. It pins the reusable read, the `dayRef` guard, and *both* exit paths - including
`expect(escapeHandler()).not.toContain('setSelected(null)')`, so the keyboard path
cannot quietly regress back to the direct call.

- `bun test` - **831 pass / 0 fail** (58 files)
- `tsc --noEmit` - clean

## Scope

One component and one new test file. No Rust, no schema, no reader changes - the
backend was already returning the right answer.

## Summary by CodeRabbit

- **Bug Fixes**
  - Draft badges now refresh when closing a task and returning to the day summary.
  - Back-button and Escape-key navigation now provide consistent task-closing behavior.
  - Initial draft-state loading and subsequent refreshes are handled consistently.

### DIFF
--- a/ui/__tests__/summary-draft-badges-refresh.test.ts
+++ b/ui/__tests__/summary-draft-badges-refresh.test.ts
@@ -1,0 +1,97 @@
+//ambient dev tool that watches what you do and updates your PM tickets automatically, boosting developer productivity
+//
+// Guards that the daily summary's worklog badges are re-read after the user acts on
+// one, rather than staying frozen at whatever they were when the overlay opened.
+//
+// THE BUG: `DaySummaryOverlay` read `get_day_draft_states` exactly once, inside the
+// day-change effect, and never again for the life of the overlay. Every path that
+// CHANGES a draft's state - approve, post, dismiss, retarget - lives in
+// `SummaryTaskView`, which the summary opens over itself and then returns from. So
+// the user opened a row, posted the update, came back, and the row they had just
+// filed still read DRAFT READY TO POST, in the loud accent treatment reserved for
+// "this needs you". Posting three rows left three rows still shouting.
+//
+// Measured on a real DB while it was happening: `day_task_worklogs` had flipped to
+// `posted` for the rows in question. The write was never the problem - the screen
+// simply never asked again. And because the tint and the `›` chevron key off the
+// same badge, the whole list still looked like a queue of outstanding work.
+//
+// TWO EXIT PATHS, not one. The task view is left by its own Back button AND by
+// Escape (handled in the overlay's own key listener, which nulls `selected`
+// directly). A refresh hung off `onBack` alone would leave the Escape path exactly
+// as broken as before, which is why both must go through one helper.
+//
+// No React render harness in this repo (see summary-plan-row-drafts / plan-store),
+// so this is source-scanned: the assertions are about the wiring, which is the half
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

PR #784, opened 2026-08-16
fix(tray): fail closed when the setup window's state cannot be read

Akarsh-Hegde:
Closes a CodeRabbit finding from #777 (*"Fail closed when window-state queries fail"*, Major) that was left unresolved when that PR was cut. Raising it here rather than letting it lapse.

## The problem

`resize_setup_window` skips the resize while the wizard window is full-screen or maximized, because **resizing in either state is the hazard** - on Win32, `SetWindowPos` on a `WS_MAXIMIZE` window clears the maximized state, so the window visibly un-maximized itself when the user pressed Continue mid-wizard.

Both guards read `.unwrap_or(false)`:

```rust
if win.is_fullscreen().unwrap_or(false) { ... return Ok(()); }
if win.is_maximized().unwrap_or(false) { ... return Ok(()); }
```

So a query that **errored** was treated as "neither full-screen nor maximized" and the resize went ahead - performing the exact action the guards exist to prevent, in the one situation where we have no idea what the window is doing.

## Why fail-closed is free here

The two wrong guesses are not symmetric:

- Guessing **"not maximized"** when it is → the bug, in full.
- Guessing **"maximized"** when it isn't → nothing. The card's own `transform: scale(...)` in `ui/app/setup/page.tsx` already grows it to fill whatever size the window actually is. That is *why* both guards can `return Ok(())` instead of erroring.

An unknown window state is not a licence to resize it. Both guards now read `.unwrap_or(true)`, and the debug lines say "or unreadable" so the skipped-because-errored case is distinguishable in a trace from the skipped-because-maximized one.

## Test

Source-scanned - the guards need a live window, and the defect is in *which way the `Result` collapses*, which no unit test can reach. Two things keep the scan honest:

- **Truncated at `#[cfg(test)]`**, so it cannot match its own literals and pass forever. (That trap has been hit in this repo twice.)
- **Bounded at the function's closing brace**, so it does not drag in the next function - `dismiss_popover`'s `is_visible().unwrap_or(false)` gates only a debug log (the `hide()` runs either way) and is not the same hazard. Without the bound the test failed on innocent code.

Verified in **both** directions: red before the change, and red again when `is_maximized` is mutated back to `unwrap_or(false)`.

## Gates

- `cargo fmt --all --check` - clean
- `cargo clippy --workspace --all-targets -- -D warnings` - clean
- `cargo test --workspace` - **954 pass / 0 fail**

Two lines of behaviour change, one new test.

## One follow-on commit

`reopen.rs`'s `window_resizes_check_for_full_screen_first` pinned the whole expression - `win.is_fullscreen().unwrap_or(false)` - so fixing the unwrap broke a test with no stake in the question. Caught by the pre-push hook.

That test exists to protect that something **asks** before it resizes (the dashboard copies the pattern, which is why it lives in `reopen.rs`). *Which way the answer collapses* is the new test's claim. The needle is now the query alone, so the two tests hold two separate claims and neither fails on the other's change. Verified it still fails when the guard is removed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window resizing safeguards when fullscreen or maximized state checks fail.
  * Prevented unintended resizing when the app cannot reliably determine the current window state.

### DIFF
--- a/tray/src-tauri/src/commands/system.rs
+++ b/tray/src-tauri/src/commands/system.rs
@@ -151,8 +151,15 @@ pub async fn resize_setup_window(
     // There's nothing to fix here anyway while full-screen — the card's own
     // `transform: scale(...)` in page.tsx already grows it to fill whatever
     // size the window actually is.
-    if win.is_fullscreen().unwrap_or(false) {
-        tracing::debug!("setup window is full-screen - resize skipped");
+    //
+    // `unwrap_or(TRUE)`, deliberately: the guards fail CLOSED. An errored query
+    // tells us nothing about the window, and the two outcomes are not
+    // symmetric - guessing "not full-screen" performs the very resize the
+    // guard exists to prevent, while guessing "full-screen" skips a resize
+    // that was never needed anyway (the card's `transform: scale(...)` fills
+    // whatever size the window is). Same reasoning on both guards below.
+    if win.is_fullscreen().unwrap_or(true) {
+        tracing::debug!("setup window is full-screen or unreadable - resize skipped");
         return Ok(());
     }
     // Same reasoning for a MAXIMIZED window, which is the shape the hazard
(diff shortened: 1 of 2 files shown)


Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

Akarsh-Hegde:
@coderabbitai review

PR #828, opened 2026-08-20
fix(tray): repair autostart on every launch, and relaunch each morning

adityaharishch:
## Why

Users reported Meridian not starting itself. Capture runs **in-process in the tray**, so a tray that does not come back records nothing at all — and in PostHog that is indistinguishable from churn, because `app_active` can only fire from a running tray. Autostart reliability and the returning-user/churn metrics are the same problem.

## Three defects, one reproduced live

1. **`autostart.rs` registered once ever** and never asked the OS whether the login item still existed. The `~/.meridian/autostart_configured` marker survives the app being trashed, reinstalled or moved, so it routinely diverged from reality with nothing able to notice. Measured on a dev Mac: **marker present, no Meridian login item at all** in `~/Library/LaunchAgents/`. That install could never autostart again.
2. **`tauri-plugin-autostart` named its plist after `productName`** (`Meridian.plist`), not `com.meridiona.*`. `src/uninstall.rs` asserted in a comment that its `com.meridiona.*` sweep covered it. It did not — every uninstall left the login item behind, pointing at a trashed app.
3. **Path drift was undetectable.** `path_is_stable` only rejects `/Volumes` and AppTranslocation, so a first launch from `~/Downloads` pinned that path; moving the app to `/Applications` later killed autostart permanently.

## What changed

**The plugin is removed, not worked around.** It can only emit `RunAtLoad` (verified against `auto-launch-0.5.0/src/macos.rs`), so it cannot express a morning relaunch. No new dependency replaces it — `autostart.rs` reuses `backend_install`'s `launchctl`/`schtasks` shell-outs.

| | mechanism |
|---|---|
| macOS | `~/Library/LaunchAgents/com.meridiona.tray.plist`: `RunAtLoad` + `StartCalendarInterval` 09:00, **no `KeepAlive`** so Quit sticks for the day |
| Windows | ONE scheduled task with a `LogonTrigger` **and** a daily `CalendarTrigger`, from XML, `MultipleInstancesPolicy: IgnoreNew` |

- **Verify-and-repair on every launch**, with the repair reason logged at WARN so the field failure rate is finally visible.
- **macOS is deliberately not `bootstrap`ped in-session**: `bootstrap` honours `RunAtLoad` and would start a second tray against the same SQLite file. The job goes live at the next login. Bounded cost: in the one session where the plist is first written the calendar trigger is not yet armed.
- **Windows needs XML** because the command form takes one `/SC`; two tasks would be two instance policies and the 09:00 one would stack a second tray. Falls back to a logon-only task, then a Startup-folder VBScript, each logged at WARN. Also overrides Task Scheduler's battery defaults, which would otherwise make Meridian desktop-only.
- **Unattended launches stay windowless.** The job passes `--autostart`; `whats_new_auto_open` defers rather than opening a changelog on a machine nobody touched. (`plan_auto_open` is left alone — "first activity of the day" is its documented purpose.)
- **Both platform modules compile on every target**, so the plist/XML generators are linted and tested by macOS/Linux CI. Otherwise the Windows path would have had zero coverage until a user reported it.
- **The opt-out has a home.** Repair now overrides the OS's own login-item toggle, so `autostart_enabled` (settings.json, opt-out, default true) plus a Startup switch in Settings → Capture & Privacy is the supported no. Writing it registers or **removes** the job immediately — declining to write is not enough, since the job already on disk is what the OS acts on.
- **Uninstall fixed**: the new label is swept by the existing `com.meridiona.*` glob; the legacy `Meridian.plist`, the legacy HKCU `Run` value, the tray task and both Startup launchers are removed by name.

## Metrics

`daily_usage` now carries `health_autostart_registered`, `health_autostart_path_ok`, `health_autostart_repaired`, `health_autostart_action`, `launched_by_autostart`, `tray_uptime_s`. That is what separates **churn from a dead login item** — today they are one number. Installs (`app_installed`) and returning users (`app_active`) already existed; they just were not trustworthy while the tray could silently stop coming back.

Known gap, unchanged and flagged rather than papered over: every analytics event is gated on a signed-in account email, so an install that never signs in is invisible.

## Tests

27 new unit tests. The pure builders and the repair decision are covered in both directions (absent / path drift / missing morning trigger / already correct / user opt-out / transient path), plus UTF-16 encode+decode for `schtasks`, and a source-scan pinning that the unattended gate precedes the window opener.

`cargo fmt` · `cargo clippy --workspace --all-targets -D warnings` · `cargo test --workspace` · `bun test` (894) · `tsc --noEmit` · `npm run build` — all green via the pre-push gate.

## Still needs a real machine

The Windows path cannot be exercised here. Needs a staging check of the ONLOGON + DAILY task, the fallback chain, and a reboot; on macOS, a reboot plus a temporarily-advanced calendar entry to watch the tray return with no window.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -554,17 +554,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "auto-launch"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1f012b8cc0c850f34117ec8252a44418f2e34a2cf501de89e29b241ae5f79471"
-dependencies = [
- "dirs 4.0.0",
- "thiserror 1.0.69",
- "winreg 0.10.1",
-]
-
-[[package]]
 name = "autocfg"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2273,15 +2262,6 @@ dependencies = [
 
 [[package]]
 name = "dirs"
-version = "4.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ca3aa72a6f96ea37bbc5aa912f6788242832f75369bdfdadcb0e38423f100059"
-dependencies = [
- "dirs-sys 0.3.7",
-]
-
-[[package]]
-name = "dirs"
 version = "5.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "44c45a9d03d6676652bcb5e724c7e988de1acad23a711b5217ab9cbecbec2225"
@@ -2296,17 +2276,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3e8aa94d75141228480295a7d0e7feb620b1a5ad9f12bc40be62411e38cce4e"
 dependencies = [
  "dirs-sys 0.5.0",
-]
-
-[[package]]
-name = "dirs-sys"
-version = "0.3.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b1d1d91c932ef41c0f2663aa8b0ca0342d444d842c06914aa0a7e352d0bada6"
-dependencies = [
- "libc",
- "redox_users 0.4.6",
- "winapi",
 ]
 
 [[package]]
@@ -2634,7 +2603,7 @@ dependencies = [
  "rustc_version",
  "toml 1.1.2+spec-1.1.0",
  "vswhom",
- "winreg 0.55.0",
+ "winreg",
 ]
 
 [[package]]
@@ -5277,7 +5246,6 @@ dependencies = [
  "sysinfo",
  "tauri",
  "tauri-build",
- "tauri-plugin-autostart",
  "tauri-plugin-clerk",
  "tauri-plugin-http",
  "tauri-plugin-notifications",
@@ -10108,20 +10076,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "tauri-plugin-autostart"
-version = "2.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "459383cebc193cdd03d1ba4acc40f2c408a7abce419d64bdcd2d745bc2886f70"
-dependencies = [
- "auto-launch",
- "serde",
- "serde_json",
- "tauri",
- "tauri-plugin",
- "thiserror 2.0.18",
-]
-
-[[package]]
 name = "tauri-plugin-clerk"
 version = "0.1.1"
 dependencies = [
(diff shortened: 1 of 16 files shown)


adityaharishch:
## Upgrade-safety audit — three self-inflicted regressions found and fixed (77685824)

Went back over the diff specifically asking "what breaks for someone who already has Meridian installed". Three ways it could have, none of which a fresh-install test would have caught.

**1. `launchctl bootout` kills the tray.** `bootout` terminates a job's processes — and on an existing install the tray *is* that job's process, because the plugin's login item is what started it. `migrate_off_plugin` would have killed the tray seconds after the update landed, with no `KeepAlive` to bring it back. The app just vanishes at login. Deleting the plist alone is safe and sufficient: the plugin's plist carries only `RunAtLoad`, which fires at session start and never again.

Same hazard, more obviously wrong, in `unregister` — untick "Start Meridian automatically" and the app you're looking at quits. Accepted cost of not booting out: our plist *does* carry a morning trigger and stays loaded until logout, so turning autostart off and then quitting later the same day could still relaunch once at 09:00. That beats quitting the app out from under someone changing a preference.

**2. The same thing in `src/uninstall.rs`, where it was previously invisible.** The tray's login item used to be labelled `Meridian`, which the plist loop's `com.meridiona.*` glob never matched. Now that `autostart.rs` uses the conventional label it *does* match — and the in-app wizard shells out to `meridian uninstall` and **waits on its output**. Booting that label out kills the wizard mid-run: window gone, itemized result never rendered, and the permissions guidance that is the entire reason the wizard exists never appears. Tray labels are now deleted without a bootout. The daemon still gets one — it's headless, nothing waits on it, and it must actually stop.

**3. Windows `schtasks /Delete` terminates running instances**, so the same self-kill applied to the Settings toggle. `unregister` now uses `/Change /DISABLE`. That made the switch one-way, so a disabled task is now treated as absent by both the repair path and the analytics probe — otherwise turning autostart back on would report `already_correct` over a permanently disabled task, silently. Bonus: it also self-heals a task someone disabled by hand in Task Scheduler.

### Two correctness fixes while in there

- **`autostart::status()` is now async.** It was shelling out to `schtasks` *blocking*, from `analytics::health::snapshot`, which runs on the tray poll loop — the loop that also drives the tray icon, the health banner and the notification drain. A stalled runtime worker isn't a fair trade for a telemetry field. macOS uses `tokio::fs` too, and Windows now reuses `read_registration` so the read and repair paths can't disagree about what "registered" means.
- **`mark_process_start()` moved after `crash::init_client()`**, which is documented as having to be first: the minidump reporter relaunches this exe in a mode that short-circuits before any app work, and a process that exists only to upload a crash dump has no uptime worth recording.

### Verified, not changed

- `RuntimeSettings` carries a **struct-level `#[serde(default)]`** — an existing `settings.json` without `autostart_enabled` loads as `true` rather than failing to deserialize. This was the one that could have bricked every install.
- `read_settings_value` serializes the defaults **before** overlaying the file, so the new Settings switch renders correctly (checked) for an install whose `settings.json` predates the key — rather than rendering unchecked and telling the user autostart is off when it isn't.
- No JS or capability still references the removed plugin. `tauri`'s build script validates `capabilities/*.json` against installed plugins, so a stale grant would have failed the build — it compiles, which is the proof.
- Nothing here touches `meridian.db`. No migration, no rollback risk.

4 new tests; 378 passing in the tray crate.

PR #849, opened 2026-08-24
wip: worklog-time onboarding dialog not saving via the tour (root cause not found)

Akarsh-Hegde:
## Status: investigation parked, not a fix

Draft on purpose — this does NOT fix the bug. It's the diagnostic instrumentation and everything confirmed/ruled out so far, kept together so a future session (or a fresh rewrite of this beat) doesn't have to re-derive it. See the commit message for the full writeup: confirmed facts, what's been ruled out, and suggested next steps.

**The bug:** picking a time in the onboarding tour's "when do you want these ready?" step and clicking "Turn on" does not save — Settings → Worklogs still shows the old default afterward. Reproduces on every attempt.

**What's NOT the cause** (confirmed via instrumentation): the backend save path (manual Settings → Worklogs → Save works every time), a disabled button being clicked, duplicate dialog instances, capability/ACL denial, and a ~900ms fence-race window that looked promising but reordering around it didn't change the outcome.

**What's genuinely still unexplained:** the tour's click-wait resolves as if a click happened, but zero click of any kind (button, backdrop, or otherwise) is observed in that window by a document-level capture-phase listener.

All temporary diagnostic code is tagged `TEMPORARY DIAGNOSTIC` for easy removal. None of it should ship as-is.


Akarsh-Hegde:
Superseded by #852, which sidesteps the click-detection mystery instead of resolving it — verified working live. Leaving this draft open for the diagnostic trail.

Akarsh-Hegde:
Closing without merging — this is temporary diagnostic instrumentation only, never meant to ship. Its purpose (documenting the investigation) is preserved in #852's description/commit message. Branch stays around for reference if anyone needs to see the raw traces.

Issue #867, opened 2026-08-24
Redaction strips the exact fields three WARN sites exist to carry


Akarsh-Hegde:
Found while debugging a fleet incident from central OO on 2026-08-24. Several `warn!` sites were authored specifically to carry a diagnostic, the author reasoned about it in a comment, and it never reached the backend.

> **Corrected after measuring.** This issue originally claimed `cli_exec.rs`'s `bin`, `cwd` **and `timeout_s`** were all dropped by `SAFE_STRING_KEYS`. That was wrong about `timeout_s` and wrong about the mechanism. `timeout_s` is dropped for a completely different reason, described below as (1), and numeric attributes are otherwise kept unconditionally. The table below is the measured version.

| site | field | why it doesn't ship |
|---|---|---|
| `cli_exec.rs:75` | `timeout_s` | **`u64` stringified by the otel bridge** (1) |
| `cli_exec.rs:75` | `bin`, `cwd` | not allowlisted - both are home-dir paths (2) |
| `clerk/lib.rs:203` | `path` | key denied for an unrelated meaning (3) |
| `main.rs`, `repair_boot.rs` | `error` | ships, but carries only the outer `.context()` (4) |

## 1. A `u64` field is not shipped as a number

`tracing-opentelemetry` 0.28's `Visit` impl covers `record_bool`/`record_f64`/`record_i64`/`record_str`/`record_debug` and **not `record_u64`**, so `tracing::field::Visit`'s default applies and forwards to `record_debug`, which emits a **`StringValue`**. It then misses the allowlist - nobody lists `timeout_s` as a *string* key.

Same field name, same severity, same allowlist:

| site | expression | type | shipped |
|---|---|---|---|
| `cli_exec.rs` | `timeout.as_secs()` | `u64` | **`null`** on all 25 records |
| `distiller/mod.rs` | `EMBED_TIMEOUT.as_secs_f64()` | `f64` | `210.0` on 3,329 records |

Blast radius is every `u64`/`usize` field on a WARN/ERROR site.

## 2. `bin` / `cwd`

Correctly dropped - they are home-directory paths. But `meridian_bin()`'s doc explains the failure `bin` exists to answer: a release tray calls the **installed** `meridian`, which can be older than the tray asking it for a subcommand. 65 shipped `cli_exec.rs` records answer nothing. What's needed is a coarse identity naming our component, not the user's disk.

## 3. `path` is denied for a different meaning of the word

[...]

Issue #866, opened 2026-08-24
plan-task-draft reports "timed out" instead of the provider error it already has


Akarsh-Hegde:
## What the user sees

"plan-task-draft timed out" in the Daily plan composer, with a **Try again** that fails identically. Observed on `mac_31187a3832039bc9` (1.90.0-staging.5), three times in nine minutes.

## What is actually wrong

The tray gives the CLI a **150s** budget (`LLM_TIMEOUT`, `commands/plan_tasks.rs:39`) and the CLI has **no timeout of its own below it**. When the user's provider hangs rather than refusing, `meridian plan-task-draft` blocks past the budget, `cli_exec.rs`'s `tokio::time::timeout` kills it, and the user is told our subprocess timed out.

**The mechanism is a hang, not a rejection.** In the window containing all three timeouts (`1787553100`-`1787553760`) the only LLM-related lines are tray-side and both `error = null`:

```
llm: connectivity probe timed out - retrying once before reporting failure
in-use LLM provider is unavailable
```

The matching error text elsewhere in the 48h window is `custom provider unreachable: error sending request for url (...) operation timed out`, which is what the code should have surfaced.

The same host also logs `custom provider rejected the API key (401 Unauthorized) - check it in Settings` 40 times across the 48h window - a **separate concurrent symptom**, and specifically *not* the cause of these three events: a 401 returns fast and cannot burn 150s.

## Why the fast-refuse gate does not save it

`resolver.rs:495`'s health gate is supposed to refuse without dialling out, but reaching a verdict calls `detect::in_use_provider_health`, which performs its own network probe **plus one retry** (`llm: connectivity probe timed out - retrying once`, logged between timeouts #2 and #3). On a hanging endpoint the refusal path is itself slow.

## Ruled out: this is not the DB

The same machine was carrying `code: 11 database disk image is malformed` at the time, which makes "timed out" look like a DB stall. It is not:

- pool acquire timeout is **10s** (`meridian-core/src/db_crypto.rs:103`), so a DB-blocked CLI exits non-zero in ~10s and the tray logs `non-zero`, not `timed out`. It ran the full 150s.
- `plan-task-draft` **never opens meridian.db**: `cli::run_draft` -> `draft::draft` -> `llm::complete`, and there is no pool or `sqlx` reference in `resolver.rs`, `openai_compat.rs`, `runtime_health.rs` or `detect.rs`.

## Suggested fix

[...]

PR #868, opened 2026-08-24
fix(clerk): default every required key on the sign_in/sign_up scratch objects

Akarsh-Hegde:
Fixes the fourth production failure in a row on the same Clerk object. Found in central OO on `mac_31187a3832039bc9` (1.90.0-staging.5), firing **every ~43 seconds**, 240 occurrences:

```
clerk: auth event JSON still failed to deserialize after field backfill
error = missing field `id`
```

Consequence is in the log line itself: `set_client` never persists, so the offline session cache stays empty and **a cold start without network shows the user signed out** - on a machine whose user was signed in the whole time.

## Root cause

`client.sign_in` / `client.sign_up` are **scratch objects**. clerk-js keeps them on every client whether or not an attempt is in flight, and when none is, their fields are `undefined`. `JSON.stringify` drops undefined-valued keys entirely, so what reaches Rust for the ordinary signed-in user is close to:

```json
{ "object": "sign_up", "status": null }
```

- nearly every key `clerk-fapi-rs` requires, absent at once.

## Why this object keeps failing

Four incidents, each fixing the one field serde reported, deserialization then advancing by one field and failing on the next:

| # | field | outcome |
|---|---|---|
| 1 | `status` | `invalid type: null, expected string or map` |
| 2 | `verifications` | `invalid type: null` |
| 3 | `password_enabled` / `custom_action` | `missing field` |
| 4 | `id` | `missing field \`id\`` - **this one** |

This PR lists **every** field both models require, so there is no next field. Values match each model's own `impl Default` (`""` for the `String` id, `abandoned` for the status enum, `[]` for the `Vec`s), and every entry is an `or_insert`, so a key guest-js did send always wins - `backfill_never_overwrites_a_value_guest_js_did_emit` and the idempotency tests still pass unchanged.

## Why no fixture caught it

The module already learned once that hand-written fixtures hide defects, and switched to transcribing guest-js's serializer field-for-field. That was the right move and **still could not catch this**: a transcription is faithful to the *code* and silently wrong about the *runtime*. The serializer reads `"id": signUp.id`, so the fixture shows an `id` - and a human filling it in writes `"sua_abc123"`. At runtime that value is `undefined` and the key does not exist.

So the two new completeness tests feed the **bare discriminator and nothing else**:

```rust
let mut value = json!({ "object": "sign_in" });
backfill_clerk_client_json(&mut value);
serde_json::from_value::<ClientSignIn>(value).expect(...);
```

That cannot rot the same way, and a newly-required upstream field breaks it immediately rather than six weeks later on a user's machine.

## Tests

Four added, all confirmed failing before the fix with the exact production error `missing field \`id\``:

- `without_the_fix_...` -> `a_sign_up_whose_only_remaining_defect_is_a_missing_id_names_that_field` - drops `id` **after** the backfill, so every other defect is already repaired and `id` is provably the sole cause. (Dropping it from the raw fixture reports the float `abandon_at` instead, which is what the first draft of this test did.)
- `a_bare_sign_in_scratch_object_deserializes_after_backfill`
- `a_bare_sign_up_scratch_object_deserializes_after_backfill`
- `a_client_whose_scratch_objects_have_no_id_survives_the_round_trip` - end-to-end through `ClientClient`, which is the only shape parsed at runtime

`cargo test --workspace` green (41 in this module, 395 tray, 297 core). fmt + clippy clean.

## Note on release sequencing

1.90.0 is already the `minimum-version` floor (#865, merged), so the fleet is being force-updated onto the build carrying this bug. It is not a regression - 1.89.0 failed at the same site on a different field - but this is worth landing on the next staging cut rather than waiting.

Related: #866, #867 (both opened from the same OO investigation).

### DIFF
--- a/tray/src-tauri/vendor/tauri-plugin-clerk/src/json_backfill.rs
+++ b/tray/src-tauri/vendor/tauri-plugin-clerk/src/json_backfill.rs
@@ -64,7 +64,24 @@ fn defaults_for(object_type: &str) -> &'static [FieldDefault] {
         // included `abandon_at`, which guest-js never sends, so it round-tripped
         // and hid this. `0` matches how every other missing timestamp is
         // defaulted here.
-        "sign_in_attempt" => &[("abandon_at", || json!(0))],
+        //
+        // EVERY OTHER FIELD ON THE MODEL IS LISTED TOO, and that is the point —
+        // see the `sign_up_attempt` note below for why this object type is
+        // defaulted exhaustively rather than field-by-field as each one is
+        // observed failing.
+        "sign_in_attempt" => &[
+            ("id", || json!("")),
+            ("status", || json!("abandoned")),
+            ("supported_identifiers", || json!([])),
+            ("supported_first_factors", || Value::Null),
+            ("supported_second_factors", || Value::Null),
+            ("first_factor_verification", || Value::Null),
+            ("second_factor_verification", || Value::Null),
+            ("identifier", || Value::Null),
+            ("user_data", || Value::Null),
+            ("created_session_id", || Value::Null),
+            ("abandon_at", || json!(0)),
+        ],
         // `clerkSignUpToSignUpJSON` is missing more than one required field:
         // it emits `has_password` where the model wants `password_enabled`,
         // and never emits `custom_action` — both bare `bool`s upstream. It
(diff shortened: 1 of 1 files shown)


Akarsh-Hegde:
**Checked what happens *after* deserialization succeeds**, since that is the part the tests don't cover.

Before this fix `set_client` persisted nothing at all. After it, the restored cache contains a scratch `sign_in`/`sign_up` carrying values this PR invents - `id: ""`, `status: "abandoned"`. The justification given for each ("matches the model's own `impl Default`") is a serde argument, not a semantic one, and identity fields are exactly where a synthetic value tends to get branched on.

So: **nothing on our side reads them.** No field access to `sign_in`/`sign_up` anywhere in the plugin (`tauri_store.rs`, `lib.rs`, `commands.rs`) or in `tray/src-tauri/src/`. The client is serialized and restored wholesale; we never inspect the scratch objects, so a present-but-abandoned one cannot be mistaken for a resumable attempt by our code.

`status: "abandoned"` additionally has the semantic argument the module already makes for it - a null status means no attempt is in flight. `id: ""` has only the `Default` argument, which is sound here precisely because no consumer exists.

PR #885, opened 2026-08-25
fix(jira): fetch every assigned type except Epic, subtasks included

adityaharishch:
## What changes

```diff
- assignee = currentUser() AND statusCategory != Done AND type IN (Task, Feature, Bug)
+ assignee = currentUser() AND statusCategory != Done AND type != Epic
```

Sub-tasks are now fetched. So are Stories, and so is any custom type a user's
board defines.

## Why a denylist, not "add Story and Subtask to the list"

This JQL is the **only** door a Jira ticket enters `pm_tasks` through, which makes
an omission invisible rather than noisy. A type left out is absent from the Tasks
page, absent from the worklog matcher's candidates (hours against it can only ever
come back "no match"), and if a ticket of that type is filed through the plan
composer it becomes a **permanent `'local'` shadow row** - `plan_tasks::create`
reads "not in `pm_tasks` after a sync" as "self-assign failed", and for an
unfetchable type the heal never arrives.

That already cost us once, when `Bug` was missing. The list still omitted
**`Story`** - the default primary work type in every Jira Scrum project - so any
user on a standard board was syncing almost nothing. It produces an empty board
rather than an error, which is why nobody reported it.

Appending `Story` fixes today's known gap and leaves the trap armed for the next
custom type. Naming only what stays out closes the class.

Epics stay excluded because they are containers, not work: they already reach
Meridian as the `parent_key`/`epic_title` of their children.

## Sub-tasks

Reverses the earlier "tasks/features and their epics, never subtasks" rule, at the
user's request: on many boards the subtask is the unit of work someone actually
spends a day on, and excluding it meant that day could not be logged against
anything.

`type != Epic` admits them **without naming them**, which matters - the type name is
site-specific. Jira ships both `Subtask` and `Sub-task` as defaults and custom
subtask types are common, so a hardcoded name would work on some installs and
silently fail on others. (`subtaskIssueTypes()` is the name-independent function if
subtasks ever need selecting specifically.)

One consequence worth knowing: for a subtask Jira's `parent` is its Story/Task, not
its Epic, so `epic_title` holds the parent task's summary. Harmless where it is
consumed - `task_triage` uses it only as a context-anchor presence check
(`mod.rs:486`), and a subtask always has a parent, so it never trips
`NoContextAnchor`.

## Consistency

Jira was the **only** provider filtering by type at all. Linear, GitHub and Trello
fetch everything assigned to you; Azure's WIQL is `AssignedTo = @me` alone. This
brings Jira in line.

## Verification

Both JQL forms run against a live Jira Cloud site before committing - the new one
returns the same issues as the old one on a board with no Story type, confirming no
silent narrowing. `subtaskIssueTypes()` also confirmed valid there.

Tests now assert the **shape**: `type != Epic` present, `type IN` absent, and no
subtask exclusion. Tightening it back into an allowlist fails the build.

`cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`,
`cargo test --workspace` all clean.

## Not covered

No live test against a board that actually has Stories or subtasks assigned - the
site I could reach has neither. The JQL is verified valid; that it surfaces those
rows is inference from the semantics of `!=`.

### DIFF
--- a/src/intelligence/providers/jira/fetch.rs
+++ b/src/intelligence/providers/jira/fetch.rs
@@ -118,6 +118,51 @@ pub(super) async fn discover_start_date_field(ctx: &JiraReqCtx) -> Option<String
 const ACTIVE_TASK_JQL: &str =
     "assignee = currentUser() AND statusCategory != Done AND type != Epic ORDER BY updated DESC";
 
+/// One active-task fetch: the work items to upsert, plus how many rows the
+/// server actually returned before [`is_work_item`] filtered any out.
+///
+/// These two counts must not be conflated, and that is the entire reason this
+/// type exists rather than a bare `Vec`. The caller uses the count to decide
+/// whether the page may have been TRUNCATED at [`MAX_RESULTS`], and it only
+/// prunes `pm_tasks` when it was not — pruning against a partial list deletes
+/// live tickets. Using the post-filter length there would make a full page look
+/// partial the moment a single container row was dropped, silently turning the
+/// truncation guard off on exactly the boards this filter was added for.
+pub(super) struct ActiveFetch {
+    pub(super) issues: Vec<(JiraIssue, serde_json::Value)>,
+    pub(super) returned_by_server: usize,
+}
+
+/// The rung at and above which a Jira issue type is a CONTAINER rather than
+/// work: `1` is Epic, `2`+ are the Premium/Advanced-Roadmaps tiers (Initiative,
+/// Capability, and custom ones an org invents).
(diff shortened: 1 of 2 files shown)


adityaharishch:
Closed in favour of #888 - same work, reviewed end to end and squashed into one commit, plus two latent bugs the review turned up (the prune truncation guard being silently disabled by client-side filtering, and `NOT IN ()` on an empty fetch in both `prune` and the shared `mark_retained_offboard`).

PR #897, opened 2026-08-25
fix(oauth): retry a rotating refresh promptly, never late

adityaharishch:
## The bug

A Jira connection dies **permanently** - no user action, no way back except a manual re-authorisation:

> Jira sync failing - `403 Forbidden: {"error":"unauthorized_client","error_description":"refresh_token is invalid"}`

## Root cause

The daemon's own trace, from an affected install:

```
18:26:55.197  SPAN  post_token_retrying         <- the refresh POST leaves
18:55:29.574  WARN  ...error sending request    <- 28m34s later
18:55:31.675  WARN  403 unauthorized_client: refresh_token is invalid
```

The first attempt didn't fail fast. It hung for **28 and a half minutes**, because the laptop suspended mid-request. The 8 s timeout never fired: tokio's timers are monotonic, and a monotonic clock doesn't advance while the machine is asleep. So the retry 400 ms after the wake re-presented a refresh token Atlassian had rotated 29 minutes earlier.

That distance is the whole bug. [Atlassian documents a **reuse interval**](https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/) (10 min by default) in which exchanging the same rotating refresh token again is explicitly forgiven - *"this interval helps avoid network concurrency issues"* - and returns a fresh, valid pair. **Inside** it, that retry would have recovered the lost response and nobody would have noticed. **Outside** it, the identical request is breach detection's textbook signal.

A laptop suspending, or an app update restarting the daemon, while a background refresh is in flight is ordinary - which is why this is hitting many installs rather than a few.

## The fix

Retrying **is** the recovery mechanism, so switching it off would have left this user just as disconnected - the token was already spent and the ladder that could have retrieved its replacement would be gone. The rule is the window, not the failure class: **retry promptly, never late.**

- **`REUSE_GRACE_BUDGET` (5 min)** bounds how long after the first send a retry may still re-present the grant - half of Atlassian's documented interval. Deliberately not lower: the 1-5 minute band is where the ordinary interruptions land (a short lid-close, a slow restart), all of them forgiven, and declining those abandons grants one request would have restored. The clocks err safely - we count from *send*, the provider from *process*, so our elapsed is never the smaller.
- **`should_retry(kind, failure, wall_elapsed)`** - open window: retry (recovery). Closed: never (protection). Terminal never retries; interactive code exchanges are unaffected.
- **`wall_elapsed_since` uses `SystemTime`, not `Instant`.** An `Instant` would report that 28-minute request as milliseconds and wave the stale retry straight through - the same blindness that stopped the timeout firing. The clock we need is the one the provider is using.
- **The ladder re-checks the window after its backoff sleep**, since `tokio::time::sleep` is monotonic too.
- **Timeouts split:** one 8 s per-request cap becomes a 10 s connect cap inside a 30 s request cap. The old cap was its own way to lose a rotated token, and splitting connect out is what makes "never delivered" classifiable at all.
- **`ensure_fresh` no longer refreshes without the cross-process lock** - proceeding anyway double-spends the token. `LOCK_WAIT` 10 s -> 150 s, bounded by the ladder's own worst case and pinned by a test that recomputes it from flow's constants.
- **A failed `store::save` is no longer shrugged at.** The token in hand is the only one the provider accepts and the one on disk is already spent, so losing it costs the same reconnect an hour later. `save_rotated` retries first.

### Recovering an interrupted exchange (`refresh_in_flight_at`)

The above fixes the case where our process is alive to retry. It isn't always - an app update stops the daemon whenever it likes, and the provider then completes an exchange and answers into a socket with no process behind it.

That is repairable inside the reuse window, but nothing was reaching for it in time: the daemon refreshes **lazily**, when a sync finds its task cache stale, up to `SYNC_INTERVAL_MINS` away. An update interrupting a refresh five minutes after a successful sync got its next attempt twenty-five minutes later - past the window, grant gone.

- `OAuthTokens::refresh_in_flight_at` is stamped and flushed **before** the grant leaves, so a process that dies mid-exchange leaves evidence. Cleared on any outcome that resolves it (a 4xx: dead, nothing to repair; a connect failure: it never left). Only an *ambiguous* failure leaves it standing. `#[serde(default)]`, so every existing token file reads as "nothing outstanding".
- `flow::may_have_spent_the_grant` is the bookkeeping question, separate from `is_transient`'s user-facing one.
- `jira::recover_interrupted_refresh()` runs once at daemon startup, before the first sync, beside the two crash-recovery steps it's a sibling of. One file read when there's nothing to do.

## Tests

36 in `meridian-oauth`, all hermetic. The wire ones drive a scripted loopback token endpoint that counts requests it fully read, because the property that matters is how many times the grant reaches the provider.

- `a_lost_response_is_recovered_by_a_prompt_retry` - stub reads the grant then hangs up; the retry returns the pair whose first delivery was lost. **This is the case the incident should have been.**
- `a_rotating_grant_is_never_retried_once_the_reuse_window_has_passed` - the incident as it happened
- `a_short_interruption_is_still_inside_the_window` - the recoveries the budget must not decline
- `a_5xx_is_retried_inside_the_window_and_recovers`, `a_rotating_grant_still_retries_a_429_and_succeeds`
- `a_403_is_one_request_and_a_dead_grant` - the literal response, and the one case that *does* surface the reconnect remedy
- `a_lost_response_that_stays_lost_is_not_reported_as_a_dead_grant` - no false "Reconnect Jira" for a passing network fault
- `a_refused_connection_is_retry_safe` - a real `reqwest` connect error against a closed port, pinning the shipped version's `is_connect()` rather than our belief about it
- `the_in_flight_marker_round_trips_and_defaults_to_absent` - including a token file predating the field
- `lock_wait_outlasts_the_worst_case_refresh`

`cargo test -p meridian -p meridian-core -p meridian-oauth`: 1079 + all suites green. `cargo clippy` on the three, `-D warnings`, clean.

`cargo clippy --workspace` / `cargo test --workspace` can't run on this machine - `cidre`'s build script needs full Xcode, not Command Line Tools. Verified pre-existing: identical failure on a stashed tree.

## What this does and doesn't cover

Covered, silently, with no user involvement: a lost response with the process alive; an exchange interrupted by an app update, restart or crash; a double-spend between daemon and tray; a failed token write.

**Not covered: grants already revoked.** Those installs need one Reconnect in Settings - Integrations, and nothing shipped in an update can change that - OAuth requires the user's browser session to mint a new grant. (`prompt=none` was probed against `auth.atlassian.com`: with no live session it 302s to `id.atlassian.com/login` rather than returning `error=login_required` to the callback, so a "silent" re-auth attempt can strand a user on a login page we can't close. Worse than the banner.)

**Also not covered: a suspend that outlasts the reuse window** while a refresh is in flight - no code of ours runs while the machine is asleep. Eliminating that class entirely means refreshing server-side, which is a product decision, not a patch.

https://claude.ai/code/session_01Pw24sztR6WcCF1JEedMjzt

### DIFF
--- a/meridian-oauth/src/flow.rs
+++ b/meridian-oauth/src/flow.rs
@@ -754,6 +754,8 @@ fn decode(s: &str) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::sync::Arc;
 
     /// `is_transient` must find the [`TokenError`] even under the `.context()`
     /// layers `refresh`/`ensure_fresh` wrap it in — that chain walk is the whole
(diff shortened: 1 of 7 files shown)


adityaharishch:
Closing at the author's request - not proceeding with this approach for now. The branch `fix/jira-oauth-refresh-token-revocation` is retained if any of it is wanted later.

Issue #851, opened 2026-08-24
Meridian deleted all my past history and is stuck on 13m of activity recorded after 8hrs.


techinsf:
### What happened?

Few issues:
- Meridian duplicated all of my tasks x 20
- Meridian isnt recording anything its stuck on 13m of activity.
- Previous days status and breakdown are gone now.
- Past days it did work only showed maybe 2hrs out of 10hrs recorded.
- Meridian keeps freezing

### What did you expect?

- Not to duplicate all my tasks
- For it to continue recording my data and day
- To be able to view previous days regardless if only showed 2hrs
- To work as intended

### Steps to reproduce

1. Start meridian
2. Add 1 task
3. Work and let it record
4. Come back after some time
5. See nothing and last days wiped and when reviewing tasks it shows all of them 20x 

### Meridian version

latest

### macOS version

latest

### Chip

m4 max 48gb

### Relevant logs

```text
Meridian not working in terminal
```

### Doctor output

```text
Command doesnt work along with meridian by itself
```

Akarsh-Hegde:
Thanks for reporting this. We have identified a database corruption path and merged an automatic repair fallback in #859. We are also investigating the duplicated tasks, missing history, incomplete activity recording, freezing, and nonfunctional CLI commands as urgent issues.

Please install the next release once available and let us know whether the repair restores normal recording.

Akarsh-Hegde:
@techinsf The database repair fix has shipped. Please update Meridian and let us know whether activity recording and history are restored. If task duplication, freezing, or CLI failures continue after updating, please reopen this issue.

Thanks

PR #911, opened 2026-08-26
fix(sync): a transient SQLite fault must not fail a queued sync

adityaharishch:
## The bug

"Sync now" reported failure for syncs that succeeded, on the machine that had been reproducing the sync issues all along.

Measured on `1.91.0-staging.3`, from that machine's own logs:

```
18:19:02.510  SIGHUP received — reloading (graceful restart) pid=54835
18:19:02.510  shutting down pid=54835              ← daemon truncates the WAL
              ← tray's outcome() read fails, red error shown to the user
18:19:05.600  servicing PM sync request mode=force reason=sync_now seq=17
18:19:08.168  PM sync request completed task_count=Some(4)
```

The user was shown:

> could not read the sync outcome: reading the PM sync outcome: error returned from database: (code: 522) disk I/O error

`522` is `SQLITE_IOERR_SHORT_READ` (`10 | (2 << 8)`) — SQLite asked for a page and was handed fewer bytes than it expected, because the daemon's shutdown `PRAGMA wal_checkpoint(TRUNCATE)` truncated the WAL underneath a live reader. Connecting a tracker reloads the daemon, so a "Sync now" pressed right after connecting lands in that window.

**The daemon did the sync.** It completed 5.6 s later, with 47 poll turns still unspent in the 30 s budget. The failure was entirely in how the tray reacted to a read that didn't land.

## The fix

Once the request row is written, the daemon owns the work. A failed *read* of the outbox is the absence of an answer, not an answer — the only honest terminal states are "the daemon reported a result" or "the budget elapsed, it's still going".

Both SQLite call sites in `ask_daemon_to_sync` treated any `Err` as terminal. The wait loop's own comment already said it should keep polling, and it only did so when `recover_if_corrupt` returned true — which `522` is not.

- **`db::integrity::is_transient_sqlx` / `is_transient_error`** — classify the `IOERR` (10), `BUSY` (5) and `LOCKED` (6) families by low byte, so extended variants like `522` are covered. Deliberately **disjoint from corruption**: retrying a damaged page is precisely what the ETL latch exists to prevent, and `corruption_is_never_transient` pins that. No message fallback either, unlike `is_corrupt_error` — a stringified error stays terminal, which is the safe direction.
- **Wait loop** — retry a transient fault instead of surfacing it.
- **Write path** — retry a transient fault, and treat a momentarily closed pool as the normal state it is. A "Sync now" pressed while `reload_with_pool_cycle` had the pool closed returned *"the database is not open - Meridian may be repairing it"*, naming a repair that was not happening.

Applied to the trace above, the read fails at 18:19:02.5, the loop keeps polling, and the poll at 18:19:08.2 reads `completed_seq=17` and renders **"synced 4 task(s)"**.

## Relationship to #910

#910 fixed outcomes being *destroyed* by an overlapping request (the sequence watermark). That was real and this PR does not touch it — the coalescing is visibly working in the same logs (`seq=10` was absorbed into the `seq=11` sync rather than triggering its own).

But #910 did **not** fix the reported symptom, and I presented it as though it had. This is the actual cause of the timeout and the error text the user kept seeing.

## Explicitly not fixed here

**The tray can still open a connection during the daemon's shutdown checkpoint.** `reload_with_pool_cycle` closes the tray's pool, signals, and reopens *lazily and immediately* — so no connection spans two daemon generations, which is what it was built for. It does nothing to stop a brand-new connection reading while the daemon spends a few hundred ms flushing and truncating. This PR makes that window survivable rather than closed.

Closing it is a change to the reload lifecycle that two production incidents already live in (#851's corruption, and the `ThrottleInterval` outage), so it wants its own PR and its own reasoning.

**One loose end I have not proven.** An earlier failure on the same machine showed the tray handed `seq=9` while the daemon read the same row as `seq=8` two seconds later. I hypothesised a lost commit across the checkpoint. A short read during the `SELECT seq` read-back explains it just as well and requires no data loss, and I could not distinguish the two. Under this fix that case degrades to a neutral "still running in the background" rather than a red error, which is truthful — but I am not claiming a lost write is impossible.

## Testing

- `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo test --workspace` — 1120 + all members, green
- New: `transient_codes_match_primary_and_extended`, `corruption_is_never_transient`, `a_stringified_transient_error_is_not_retried`, `transient_faults_are_retried_before_any_terminal_failure`, `the_request_retry_budget_covers_a_daemon_reload`
- The ordering guard was **verified to fail** with the wait-loop check removed, rather than assumed to work

The ordering guard is a source scan, for the reason `no-native-dialogs.test.ts` is: provoking a short read needs two real processes racing a `wal_checkpoint(TRUNCATE)` on a real file, which an in-memory pool cannot reproduce. **No test here exercises the real race** — the verification that matters is on the staging DMG.

## Manual verification needed on staging

1. Connect Jira, then press **Sync now immediately** — expect a real result, not `(code: 522)` and not "the database is not open"
2. Repeat 5+ times, pressing as fast as possible after connect, since the failure is timing-dependent (it passed 2 of 3 hand-run attempts before)
3. Watch `~/.meridian/bin/meridian logs -f | grep -iE "transient fault|servicing PM sync|did not report"` — a `transient fault ... retrying` line followed by a successful sync is the fix working

### DIFF
--- a/src/db/integrity.rs
+++ b/src/db/integrity.rs
@@ -145,6 +145,56 @@ pub fn is_corrupt_sqlx(err: &sqlx::Error) -> bool {
     }
 }
 
+/// SQLite primary result codes that mean "this attempt failed, the database is
+/// fine" - the file was locked, or a read came up short because another process
+/// moved the ground under it.
+///
+/// `SQLITE_IOERR` (10) is the one that matters here and the reason this exists.
+/// Its extended variant `SQLITE_IOERR_SHORT_READ` (522, `10 | (2 << 8)`) is
+/// SQLite asking for a page and being handed fewer bytes than it expected -
+/// which is exactly what a reader sees when the daemon's shutdown
+/// `PRAGMA wal_checkpoint(TRUNCATE)` truncates the WAL underneath it. Measured
+/// on 1.91.0-staging.3: the tray's "Sync now" read of the outbox raced a daemon
+/// reload and surfaced `(code: 522) disk I/O error` to the user, for a sync the
+/// daemon then completed successfully 5 s later.
+///
+/// `SQLITE_BUSY` (5) and `SQLITE_LOCKED` (6) belong to the same class: two
+/// processes write this file, so losing a lock race is a normal event.
+const SQLITE_TRANSIENT: [u32; 3] = [10, 5, 6];
+
+/// True when `err` is SQLite reporting a **retryable** failure - the operation
+/// did not happen, but nothing is wrong with the data.
+///
+/// # Never confuse this with corruption
+///
+/// Corruption (11) and "file is not a database" (26) are deliberately absent.
+/// Those are latching conditions that need an operator, and retrying them just
+/// re-reads damaged pages - the daemon's ETL latch exists for precisely that
+/// reason. This predicate is only for faults where trying again is the correct
+/// and sufficient response.
+///
+/// The low byte is compared rather than the whole code, so extended variants
+/// (`_SHORT_READ`, `_READ`, `_WRITE`, `_FSYNC`, …) are all covered - the same
+/// `primary | (n << 8)` scheme [`is_corrupt_sqlx`] relies on.
+pub fn is_transient_sqlx(err: &sqlx::Error) -> bool {
+    let sqlx::Error::Database(db) = err else {
+        return false;
+    };
+    db.code()
+        .and_then(|c| c.parse::<u32>().ok())
+        .is_some_and(|code| SQLITE_TRANSIENT.contains(&(code & 0xFF)))
+}
+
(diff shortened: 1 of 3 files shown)


adityaharishch:
Closing - superseded by #912, which reverts #909 and #910 wholesale. The diagnosis this PR was based on is preserved in #912's description so it isn't lost: the remaining cause of the staging.3 failure was a `(code: 522) SQLITE_IOERR_SHORT_READ` from the daemon's shutdown `wal_checkpoint(TRUNCATE)` truncating the WAL under a tray read, which the tray treated as terminal.

